Issue #166, opened 2019-10-14
Fine grain authorization


jankaspar:
- [x] Implement queue ownership by user or group
- [x] Add permissions for all possible user actions

PR #187, opened 2019-10-22
Implement basic permission checks and Improve OIDC support.

jankaspar:


config/armada/config.yaml
```diff
@@ -1,3 +1,4 @@
+development: false
```
JamesMurkin: Should development be default false?

internal/common/startup.go
```diff
@@ -44,6 +45,8 @@ func LoadConfig(config interface{}, defaultPath string, overrideConfig string) {
 	viper.SetEnvPrefix("ARMADA")
 	viper.AutomaticEnv()
 
+	fmt.Print(os.Getenv("ARMADA_DEVELOPMENT"))
```
JamesMurkin: Is this intended?

config/armada/config.yaml
JamesMurkin: Add a brief note saying to set this for local development in docs

jankaspar:
Fixes #166.

Issue #105, opened 2019-09-13
Write commandline formatter so we can use log in commandline tools rather than fmt


JamesMurkin:
**Problem**

All of our code in commandline tools (armadactl/armada-load-tester) have to use fmt.Println to log messages to the user.

This is inconsistent with the other code and less flexible (if we say wanted to provide a log of what user had done to a file etc).

**Suggested solution**
We could implement a custom logrus formatter, so it would printout like we are using fmt.Println but actually runs through logrus. 

Some details on how to do that:
https://stackoverflow.com/questions/43022607/how-do-i-disable-field-name-in-logrus-while-logging-to-file

PR #195, opened 2019-10-23
Add a log CommandLineFormatter

JamesMurkin:
This allows us to format log messages how you'd expect to see them on the command line

So we can use log.Info etc everywhere (rather than a mix of fmt/log)

We can then just swap the formatter for the commandline tools


JamesMurkin:
#105

PR #204, opened 2019-10-24
Do not try to push images from fork builds.

jankaspar:


.circleci/config.yml
```diff
@@ -67,6 +67,12 @@ jobs:
           name: Push Image
           command: |
 
+            if [ -z "${DOCKERHUB_USER}" ]
+              then
+                echo "Do not push image inside PR"
```
jankaspar:
```suggestion
                echo "Do not push image inside fork."
```

Issue #199, opened 2019-10-23
Some improvements for the README


itamarst:
I went through the quickstart, and got through it, which was pretty nice. 

Here are some small changes that might improve it:

* Mention need for 5GB+ disk space. I ran out.
* Maybe add a sentence explaining what `kind` is, for those who haven't heard of it.
* Installing `kind` via the official install instructions is probably faster (and gets you an official release rather than work-in-progress off master branch).
* `--expose=6379 --network=host` only works on Linux. You might want to change that to `-p 6379:6379` or that'll break for Mac users.
* Default configuration on Linux is requiring root to run `docker`, which you do somewhat address, but I missed it until I went looking to see if it was there. Might be useful to remind people to run all commands in steps 3 and on as root if they have that setup (I tried using `sudo`, and that let to unexpected issues).
* Maybe split the two executor runs into different bullet points.

PR #206, opened 2019-10-24
Adding improvements to README

JamesMurkin:
This should make it easier to install, by making the requirements more explicit.

Hopefully also it'll be being more reliable by pointing out common problems early and using release versions + compatible versions of commands.


JamesMurkin:
#199

PR #211, opened 2019-10-28
Allowing TLS for unauthenticated except when on localhost

JamesMurkin:


JamesMurkin:
Should we add a flag to force no-tls? Or some other flag to allow control over TLS in general? 

Currently this would prevent connecting to an unauthenticated server no on local host. Which could happen via:

- NodePort installation
- Hosting on docker on a different machine

Both not expected for now, so we could leave it until someone requests.

Issue #201, opened 2019-10-23
Broken jobs aren't flagged in useful way by `armadactl submit`


itamarst:
### Example 1: Completely missing everything

If the JSON/YAML is completely broken, no job is submitted, but also no useful error is printed:

```shell
$ cat empty.json 
{}
$ ./bin/armadactl submit empty.json 
$ echo $?
0
```

I would've expected an error about missing fields, and a non-zero exit code.

### Example 2: Bad pod spec

If the pod spec is no good, the job is successfully submitted, and then it get stuck in a loop between `Leased` and `Queued` state:

```
$ ./bin/armadactl submit badpodspec.yaml 
Submitted job id: 01dqwxzkxfwrrfhmr4hvh5zw5a (set: job-set-2)
$ ./bin/armadactl watch job-set-2
Watching job set job-set-2
Oct 23 18:25:44 | Queued:   0, Leased:   0, Pending:   0, Running:   0, Succeeded:   0, Failed
:   0, Cancelled:   0 | event: *api.JobSubmittedEvent, job id: 01dqwxw5b06tqc909ebxd2029f    
Oct 23 18:25:44 | Queued:   1, Leased:   0, Pending:   0, Running:   0, Succeeded:   0, Failed
:   0, Cancelled:   0 | event: *api.JobQueuedEvent, job id: 01dqwxw5b06tqc909ebxd2029f       
Oct 23 18:25:46 | Queued:   0, Leased:   1, Pending:   0, Running:   0, Succeeded:   0, Failed
:   0, Cancelled:   0 | event: *api.JobLeasedEvent, job id: 01dqwxw5b06tqc909ebxd2029f       
Oct 23 18:27:46 | Queued:   1, Leased:   0, Pending:   0, Running:   0, Succeeded:   0, Failed
:   0, Cancelled:   0 | event: *api.JobLeaseExpiredEvent, job id: 01dqwxw5b06tqc909ebxd2029f 
Oct 23 18:27:46 | Queued:   0, Leased:   1, Pending:   0, Running:   0, Succeeded:   0, Failed
:   0, Cancelled:   0 | event: *api.JobLeasedEvent, job id: 01dqwxw5b06tqc909ebxd2029f 
```

Errors in the pod spec are quite likely to occur, and the debug cycle is difficult if I now have to dig deep into ... the logs somewhere, not sure where as a new user, to figure out what I did wrong.

[...]

PR #214, opened 2019-10-29
Handling pod submission failure in Executor

JamesMurkin:
When a pod is submitted to Kubernetes and it fails, it is either because:
 - The pod and therefore job's podSpec are invalid
 - Some transient issue

We identify which of these it is, based on if it returns unprocessable entity status code (422)

If invalid podSpec
 - Send fail event
 - Report job done

else
 - Send lease returned event
 - Return lease


JamesMurkin:
#201

Issue #160, opened 2019-10-10
Make watch client code usage more efficient


JamesMurkin:
**Problem**
When watching a very large jobset (100000+ jobs) it takes a very long time to catch up to the most recent events in the stream.

**Cause**
Currently the code that uses watch generates a summary of the current state each time an event comes in.

This involves looking through all currently seen jobs and reporting what state they are in. 

For large jobsets this is highly inefficient and means the application uses a LOT of CPU along with being slow. 

- As every event that comes in, loops over 100000 objects and prints a line

**Suggested solution**

We should just make a state struct that holds the current state. Each new event would just add to the state, rather then regenerating a state summary each time.

PR #218, opened 2019-10-30
Changing watch to maintain a context rather than recalculating state summary each time

JamesMurkin:
Currently whenever we want to summary of the events we watch. We loop through every job we have seen events for and sum up which state they are currently in.

This becomes very slow when you have 50000+ jobs, as each event you'll process all 50000 jobs

Now we hold a summary of the current state, so it isn't recalculated on each event.
 - Each event just updates the current state
 - Getting a summary of the current state is just a case of looking it up, rather than recalculating

We could do more optimisations to make starting watching a big jobset faster:
 - Start from current (This may make the summary confusing however, as it won't include all jobs in the jobset)
 - We could not print anything until we catch up with the current state
   - Or just print every 1s rather than every event

Printing takes most of the time now. So skipping past until current saves a LOT of time


JamesMurkin:
#160

PR #219, opened 2019-10-30
Current usage scheduling

jankaspar:


internal/armada/scheduling/resources_test.go
```diff
@@ -0,0 +1,57 @@
+package scheduling
+
+import (
+	"testing"
+
+	"github.com/magiconair/properties/assert"
```
JamesMurkin: This should be testify

Issue #161, opened 2019-10-10
Allow submitting jobs to the api in bulk


JamesMurkin:
**Problem**
When submitting 10000 - 100000 jobs at a time, it takes a long time to complete submission.

This is because the jobs are submitted one at a time.

**Suggested solution**
- The API should be updated to support bulk submission, 

- The client should be updated to submit in groups of say 100 at a time

This should significantly speed up submitting many jobs at once.

PR #224, opened 2019-11-04
Replacing SubmitJob with SubmitJobs

JamesMurkin:
Currently we have an endpoint SubmitJob

However this is very inefficient when submitting many jobs (which we expect to be common for this application)
Also many API calls (one for each SubmitJob) can cause load balancers to throttle connections to the server.

I have replaced SubmitJob with SubmitJobs, so you can submit many jobs in a single call
 - All jobs must belong to the same queue/jobsetid, as specified in the JobSubmitRequest struct)

I have updated client code to use the new SubmitJobs endpoint


JamesMurkin:
#161

PR #226, opened 2019-11-05
Remove using root user in docker + helm

JamesMurkin:


build/armada/Dockerfile
```diff
@@ -7,4 +7,10 @@ RUN apk update && apk add --no-cache ca-certificates
 
 WORKDIR /app
 
+RUN addgroup -S armada && adduser -S armada -G armada
```
JamesMurkin: Is this excessive. We could just use USER 1000 or something

deployment/armada/templates/deployment.yaml
```diff
@@ -18,6 +18,10 @@ spec:
     spec:
       terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
       serviceAccountName: {{ include "armada.name" . }}
+      securityContext:
```
JamesMurkin: This could be configurable and this just be the default

PR #230, opened 2019-11-07
Add DotNet client & gRPC REST Gateway.

jankaspar:


cmd/armada/main.go
```diff
@@ -38,6 +38,9 @@ func main() {
 	shutdownMetricServer := common.ServeMetrics(config.MetricsPort)
 	defer shutdownMetricServer()
 
+	shutdownGateway := armada.ServeGateway(8080, config.GrpcPort)
```
JamesMurkin: Could this port be configurable, for consistency with other ports?

Issue #229, opened 2019-11-07
Add rolling policy to helm chart


JamesMurkin:
**Problem**
When you helm upgrade the armada server component, all pods restart at once, potentially causing an outage.

**Suggestion**
Add to armada server helm chart, to roll pod restarts out one at a time rather than all at once

This way all pods get updated, but there is no outage due to no pods being alive when the deployment is updated.

PR #236, opened 2019-11-12
Improving helm chart roll out process

JamesMurkin:
Currently when the deployment is updated, all pods are deleted at once and new pods are created.

This can cause short downtime when a new version goes out.

Adding:
- RollingUpdate strategy which makes sure a maximum of one pod can be unavailable at once
   - Made the default replicas 3, so the rollingUpdate strategy causes no downtime
- Increased terminationGracePeriodSeconds to 5 seconds, to give components a few seconds to stop gracefully


JamesMurkin:
#229

deployment/armada/values.yaml
```diff
@@ -9,9 +9,12 @@ resources:
     memory: 512Mi
     cpu: 200m
 additionalLabels: {}
-terminationGracePeriodSeconds: 0
-replicas: 1
-
+terminationGracePeriodSeconds: 5
+replicas: 3
```
jankaspar:
```suggestion
replicas: 1
```
Lets keep this 1 for quick start.

Issue #237, opened 2019-11-13
Make armadactl submit validation work offline


JamesMurkin:
**Problem**

Currently armadactl uses kubeval to validate a pod spec being submitted.

kubeval assumes it has access to the internet. When it cannot access the internet, it gets:
stdin: Failed initalizing schema https://kubernetesjsonschema.dev/master-standalone-strict/pod-v1.json: Get https://kubernetesjsonschema.dev/master-standalone-strict/pod-v1.json: dial tcp: lookup kubernetesjsonschema.dev: no such host

**Solutions**
* Replace kubeval with alternative
* Use offline validation with kubeval:
   * https://github.com/instrumenta/kubeval/issues/21
   * https://github.com/instrumenta/kubeval/pull/25
* Remove armadactl validation as it is only helpful, not mandatory

PR #249, opened 2019-11-18
Workaround to handle submitting jobs when offline

JamesMurkin:
Now if validation cannot be performed due to being unable to download the schema (likely due to being offline), we'll warn users but not block

Alternatives:
- We build in a version of the pod schema into our code and use this always
- We build in a version of pod schema into our code and fallback to this

Both of the above are OK however building static files into binaries is a bit rubbish in Go.
Also it'd mean we are "locked" to the version of pod spec we happen to use. Second alternative improves on this a bit, by only using it when the current version is unavailable.


internal/client/validation/submit_file.go
```diff
@@ -46,10 +47,15 @@ func ValidateSubmitFile(filePath string) (bool, error) {
 		return false, errors.New("Warning: You have provided no jobs to submit.")
 	}
 
+	noSchema := false
 	for i, job := range submitFile.Jobs {
 		rawPod := rawPod(job.PodSpec)
 		result, err := validate(rawPod)
 
+		if !result[0].ValidatedAgainstSchema {
+			noSchema = true
```
jankaspar:
```suggestion
			noSchema = true
			break
```

JamesMurkin:
#237

Issue #240, opened 2019-11-14
Investigate tidying helm charts to use better formatting


JamesMurkin:
**Problem**
Currently our helm charts format "special operation", such as if or toYaml, to the left

This breaks the natural flow of indenting.

This makes it harder to read and programs that help you format the yaml (such as Goland) don't like that fact the formatting doesn't have proper indenting.

**Possible solution**
Here is an example of a chart that looks tidier:
https://github.com/codecentric/helm-charts/blob/master/charts/jenkins/templates/deployment.yaml

It uses with and nindent

PR #256, opened 2019-11-22
Using nindent to allow formatting helm with proper indentation 

JamesMurkin:
Our helm charts were formatted with non-standard indentation to allow the templating to result in the correct output

This is not ideal, as:
 - Tools expect certain indentation and when you go against this, they no longer help you and/or point out things as errors when they aren't
 - Makes it hard to read and understand the actual resulting indentation

Now we helm charts can be written with the same indentation that the resulting yaml will have, making it much easier to work with


JamesMurkin:
Fixes #240

PR #260, opened 2019-11-28
Making executor consider all pending pods as requiring resource

JamesMurkin:
Currently we consider pods requiring resource as:
 - Those running on "processing nodes"
 - Those that are "managed" by Armada and not assigned a node (i.e Pending)

The issue with this is that pods that are pending and not "managed" are not considered by Armada

So Armada keeps scheduling pods to take up the space this pending pod wants to use

This change should make it easier for non-Armada pods to get on the cluster (Pending -> Scheduled)
 - As we now leave space for it to get on


jimbobby5:
Closing old PR

Issue #261, opened 2019-11-28
Fail jobs when they are rejected by an admission webhook


JamesMurkin:
*Problem*

When jobs are submitted. If they are rejected as they don't meet the requirements of an admission webhook they have their lease returned.

This causes them to constantly bounce between being submitted and being returned.

If they are rejected by an admission webhook, likely they'll always be rejected by it.

*Solution*
When they are rejected due to admission webhook, the job should be failed rather than having lease returned.

PR #267, opened 2019-11-29
Failing jobs when they are rejected by an admission webhook

JamesMurkin:
When a job is rejected by an admission webhook, this means it'll keep getting rejected so we should fail it so the user fix it.

The "best" way to determine if it was a admission webhook that rejected it, is by looking for "admission webhook" in the message

This is because:
 - Users can define their own return codes from webhooks (Although we could just fail all in the 4xx range
 - By default Kubernetes returns a 500 on validating webhook rejection. So this can't be differentiated from a genuine kubernetes failure (which we would want to retry on)

Ideally Kubernetes changes the default to 4xx and we can just fail any pods when a 4xx response happens


JamesMurkin:
Fixes #261

PR #285, opened 2019-12-18
Support watching multiple job IDs.

### LINKED ISSUE #162: Add to watch to allow watching multiple job sets
**Problem**
If you run a load test with multiple jobsets. 

The loadtest will watch these multiple jobsets.
 
However there is no way to stop/restart this watching of multiple jobsets. meaning you have to leave it running and/or can't rewatch an experiment how it happened.

**Suggested solution**
- Armadactl watch should be updated to support watching multiple jobsets 
- armada-load-tester should be updated to support watching multiple jobsets (potentially on multiple connections, as is the case in loadtest)

itamarst:
Fixes part of #162, specifically the bit that covers `armadactl watch`.

### DIFF
--- a/cmd/armadactl/cmd/watch.go
+++ b/cmd/armadactl/cmd/watch.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


jankaspar:
To change the api you need to change `https://github.com/G-Research/armada/blob/master/internal/armada/api/event.proto#L148` and run `./scripts/proto.sh`

itamarst:
Ah, I'd implemented this before your comment, and I thought that comment was about the second half of #162, the load balancer.

jimbobby5:
Closing old PR

PR #286, opened 2019-12-19
Merging Cancel/Remove into DeleteJobs and making it clean up job fully 

JamesMurkin:
Cancel and Remove were doing the same thing with subtle differences around how they were reporting

Now there is just one DeleteJobs, which cleans all information from the database about the job

The DeleteJobs function only returns Jobs that were actually deleted and omits untouched jobs:
 - If the job doesn't exist or nothing was deleted, it won't be in the returned
 - If information was deleted for the job, it will be returned with the value of nil
 - If an error occurred while deleting, it will be returned with a value of the err message

Finally the job object is only marked to be cleaned up by redis in 7 days, rather than deleted immediately. This gives people time to investigate issues with the original data before it is gone. 
 - In the future this won't be needed. As all information is in the events, it is just hard to view currently


internal/armada/repository/job.go
```diff
@@ -169,88 +168,82 @@ func (repo *RedisJobRepository) ReturnLease(clusterId string, jobId string) (ret
...
+func processDeletionResponse(deletionResponse *deleteJobRedisResponse) (int64, error) {
+	var totalUpdates int64 = 0
+	var errorMessage error = nil
 
-	jobs, e := repo.getJobIdentities(jobIds)
+	modified, e := deletionResponse.removeFromLeasedResult.Result()
+	totalUpdates += modified
 	if e != nil {
-		return nil, e
+		errorMessage = e
```
jankaspar: It might be worth to wrap the error here with description of what actually failed.

internal/armada/repository/job.go
```diff
@@ -282,19 +290,27 @@ func (repo *RedisJobRepository) TryLeaseJobs(clusterId string, queue string, job
...
-	}
+	_, _ = pipe.Exec() // ignoring error here as it will be part of individual commands
 
 	var jobs []*api.Job
-	for _, cmd := range cmds {
+	for index, cmd := range cmds {
+		_, e := cmd.Result()
+		if e != nil {
+			if e == redis.Nil {
+				log.Warnf("No job found with with job id %s", ids[index])
```
jankaspar: Curious how many of these messages will we get, not sure if I would consider it warning, its expected.

JamesMurkin:
Fixes #192

Issue #290, opened 2020-01-06
Handle cancellation in C# Client watch


JamesMurkin:
In the C# client, when watch it called, we use reader.EndOfStream.

This blocks waiting for events (which is good from an efficiency point of view). However there is no way to cancel this and if no events come, it'll block forever.

It would be great if we could respect the cancellationToken getting cancelled while also blocking waiting for events.

The only reason this isn't completely tragic is:
 - If you are using NGINX ingress, it drops idle connections every 30s seconds. Otherwise it'd block forever with no way out.

PR #291, opened 2020-01-06
Preventing dotnet client from hanging unexpectedly when it is cancelled

JamesMurkin:
To exit the watch of the dotnet client, we cancel the CancellationToken

However if you cancel when at the end of the stream (i.e on last message) then the watch will hang while EndOfStream looks for the next line

Swapping the order of these 2 in the if statement causes it to exit as expected
 - As the cancelled token is checked first, which fails the if and breaks out of the watch

We should also consider doing some work so a cancel in general is not blocked on EndOfStream check
 - This would require a different way to watching, so I am not doing it in this commit


JamesMurkin:
#290

Issue #13, opened 2019-07-05
Create component performance test


jankaspar:
- [x] Implement fake executor
- [ ] Stress test

PR #292, opened 2020-01-14
Implement fake executor

jankaspar:
This is part of #13.
Fake executor component pretends to be executor, but instead of creating pods on Kubernetes cluster it produces expected job events without actually running anything. It can be used to stress test the Armada server component.

### DIFF
--- a/internal/executor/fake/context.go
+++ b/internal/executor/fake/context.go
@@ -1,0 +1,143 @@
+package fake
+
+import (
+	"math/rand"
+	"regexp"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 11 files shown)


internal/executor/fake/context.go
```diff
@@ -0,0 +1,143 @@
...
+	}()
+
+	return pod, nil
+}
+
+func extractSleepTime(pod *v1.Pod) float32 {
+	command := append(pod.Spec.Containers[0].Command, pod.Spec.Containers[0].Args...)
+	commandString := strings.Join(command, " ")
+
+	// command needs to be in the form: sleep sleep $(( (RANDOM % 60) + 100 ))
```
JamesMurkin:
```suggestion
	// command needs to be in the form:  sleep $(( (RANDOM % 60) + 100 ))
```

PR #295, opened 2020-01-17
Expirying events after a given period of time

JamesMurkin:
Currently we retain events forever, however this will not be sustainable in production, especially as we use Redis which is in memory

Eventually we'll run out of space (memory in Redis's case) to hold events

Now you can configure:
 - If Armada should expire events (On by default)
 - How long it should retain events for before expiring them (2 weeks by default)

This should prevent Armada grinding to a halt eventually due to hitting storage limits

### DIFF
--- a/config/armada/config.yaml
+++ b/config/armada/config.yaml
@@ -33,3 +33,6 @@ scheduling:
     memory: 100000000 # 100Mb
     cpu: 0.25
   maximalClusterFractionToSchedule: 0.25
+eventRetention:
+  expiryEnabled: true
+  retentionDuration: 336h # Specified as a Go duration, values less than 0 (i.e -1) will result in infinite retention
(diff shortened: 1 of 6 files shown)


config/armada/config.yaml
```diff
@@ -32,4 +32,7 @@ scheduling:
   minimumResourceToSchedule:
     memory: 100000000 # 100Mb
     cpu: 0.25
-  maximalClusterFractionToSchedule: 0.25
\ No newline at end of file
+  maximalClusterFractionToSchedule: 0.25
+eventRetention:
+  expiryEnabled: true
+  retentionDuration: 336h # Specified as a Go duration, values less than 0 (i.e -1) will result in infinite retention
```
jankaspar:
```suggestion
  retentionDuration: 336h # Specified as a Go duration
```

PR #311, opened 2020-02-18
Implement minimum pod lifetime

jankaspar:
For many log aggregators pods needs to be present on the cluster long enough to get logs scrapped.


internal/executor/service/job_lease.go
```diff
@@ -71,29 +78,30 @@ func (jobLeaseService *JobLeaseService) ReturnLease(pod *v1.Pod) error {
...
 func (jobLeaseService *JobLeaseService) ManageJobLeases() {
-	podsToRenew, err := jobLeaseService.clusterContext.GetBatchPods()
+	batchPods, err := jobLeaseService.clusterContext.GetBatchPods()
 	if err != nil {
 		log.Errorf("Failed to manage job leases due to %s", err)
 		return
 	}
 
-	chunkedPods := chunkPods(podsToRenew, maxPodRequestSize)
+	podsTopRenew := util.FilterPods(batchPods, isToRenew)
```
JamesMurkin: Should be podsToRenew?

internal/executor/service/job_lease.go
```diff
@@ -136,16 +145,49 @@ func (jobLeaseService *JobLeaseService) renewJobLeases(pods []*v1.Pod) {
...
 		}
 	}
+}
 
-	return finishedPods
+func isToRenew(pod *v1.Pod) bool {
+	return !isReportedDone(pod)
+}
+
+func isToReportDone(pod *v1.Pod) bool {
```
JamesMurkin: Maybe isReadyToBeReportedDone? Or something. Not sure, isToReportDone sounds odd to me

internal/executor/util/pod_util.go
```diff
@@ -125,3 +128,52 @@ func MergePodList(list1 []*v1.Pod, list2 []*v1.Pod) []*v1.Pod {
...
+		maxStatusChange = maxTime(maxStatusChange, condition.LastTransitionTime.Time)
+	}
+
+	if maxStatusChange.IsZero() {
+		return maxStatusChange, errors.New("cannot determine last status change")
+	}
+	return maxStatusChange, nil
+}
+
+func FindLastContainerStartTime(pod *v1.Pod) time.Time {
```
JamesMurkin: This func doesn't really do what it says it does, at least not entirely. Also it overlaps nicely with the above func. Maybe we could improve naming or something here.

Issue #314, opened 2020-02-20
Make lease expiry configurable and run in background loop


JamesMurkin:
Currently leases expire on server side after a hard coded 2 minutes.

Also lease expiry only occurs on LeaseJobs being called. Which is has a few disadvantages:
 - No leases expire if LeaseJobs is never called
 - It slows down LeaseJobs call
 - The more clusters you have, the more often we'll try to expire leases, which is likely wasteful as we don't need to attempt to expire leases super frequently.

**Actions**
 - We should make lease expiry duration configurable
 - We should make lease expiry happen in a background loop, independent of api calls.

PR #318, opened 2020-02-26
Making lease expiry configurable and moving it to background loop

JamesMurkin:
You can now configure:
 - lease expireAfter - The amount of time for leases to expire
 - lease expiryLoopInterval - How frequently the background loop expiring leases runs

Additionally I have moved remove all lease expiry from LeaseJobs and made it a background job
 - This should remove potential slowdown from LeaseJobs
 - Have a better separation of responsibilities
 - Lets us control how often lease expiry is checked, rather than relying on executor calls to the server to call expire leases


JamesMurkin:
Fixes #314

Issue #320, opened 2020-02-26
Refactor background tasks in Server and Executor to use common code


JamesMurkin:
Currently we do similar things to register background tasks in the Server and Executor 

It would be good to make something in common that manages background tasks.

Then they could be register and manager background tasks in a similar ways and reuse the code.

PR #321, opened 2020-02-26
Refactoring background task code to common

JamesMurkin:
Now server and executor components use the shared BackgroundTaskManager for managing their background tasks

This is better code reuse and makes sure improvements are shared between these components


JamesMurkin:
Fixes #320

PR #347, opened 2020-03-19
Quickstart improvements

jgiannuzzi:
* Fix Prometheus installation warnings
* Make it work on macOS and Windows
* Remove obsolete and duplicate commands
* Do not rely on the Docker host IP

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -27,6 +27,29 @@ brew install helm kind kubernetes-cli
 
 Ensure at least 5GB of RAM are allocated to the Docker VM (see Preferences -> Resources -> Advanced).
 
+#### Windows
+
+You can install the pre-requisites with [Chocolatey](https://chocolatey.org):
+
+```cmd
+choco install git docker-desktop kubernetes-helm kind kubernetes-cli
+```
+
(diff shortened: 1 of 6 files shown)


docs/quickstart.md
```diff
@@ -3,14 +3,60 @@
...
+
+You can install the pre-requisites with [Chocolatey](https://chocolatey.org):
+
+```cmd
+choco install git docker-desktop kubernetes-helm kind kubernetes-cli
+```
+
+Ensure at least 5GB of RAM are allocated to the Docker VM (see Settings -> Resources -> Advanced).
+
+In order to follow the instructions below as-is, you should also install [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/about).
```
jankaspar: I wonder if this would all just work with bash on windows (it is usually installed with git).

jgiannuzzi: Good point, I'll try it out.

jgiannuzzi: Indeed, it works fine with Git Bash, I just had to adapt the downloading of `armadactl`

jgiannuzzi:
@jankaspar I dropped the WSL requirement. Can you please review again? Thanks!

Issue #353, opened 2020-03-25
Fix Flaky tests


JamesMurkin:
Current known Flakey tests: 
 - server_test.go
   - TestCancelJob
 - cluster_context_test.go
   - All tests that rely on waitForCacheSync

These are tests that occasionally fail as part of the CI pipeline and cause the build to fail. However if rerun, most will will succeed.

PR #355, opened 2020-03-27
Making cluster_context_test more reliable 

JamesMurkin:
The tests in this file occasionally fail
 - This is due to "transient" pods sometimes not being truly transient. This has been fixed by preventing the context from syncing when adding these "transient" pods, making them truly transient
 - The tests relied on the order of events between informer receiving new pod event, and then propagating it to the functions watching these events having finished before the asserts should happen

The tests no longer rely on any ordering and instead wait until the state becomes the expected one or fail if the desired state is not reached in a reasonable amount of time

### DIFF
--- a/internal/executor/context/cluster_context_test.go
+++ b/internal/executor/context/cluster_context_test.go
@@ -177,16 +177,11 @@ func TestKubernetesClusterContext_AddAnnotation(t *testing.T) {
 	err := clusterContext.AddAnnotation(pod, annotationsToAdd)
 	assert.Nil(t, err)
 
-	updateOccurred := false
-	for i := 0; i < 20; i++ {
+	updateOccurred := waitForCondition(func() bool {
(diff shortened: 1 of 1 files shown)


JamesMurkin:
#353

internal/executor/context/cluster_context_test.go
```diff
@@ -345,18 +332,24 @@ func TestKubernetesClusterContext_GetNodes(t *testing.T) {
...
 	assert.Nil(t, err)
 
-	waitForCacheSync(t, eventChannel, 1)
-	nodes, err := clusterContext.GetNodes()
-
-	assert.Nil(t, err)
-	assert.Equal(t, len(nodes), 1)
-	assert.Equal(t, nodes[0].Name, node.Name)
+	nodeFound := false
+	for i := 0; i < 20; i++ {
```
jankaspar: It might be useful to create `waitForCondition(func() bool)`.

PR #359, opened 2020-04-10
Assume helm3 and change all 'helm template' commands to 'install'

stackedsax:
It felt a little weird to require Helm v3 and then do `helm template` piped to `kubectl` -- the old v2 way of avoiding using tiller.  Doubly weird since half the instructions were to `helm install` and half were `helm template`.

So I changed everything to `helm install` on the assumption that new users will use Helm v3, like we require.

If we find that people are sending us lots of issues where we have to reply "please use helm v3, like the pre-reqs suggest," then we can always change everything back to `helm template`.

Oh, I also moved a few parameters to their respective values.yaml files.  They didn't seem terribly interesting to me as a new user and just made the commands unnecessarily long.  If anyone feels strongly that those parameters should be more visible to first time users, let me know.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -71,14 +71,16 @@ helm install redis stable/redis-ha -f docs/quickstart/redis-values.yaml
 helm install prometheus-operator stable/prometheus-operator -f docs/quickstart/server-prometheus-values.yaml
 
 # Install Armada server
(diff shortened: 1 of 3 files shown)


stackedsax:
I think CircleCI hates the slash in my branch name.  Opening a new PR for a branch with dashes and underscores, not slashes.

PR #360, opened 2020-04-13
Add image tag in values.yaml for server and executor

stackedsax:
I was surprised that the `values.yaml` file for the server and the executor didn't include an image tag by default.  It's customary to include the version of the image in the `values.yaml` so that the chart knows what image it should use.  A user can always override it on the command-line with `--set image.tag=$ARMADA_VERSION`, but I think that would be a rare situation for an end-user rather than the norm.

Regarding documentation, I would vote to take all `$ARMADA_VERSION` references out from our examples.  It might be useful to use the command-line override in a development context when you want to push out new images often, but most end-users will expect to have the image set correctly within the chart they're using.  The docs are looking pretty good already, but I think our instructions would look even cleaner if we removed this parameter.

If there's some logic to leaving the image tag out of the values file that I'm missing, please let me know.  I just haven't seen this done before in all the helm charts I've seen.

### DIFF
--- a/deployment/armada/values.yaml
+++ b/deployment/armada/values.yaml
@@ -1,5 +1,6 @@
 image:
   repository: gresearchdev/armada-server
+  tag: v0.1.2
 resources:
   limits:
     memory: 1Gi
--- a/deployment/executor/values.yaml
+++ b/deployment/executor/values.yaml
@@ -1,5 +1,6 @@
 image:
   repository: gresearchdev/armada-executor
+  tag: v0.1.2
 resources:
   limits:
     memory: 1Gi


jankaspar:
Agreed, the only tricky part of this is how to do the release. We have circleci hooked up to tagging commit to create the release, version is not known before tagging, any idea how to automate the update in this file easily?

One option would be to generate helm chart tarball as part of the release with the correct version replaced in it.

stackedsax:
Gotcha, I knew there had to be a good reason for `$ARMADA_VERSION` everywhere.

Does CircleCi also tag the docker images as part of the release?  Is that why they're currently all in sync with each other?

jankaspar:
Yes, circleci discovers release git tag and tags docker images + rebuilds and publishes the armadactl binaries for multiple architectures (https://github.com/G-Research/armada/blob/master/.circleci/config.yml#L139). If we would publish helm chart the same way in tarball (adding proper version number inside) everything would just work and user would not even have to checkout the repo, just get the released helm chart tar.

stackedsax:
Hm, yeah, I'm not so deep with CircleCI, but it looks like some people have tackled the same kind of thing... almost too many examples...  There might be even more orbs than this, but these were the ones that I could see that might be relevant:

* https://circleci.com/orbs/registry/orb/andsafe/update-helm-chart
* https://circleci.com/orbs/registry/orb/banzaicloud/helm
* https://circleci.com/orbs/registry/orb/payfit/helm-package-chart
* https://circleci.com/orbs/registry/orb/paperspace/helm-tools
* https://circleci.com/orbs/registry/orb/dialogue/helm-app
* https://circleci.com/orbs/registry/orb/nukengprod/helm
* https://circleci.com/orbs/registry/orb/bizzabo/helm
* https://circleci.com/orbs/registry/orb/pennlabs/helm-tools
* https://circleci.com/orbs/registry/orb/equisoft/chartmuseum

I'll check Armada's CircleCi config and see if I can see a way through.

jankaspar:
So many options :), I was thinking that few lines of bash could do it (maybe in makefile), something like:
```
sed -i 's/latest/$ARMADA_VERSION/g' /deployment/armada/Chart.yaml
sed -i 's/latest/$ARMADA_VERSION/g' /deployment/armada/values.yaml
helm package ./deployment/armada
```

stackedsax:
@jankaspar yep, we could do it that way, too.

I'm thinking we would need to update three values on every build:

* `Chart.yaml`
  * `appversion`
  * `version`
* `values.yaml`
  * `image:tag`

And we will need to do that for all three armada, executor, and executor-cluster-monitoring charts.

Just making sure this is what you want.  I'll mess around with CircleCI to make it happen, if so.

jankaspar:
Agreed, although I would avoid any commits from circle ci, we can have placeholder (like latest) or empty values in git and publish artefacts to download the same way as we publish armadactl.

stackedsax:
Since we're going with the 'replacing the "LATEST" keyword' approach in PR #401, I'll close this sucker.

PR #361, opened 2020-04-13
Assume helm3 and change all 'helm template' commands to 'install'

stackedsax:
It felt a little weird to require Helm v3 and then do helm template piped to kubectl -- the old v2 way of avoiding using tiller. Doubly weird since half the instructions were to helm install and half were helm template.

So I changed everything to helm install on the assumption that new users will use Helm v3, like we require.

If we find that people are sending us lots of issues where we have to reply "please use helm v3, like the pre-reqs suggest," then we can always change everything back to helm template.

Oh, I also moved a few parameters to their respective values.yaml files. They didn't seem terribly interesting to me as a new user and just made the commands unnecessarily long. If anyone feels strongly that those parameters should be more visible to first time users, let me know.

-- I reopened this PR without a slash in the branch name, sorry for any confusion --

### DIFF
--- a/docs/quickstart/executor-values.yaml
+++ b/docs/quickstart/executor-values.yaml
@@ -1,5 +1,8 @@
 applicationConfig:
   apiConnection:
+    ## Please note that this setting is insecure
+    ## Do not use this setting in a production environment
+    ## This should only be used for the quickstart and local testing
     forceNoTls: true 
   kubernetes:
     minimumPodAge: 0s
--- a/docs/quickstart/executor-values.yaml
+++ b/docs/quickstart/executor-values.yaml
@@ -1,0 +1,8 @@
+applicationConfig:
+  apiConnection:
+    forceNoTls: true 
+  kubernetes:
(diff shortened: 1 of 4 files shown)


docs/quickstart/executor-values.yaml
```diff
@@ -0,0 +1,8 @@
+applicationConfig:
+  apiConnection:
+    forceNoTls: true
```
jankaspar, line 3: This is unsafe and suitable only for local testing, maybe we should put comment with big warning here? If it is in file people are more likely to copy it....

stackedsax: People copying insecure code for their production install is a completely reasonable assumption.... I'll add some harsh comment in the yaml and see if we feel comfortable with that.

jimbobby5: It's probably OK for the quickstart

stackedsax: @jankaspar This work for you or should I make it more severe?

PR #367, opened 2020-04-22
Adding limits on how much resource can be allocated to a queue at a time 

JamesMurkin:
It is now possible to configure how much resource is allocated to a queue (as a percentage of total resource)

It can be configured:
 - At a global level, so all queues have a the same limit
 - At a queue level, so an individual queue can have its own limit

There is also the ability to configure how much resource can be allocated to a queue in a scheduling round
 - This is a workaround to reduce the error of exceeding the limit in the case of multi clusters requesting jobs at the same time


internal/armada/scheduling/lease.go
```diff
@@ -38,20 +38,39 @@ func LeaseJobs(
...
 	resourcesToSchedule := common.ComputeResources(request.Resources).AsFloat()
 	currentClusterReport, ok := activeClusterReports[request.ClusterId]
 
+	totalCapacity := &common.ComputeResources{}
+	for _, clusterReport := range activeClusterReports {
+		totalCapacity.Add(clusterReport.ClusterAvailableCapacity)
+	}
+
+	resourceAllocatedByQueue := map[string]common.ComputeResources{}
+	for _, job := range activeClusterAllocatedJobs {
```
jankaspar, line 54: Could we use `currentClusterReport.Queues` here instead of loading all leased jobs?

internal/common/resource.go
```diff
@@ -158,6 +158,12 @@ func (a ComputeResourcesFloat) LimitWith(limit ComputeResourcesFloat) ComputeRes
 	return targetComputeResource
 }
 
+func (a ComputeResourcesFloat) LimitTo0() {
```
jankaspar, line 161: There might be better name for this, maybe spelling Zero is better, or `MakePositive()`?

PR #377, opened 2020-05-22
Validating if jobs can be scheduled on submission

JamesMurkin:
When you submit a job, if it cannot be scheduled on any cluster, an error will be returned.

This is so jobs don't sit in the queue forever as they cannot be scheduled anywhere


JamesMurkin:
Should there be a config option to turn this off? To allow easier local test of the Server in isolation.

If you could turn off the validation, you could submit jobs to a server that has no executors for development purposes.

PR #378, opened 2020-05-26
Add minimum job size requirement for Executor.

jankaspar:
This allows executor to specify minimum size of jobs it accepts.


internal/executor/application.go
```diff
@@ -74,7 +74,8 @@ func StartUpWithContext(config configuration.ExecutorConfiguration, clusterConte
 		clusterContext,
 		queueClient,
 		config.Kubernetes.MinimumPodAge,
-		config.Kubernetes.FailedPodExpiry)
+		config.Kubernetes.FailedPodExpiry,
+		config.Kubernetes.MinimumJobSize)
```
JamesMurkin, line 78: Maybe we should just pass in KubernetesConfiguration here?

jankaspar: Not sure about this one, there are still few unrelated things in the Kubernetes config.

PR #398, opened 2020-06-26
Add event reporting for job resource utilisation.

jankaspar:
This PR introduces new `JobUtilisationEvent` which contains information about job CPU and memory usage.

To prevent panic caused by bug in gogo protoguf text message serialization on privete fields, protobuf code generation was changed to use generated String() methods instead.


internal/executor/service/job_utilisation_reporter.go
```diff
@@ -0,0 +1,146 @@
...
+	if !util.IsManagedPod(pod) {
+		return
+	}
+	delete(r.podInfo, pod.Name)
+}
+
+func (r *UtilisationEventReporter) reportUsage(info *podUtilisationInfo) {
+	event := reporter.CreateJobUtilisationEvent(info.pod, info.utilisationMax, r.clusterContext.GetClusterId())
+	r.queueEventWithRetry(event, 3)
+	info.utilisationMax = common.ComputeResources{}
```
JamesMurkin, line 131: Should resetting info fields be done together. Currently we set reported time in ReportUtilisationEvents and utilisationMax here.

Would be nice if they could be together, so state isn't modified as much in various places.

internal/executor/configuration/types.go
```diff
@@ -29,6 +29,8 @@ type TaskConfiguration struct {
 	PodDeletionInterval                   time.Duration
 	PodMetricsInterval                    time.Duration
 	QueueUsageDataRefreshInterval         time.Duration
+	UtilisationEventReportingInterval     time.Duration
+	UtilisationEventReportingPeriod       time.Duration
```
JamesMurkin, line 33: Not totally against this. However everything else in this struct is a task interval. This isn't a task interval.

PR #401, opened 2020-07-07
Add CircleCI config to push to GR charts repository

stackedsax:
I wanted to follow up on #360 and set up CircleCI to push new releases to https://github.com/G-Research/charts.

I did a few things here:
* Added commands for:
   * installing helm
   * checking out the charts repo
   * updating the charts and pushing the packaged charts  back to the charts repo
* Put those new commands in the release job
* Changed the `index.yaml`, `Chart.yaml` and `values.yaml` files to use 'LATEST' for version/appVersion to we can replace the value when we run the release.

I've run my steps in the local CircleCI environment and also in a test job in the proper CircleCI environment, so unless I've made some bonehead mistake moving my code into Armada's `.circleci/config`, it should all run fine.

The git clone and push commands rely on a deploy key from the charts repository.  I added the key as the `ARMADA_CHART_UPDATE_KEY` environment variable to the armada project in CircleCI so that it will run happily.  There was a problem with CircleCI eating line breaks in the key and I can't tell if I've put it in the right way or not.  If it breaks, I'll just have to change it to the version with `\n`'s.

Also, just register my feelings about doing a replace on 'LATEST': I'd really prefer to check the updated code back in somewhere.  It feels wrong that someone couldn't just grab the code for a specific version and build it immediately.  I understand the reticence to have a CI job check code in, but I'd like to continue the conversation and see if there isn't a way that feels ok for everyone.

Would love any feedback you have for this.


jankaspar:
Thanks, this looks great.

For the replace functionality, yes it unfortunate that you can't just clone the repo, we could maybe publish "latest" tag for all images so you could clone it and use latest? Other option would be to completely rework the release process. Currently it is initiated by tagging the repository, that effectively prevents you from doing any changes to the repo. You could reverse it and initiate the release from manual circle ci job, in that case I would not be oposed to commiting changes and tagging from the circleci job.

jankaspar:
Actually one more thing, could you please change quick start accordingly? I think this change might break it.

stackedsax:
> we could maybe publish "latest" tag for all images 

Nah, at least this way the charts that we publish will refer to a specific version.  It's halfway to what I'd like.  If we only ever point these charts at "latest" it will be a mess to go backwards and use an old chart, or maybe someone continues to use an old chart but starts pulling in a newer "latest" image... eh, it's not the way we want to go.

> Other option would be to completely rework the release process. 

Eventually, this is what I'd like to look at doing.  Do we have any allegiance to CircleCI?  I feel like Github Actions might be a more contained way of doing CI these days, but I don't know if we can do the whole multi-cluster setup like we need with GH Actions.  @jgiannuzzi would know.  If we decide to rework the process, I like your idea of "commiting changes and tagging from the [circleci/gh actions] job".

> Quickstart needs an update before merging this.

Totally -- my whole goal with this change was to get back to simplifying the Quickstart guide, so I'm glad you asked!  Hopefully, this allows us to make it even simpler for new users.  I'll try and get to it tonight, or over the weekend at the worst.

stackedsax:
I added a couple of commits that I believe bring the Quickstart guide up to speed with the changes in this PR.  

I ran through the steps and everything seemed to work.  The only thing I had to do beyond just copying and pasting the steps in the quickstart was `--set image.tag=v.0.1.9` when helm installing the server and the executor.  I've created a PR to add the v0.1.9 charts to the charts repo so that we can test without needing to add this command line parameter.  Once someone +1's that, the changes in this PR should be really straightforward to test.

Other thoughts:

That chunk of bash to download and install armadactl... I kinda hate to suggest it, but should we be looking at homebrew and chocolatey packages to make that solid chunk of ugliness go away?

Also, I could maybe simplify the helm install steps a little by making the redis, prometheus and executor-monitoring charts all dependencies for the server and executor charts.  However, because a lot of the parameters in the values files for the server and the executor have the same names, I can't just simply `-f docs/quickstart/*.yaml`.  The alternative is having a bunch of `-f`'s for the `helm install` command or merging some of the separate values files into just a server and executor values file.  I'll muse on it a bit more, but any input is welcome.

jankaspar:
I am all for creating packages for armadactl, we just have not prioritized it yet.

For helm, we are at a point where complete refactor of values is annoying but manageable, so if there is a need to nest all server and executor value under some key, lets do it now. Is there a way to have optional chart dependencies? Most production deployments will probably not want one chart deploying everything.

PR #405, opened 2020-07-22
Node taints support

jankaspar:
- Armada server now supports taints toleration and node selector from pods spec and considers tainted nodes in scheduling. Pod tolerating taints are still offered to any cluster with or without the specific taint applied to any node.
- Fake executor cluster can be customized through configuration. It now simulates multiple nodes and implements very very simple node pod assignment algorithm. 
- Deprecates `requiredNodeLabels` in favor of `podSpec.nodeSelector`. This property will be dropped in future versions.


internal/armada/scheduling/node_matching_test.go
```diff
@@ -0,0 +1,146 @@
...
+		{Labels: map[string]string{"armada/region": "eu", "armada/zone": "2"}},
+	}}))
+
+	assert.True(t, MatchSchedulingRequirements(job, &api.ClusterSchedulingInfoReport{NodeTypes: []*api.NodeType{
+		{Labels: map[string]string{"x": "y"}},
+		{Labels: map[string]string{"armada/region": "eu", "armada/zone": "1", "x": "y"}},
+	}}))
+}
+
+func Test_atchSchedulingRequirements_isAbleToFitOnAvailableNodes(t *testing.T) {
```
JamesMurkin, line 32: Type in name

internal/armada/scheduling/node_matching.go
```diff
@@ -0,0 +1,164 @@
...
+		resourceRequest := common.TotalResourceRequest(job.PodSpec).AsFloat()
+		nodeResources := common.ComputeResources(nodeType.AllocatableResources).AsFloat()
+		if fits(resourceRequest, nodeResources) && matchNodeSelector(job, nodeType.Labels) && tolerates(job, nodeType.Taints) {
+			return true
+		}
+	}
+	return false
+}
+
+func matchAnyNodeTypeAllocation(job *api.Job, nodeAllocations []*nodeTypeAllocation, alreadyConsumed map[*nodeTypeAllocation]common.ComputeResourcesFloat) (*nodeTypeAllocation, bool) {
```
JamesMurkin, line 58: Does this function need to modify alreadyConsumed?

PR #426, opened 2020-09-09
Implement delete queue functionality.

jankaspar:
- Deletion of queue is idempotent - when the queue does not exists call just succeeds
- On priority update all non-existent queues priorities are removed


cmd/armadactl/cmd/deleteQueue.go
```diff
@@ -0,0 +1,37 @@
...
+		apiConnectionDetails := client.ExtractCommandlineArmadaApiConnectionDetails()
+
+		client.WithConnection(apiConnectionDetails, func(conn *grpc.ClientConn) {
+			submissionClient := api.NewSubmitClient(conn)
+			e := client.DeleteQueue(submissionClient, queue)
+			if e != nil {
+				log.Error(e)
+				return
+			}
+			log.Infof("Queue %s deleted or did not existed at all.", queue)
```
JamesMurkin, line 34: Improve wording of the second half. Maybe "Queue %s deleted or did not exist."

Or just "Queue %s deleted."

JamesMurkin (approved):

PR #428, opened 2020-09-10
armada-load-tester loadtest not submitting without --watch fix

### LINKED ISSUE #419: Armada-load-tester doesn't submit jobs if you don't use --watch
**What happens**
If you run:

armada-load-tester loadtest filename.yaml

The program instantly exits

**What should happen**

It should create queues and submit jobs specified in the filename.yaml file.

**Workaround**

If you do:
armada-load-tester loadtest filename.yaml --watch

It successfully creates the queues and submits the jobs.

carlocamurri:
fix #419 

Use a nested WaitGroup to ensure level 1 goroutines wait for level 2 goroutines.

Can alternatively be done using same WaitGroup and expecting two routines to finish (`Add(2)`), as there is a 1-1 mapping between level 1 and level 2 goroutines. Thought this way was more explicit.


pkg/client/load-test.go
```diff
@@ -48,14 +48,20 @@ func (apiLoadTester ArmadaLoadTester) RunSubmissionTest(ctx context.Context, spe
 			wg.Add(1)
 			go func(i int, submission *domain.SubmissionDescription) {
 				defer wg.Done()
-				jobIdChannel, jobSetId := apiLoadTester.runSubmission(ctx, submission, i)
+
+				nestedWg := &sync.WaitGroup{}
```
jankaspar, line 52: I would name this according to what it is for i.e `submissionWg`, `jobSubmissionWg`, or something like that.

jankaspar (changes requested):
Looks alright, just small naming change.

JamesMurkin:
You could even have runSubmission function create the waitGroup and return it. Call it something like submissionComplete.

Then callers to runSubmission can use the waitGroup to wait for submission to finish.

carlocamurri:
Thanks for the comments, will proceed to fix

JamesMurkin (approved):

jankaspar (approved):

PR #435, opened 2020-10-12
Updating documentation

JamesMurkin:
There is likely still plenty to document but this covers some obviously missing features.

### DIFF
--- a/docs/helm/executor.md
+++ b/docs/helm/executor.md
@@ -148,9 +148,11 @@ It will then either have:
 applicationConfig:
   kubernetes:
     trackedNodeLabels:
-    - somevalue
+    - label1
+    - label2
     toleratedTaints:
-    - somevalue
+    - taintName1
+    - taintName2
     minimumJobSize:
       memory: 0.25
       cpu: 0.25
@@ -163,7 +165,11 @@ Armada-executor will report these labels back to armada-server to allow jobs set
 
 **toleratedTaints**
 
-This is a list of node taints that armada-executor will pay attention to, any other node taints will be ignored. 
+This is a list of node taints that armada-executor will consider usable by jobs. 
+
+By default tainted nodes will be ignored by armada-executor and armada will not try to schedule jobs on them.
+
+This is an opt-in for tainted nodes that you want armada to use and schedule jobs on.
 
 It will report these taints back to armada-server so it use them when schedule jobs onto clusters. (i.e only jobs which tolerate these taints will be scheduled onto these nodes)
 
--- a/docs/helm/executor.md
+++ b/docs/helm/executor.md
@@ -4,22 +4,43 @@ This document briefly outlines the customisation options of the Executor helm ch
 
 ## Values
 
-| Parameter                         | Description                                                                                                                                                                      | Default                                                                          |
-|-----------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------|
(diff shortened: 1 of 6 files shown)


docs/helm/server.md
```diff
@@ -134,3 +134,131 @@ permissionScopeMapping:
...
+  queueLeaseBatchSize: 200
+  minimumResourceToSchedule:
+    memory: 100000000 # 100Mb
+    cpu: 0.25
+  maximalClusterFractionToSchedule:
+    memory: 0.25
+    cpu: 0.25
+```
+
+`queueLeaseBatchSize` This is the number of jobs at the top of a queue the scheduler will look at when trying to find one that'll fit in the available resource.
```
jankaspar, line 192: I think this is just size of batch, if whole batch is scheduled, we will load additional `queueLeaseBatchSize` amount until we run out of time or global scheduling limit or the queue is empty.

docs/helm/server.md
```diff
@@ -134,3 +134,131 @@ permissionScopeMapping:
...
+```yaml
+scheduling:
+  maximalResourceFractionToSchedulePerQueue:
+    memory: 0.05
+    cpu: 0.05
+  maximalResourceFractionPerQueue:
+    memory: 0.25
+    cpu: 0.25
+```
+
```
jankaspar, line 226:
```suggestion
All limits are proportional to overall amount of resources in the system. In this example, the queue can use at most 5% of all available cpu and memory.
```

jimbobby5 (approved):

docs/helm/server.md
```diff
@@ -134,3 +134,131 @@ permissionScopeMapping:
...
+`maximalResourceFractionPerQueue` Is the maximum resource a queue can hold as a percentage of all resource over all clusters.
+
+`maximalResourceFractionToSchedulePerQueue` Is the percentage of total resource a queue can be allocated in a single scheduling round.
+
+Currently scheduling is done in parallel, so it can happen that we exceed the resource limit set in `maximalResourceFractionPerQueue` due to two clusters requesting new jobs at the exact same time.
+
+To migate this, `maximalResourceFractionToSchedulePerQueue` specifies how much can be scheduled in a single round and can be thought of as the margin for error.
```
jankaspar, line 233:
```suggestion
To mitigate this, `maximalResourceFractionToSchedulePerQueue` specifies how much can be scheduled in a single round and can be thought of as the margin for error.
```

docs/helm/executor.md
```diff
@@ -75,26 +100,113 @@ If the username/password are not present in the server components secret, commun
...
+
+**stuckPodExpiry**
+
+This is how long the executor will let a pod will sit in `Pending` state before it considers the Job stuck.
+
+Once a Job is considered stuck a JobUnableToScheduleEvent will be reported for the job. 
+
+It will then either have:
+ - If the problem is deemed unretryable (for example the image is getting `InvalidImageName`) the job will get a JobFailedEvent and be considered Done
+ - If the probelm is deemed retryable, the job will Have its lease returned to armada-server (JobLeaseReturnedEvent) and the job will be rescheduled
```
carlocamurri, line 145:
```suggestion
 - If the problem is deemed retryable, the job will have its lease returned to armada-server (JobLeaseReturnedEvent) and the job will be rescheduled 
```

jankaspar (dismissed):

docs/helm/server.md
jankaspar: Never mind the effect is actually what is described.

jankaspar (approved):

PR #439, opened 2020-10-27
Limit maximum number of retries for stuck jobs

carlocamurri:
Currently, if a retryable stuck job is found, the executor will schedule the job again.

There have been situations in which pods identified as retryable are rescheduled and get stuck again. This cycle continues indefinitely.

These changes limit the number of times a job get retried to a maximum of five times per Armada executor, as job retries are persisted in an in-memory cache.

### DIFF
--- a/internal/executor/service/job_lease.go
+++ b/internal/executor/service/job_lease.go
@@ -29,6 +29,7 @@ type LeaseService interface {
 type JobLeaseService struct {
 	clusterContext  context2.ClusterContext
 	queueClient     api.AggregatedQueueClient
+	retryCache      RetryCache
 	minimumPodAge   time.Duration
 	failedPodExpiry time.Duration
 	minimumJobSize  common.ComputeResources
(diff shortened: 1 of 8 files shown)


internal/executor/service/retry_cache.go
```diff
@@ -0,0 +1,39 @@
...
+	GetNumberOfRetryAttempts(jobId string) int
+	Evict(jobId string)
+}
+
+type InMemoryRetryCache struct {
+	mutex *sync.RWMutex
+	cache map[string]int
+}
+
+func NewInMemoryRetryCache() RetryCache {
```
ingvarm-gr, line 16: It would be good if this (and the other) exported function(s) could have a comment explaining what the function actually does.

internal/executor/service/job_lease.go
```diff
@@ -237,3 +242,11 @@ func chunkPods(pods []*v1.Pod, size int) [][]*v1.Pod {
 	}
 	return chunks
 }
+
+func evictRetriedJobs(retryCache RetryCache, jobIds []string) {
+	for _, jobId := range jobIds {
+		if retryCache.GetNumberOfRetryAttempts(jobId) > 0 {
+			retryCache.Evict(jobId)
```
ingvarm-gr, line 249: Since `delete(map, key)` is safe no matter what, do we actually need this check here?

carlocamurri: The only reason I put the check there is because `GetNumberOfRetryAttempts` only requires a `RLock`, whilst `Evict` uses a  write `Lock` (i.e. to avoid having unnecessary writer locks). Do you think this is unnecessary?

PR #441, opened 2020-10-28
Limit job retries

carlocamurri:
Currently, if a retryable stuck job is found, the executor will schedule the job again.

There have been situations in which pods identified as retryable are rescheduled and get stuck again. This cycle continues indefinitely.

These changes ensure that job retries are recorded in Redis, and jobs are not retried if a certain number (parametrised in SchedulingConfig) is exceeded.
The Redis key pattern used is `Job:Retries:{jobId}`.


internal/armada/server/lease.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/G-Research/armada/pkg/api"
 )
 
+const maxRetries = 5
```
carlocamurri, line 19: Should this be part of the scheduling config?

internal/armada/repository/job.go
```diff
@@ -526,6 +541,33 @@ func (repo *RedisJobRepository) ExpireLeases(queue string, deadline time.Time) (
...
+func (repo *RedisJobRepository) AddRetryAttempt(jobId string) error {
+	exists, err := repo.db.Exists(jobRetriesPrefix + jobId).Result()
+	if err != nil {
+		return err
+	}
+	if exists == 0 {
+		return fmt.Errorf("RedisJobRepository.AddRetryAttempt: job with id %q does not exist", jobId)
+	}
+
+	_, err = repo.db.Incr(jobRetriesPrefix + jobId).Result()
```
jankaspar, line 553: Increment creates the key if it does not exist, so you don't need the logic above.

internal/armada/repository/job.go
```diff
@@ -144,6 +148,7 @@ func (repo *RedisJobRepository) AddJobs(jobs []*api.Job) ([]*SubmitJobResult, er
 
 		submitResult.saveJobResult = pipe.Set(jobObjectPrefix+job.Id, jobData, 0)
 		submitResult.jobSetIndexResult = pipe.SAdd(jobSetPrefix+job.JobSetId, job.Id)
+		submitResult.initJobRetriesResult = pipe.Set(jobRetriesPrefix+job.Id, 0, 0)
```
jankaspar, line 151: You might not need this, just assume there were no retries when the key is missing.

internal/armada/server/lease.go
jankaspar: yes

jankaspar (changes requested):
Looks alright, just few small details.

internal/armada/server/lease.go
```diff
@@ -134,10 +135,39 @@ func (q *AggregatedQueueServer) ReturnLease(ctx context.Context, request *api.Re
...
+
+	maxRetries := int(q.schedulingConfig.MaxRetries)
+	if retries >= maxRetries {
+		_, err := q.ReportDone(ctx, &api.IdList{Ids: []string{request.JobId}})
+		if err != nil {
+			return nil, err
+		}
+
+		failureReason := fmt.Sprintf("Exceeded maximum number of retries: %d", maxRetries)
+		err = q.reportFailure(request.JobId, request.ClusterId, failureReason)
```
jankaspar, line 153: I think it would be better to do this before call to ReportDone. That way there is no chance of deleting job info without creating the event.

internal/armada/server/lease.go
```diff
@@ -162,3 +192,29 @@ func (q *AggregatedQueueServer) ReportDone(ctx context.Context, idList *api.IdLi
...
+	failureReason := reason
+	err = reportFailed(q.eventStore, clusterId, failureReason, job)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getJobById(jobRepository repository.JobRepository, jobId string) (*api.Job, error) {
```
jankaspar, line 211: maybe make this a method on AggregatedQueueServer or extend the repository to do that.

jankaspar (changes requested):

internal/armada/server/lease.go
carlocamurri: I wouldn't think it worth to extend the JobRepository interface with that method, let me know if you wish me to

internal/executor/service/stuck_pod_detector_test.go
```diff
@@ -0,0 +1,261 @@
+package service
+
+import (
+	"testing"
+	"time"
+
+	"github.com/G-Research/armada/internal/common"
```
ingvarm-gr, line 7: This import block and the next one shoudl swap places ("standatd library", "third-party", "internal").

jankaspar (approved):

PR #452, opened 2020-11-10
Add Lookout endpoint to query jobs in a given queue

carlocamurri:
Adds endpoint to Lookout API. Returns job information for jobs in a given queue. This includes all runs (i.e. pods created) for a given job.

Supports:
- Filtering by queue (exact)
- Filtering by multiple job states (Queued, Pending, Running, Succeeded, Failed, Cancelled)
- Filtering by multiple job sets (begins with)
- Ordering by job creation time
- Skip/take for pagination


internal/lookout/repository/query.go
```diff
@@ -50,3 +57,317 @@ func (r *SQLJobRepository) GetQueueStats() ([]*lookout.QueueInfo, error) {
...
+			&row.Cluster,
+			&row.Node,
+			&row.Created,
+			&row.Started,
+			&row.Finished,
+			&row.Succeeded,
+			&row.Error,
+		)
+		if err != nil {
+			return nil, err
```
ingvarm-gr, line 118: Idle question, is it better to return what you have gathered and the error, or nil and the error?

internal/lookout/repository/query.go
```diff
@@ -50,3 +57,317 @@ func (r *SQLJobRepository) GetQueueStats() ([]*lookout.QueueInfo, error) {
...
+			   job_run.run_id as run_id,
+			   job_run.cluster as cluster,
+			   job_run.node as node,
+			   job_run.created as created,
+			   job_run.started as started,
+			   job_run.finished as finished,
+			   job_run.succeeded as succeeded,
+			   job_run.error as error
+		FROM job LEFT JOIN job_run ON job.job_id = job_run.job_id
+		WHERE job.job_id IN (%s)`
```
ingvarm-gr, line 143: Does this open us up to SQL injection? What happens if I request a job ID of `3); DROP TABLE job; --`?

internal/lookout/repository/query.go
```diff
@@ -50,3 +58,285 @@ func (r *SQLJobRepository) GetQueueStats() ([]*lookout.QueueInfo, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	rows, err := r.db.Query(query, args...)
+	if err != nil {
+		return nil, err
+	}
+
+	joinedRows := make([]*jobsInQueueRow, 0)
```
jankaspar, line 102: When we are using goqu now, maybe we should utilise the scanning functionality as well (https://github.com/doug-martin/goqu/blob/master/docs/selecting.md#scan-structs) ?

internal/lookout/repository/query.go
```diff
@@ -50,3 +58,285 @@ func (r *SQLJobRepository) GetQueueStats() ([]*lookout.QueueInfo, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		joinedRows = append(joinedRows, row)
+	}
+	return joinedRows, nil
+}
+
+func makeGetJobsInQueueQuery(opts *lookout.GetJobsInQueueRequest) (string, []interface{}, error) {
+	dialect := goqu.Dialect("postgres")
```
jankaspar, line 131: Can you get dialect from db instance? Or create it in constructor.

internal/lookout/repository/query.go
```diff
@@ -50,3 +58,285 @@ func (r *SQLJobRepository) GetQueueStats() ([]*lookout.QueueInfo, error) {
...
+
+	subDs := dialect.
```
jankaspar, line 133: what `subD` stands for?

internal/lookout/repository/query.go
```diff
@@ -50,3 +58,285 @@ func (r *SQLJobRepository) GetQueueStats() ([]*lookout.QueueInfo, error) {
...
+		LeftJoin(goqu.T("job_run"), goqu.On(goqu.Ex{
+			"job.job_id": goqu.I("job_run.job_id"),
+		})).
+		Select(goqu.I("job.job_id"))
+
+	subDs = addWhere(subDs, opts.Queue, opts.JobSetIds)
+
+	subDs = subDs.GroupBy(goqu.I("job.job_id"))
+
+	subDs = addHaving(subDs, opts.JobStates)
```
jankaspar, line 144: These `addHaving` and `addWhere` are quite hard to follow. Could you change it to have method like `CreateJobStateFilter(opts.JobSetIds)` and call it here like `ds.Having(CreateJobStateFilter(opts.JobStates))`. That way it will be much easier to see how the query actually looks like just from this method.

internal/lookout/repository/query.go
```diff
@@ -50,3 +58,285 @@ func (r *SQLJobRepository) GetQueueStats() ([]*lookout.QueueInfo, error) {
...
+				Created:   parseNullTime(row.Created), // Pod created (Pending)
+				Started:   parseNullTime(row.Started), // Pod running
+				Finished:  parseNullTime(row.Finished),
+			})
+		}
+	}
+	return result
+}
+
+func parseNullString(nullString sql.NullString) string {
```
jankaspar, line 309: It might worth moving these methods to separate file (maybe `util.go`?).

internal/lookout/repository/query.go
carlocamurri: `ds` stands for dataset, from goqu's `SelectDataset`

internal/lookout/repository/query.go
carlocamurri: Can create goqu db in constructor and use it for scanning as well

internal/lookout/repository/store.go
```diff
@@ -40,9 +40,10 @@ func (r *SQLJobStore) RecordJob(job *api.Job) error {
 }
 
 func (r *SQLJobStore) MarkCancelled(event *api.JobCancelledEvent) error {
-	_, err := upsert(r.db, "job",
-		"job_id", []string{"queue", "jobset", "cancelled"},
-		[]interface{}{event.JobId, event.Queue, event.JobSetId, event.Created})
+	_, err := r.db.Exec(`
```
jankaspar, line 43: Is it intentional to drop the upsert operation here? Theoretically the events might be processed out of order and cancellation can be seen before the job is created - that's why the upsert was used.

jankaspar (approved):

PR #458, opened 2020-11-13
Fix problem on ci with slashes in branch name.

jankaspar:


JamesMurkin (approved):

JamesMurkin (approved):

stackedsax:
oh, right, it's slashes, not underscores that are the problem.  I keep forgetting!  Must change my habits! 

Thanks for fixing it.

PR #459, opened 2020-11-18
Add temporary fix for python sspi with kerberos auth.

jankaspar:
Python request-negotiate-sspi module is failing when used with armada api when passing 
"toke accepted" response header to windows, until there is a fix available in the module or other workaround found, removing the header fixes the issue.


ingvarm-gr (approved):
I guess the "why" is actually in the comment, but it might be worth putting a fuller description in the initial comment on the PR.

jimbobby5 (approved):

PR #461, opened 2020-11-20
Initial job table in Lookout UI

carlocamurri:
Infinite list of jobs in specified queue, sortable by submission date.

image


internal/lookout/ui/src/utils.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  end = start + nValuesToAdd
+  if (end >= array.length) {
+    const padding = new Array(end - array.length).fill(null)
+    array.push(...padding)
+  }
+  array.splice(start, nValuesToAdd, ...newValues)
+}
+
+export function getOrDefault<T>(value: T | undefined | null, defaultValue: T): T {
```
jankaspar, line 15: I think, there is `??` operator now in typescript (and js as well).

internal/lookout/ui/make-test-jobs.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
jankaspar, line 1: It might better to just use load tester we have instead of this script. (https://github.com/G-Research/armada/tree/master/cmd/armada-load-tester, input https://github.com/G-Research/armada/blob/master/example/demo.yaml). We definitely lack documentation in this place, sorry.

jankaspar (approved):

PR #463, opened 2020-11-24
Use strings instead of enum for job state in lookout API

carlocamurri:
The current grpc-gateway API does not accept the string representation of job states enum in requests. This is an issue as the generated clients rely on the API accepting strings.

A few approaches were attempted by adding gogoproto options/other marshallers (e.g. https://jbrandhorst.com/post/gogoproto/, https://github.com/cockroachdb/cockroach/blob/master/pkg/util/protoutil/jsonpb_marshal.go), though these break previous compatibility with K8s protobuf messages.

These changes replace the enum with raw string values.


internal/common/grpc/gateway.go
```diff
@@ -8,12 +8,10 @@ import (
...
 
 	"github.com/go-openapi/runtime/middleware"
-	gwruntime "github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/jcmturner/gokrb5/v8/spnego"
 	"google.golang.org/grpc"
 
-	"github.com/grpc-ecosystem/grpc-gateway/runtime"
-
 	protoutil "github.com/G-Research/armada/internal/armada/protoutils"
```
ingvarm-gr, line 15: Not for this PR, but for some nebulous list of minor cleanups, is it really that useful to have `protoutil` instead of `protoutils`?

ingvarm-gr (approved):

internal/common/grpc/gateway.go
carlocamurri: Should probably be fixed too

PR #465, opened 2020-11-26
Add job set and job state filtering to Lookout UI

carlocamurri:
image

image

### DIFF
--- a/internal/lookout/ui/src/components/JobTableHeader.tsx
+++ b/internal/lookout/ui/src/components/JobTableHeader.tsx
@@ -81,7 +81,7 @@ export default function JobTableHeader(props: JobTableHeaderProps) {
             value={props.jobStates}
             onChange={(event) => {
               const newJobStates = (event.target.value as string[])
-                .filter(jobState => VALID_JOB_STATE_VIEW_MODELS.includes(jobState as JobStateViewModel))
+                .filter(jobState => (VALID_JOB_STATE_VIEW_MODELS as string[]).includes(jobState))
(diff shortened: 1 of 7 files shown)


internal/lookout/ui/src/components/JobTableHeader.tsx
```diff
@@ -28,10 +63,45 @@ export default function JobTableHeader(props: JobTableHeaderProps) {
...
+        </div>
+        <div className="default-horizontal-margin">
+          <InputLabel id="job-table-state-select-label">Job states</InputLabel>
+          <Select
+            labelId="job-table-state-select-label"
+            id="job-table-state-select"
+            multiple
+            value={props.jobStates}
+            onChange={(event) => {
+              const newJobStates = (event.target.value as string[])
```
jankaspar, line 83: We do lot of casting here, can it be simplified?

internal/lookout/ui/src/services/JobService.ts
```diff
@@ -1,56 +1,46 @@
...
-
-  return {
-    jobId: jobId,
-    queue: queue,
-    owner: owner,
-    jobSet: jobSet,
-    submissionTime: submissionTime,
-    jobState: jobState,
-  }
+export enum JobStateViewModel {
```
jankaspar, line 13: Could you potentially combine the 2 enums? Maybe using union of string literals as a type would make it work better (I haven't tried).

jankaspar (dismissed):

jankaspar (approved):

PR #468, opened 2020-12-02
Add Armada logo and favicon

carlocamurri:
image


internal/lookout/ui/public/logo.svg
```diff
@@ -0,0 +1,75 @@
...
+     inkscape:cx="175"
+     inkscape:cy="300"
+     inkscape:window-x="0"
+     inkscape:window-y="0"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="Layer_1" />
+  <defs
+     id="defs3851">
+    <style
+       id="style3849">.cls-1{fill:#00aae1;}.cls-2{fill:#383838;}</style>
```
jankaspar, line 52: I think if we want to keep the blue banner (Maybe we could also switch the blue selected for logo text `#00aae1`), we should go for white logo variant. The gray & black does not match it well.

carlocamurri:
Update:
image

jankaspar (approved):

PR #473, opened 2020-12-07
Batch jobs and fix fetch sync issues

carlocamurri:
- Batching of jobs in the job table (set to 100)
- Fix filtering sync issues when selecting new filter when already loading data from API
- Fix job table shaking on resize


jankaspar (approved):
Looks alright!

PR #474, opened 2020-12-08
Setup indexes for Lookout database and better monitoring.

jankaspar:


internal/lookout/application.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/doug-martin/goqu/v9"
```
ingvarm-gr, line 9: Should this be `goqu "github.com/doug-martin/goqu/v9"`? For clarity, if nothing else (as-is, I would expect that import to give me a module prefix of `v9`).

internal/lookout/repository/store.go
```diff
@@ -1,10 +1,10 @@
 package repository
 
 import (
-	"database/sql"
 	"encoding/json"
 	"fmt"
 
+	"github.com/doug-martin/goqu/v9"
```
ingvarm-gr, line 7: Same comment applies here.

ingvarm-gr (dismissed):

internal/lookout/application.go
jankaspar: Apparently `/v{n}` is special and it actually defaults to  `goqu "github.com/doug-martin/goqu/v9"`.

ingvarm-gr (approved):

PR #480, opened 2020-12-23
Implement support for multi node jobs

jankaspar:
This change introduce multi node jobs. Job can now have multiple `podSpecs` which will be used to create multiple pods in the cluster at once.
- Any event concerning running pod now have `podNumber` which identifies corresponding `podSpec` in job definition.
- Armadactl and Lookout determines job status based on events from all pods (job is finished only when all its pods are). 
- `podSpec` property of job is deprecated in favor of `podSpecs`, but will be supported for some time.


internal/executor/service/cluster_allocation.go
```diff
@@ -141,28 +150,38 @@ func (allocationService *ClusterAllocationService) returnLease(pod *v1.Pod, reas
 	}
 }
 
-func createPod(job *api.Job) *v1.Pod {
+func (allocationService *ClusterAllocationService) remove(pods []*v1.Pod) {
```
samclark, line 153: is this function used?

internal/lookout/repository/sql_repository.go
```diff
@@ -70,6 +73,7 @@ type JobRow struct {
 	Cancelled pq.NullTime     `db:"cancelled"`
 	JobJson   sql.NullString  `db:"job"`
 	RunId     sql.NullString  `db:"run_id"`
+	PodNumber sql.NullInt64   `db:"pod_number"`
```
samclark, line 76: could be `sql.NullInt32`

internal/executor/service/stuck_pod_detector.go
```diff
@@ -63,104 +67,119 @@ func (d *StuckPodDetector) onStuckPodDetected(pod *v1.Pod) (err error, retryable
...
 		// Reporting failed even can fail with unfortunate timing of executor restarts, in that case lease will expire and job can be retried
 		// This is preferred over returning Failed event early as user could retry based on failed even but the job could be running
-		event := reporter.CreateJobFailedEvent(record.pod, record.message, map[string]int32{}, d.clusterContext.GetClusterId())
-		err := d.eventReporter.Report(event)
-		if err != nil {
-			return false
+		for _, pod := range record.job.Pods {
+			message := record.message
+			if pod != record.pod {
+				message = "Peer pod stuck."
```
samclark, line 94: what happens if multiple pods in same job all get stuck at the same time?

samclark (dismissed):

internal/lookout/repository/sql_repository.go
jankaspar: `NullInt32` was introduced in golang 1.13, will update as soon as we start using newer go.

internal/executor/service/stuck_pod_detector.go
jankaspar: Actually there is always one running job record, effectively first pod to go wrong in the job is selected (https://github.com/G-Research/armada/pull/480/files/0cf26b9ebfe87e9b12b73392b80b8c4c8cbda6f7#diff-c2b7834011d307455bdc4ffc73d04123214ccd7afb6c7a9c3cf74725082c8f62R121).

samclark (approved):

PR #481, opened 2020-12-23
Reorganize file structure and change page template

ericjohnsohnisc:
Found that Jekyll wanted pages in the root directory rather than _pages.


stackedsax (approved):
Looks good to me.

PR #482, opened 2020-12-23
Fixing Github workflow to work with Jekyll

ericjohnsohnisc:
Updated Github workflow to work with Jekyll and not error out when there is nothing to update.


stackedsax (approved):
lgtm 👍

PR #489, opened 2021-01-14
Lookout API Job sets endpoint

carlocamurri:
Add API endpoint to count number of jobs in different states for each job set in a given queue.


devlounge (approved):

jankaspar (approved):
Looks alright. Maybe we could reuse some of the logic between counting jobs for queues and jobsets.

PR #501, opened 2021-01-26
Lookout UI Job Details

carlocamurri:
Modal to display Job details given a Job id.

image

If a Job is scheduled multiple times (multiple runs), the relevant information is displayed below

image

Indivudual Job Ids can be queried from the Job table.


ingvarm-gr (dismissed):

jankaspar:
Just a suggestion, maybe it would be more convenient to have job detail open as popup (or slide from side) on job listing. You might want to inspect multiple jobs without loosing context of your search.

jankaspar (approved):

PR #508, opened 2021-02-05
Lookout API make time-based tests deterministic

carlocamurri:
Some tests could fail due to off by one errors in rounding to the nearest millisecond. To fix this, we round to the nearest second and use the same timestamp for every test run.

### DIFF
--- a/internal/lookout/repository/queues_test.go
+++ b/internal/lookout/repository/queues_test.go
@@ -134,12 +134,12 @@ func TestGetQueueInfos_IncludeOldestQueuedJob(t *testing.T) {
 
 		NewJobSimulator(t, jobStore).
 			CreateJobAtTime(queue, someTime).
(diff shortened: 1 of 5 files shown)


internal/lookout/repository/queues_test.go
```diff
@@ -72,39 +72,37 @@ func TestGetQueueInfos_OldestQueuedJobIsNilIfNoJobsAreQueued(t *testing.T) {
 		jobStore := NewSQLJobStore(db)
 		jobRepo := NewSQLJobRepository(db, &DefaultClock{})
 
-		startTime := time.Now()
+		NewJobSimulator(t, jobStore).
+			CreateJobAtTime(queue, *Increment(someTime, 5)).
```
jankaspar, line 76: Maybe `someTime.Add(5 * time.Second)` would make it make even more obvious.

jankaspar (approved):

jankaspar (approved):

PR #509, opened 2021-02-08
Fix deadlock in fake executor.

jankaspar:


carlocamurri (approved):

ingvarm-gr (commented):
I would (unless the body of the for loop shoudl modify state ouside itself) consider breaking the for loop body out into a helper function and start that function with "acquire the lock, defer an unlock". If there's a need to reference lexical environment, an anonymous "func() { ...}()" can be used.

jankaspar:
I will refactor this in separate PR.

PR #517, opened 2021-02-18
Add support for separate resource pools with independent queue priorities.

jankaspar:
- Each executer can specify resource pool it belongs to in the configuration
- Priorities are newly tracked per pool identifiers, in the beginning of scheduling priorities corresponding to the pool are loaded.
- Metrics are also produced broken down per pool. For queued resources metrics we count the job under each pool it could run in based on its requirements.


internal/armada/scheduling/clusters.go
```diff
@@ -42,6 +65,19 @@ func FilterActiveClusterSchedulingInfoReports(reports map[string]*api.ClusterSch
 	return result
 }
 
+func GroupSchedulingInfoByPool(reports map[string]*api.ClusterSchedulingInfoReport) map[string]map[string]*api.ClusterSchedulingInfoReport {
+	result := map[string]map[string]*api.ClusterSchedulingInfoReport{}
+	for id, report := range reports {
+		poolReports := result[report.Pool]
```
ingvarm-gr, line 71: If I am reading this loop body correctly, could it be written as:
```
  poolReports, ok := result[report.Pool]
  if !ok { 
    poolReports = map[string]*api.ClusterSchedulingInfoReport{}
    result[report.Pool] = poolReports
  }
  poolReports[id] + report
```

Not suggesting that it gets rewritten that way, primarily checking my understanding of the code.

ingvarm-gr (approved):

internal/armada/scheduling/clusters.go
jankaspar: yes, your version is equivalent and more idiomatic, will update it.

JamesMurkin (dismissed):

JamesMurkin (approved):

PR #518, opened 2021-02-19
Lookout API record Job state in database

carlocamurri:
Currently, each Job event received results in multiple fields being updated in the database (e.g. event firing time). A Job's state is computed at query time. This may result in some Jobs being recognised as having the wrong state, and/or slower database queries.

The proposed solution is to add Job state as a field in the `job` table, and update it accordingly for individual Jobs as events are received. As a result, most queries are simplified. 
Note that multiple updates may be performed for some events.


internal/lookout/repository/schema/007_job_states.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE job ADD COLUMN state smallint NULL;
```
jankaspar, line 1: It might be worth to set correct state for currently existing jobs as part of the migration, unless it is supper hard. You could then make it not null.

JamesMurkin (approved):

jankaspar (approved):

PR #527, opened 2021-03-04
Add job queue cache tracking where individual jobs can be scheduled.

jankaspar:
We already scan through queued jobs to provide metrics. This PR adds tracking of which job could not be scheduled on specific cluster because of missing labels or type of resources. This info is then used while leasing jobs for particular cluster to skip jobs which are not relevant.


internal/armada/cache/queue.go
```diff
@@ -0,0 +1,155 @@
...
+		queuedResources:          map[string]map[string]common.ComputeResourcesFloat{},
+		queueNonMatchingJobIds:   map[string]map[string]stringSet{}}
+
+	return collector
+}
+
+func (c *QueueCache) Refresh() {
+	queues, e := c.queueRepository.GetAllQueues()
+	if e != nil {
+		log.Errorf("Error while getting queue metrics %s", e)
```
JamesMurkin, line 45: Incorrect error message?

internal/armada/cache/queue.go
```diff
@@ -0,0 +1,155 @@
...
+				}
+			}
+			nonMatchingJobs[job.Id] = nonMatchingClusters
+		})
+
+		if err != nil {
+			log.Errorf("Error while getting queue %s resources %s", queue.Name, err)
+		}
+
+		c.updateQueuedNonMatchingJobs(queue.Name, nonMatchingJobs)
```
JamesMurkin, line 92: Maybe better to update these 2 in the same mutex?

jankaspar: This way the locking is cleaner in update methods, the updates are not frequent enough to justify making it more complicated.

PR #530, opened 2021-03-12
Prevent job double submit using job client id.

jankaspar:
Users can set client id for every job, this id needs to be unique and is used to detect double submit. When job is submitted twice second operation just returns existing job id. Supplied clients (c#, go) automatically creates unique clients ids.


ingvarm-gr (approved):

internal/armada/repository/job_test.go
```diff
@@ -14,6 +14,14 @@ import (
...
 )
 
+func TestJobDoubleSubmit(t *testing.T) {
+	withRepository(func(r *RedisJobRepository) {
+		job1 := addTestJobWithClientId(t, r, "queue1", "my-job-1")
+		job2 := addTestJobWithClientId(t, r, "queue1", "my-job-1")
+		assert.Equal(t, job1.Id, job2.Id)
+	})
+}
+
```
ingvarm-gr, line 24: We don't seem to have a test for "two separate queues can have the same job ID"?

ingvarm-gr (approved):

JamesMurkin (approved):

PR #546, opened 2021-04-12
Add Lookout to quickstart

carlocamurri:


jankaspar (approved):
Nice, it might be worth adding screenshot of Lookout next to the Grafana one.

jankaspar (approved):

PR #549, opened 2021-04-14
Lookout API add support for Job annotations

carlocamurri:
Adds a new table to Lookout, named `annotation`, to store Job annotations as key/value pairs.

Only annotations that are prefixed with a specified configurable prefix will be added to the table.

Jobs can be looked up by specifying annotation key-value pairs.

Note that the annotations are saved without preserving the prefix. Searches are to be performed without specifying the annotation key prefix.


config/lookout/config.yaml
```diff
@@ -4,6 +4,7 @@ metricsPort: 9009
 
 uiConfig:
   armadaApiBaseUrl: "http://localhost:8080"
+  annotationPrefix: "armada/"
```
jankaspar, line 7: lets go for `armadaproject.io/` to follow the domain pattern other things use with annotations.

internal/lookout/repository/store.go
```diff
@@ -257,6 +261,23 @@ func (r *SQLJobStore) upsertContainers(k8sId string, exitCodes map[string]int32)
...
+				"key":    trimmedKey,
+				"value":  value,
+			})
+		}
+	}
+
+	return upsert(r.db, annotationTable, []string{"job_id", "key"}, annotationRecords)
+}
+
 func (r *SQLJobStore) determineJobState() exp.CaseExpression {
```
robertdavidsmith, line 281: It's a bit confusing having a data store that filters and trims values. Ideally datastores wouldn't change the data. Could this logic be moved elsewhere? What is the cost of just ditching this logic and storing all annotations including prefix? @jankaspar may have views.

robertdavidsmith (dismissed):
Approving, but suggest @jankaspar design reviews this area once back

internal/lookout/repository/store.go
jankaspar: All the annotations are saved in the job table already in json format, this table is intended for lookup by user specified key value pairs. For that saving just prefixed annotations is much easier. We want to for example create a dropdown with all keys in ui. Maybe the name of the table could be improved to reflect this, any suggestions?

jankaspar: Maybe `user_annotations`?

robertdavidsmith: user_annotations?
user_annotations_lookup?
user_annotations_index?

jankaspar: Lets go with `user_annotations_lookup` and rename the methods to match the idea.

jankaspar (approved):

PR #558, opened 2021-04-27
Add queue + running job metrics to server

JamesMurkin:
Adding metrics for:
 - Queued job time in queue (min/max/median)
 - Queue job resource request per resource type (min/max/median)
 - Running job run time (min/max/median)
 - Running job resource request per resource type (min/max/median)

To get running job start times, the approach is to watch the event stream and mark in redis when the job has started
 - This should be reliable enough, we could have the executor actually report running job ids + start times (+ others). However current thinking is we want to move to approach where we watch the event stream more so this is the start of that. We can always change our mind later.

The aim of these metrics to get a good idea if jobs are:
 - Sitting in queue forever or running for a very long time (or forever) or if they are regularly ending/new ones starting
   - Allowing you to see queues which have very long jobs to talk 
 - Identify which users are making odd resource requests (extremely large) and causing bottlenecks

### DIFF
--- a/internal/armada/metrics/metrics.go
+++ b/internal/armada/metrics/metrics.go
@@ -65,6 +65,34 @@ var queueResourcesDesc = prometheus.NewDesc(
 	nil,
 )
 
+var minQueueResourcesDesc = prometheus.NewDesc(
+	MetricPrefix+"queue_resource_queued_min",
+	"Min resource required by queued job",
+	[]string{"pool", "queueName", "resourceType"},
+	nil,
+)
+
+var maxQueueResourcesDesc = prometheus.NewDesc(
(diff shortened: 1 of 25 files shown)


internal/armada/metrics/metrics.go
```diff
@@ -61,13 +64,118 @@ var queueResourcesDesc = prometheus.NewDesc(
 	nil,
 )
 
+var minQueueResourcesDesc = prometheus.NewDesc(
```
carlocamurri, line 67: Wondering, would it make sense to use a Prometheus histogram rather than multiple metrics for each interested quantile?

carlocamurri (approved):

PR #560, opened 2021-04-30
Lookout UI make logo link to Overview

carlocamurri:


jankaspar (approved):

jankaspar:
Actually, why not make it an link? that way middle click will open new tab.

jankaspar (changes requested):
Making it a link would be better.

jankaspar (approved):

PR #562, opened 2021-05-04
Allow jobs to expose nodeport services

JamesMurkin:
Jobs can now configure that container ports are exposed via http
- Which cases Armada to create a nodeport service for that pod, exposing the configured ports

The job must configure the ports on the containers and then any it wants to expose via nodePort need to also be specified in the ingress section of the job.

See example:
```yaml
queue: test
jobSetId: service
jobs:
  - priority: 0
    ingress:
      - type: NodePort
        ports:
          - 5050
    podSpec:
      restartPolicy: Never
      containers:
        - name: sleep
          imagePullPolicy: IfNotPresent
          image: vad1mo/hello-world-rest:latest
          resources:
            limits:
              memory: 200Mi
              cpu: 0.5
            requests:
              memory: 200Mi
              cpu: 0.5
          ports:
            - containerPort: 5050
              protocol: TCP
              name: http
```


internal/executor/reporter/job_event_reporter.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/G-Research/armada/internal/common"
 	clusterContext "github.com/G-Research/armada/internal/executor/context"
+	domain2 "github.com/G-Research/armada/internal/executor/domain"
```
jankaspar, line 13: executorDomain?

jankaspar (approved):

PR #571, opened 2021-05-18
Add job owner as pod annotation.

jankaspar:


JamesMurkin (approved):

devlounge (changes requested):
please provide us with some example rendered pod annotations in the PR description

jankaspar:
All annotation keys in form `armada_{something}`, we should consider migration to `armadaproject.io\{something}`, but not as part of this PR.

PR #576, opened 2021-05-19
disable execution of test job for gh-pages branch

ljubon:
disable the execution of `test` job in CircleCI for `gh-pages` branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -315,7 +315,11 @@ workflows:
   build_and_deploy:
     jobs:
       - code_style
-      - test
+      - test:
+          filters:
+            branches:
+              ignore: 
+                - gh-pages
       - build
       - release:
           filters:
@@ -329,4 +333,3 @@ workflows:
               only: /.*/
             branches:
               ignore: /.*/
-


ljubon:
> Hi @ljubon, I have added empty Circle CI config into gh-pages branch to stop the builds.

Hi @jankaspar, thank you for your fix! 
I'm having trouble testing my changes here since `circleci/tests` are still appearing on my jobs list 🤔 
Is this expected? Let me know if you have some suggestions.

jankaspar:
I don't think its easy to avoid running test for any PR merging things to `master`. The only thing we could do here is to pass the tests quickly if there are no changes except in docs folder. 

When you say you have `circleci/tests` on your list, what do you mean? Do you want to make PR against `gh-pages` branch and not run any of the test? For that your branch needs to have empty circleci config file as well (https://github.com/G-Research/armada/blob/gh-pages/.circleci/config.yml). The required checks are set to protect master branch so any PR against `gh-pages` should be fine.

ljubon:
> I don't think its easy to avoid running test for any PR merging things to `master`. The only thing we could do here is to pass the tests quickly if there are no changes except in docs folder.
> 
> When you say you have `circleci/tests` on your list, what do you mean? Do you want to make PR against `gh-pages` branch and not run any of the test? For that your branch needs to have empty circleci config file as well (https://github.com/G-Research/armada/blob/gh-pages/.circleci/config.yml). The required checks are set to protect master branch so any PR against `gh-pages` should be fine.

HI @jankaspar, I tested now this fix on my fork, let me know if you can see pipelines now: https://app.circleci.com/pipelines/github/ljubon/armada?filter=ljubon
I was missing that part before, thanks for pointing it out :)

ljubon:
closing PR as fix has been applied directly to `gh-pages` [commit link](https://github.com/G-Research/armada/commit/c302e957dfb7f3135088f37264bb830924d692f2)

PR #598, opened 2021-06-09
enable scanning dependency with dependabot

ljubon:


ljubon:
FYI at the moment on my forked repo, I have 6 PRs opened and waiting for resolution 🤞 
Check it out here how it works: https://github.com/ljubon/armada/pulls
If we merge this, we can expect same PRs opened

jankaspar (approved):
Great, dependabot already nags us about npm, I bet the go updates will be more useful. :)

PR #600, opened 2021-06-09
enable gitpod

ljubon:
Really cool tool to help contributors contribute even faster & from anywhere but still using their favorite IDE 
(browser required still 🤣)
More about this here: https://www.gitpod.io/features & https://www.gitpod.io/docs/quickstart


jankaspar (approved):
Looks good, just out of curiosity, why do we build all the xecutables as part of the init?

ljubon:
> Looks good, just out of curiosity, why do we build all the xecutables as part of the init?

Hey @jankaspar , thanks for reviewing and merging this! 🙏 
To answer your question, in `init` section we want to do heavy lifting (e.g download & install dependencies, compile your source code..etc)
The benefit of that is that when someone open GitPod env, he will have everything ready and prepared for development
You can take a look official docc for flow of actions and where inits is running: https://www.gitpod.io/docs/config-start-tasks/#prebuild-and-new-workspaces

ljubon:
@jankaspar please merge this when you can, all checks are passed 🚀 I'm available if anything else is needed 💪

PR #623, opened 2021-07-08
Make developer instructions work with kind v0.11.1

robertdavidsmith:
Newer kind doesn't support that api version - so move to newer version already used by the quickstart
Newer kind also wants names in lowercase


samclark (approved):
LGTM

PR #624, opened 2021-07-08
Better error messages/status code on failure to load queue

robertdavidsmith:
1) Replace message `Could not load queue: redis: nil` with `Queue "myqueue"  not found`

2) Only return NotFound status if genuinely not found. For other errors (e.g. redis down) return Unavailable

### DIFF
--- a/internal/armada/repository/queue.go
+++ b/internal/armada/repository/queue.go
@@ -1,6 +1,8 @@
 package repository
 
 import (
+	"errors"
+
 	"github.com/go-redis/redis"
 	"github.com/gogo/protobuf/proto"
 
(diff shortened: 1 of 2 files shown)


internal/armada/repository/queue.go
```diff
@@ -9,6 +11,8 @@ import (
 
 const queueHashKey = "Queue"
 
+var QueueNotFound = errors.New("Queue does not exist")
```
samclark, line 14: The naming convention in go is to prefix well known errors with 'Err'. So 'ErrQueueNotFound' might be a better name to use here.

samclark (dismissed):
Approved with minor comments

samclark (approved):

PR #629, opened 2021-07-16
Experimental auto-scaling support

jankaspar:
This change allows to specify auto-scaling information in executor configuration. Server component then figures out which machines could be provisioned in the cluster and tries to schedule for these dormant resources too. The jobs can be placed to these machines only if old enough (this is controlled by `AutoscalingNodePlacementDelay` server config value). The intention is that only if the jobs can't be placed anywhere else, or are sufficiently behind schedule, the auto-scaling will happen.

TODO:
- [ ] Test this in public cloud cluster with autoscaling group.


jimbobby5:
Closing stale PR

PR #640, opened 2021-07-28
Fix kind build

carlocamurri:


JamesMurkin:
Are we sure `kubernetes-ready` needs to be changed to `config`?

I think it was originally used because the docs/example suggests that endpoint:
https://github.com/bsycorp/kind-circleci-example/blob/master/.circleci/config.yml

Although its not totally clear from the docs which is the right one to check and which order they become ready in.

JamesMurkin:
Also I think you can use `sed -i` to do it in place, but not totally familiar. However seems cleaner if it works.

carlocamurri:
> Are we sure `kubernetes-ready` needs to be changed to `config`?
> 
> I think it was originally used because the docs/example suggests that endpoint:
> https://github.com/bsycorp/kind-circleci-example/blob/master/.circleci/config.yml
> 
> Although its not totally clear from the docs which is the right one to check and which order they become ready in.

The problem with that is the endpoint `/kubernetes-ready` now, for some reason, gives 404. Thought `/config` would be the next best approximation for this.

PR #654, opened 2021-08-12
No longer send completely empty utilisation events

JamesMurkin:
There is no need to send utilisation events that don't contain any utilisation data
 - This may be because kubelet isn't providing the data, or we haven't scraped the kubelet yet (very short jobs)


robertdavidsmith (changes requested):
I'd add a test to https://github.com/G-Research/armada/blob/master/internal/executor/utilisation/job_utilisation_reporter_test.go

robertdavidsmith (approved):

PR #719, opened 2021-09-28
Support required node affinity, break out pod_matching_context.go, numerous library upgrades

robertdavidsmith:
Two main changes. This becomes clearer if you view the individual commits (should have done 2 PRs with hindsight)
1) Add support for node affinity
2) Break pod_matching_context out from node_matching (node_matching was getting too big, and this is also a performance win as totalPodResourceRequest/requiredNodeAffinitySelector are created once per pod now, not once per pod per nodeType


JamesMurkin (approved):

internal/common/validation/podspec_test.go
```diff
@@ -77,3 +77,99 @@ func Test_ValidatePodSpec_checkForPortConfiguration(t *testing.T) {
...
+	podSpec.Affinity = &v1.Affinity{
+		NodeAffinity: &v1.NodeAffinity{
+			RequiredDuringSchedulingIgnoredDuringExecution: nodeSelector,
+		},
+	}
+
+	assert.Nil(t, ValidatePodSpec(podSpec))
+}
+
+func Test_ValidatePodSpec_WhenInValidRequiredAffinitySet_Fails(t *testing.T) {
```
JamesMurkin, line 134: Minor point, Invalid is one word

robertdavidsmith: Fair point, will fix in next PR

PR #725, opened 2021-10-01
Armada Executor: When returning lease include AvoidNodeLabels

robertdavidsmith:
Add an executor config option - AvoidNodeLabelsOnRetry

When a job fails, ask the server to avoid any nodes where all of these labels match the labels on the node that failed.

For small clusters, AvoidNodeLabelsOnRetry can just be set to a label giving hostname

For larger clusters, where we don't want a nodeType per node to be created on the server, can set to something like cluster name.

A separate PR will be needed to make the server do the right thing when it receives AvoidNodeLabels.


internal/executor/service/job_lease.go
```diff
@@ -127,3 +138,25 @@ func (jobLeaseService *JobLeaseService) RenewJobLeases(jobs []*job.RunningJob) (
...
 }
+
+func getAvoidNodeLabels(pod *v1.Pod, avoidNodeLabelsOnRetry []string, clusterContext context2.ClusterContext) (map[string]string, error) {
+	if len(avoidNodeLabelsOnRetry) == 0 {
+		return map[string]string{}, nil
+	}
+
+	nodeName := pod.Spec.NodeName
+	if nodeName == "" {
+		return nil, errors.New("Empty node name in pod")
```
JamesMurkin, line 149: Not sure this should be an error.

This would happen on webhook failures (i.e not never got submitted) OR even pod never got assigned to a node due to fragmentation so we are returning it.

JamesMurkin (approved):

JamesMurkin (approved):

internal/executor/service/job_lease.go
robertdavidsmith: I've pushed another commit to improve log messages

PR #738, opened 2021-10-14
Report job updates as an event

robertdavidsmith:


internal/lookout/repository/store.go
```diff
@@ -121,6 +122,35 @@ func (r *SQLJobStore) RecordJobReprioritized(event *api.JobReprioritizedEvent) e
 	return err
 }
 
+func (r *SQLJobStore) RecordJobUpdated(event *api.JobUpdatedEvent) error {
+	updatedJobJson, err := r.getUpdatedJobJson(event)
+	if err != nil {
+		return err
+	}
+
+	ds := r.db.Update(jobTable).
```
carlocamurri, line 131: Minor thing: although unlikely in this case, we could be getting events in the wrong order. Probably worth to change this to an `Insert` with `OnConflict`, see the other `Report*` methods.

PR #761, opened 2021-11-08
Update Developer guide

d80tb7:
* Specify that we need Go 1.1.6 as per https://github.com/G-Research/armada/blob/master/go.mod#L3
* Move to latest version of Kind.
* Update commands to reflect the fact that Kind have changed the way they do contexts
* `armadactl create queue` now needs a `--queueName`


docs/developer.md
```diff
@@ -35,12 +35,9 @@ This is recommended when working on features that are purely Armada specific or
 
 1. Get kind (Installation help [here](https://kind.sigs.k8s.io/docs/user/quick-start/))
     ```bash
-    GO111MODULE="on" go get sigs.k8s.io/kind@v0.5.1
+    GO111MODULE="on" go get sigs.k8s.io/kind@v0.11.1
     ``` 
-2. Create kind clusters (you can create any number of clusters)
-
-    As this step is using Docker, it will require root to run
```
d80tb7, line 42: I don't think this requires root? At least not if you've done https://docs.docker.com/engine/install/linux-postinstall/?

robertdavidsmith: In practice it does require root on many machines. Maybe change to "it might require root (depending on how your machine is set up)" or something

docs/developer.md
```diff
@@ -144,7 +141,7 @@ When you're done, you can run
 
 1. Create queue & Submit job
 ```bash
-go run ./cmd/armadactl/main.go create queue test --priorityFactor 1
+go run ./cmd/armadactl/main.go create queue --queueName test --priorityFactor 1
 go run ./cmd/armadactl/main.go submit ./example/jobs.yaml
```
robertdavidsmith, line 145: Good spot, but note this is broken elsewhere under docs/ too.

d80tb7: ah yes- they should be fixed now

docs/developer.md
d80tb7: done

robertdavidsmith (approved):

PR #768, opened 2021-11-10
Update user.md to be more concise and clear

severinson:
This updated version is edited for clarity and is about 100 lines, whereas the old was about 170 lines. The new version has the same information content as the old.


docs/user.md
```diff
@@ -1,19 +1,23 @@
-# User Guide
+# User guide
 
-## How to use Armada
+Armada is a cluster manager, and is responsible for scheduling and running user jobs (e.g., a batch compute job for training a machine learning model) over one or more Kubernetes clusters. This document is meant to be an overview of Armada for new users. A basic understanding of Docker and Kubernetes is required; see, e.g., the following links:
```
d80tb7, line 3: I *think* we can remove all the commas in this paragraph

devlounge: What about this instead?

_Armada is a Kubernetes native scheduler responsible for scheduling and running user jobs (e.g., a batch compute job for training a machine learning model) over one or more Kubernetes clusters in a federated way.
This document is meant to be an overview of Armada for new users. A basic understanding of Docker and Kubernetes is required; see, e.g., the following links:_

severinson: I think federated is unnecessary here. The fact that jobs may be scheduled to any of the underlying Kubernetes clusters is explained further down and may not need to be part of the first sentence.

severinson: I've updated the file to refer to Armada as a scheduler instead of as a cluster manager.

severinson: This change also addresses @d80tb7 comment.

devlounge (dismissed):

docs/user.md
devlounge: In the K8S world, the word "federated" expresses the fact that it spans over multiple clusters.

severinson: Yes, I agree. But this is already made clear without needing to use the word federated, the meaning of which may not be clear to users of Armada.

samclark (approved):
LGTM

devlounge (approved):

samclark (approved):
LGTM

PR #769, opened 2021-11-10
[Issue-766] Update developer guide for kind version v0.11.1+

### LINKED ISSUE #766: [Documentation] Developer guide for using kind is out of date
Developer guide for running armada executors locally is out of date.

[Kind version](https://github.com/G-Research/armada/blob/8a2e87da9dfa74d08d4ccd90439ad3ac2e64ac63/docs/developer.md?plain=1#L38) that should be used should be updated. In quick start guide v0.10.0+ version is used.

These [commands](https://github.com/G-Research/armada/blob/8a2e87da9dfa74d08d4ccd90439ad3ac2e64ac63/docs/developer.md?plain=1#L59-L66) no longer work with kind version v0.10.0+

steffnova:
[Problem]
Kind version used by armada is v0.10.0+. Command specified in the developer
guide `kind get kubeconfit-path` no longer exists

[Solution]
Updated the guide to use `kind export kubeconfig` command

Close #766


d80tb7:
Hi,

just wondering if we need this?  My understanding is that Kind will, by default, write out the cluster config to your KUBECONFIG file ([link to relevant docs](https://kind.sigs.k8s.io/docs/user/quick-start/#interacting-with-your-cluster)) so we shouldn't have to export it separately. I tried this on my system and the existing doc seemed to work fine.  I'm quite new to both Kind and Armada though so it's quite possible I've missed something.

thanks,

Chris

robertdavidsmith (approved):

steffnova:
Hi @d80tb7 

I think you maybe right. I've tested it locally. I had to update my kind version.

Regards,
Stefan

PR #773, opened 2021-11-15
[Armadactl] Authenticate via token from external cmd

d80tb7:
### What changes were proposed in this pull request?
Adds a new type of auth `ExecAuth` to ArmadaCtl that causes an external program to be invoked to generate a BeraerToken.

### Why are the changes needed?
Allows Armada to integrate with external auth systems for which it doesn't have native support

### Does this PR introduce any user-facing change?
Yes.  Users can now modify their Armada Configuration to include an ExecAuth Section.  For Example,
```
execAuth:
  cmd: some-command

  args:
    - "arg1"

  env:
    - name: "FOO"
      value: "bar"

  interactive: true
```
The stdout from the command will be used as a bearer token when making calls to armada server.
### How was this patch tested?
Unit tests


pkg/client/auth/exec/authenticator.go
```diff
@@ -0,0 +1,98 @@
+package exec
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	grpc_credentials "google.golang.org/grpc/credentials"
```
carlocamurri, line 7: Looks like this is out of order, not sure why it's not being picked up by `goimports`

pkg/exec.go
```diff
@@ -0,0 +1,25 @@
+package main
```
carlocamurri, line 1: Is this file needed?

d80tb7: ah yes good spot

pkg/client/auth/exec/authenticator.go
d80tb7: done

carlocamurri (approved):

PR #782, opened 2021-11-18
Insert build variables into and add version command to `armadactl`

severinson:
Insert build variables (release version, git commit, go version, current date and time) into `go build` when building `armadactl`. This is done using linker flags; see [this](https://blog.alexellis.io/inject-build-time-vars-golang/). These variables are stored in a new sub-package of `armadactl` and can be printed with a new `version` command of `armadactl`. Designed to resemble the `docker version` command.

Example:
```bash
> ./armadactl version
Version:    UNKNOWN_VERSION
Go version: go1.17.2 windows/amd64
Commit:     b50144f
Built:      18 November 2021 12:18:47
```

Version is set to `UNKNOWN_VERSION` if the `RELEASE_VERSION` environment variable (which is set by circleci, see [here](https://github.com/G-Research/armada/blob/master/.circleci/config.yml)) is missing.

In addition, this PR adds some checks to the `makefile` to ensure that necessary executables are available before trying to build.

### DIFF
--- a/cmd/armadactl/build/build.go
+++ b/cmd/armadactl/build/build.go
@@ -1,0 +1,9 @@
+package build
+
+var BuildTime string
+
+var GitCommit string
+
+var ReleaseVersion string
+
+var GoVersion string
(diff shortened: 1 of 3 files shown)


d80tb7 (approved):
lgtm

PR #787, opened 2021-11-25
Process Armada events in batches

carlocamurri:
Enable processing job events in batches. Modified `UpdateStartTime` in the `JobRepository` to allow for updating from multiple events at the same time.

Not modifying the logic in Lookout event processing, as the job storage logic should be modified.


internal/armada/configuration/types.go
```diff
@@ -64,6 +63,14 @@ type LeaseSettings struct {
 	ExpiryLoopInterval time.Duration
 }
 
+type EventsConfig struct {
+	StoreQueue     string // Queue group processor responsible for
```
robertdavidsmith, line 67: comment incomplete

robertdavidsmith, line 67: comment incomplete

internal/armada/repository/job.go
```diff
@@ -427,35 +427,71 @@ func (repo *RedisJobRepository) getAssociatedCluster(jobIds []string) (map[strin
...
 local clusterAssociation = KEYS[2]
+local job = KEYS[3]
 
 local clusterId = ARGV[1]
 local startTime = ARGV[2]
 local startTimeNumber = tonumber(ARGV[2])
 
+local exists = redis.call('GET', job)
+if not exists then
+	return -3
```
robertdavidsmith, line 492: It's a shame to repeat the "-3", use fmt.Sprintf to add it?

robertdavidsmith (commented):
It looks like we can lose Acked events now they are queued in memory.

internal/armada/repository/event_job_status.go
```diff
@@ -8,13 +8,53 @@ import (
...
+		}
+
+		jobErrors, err := jobRepository.UpdateStartTime(jobStartInfos)
+		if err != nil {
+			log.Errorf("error when updating start times for jobs: %v", err)
+			return err
+		}
+		for _, err := range jobErrors {
+			if err != nil {
+				log.Errorf("error when updating start time for single job: %v", err)
```
d80tb7, line 48: why doesn't this return an error?  Is it the case that the caller doesn't need to know what's gone wrong?

internal/armada/repository/event_job_status.go
```diff
@@ -35,12 +74,13 @@ func (p *EventJobStatusProcessor) handleMessage(eventMessage *api.EventMessage)
...
 	switch event := event.(type) {
 	case *api.JobRunningEvent:
-		err = p.jobRepository.UpdateStartTime(event.JobId, event.ClusterId, event.Created)
+		eventMessage, err := api.Wrap(event)
 		if err != nil {
-			log.Errorf("error while updating job start time: %v", err)
-			if err.Error() != JobNotFound {
-				return err
-			}
+			log.Errorf("error when wrapping job running event: %v", err)
```
d80tb7, line 79: What was the purpose of the deleted code which ignored JobNotFound?  Is it a hack that's no longer required?

internal/common/eventstream/eventbatcher.go
```diff
@@ -0,0 +1,121 @@
...
+	if len(b.batch) >= b.batchSize {
+		err := b.flushUnlocked()
+		if err != nil {
+			return err
+		}
+	} else {
+		if b.t != nil {
+			b.t.Stop()
+		}
+		b.t = time.AfterFunc(1*time.Second, func() {
```
d80tb7, line 60: My inexperience in Go is probably showing here but does this mean that every event will trigger a Flush, albeit one that happens a second after the event is reported and that may just try and flush am empty buffer?

If this is the case it sounds a bit hard to reason about and potentially a bit inefficient.  Do you think it would be possible to flush when the buffer is full or a second after the last message

d80tb7:
I guess we should note that this change means that we can potentially lose up to 1 second of events if an api server goes down.  I guess this is ok?

Also if we run multiple api servers- will this lead to a (much greater) chance of events being stored out of order? E.g. a job runs and completes in < 1 second (unlikely I know).  The run event is processed by server a and is queued for 1 second.  The completion event is processed immediately.  As a result, the completion event is stored before the processed event.  If this is the case, does it matter?

internal/armada/repository/event_job_status.go
```diff
@@ -25,24 +38,59 @@ func (p *EventJobStatusProcessor) Start() {
...
+			})
+		}
+	}
+
+	jobErrors, err := p.jobRepository.UpdateStartTime(jobStartInfos)
+	if err != nil {
+		log.Errorf("error when updating start times for jobs: %v", err)
+		return err
+	}
+	if len(jobErrors) < len(batch) {
```
JamesMurkin, line 82: Maybe this should be != ?
If job errors > batch, we'd still get an index out of range in the loop below.

Should it be called updateResults or something?

internal/armada/repository/job.go
```diff
@@ -50,7 +50,7 @@ type JobRepository interface {
 	DeleteJobs(jobs []*api.Job) map[*api.Job]error
 	GetActiveJobIds(queue string, jobSetId string) ([]string, error)
 	GetLeasedJobIds(queue string) ([]string, error)
-	UpdateStartTime(jobId string, clusterId string, startTime time.Time) error
+	UpdateStartTime(jobStartInfos []*JobStartInfo) ([]error, error)
```
JamesMurkin, line 53: I think typically we've made the return type something like map[string]error or map[api*Job]error. 
 - Maybe we've deleted most of these. I only see DeleteJobs matching this now

Then its clearer which job id corresponds with this error, rather than relying on index. Although not mandatory, I guess this is more "efficient" than passing as much data around

internal/armada/server.go
```diff
@@ -80,18 +80,30 @@ func Serve(config *configuration.ArmadaConfig, healthChecks *health.MultiChecker
...
+		teardown = func() {
+			err := eventRepoBatcher.Stop()
+			if err != nil {
+				log.Errorf("failed to flush event processor buffer for redis")
+			}
+			err = jobStatusBatcher.Stop()
+			if err != nil {
+				log.Errorf("failed to flush job status batcher processor")
+			}
+			err = eventStream.Close()
```
JamesMurkin, line 104: If we closed eventStream first, does it prevent new event coming in? This might make the above Stop() calls cleaner as there will be an "end" to the events with no more coming in.

internal/armada/repository/job.go
```diff
@@ -427,43 +427,79 @@ func (repo *RedisJobRepository) getAssociatedCluster(jobIds []string) (map[strin
...
 local startTimeKey = KEYS[1]
 local clusterAssociation = KEYS[2]
+local job = KEYS[3]
 
 local clusterId = ARGV[1]
 local startTime = ARGV[2]
 local startTimeNumber = tonumber(ARGV[2])
 
+local exists = redis.call('GET', job)
+if not exists then
```
JamesMurkin, line 491: What are we trying to catch here?
 - Why not use EXISTS? I imagine it must be faster than GET
 
Also if we are trying to check the job isn't already deleted. Better to check if the TTL is set on this object rather than GET/EXISTS

internal/armada/server.go
```diff
@@ -120,7 +132,7 @@ func Serve(config *configuration.ArmadaConfig, healthChecks *health.MultiChecker
 	grpcCommon.Listen(config.GrpcPort, grpcServer, wg)
 
 	return func() {
-		stopSubscription()
+		teardown()
```
JamesMurkin, line 135: What grace period do we give for a server to shut down? In theory this could have added another 20 seconds to shut down time

JamesMurkin (dismissed):

JamesMurkin (approved):

PR #790, opened 2021-11-30
Allow creation of TLS protected ingresses

theAntiYeti:
Allow creation of TLS protected ingresses through armada jobsets. Will create the yaml on the ingress but dealing with the certificate will be left to the implementation of armada on a given system.


internal/executor/util/kubernetes_object.go
```diff
@@ -89,6 +94,20 @@ func CreateIngress(name string, job *api.Job, pod *v1.Pod, service *v1.Service,
 		rules = append(rules, path)
 	}
 
+	tls := make([]networking.IngressTLS, 0, 1)
+
+	if jobConfig.TlsEnabled {
```
d80tb7, line 99: what should do the validation here? E.g. is it valid to have `jobConfig.TlsEnabled = true` but `len(service.Spec.Ports = 0`.

internal/executor/util/kubernetes_object.go
```diff
@@ -89,6 +94,20 @@ func CreateIngress(name string, job *api.Job, pod *v1.Pod, service *v1.Service,
 		rules = append(rules, path)
 	}
 
+	tls := make([]networking.IngressTLS, 0, 1)
```
d80tb7, line 97: Should there be a test for this new logic in `kubernetes_objects_test,go`?

Arguably we should also have a test for the changes to `ingres_util`, but I could accept the argument that maybe the code is trivial enough that we don't need to test this?

internal/executor/util/kubernetes_object.go
theAntiYeti: Are you worried about the case of people configuring no rules and just using a default backend?

d80tb7: I guess what I'm really thinking about is what validates that the job spec makes sense, so that we don't go to the trouble of submitting jobs to k8s only to find out that they fail or act in a way that the user doesn't expect.

In this particular case I was wondering about the case of a user submitting a job spec with zero ports but some ingress config.  To me, this wouldn't make any sense because if you haven't got any service ports it's not clear what you'd be ingressing to (although my knowledge of k8s isn't brilliant so it's possible this is in fact a meaningful config!).

theAntiYeti: if you have no rules everything gets routed to the default backend, which might have a use in general but probably won't be spun up (but I don't see a reason as to why we should prevent this considering it is in fact valid K8s).

Nonetheless, having a TLS certificate or not is unrelated to how many ports you configure routing rules for making validation of this completely out of the scope of this story. I hope this resolves this :)

internal/executor/util/kubernetes_object.go
theAntiYeti: I'll have a look at adding a test, but what logic in specific do you mean when you say this and highlight line 97?

internal/executor/util/kubernetes_object.go
JamesMurkin: I guess here would be a sensible place to add the check:
https://github.com/G-Research/armada/blob/master/internal/common/validation/job.go#L13

But yeah it can be a separate PR. 

Similarly if we wanted to be pessimistic, we could also check it on the executor somewhere and skip creation on 0 ports 
 - It looks like we do in the calling function
 
However that first check would reject at submission time so the user can see instantly its invalid

internal/executor/util/kubernetes_object.go
d80tb7: I mean we should have a test for all the new logic you've added in this function.  I.e. that if we pass a jobConfig with  `TlsEnabled = true` into the function, we get back an ingress with the correct `IngressSpec` set.

theAntiYeti: we don't test this function at all currently. I'm not entirely sure what the value of this is because generating this ingress spec by hand seems like an exercise in futility considering there might be multiple ways to represent the same fundamental k8s ingress in go structs. So I fear to actually generate this config we'd have to either use the function (which isn't really a test at that point) or follow the function through by hand which isn't any better (arguably worse)

JamesMurkin (approved):

d80tb7 (approved):

PR #791, opened 2021-11-30
Fixed incorrect jobfile examples in user guide

severinson:
Previously, it was possible to submit a jobfile consisting of just a podspec + queue name and priority, but this option has since been deprecated. Now, jobs must always be given as a list of jobs. This PR updates examples using the old (deprecated) format to use the new format.


docs/user.md
```diff
@@ -16,27 +16,28 @@ The Armada workflow is:
 
 To create a job that sleeps for 60 seconds, start by creating a yaml file with the following contents:
 
-```yaml
+```yaml
```
carlocamurri, line 19: Trailing whitespace

carlocamurri:
```suggestion
```yaml
```

carlocamurri (approved):

PR #793, opened 2021-12-01
Batch cancel jobs in Armada API

carlocamurri:
Avoids loading all jobs in memory when cancelling. Defaults to cancelling 200 jobs at a time.


internal/armada/server.go
```diff
@@ -129,3 +137,10 @@ func Serve(config *configuration.ArmadaConfig, healthChecks *health.MultiChecker
 func createRedisClient(config *redis.UniversalOptions) redis.UniversalClient {
 	return redis.NewUniversalClient(config)
 }
+
+func getCancelJobsBatchSize(config *configuration.ArmadaConfig) int {
+	if config.CancelJobsBatchSize > 0 {
```
theAntiYeti, line 142: Is it not worth putting in an upper limit? Would there not be a problem if we asked the server to cancel jobs in batches of 1000000 or so?

internal/armada/server/submit.go
```diff
@@ -161,23 +164,33 @@ func (server *SubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmitRe
...
+		for _, batch := range batches {
+			jobs, err := server.jobRepository.GetExistingJobsByIds(batch)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, err.Error())
+			}
+			result, err := server.cancelJobs(ctx, request.Queue, jobs)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, err.Error())
+			}
+			cancelledIds = append(cancelledIds, result.CancelledIds...)
```
theAntiYeti, line 191: What happens if a jobset gets half cancelled? should we not pass this through with the error information to at least have some accountability on what got cancelled and what didnt?

theAntiYeti (approved):

PR #795, opened 2021-12-01
Add a short hostname at the top of the hostnames for tls cert signing

theAntiYeti:
TLS certificates need a first host for signing that must be less than 64 characters long for our signing agent.
As all the hostnames generated for armada jobs tend to be significantly longer (~90 chars) this change introduces a new hostname to act as the first host.


internal/executor/util/kubernetes_object.go
```diff
@@ -66,8 +66,13 @@ func CreateIngress(name string, job *api.Job, pod *v1.Pod, service *v1.Service,
 	})
 
 	rules := make([]networking.IngressRule, 0, len(service.Spec.Ports))
-	tlsHosts := make([]string, 0, len(service.Spec.Ports))
+	tlsHosts := make([]string, 0, len(service.Spec.Ports)+1)
 
+	// First host used for certificate signing, needs to be less than 64 chars long
+	firstHost := fmt.Sprintf("%s.svc", name)
```
JamesMurkin, line 72: Looks good but we should make the domain extension configurable.

Currently this assumes certs can be signed for the .svc domain which I don't think is going to be true everywhere
 - Happy for the default to be .svc

Should just be a case of adding another field to executor IngressConfiguration + the executor config file to set the default 
 - You already have access to this struct in this func

theAntiYeti: what would be wrong with using the hostnameSuffix? This is already in the executorIngressConfig

internal/executor/util/kubernetes_object.go
```diff
@@ -66,8 +66,17 @@ func CreateIngress(name string, job *api.Job, pod *v1.Pod, service *v1.Service,
 	})
 
 	rules := make([]networking.IngressRule, 0, len(service.Spec.Ports))
-	tlsHosts := make([]string, 0, len(service.Spec.Ports))
+	tlsHosts := make([]string, 0, len(service.Spec.Ports)+1)
 
+	// First host used for certificate signing, needs to be less than 64 chars long
+	domain := executorIngressConfig.CertDomain
+	if domain == "" {
+		domain = "svc"
```
theAntiYeti, line 74: I'm not sure if this is the right way to set the default, but I've been looking around the codebase and haven't stumbled across how it's done for HostnameSuffix

JamesMurkin: https://github.com/G-Research/armada/blob/master/config/executor/config.yaml#L24

pkg/client/auth/kerberos/spnego_unix.go
```diff
@@ -1,3 +1,4 @@
+//go:build linux || darwin
```
carlocamurri, line 1: Is this needed here?

theAntiYeti: The goimports linter keeps complaining if it isn't there. I don't believe it's necessary though

carlocamurri: Don't see it reported as a linter error, I'd leave it out of this PR

theAntiYeti: Removed

carlocamurri (approved):

PR #800, opened 2021-12-02
[feature/issue-702] Add unit test coverage for Submit Job API [Armada Server]

steffnova:
[Problem]

Submit Job is one of main fatures, and it doesn't have proper test coverage

[Solution]

Create handler in handlers package for submit job request. Add unit tests
to cover all of handlers execution paths.


internal/armada/server/handlers/submit_jobs.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+		}
+
+		return submit(ctx, request)
+	}
+}
+
+func (submit submitJobs) Authorize(authorizeOwnership authorizeOwnership, authorize authorize, autocreate bool) submitJobs {
+	return func(ctx context.Context, request *api.JobSubmitRequest) (*api.JobSubmitResponse, error) {
+		switch owned, err := authorizeOwnership(ctx, request.Queue); {
```
robertdavidsmith, line 41: This is rather confusing, owned/err aren't referenced by any case statements. It would be better on its own line.

I think if/else is clearer here, not case.

internal/armada/server/handlers/submit_jobs_test.go
```diff
@@ -0,0 +1,282 @@
...
+			}
+
+			if status, _ := status.FromError(err); status.Code() != codes.NotFound {
+				t.Errorf("invalid status code: %s", status.Code())
+				return false
+			}
+
+			return true
+		},
+		"internal": func(request api.JobSubmitRequest, sumbitAnyJob bool) bool {
```
robertdavidsmith, line 191: clearer name for this test would be good

internal/armada/server/handlers/submit_jobs.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/G-Research/armada/internal/armada/permissions"
+	"github.com/G-Research/armada/internal/armada/repository"
+	"github.com/G-Research/armada/internal/common/validation"
+	"github.com/G-Research/armada/pkg/api"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type authorizeOwnership func(ctx context.Context, queueName string) (bool, error)
```
robertdavidsmith, line 15: Rename to queueOwnedBy ?

robertdavidsmith (changes requested):

internal/armada/server/handlers/submit_jobs_test.go
steffnova: Hm, It's the name of the status code returned by handler. Do you have any propositions?

internal/armada/server/handlers/submit_jobs_test.go
```diff
@@ -0,0 +1,282 @@
...
+				t.Errorf("failed to handle request: %s", err)
+			}
+
+			return true
+		},
+	}
+
+	for name, property := range properties {
+		t.Run(name, func(tb *testing.T) {
+			if err := quick.Check(property, nil); err != nil {
```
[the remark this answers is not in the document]
steffnova: The handler itself is a higher order function (it receives a functions as an input). testing/quick.Check allows generating test cases without manually writing them down. As it is seen from the tests themselves, they are not tied to any input specifically (at least in case of this handler), but are tied to outcome of the injected functions.

with quick package, arbitrary submit requests are generated just to ensure that the higher property of the handler is satisfied no matter how request data looks like. If structure of the request changes in the future, the test would still hold. Quick package ensures that various values for all fields of request are being generated (0 values for numeric types, empty slices and maps, nil values where possible and so on)

Every test here is written for specific outcome (status code) even if there are multiple paths that lead to it.

I would like to see an example that you head in mind though, as I am curious.

internal/armada/server/handlers/submit_jobs.go
steffnova: I've renamed it.

internal/armada/server/handlers/submit_jobs_test.go
```diff
@@ -0,0 +1,282 @@
...
+			}
+
+			if status, _ := status.FromError(err); status.Code() != codes.Internal {
+				t.Errorf("invalid status code: %s", status.Code())
+				return false
+			}
+
+			return true
+		},
+		"permissionDenied": func(request api.JobSubmitRequest, sumbitJob bool) bool {
```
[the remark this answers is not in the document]
steffnova: Done!

internal/armada/server/handlers/submit_jobs.go
steffnova: Maybe you are not familiar but in go you can define variables in switch statement. Most common place is when you do type switching like here: https://github.com/G-Research/armada/blob/1b933fe66485c8aba7112b9ab0ab4d147a0af336/pkg/api/event_util.go#L35

If you follow up with semicolon like here, it doesn't switch by any value. You can have switch statement in go without switch value:

```go
var i;
var j;

switch {
  case i == 0:
  case i == 1:
  case j == 3:
  case i%j == 0:
  default:
}

```

This is very useful to define code that needs more complex if/else statements. I hopes this makes sense.

Note: I've moved 
```go
owned, err := authorizeOwnership(ctx, request.Queue)
```
outside of switch statement.

internal/armada/server/jobs_test.go
```diff
@@ -0,0 +1,184 @@
...
+			)
+
+			if _, err := new(context.Background(), &req); err != nil {
+				t.Errorf("failed to generate jobs from request: %s", err)
+				return false
+			}
+
+			return true
+		},
+		"erorr": func(req api.JobSubmitRequest, newJobsError, schedulingError, validationError bool) bool {
```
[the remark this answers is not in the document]
steffnova: fixed

internal/armada/server/handlers/submit_jobs.go
robertdavidsmith: Yes I'm aware this is possible, but still think it's clearer to have `owned, err := authorizeOwnership(ctx, request.Queue)` outside, thanks for changing

internal/armada/repository/queue.go
```diff
@@ -103,3 +103,27 @@ func (r *RedisQueueRepository) DeleteQueue(name string) error {
 	result := r.db.HDel(queueHashKey, name)
 	return result.Err()
 }
+
+type GetQueueFn func(string) (*api.Queue, error)
+type CreateQueueFn func(*api.Queue) error
+
+func (get GetQueueFn) Autocreate(autocreate bool, priorityFactor float64, create CreateQueueFn) GetQueueFn {
```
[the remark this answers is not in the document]
steffnova: I've moved it to server package.

internal/armada/repository/job.go
```diff
@@ -991,3 +1003,105 @@ if currentClusterId == clusterId then
 end
 return 0
 `)
+
+type reportEvents func([]*api.EventMessage) error
```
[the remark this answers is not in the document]
steffnova: Fixed

steffnova:
Coding style in PR doesn't satisfy coding style of repository

PR #803, opened 2021-12-02
Added new config system with support for contexts

severinson:
This PR introduces a new config system with support for contexts. This new system isn't complete and isn't used by any part of Armada yet. The intention is to merge this into master, perform changes to other parts of the code base required to support contexts, and then finishing implementing contexts. In addition, it cleans up the existing config system somewhat.


pkg/client/command_line.go
```diff
@@ -1,11 +1,12 @@
 package client
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 
-	"github.com/mitchellh/go-homedir"
-	log "github.com/sirupsen/logrus"
+	yaml "gopkg.in/yaml.v2"
```
JamesMurkin, line 8: Why is this in its own group?

severinson:
There is a failing e2e test related to the watch command. However, we need to wait until PR 792 is merged to work on it, since it may be fixed by PR 792 already, and any work on armadactl in this branch would introduce merge conflicts.
https://github.com/G-Research/armada/pull/792

pkg/client/command_line.go
d80tb7: If you're on goland then adding a file watcher that runs gofmt and go goimports on file save is awesome.

pkg/client/command_line.go
```diff
@@ -15,62 +16,215 @@ func AddArmadaApiConnectionCommandlineArgs(rootCmd *cobra.Command) {
 	viper.BindPFlag("armadaUrl", rootCmd.PersistentFlags().Lookup("armadaUrl"))
 }
 
-func LoadCommandlineArgsFromConfigFile(cfgFile string) {
+// LoadCommandlineArgsFromConfigFile load config from at exePath/armadactl-defaults.yaml, where
```
d80tb7, line 19: don't need the `at` here

pkg/client/command_line.go
```diff
@@ -15,62 +16,215 @@ func AddArmadaApiConnectionCommandlineArgs(rootCmd *cobra.Command) {
...
-			default:
-				log.Errorf("Can't read config file %s because %s\n", viper.ConfigFileUsed(), err)
-				os.Exit(1)
-			}
+	exeDir := filepath.Dir(exePath)
+	configPath := filepath.Join(exeDir, "/armadactl-defaults.yaml")
+	viper.SetConfigFile(configPath)
+	if err := viper.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			// Config file not found; ignore
```
d80tb7, line 36: I'd be tempted to log this.  It may save someone some time one day when the defaults file ahs been incorrectly deleted and armadactl doesn't work.

pkg/client/command_line.go
JamesMurkin, line 19: This is probably just a style thing, I find this hard to understand.

Here is another attempt, which may not be any better:
// LoadCommandlineArgsFromConfigFile loads armadactl config
// armadactl-defaults.yaml - From exePath, where exePath is the path to the armadactl executable
// armada config file           - From cfgFile or defaulting to $HOME/.armadactl
// These configs are then merged

pkg/client/command_line.go
```diff
@@ -15,62 +16,215 @@ func AddArmadaApiConnectionCommandlineArgs(rootCmd *cobra.Command) {
...
+// config loaded into viper. Hence, this function must be called after loading config into viper,
+// e.g., by calling LoadCommandlineArgsFromConfigFile.
 func ExtractCommandlineArmadaApiConnectionDetails() *ApiConnectionDetails {
 	apiConnectionDetails := &ApiConnectionDetails{}
 	viper.Unmarshal(apiConnectionDetails)
 	return apiConnectionDetails
 }
+
+// ConfigYamlString returns a string representation of the merged configuration stored in viper.
+func ConfigYamlString() (string, error) {
```
JamesMurkin, line 80: Is this used?

pkg/client/command_line.go
```diff
@@ -15,62 +16,215 @@ func AddArmadaApiConnectionCommandlineArgs(rootCmd *cobra.Command) {
...
-			// This only occurs when looking for the default .armadactl file and it is not present
-			// This is not an error as users don't have to specify it, so do nothing
-		default:
-			log.Errorf("Can't read config file %s because %s\n", viper.ConfigFileUsed(), err)
-			os.Exit(1)
+	// merge in new config with those loaded from armadactl-defaults.yaml
+	// (note the call to viper.MergeInConfig instead of viper.ReadInConfig)
+	if err := viper.MergeInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			// Config file not found; ignore
```
d80tb7, line 61: again I'd log this

pkg/client/command_line.go
```diff
@@ -15,62 +16,215 @@ func AddArmadaApiConnectionCommandlineArgs(rootCmd *cobra.Command) {
...
+			if user == nil {
+				return fmt.Errorf("[armadaClientConfig.validate] user %s is nil", name)
+			}
+		}
+	}
+
+	if config.Clusters != nil {
+		for name, cluster := range config.Clusters {
+			if name == "" {
+				return fmt.Errorf("[armadaClientConfig.validate] empty cluster names not allowed")
```
d80tb7, line 134: I'm not familiar with go logging, but in logging frameworks in other languages you wouldn't need to add all the `[armadaClientConfig.Validate]`s because the logging framework could add it from the logging context.  It may be that something is possible here.  If not I'd be tempted to make it a const string and add it via the format string, but I accept that's really a matter of taste!

JamesMurkin:
Overall the above comments are minor. The main thing I'm interested in is, is this the structure we want?

```
	ApiVersion     string
	Users          map[string]*userConfig
	Clusters       map[string]*clusterConfig
	Contexts       map[string]*contextConfig
	CurrentContext string
```

How does this relate to the actual connection details structure?
https://github.com/G-Research/armada/blob/master/pkg/client/connection.go#L18


This does pretty much correlate with kubeconfig. I guess we are saying the contents of ApiConnectionDetails will be split between Users/Clusters?


Is `clusters` the right word here?

pkg/client/command_line.go
```diff
@@ -15,62 +16,215 @@ func AddArmadaApiConnectionCommandlineArgs(rootCmd *cobra.Command) {
...
+	}
+	return string(bs), nil
+}
+
+// LoadClientConfig reads in config from the file at configPath, or, if configPath is the empty
+// string, from $HOME/.armada/config, and from environment variables, using viper.
+func LoadClientConfig(configPath string) (*armadaClientConfig, error) {
+
+	// if no file is provided (i.e., if configPath is the empty string), default to $HOME/.armada/config
+	if len(configPath) == 0 {
```
d80tb7, line 192: can we log which files we resolved or does viper do that for us?

d80tb7 (approved):

pkg/client/command_line.go
JamesMurkin: This does mean if you aren't setting an `armadactl-defaults.yaml` file. We'd log an error on every command.

Could we have a verbose logging option maybe?

d80tb7: so it doesn't need to be an error-debug would be fine.  I'd assumed that `armada-defaults` would be installed as part of the client install (as we're assuming it's next to the exe) and therefore should generally be there?

JamesMurkin: Its so package managers internal to companies can put a default there along with the installed exe/binary. 
 - So their users don't need to specify any armada config like api
 
 However if the user isn't using a package manager - or their company decides not to populate a armada-defaults, it is not expected to be present.

pkg/client/command_line.go
severinson: Just a mistake on my part; fixed

pkg/client/command_line.go
severinson: Later, I want to add a `view` command to `armadactl`, which prints the merged config. This command could perhaps be made to print the source files used to produce the merged config (as suggested below), which might be the right tradeoff between clarity and verbosity.

pkg/client/command_line.go
severinson: It will be used by the `armadactl view` command when added. For now I'm using it for debugging.

pkg/client/command_line.go
severinson: See response to previous comment on this.

pkg/client/command_line.go
severinson: :)

pkg/client/command_line.go
severinson: Doesn't seem like viper does it for us. Instead, I've added code that adds the found config files to a slice.

severinson:
> Overall the above comments are minor. The main thing I'm interested in is, is this the structure we want?
> 
> ```
> 	ApiVersion     string
> 	Users          map[string]*userConfig
> 	Clusters       map[string]*clusterConfig
> 	Contexts       map[string]*contextConfig
> 	CurrentContext string
> ```
> 
> How does this relate to the actual connection details structure? https://github.com/G-Research/armada/blob/master/pkg/client/connection.go#L18
> 
> This does pretty much correlate with kubeconfig. I guess we are saying the contents of ApiConnectionDetails will be split between Users/Clusters?
> 
> Is `clusters` the right word here?

I made the struct like this because I felt it was the one most natural from the point of view of a user. As you say, `ApiConnectionDetails` needs to be split between users and clusters. I choose cluster since I think a user would find that more netural than, e.g., server. I think a user would say "let's submit this job to the Armada cluster", as opposed to "let's submit this job to the Armada server", for example. Further, each cluster has a server, which makes the config file structure logical.

Note that the structure is a bit different from kubeconfig. In kubeconfig, the users, for example, are stored as a list, where each entry (corresponding to a user) has a name field. However, I felt that storing users, clusters, contexts as maps from a name to the details of the entry was better since we have to require that names are unique between users/clusters/contexts, which we get automatically from a map but not from a list.

JamesMurkin (approved):

d80tb7 (approved):

PR #804, opened 2021-12-03
Make executor use multiple threads for pod update + delete

JamesMurkin:
On large clusters (or when there is a very high pod churn).

Doing these operations in a single thread means the executor can never keep up:
 - As it can only process things as fast as the kube-api responds.
 - Meaning at 30ms response time, the executor can only handle updating/deleting 30 pods/second

Making the executor use more than 1 thread to handle update + delete operations
 - The number of threads can be configured, to suit different needs


config/executor/config.yaml
```diff
@@ -2,6 +2,8 @@ application:
   clusterId : "Cluster1"
   pool: "default"
   submitThreadCount: 1
```
d80tb7, line 4: just curious why you would have different numbers of threads for different operations?  Is it that we don't want to ddos the k8s api and some operations are more expensive than others.

Also re the naming- is this really a thread count here?  It seems more like a concurrency limit?

d80tb7 (approved):

config/executor/config.yaml
JamesMurkin: Possibly its just a concurrency limit, as its effectively:
 - How many go routines do we use to submit/update/delete that can talk to the kube-api concurrently

Happy to change the values here.

I wanted delete to be > submit, just so we don't end up clogging the cluster up as we can submit things faster than delete them.

Also we perform operations in approximately this ratio:
 - 1 submit
 - 1 delete
 - 5 updates
Per job.

So have more update "threads" is to account for this. 
 - It could reasonably set to 5

JamesMurkin:
Set the concurrency limits in the ratio explained about - Although I don't think following the exact ratio above is 100% relevant.

We have discussed and ideally we'd move to a global concurrency limit, however that'd be a larger change than this PR.
 - This would be simpler than having several independent values that makes it hard to know what the right value is

carlocamurri (approved):

PR #812, opened 2021-12-06
armadactl cleanup

severinson:
- Re-organized cobra command setup to avoid using globals
- Moved config file CLI flag into client package (since the client package is the only consumer of this file)
- Added top-level dry-run flag
- Improved printouts and doc strings
- Return an error on not finding user-supplied config file
- Avoid exporting code internal to cmd/armadactl/cmd
- Deleted deprecated commands `queue create`, `queue delete`, `queue info`


cmd/armadactl/cmd/root.go
```diff
@@ -1,56 +1,46 @@
...
-	if err := client.LoadCommandlineArgsFromConfigFile(cfgFile); err != nil {
-		log.Error(err)
-		os.Exit(1)
-	}
-}
-
-func exitWithError(e error) {
-	log.Error(e)
-	os.Exit(1)
+username: user1
```
d80tb7, line 22: is the indentation here correct?

severinson: Yes

d80tb7 (approved):

PR #820, opened 2021-12-08
Add headless services to armada

theAntiYeti:
Big changes:
add ServiceConfig message to API.
Add different ServiceConfig struct to executor to combine differences in IngressConfig and ServiceConfig in the API (getting rid of most of IngressType enum values, potentially even the whole enum at some point)


theAntiYeti:
It might be worth replacing the internal translation datastructure (in internal/executor/util/service_config.go) with an interface in the API package. The pros would be some logical cleaness in the code but the cons would be it would massively entangle the api and util packages, as well as remove the structure from the logic it's used in.

Thoughts?

carlocamurri (approved):

PR #822, opened 2021-12-10
Implement Job Requirements Key

JamesMurkin:
This allows converting a job to a string representing the scheduling requirements of the job.
 - We can consider all jobs with the same "key" as the same for scheduling purposes. 
 - When scheduling, instead of loading the whole job object, we can just load the key
 - This is the first step towards scheduling based on requirements key

Main considerations for this key are:
 - We can convert from job -> key -> job. The job must not change from the point of view of scheduling but minor changes are acceptable. I.e conversion of "1Gi" -> "1073741824"
 - It should be compact. As long keys are slow in Redis
 - It must be usable as a Redis key


JamesMurkin:
This will need discussion. 

Main points of concern that I can see (probably more)
 - What should be use to convert Affinities to a key? I've used protobuf as its extremely compact + backwards compatible
   - We could write some custom code to do this
   - Json/message pack etc?
 - Should we assume a key can change over time?
   - This is mostly RE saving to a job to redis under its key. If the code that determines a key changes to generate the key differently, how should this work?
   -  IMO we should save the job key on job creation. From then on to work out the key, we load it from redis. Rather than calculate it again (sub-optimal but safe)
   -  Alternatively some reconciliation to keep keys "current" or make sure the key generate code never changes
- Is this too complicated?

severinson:
I have some concerns regarding this approach. Using the key field as data storage comes with several drawbacks, some of which you have already pointed out. That said, there seem to be some commonly used patterns in Redis for creating indices based on carefully constructed keys. Is this the approach you had in mind?
https://redis.io/topics/indexes

From what I can tell, using this approach, it should be possible to create efficient indices using keys with size in the order of hundreds of MBs.

JamesMurkin:
So not quite. The intent is to achieve effectively HTCondor autoclustering.

We'd be creating a secondary index, where the index contains all the jobs that look like X (same job requirements).

The field wouldn't be storage. It'd be a key, that a job belongs to:
 - i.e all jobs that request 1Cpu + 2Gi would have their job id in this index, as they have the job requirements key

It seems to be possible to create keys on the order of 100s of MB, however all recommendations I've seen is to keep them smaller
- Although we are probably debating 200 bytes vs 400 bytes

severinson:
I think we could achieve that using a Redis index, implemented as suggested on the link I posted. Does the following make sense to you?

Say that we represent the scheduling requirements of each job by a string, referred to here as the scheduling string, with the following format (we can add more fields if necessary): `memory_in_GB:number_of_cores:number_of_gpus:job_id`

Then, we can create an index of jobs with identical scheduling requirements by storing such strings in a sorted set, where all strings are assigned the same priority. For example, say that we insert the following keys:

- `key "0001:0010:0001:1234", priority: 0`
- `key "0001:0010:0001:1235", priority: 0`
- `key "0010:0010:0001:1236", priority: 0`
- `key "0010:0200:0001:1237", priority: 0`

Now, because keys with equal priority are sorted lexicographically, we can efficiently query all jobs for which the scheduling string has the same prefix (e.g., all jobs for which the scheduling string starts with `"0001:0010:0001:"`) using range queries.

We could store multiple such indexes with different format to be able to, e.g., query all jobs requiring the same number of GPUs, efficiently.

Regarding key size. The complexity of comparing two keys is linear in the length of the prefix common to both keys. However, we should only need to do `O(log n)` such comparisons to find the range of jobs that satisfy the query, and I would expect that comparing a few hundred bytes already stored in memory to be relatively fast compared to the time needed to transmit the query over the network.

severinson:
I think we could achieve that using a Redis index, implemented as suggested on the link I posted. Does the following make sense to you?

Say that we represent the scheduling requirements of each job by a string, referred to here as the scheduling string, with the following format (we can add more fields if necessary): `memory_in_GB:number_of_cores:number_of_gpus:job_id`

Then, we can create an index of jobs with identical scheduling requirements by storing such strings in a sorted set, where all strings are assigned the same priority. For example, say that we insert the following keys:

- `key "0001:0010:0001:1234", priority: 0`
- `key "0001:0010:0001:1235", priority: 0`
- `key "0010:0010:0001:1236", priority: 0`
- `key "0010:0200:0001:1237", priority: 0`

Now, because keys with equal priority are sorted lexicographically, we can efficiently query all jobs for which the scheduling string has the same prefix (e.g., all jobs for which the scheduling string starts with `"0001:0010:0001:"`) using range queries.

We could store multiple such indexes with different format to be able to, e.g., query all jobs requiring the same number of GPUs, efficiently.

Regarding key size. The complexity of comparing two keys is linear in the length of the prefix common to both keys. However, we should only need to do `O(log n)` such comparisons to find the range of jobs that satisfy the query, and I would expect that comparing a few hundred bytes already stored in memory to be relatively fast compared to the time needed to transmit the query over the network.

PR #823, opened 2021-12-10
Armada server comments, todos, and a few small cleanups

severinson:
This PR adds comments (docstrings, inside functions, and todos) to several files in `internal/armada`. The todo-comments are added to facilitate subsequent cleanup and refactoring work. There is also a small amount of cleanup (moving code around and changing whitespace).

### DIFF
--- a/internal/armada/repository/queue.go
+++ b/internal/armada/repository/queue.go
@@ -11,6 +11,20 @@ import (
 
 const queueHashKey = "Queue"
 
+// TODO These errors should be of a specific error type, e.g.:
+// type ErrQueueNotFound struct {
+// 	queueName string
+// }
+// func (err *ErrQueueNotFound) String() {
+// 	return fmt.Sprintf("Queue %s does not exist", err.queueName)
+// }
+// In doing so, if we wrap errors as they travel up the call stack
+// (using fmt.Errorf with the %w verb, i.e., fmt.Errorf("err %w", err))
+// and easily detect at the gRPC handler what the underlying issue is.
+// We need this since we need to create errors in the gRPC handler with the correct gRPC error code.
+// For more on error wrapping, see:
+// https://go.dev/blog/go1.13-errors
+
 var ErrQueueNotFound = errors.New("Queue does not exist")
 var ErrQueueAlreadyExists = errors.New("Queue already exists")
 
@@ -55,6 +69,7 @@ func (r *RedisQueueRepository) GetQueue(name string) (*api.Queue, error) {
 	} else if err != nil {
 		return nil, err
 	}
+
 	queue := &api.Queue{}
 	e := proto.Unmarshal([]byte(result), queue)
 	if e != nil {
--- a/internal/armada/repository/queue.go
+++ b/internal/armada/repository/queue.go
@@ -15,7 +15,7 @@ const queueHashKey = "Queue"
 // type ErrQueueNotFound struct {
 // 	queueName string
 // }
-// func (err *ErrQueueNotFound) String() {
+// func (err *ErrQueueNotFound) Error() {
 // 	return fmt.Sprintf("Queue %s does not exist", err.queueName)
 // }
 // In doing so, if we wrap errors as they travel up the call stack
--- a/internal/armada/server.go
+++ b/internal/armada/server.go
@@ -28,9 +28,16 @@ import (
 )
 
 func Serve(config *configuration.ArmadaConfig, healthChecks *health.MultiChecker) (func(), *sync.WaitGroup) {
+
+	// TODO Using an error group would be better.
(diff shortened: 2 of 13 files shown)


internal/armada/repository/queue.go
```diff
@@ -11,6 +11,20 @@ import (
 
 const queueHashKey = "Queue"
 
+// TODO These errors should be of a specific error type, e.g.:
+// type ErrQueueNotFound struct {
+// 	queueName string
+// }
+// func (err *ErrQueueNotFound) String() {
```
carlocamurri, line 18: I think the interface is `Error()`, not `String()`

severinson: You're right. That's an `Error()` on my part.

internal/armada/server.go
```diff
@@ -28,18 +28,30 @@ import (
 )
 
 func Serve(config *configuration.ArmadaConfig, healthChecks *health.MultiChecker) (func(), *sync.WaitGroup) {
+
+	// TODO Using an error group would be better.
+	// Since we want to shut down everything in a controlled manner in case of an irrecoverable error.
+	// Doesn't look like wg.Done() is called anywhere in this method.
```
carlocamurri, line 34: `wg.Done()` is called in `grpcCommon.Listen` below

severinson: Removed that part of the comment.

internal/armada/server.go
```diff
@@ -98,7 +113,9 @@ func Serve(config *configuration.ArmadaConfig, healthChecks *health.MultiChecker
 		jobStatusProcessor := repository.NewEventJobStatusProcessor(config.Events.JobStatusQueue, jobRepository, eventStream, jobStatusBatcher)
 		jobStatusProcessor.Start()
 
-		teardown = func() {
+		// TODO Teardown functions should return an error that can be logged/whatever by the caller.
```
d80tb7, line 114: I guess the question would be whether the caller could do anything useful with the error.

internal/armada/server.go
```diff
@@ -143,19 +163,24 @@ func Serve(config *configuration.ArmadaConfig, healthChecks *health.MultiChecker
 
 	grpc_prometheus.Register(grpcServer)
 
+	// TODO grpcCommon.Listen is supposed to call wg.Done() internally.
+	// Except, that it calls log.Fatalf first, so wg.Done() is never called.
```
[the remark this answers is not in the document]
d80tb7: ah I just stepped through the code. Log.Fatal exits there and then.  Wow, that's not what I would have expected.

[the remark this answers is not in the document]
severinson: `log.Fatal` calls `os.Exit(1)` internally, causing the program to exit immediately. For this reason, I think we should avoid using `log.Fatal`, with the possible exception of in the `main` function.
https://pkg.go.dev/log#Fatal

In this instance, closing the gRPC server with `Stop` or `GracefulStop` won't return an error, so the comment is not entirely accurate. However, if the server encounters an error, it will log at fatal, which causes the program to exit immediately without letting cleanup routines in other goroutines finish.

internal/armada/server.go
severinson: You can wrap errors on the way up, such that when you finally log an error and exit in `main` the logged error allows you to understand the sequence of function calls that caused the error. When calling `panic`, we lose all that information (`panic` is non-local return) and we may not be able to easily understand what sequence of calls were made.

d80tb7: yeah I wasn't arguing that it should panic, I was more meaning that this function could either a) return an error if it fails  or b) return void and log any errors that it encounters.  My assumption here is that this is just teardown code whereby if things go wrong, there's not much you can do other than continue on and assume you may have left some resources not cleaned up.  That being the case it *can* be better to choose b), especially if the caller can't do anything useful with the error,  although whether it is in this case her I'm less certain.

internal/armada/server.go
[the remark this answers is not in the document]
d80tb7: yeah that's insane.  I'd be tempted to  never use log fatal in that case!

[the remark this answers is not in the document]
severinson: Agreed :)

internal/armada/server.go
severinson: Ah, I see your point.

Something worth considering is integration with the `errgroups` package. If we run all services within an `ErrGroup` (as opposed to a `WaitGroup`), returning an error can trigger cancelling a context, which in turn can be used to trigger graceful shutdown of other services running in parallel. We do this by giving all services running in parallel a reference to the same context, which then acts as a channel for carrying error information between services. We could do this using both approach a) and b). I'll think about what would be best.
https://pkg.go.dev/golang.org/x/sync/errgroup

carlocamurri (approved):

PR #828, opened 2021-12-13
FIX warning CWE-117

theAntiYeti:


severinson:
Is there a reason for why we can't use the `%q` verb in `fmt.Sprintf` to safely escape user-provided strings?

theAntiYeti:
@severinson - CodeQL isn't recognising any attempt to rectify the bug. The "good example" provided by codeQL looks like this https://github.com/github/codeql-go/blob/main/ql/src/Security/CWE-117/LogInjectionGood.go, I have tried all but making it look exactly identical to this good example and will submit a bug report to them if this run fails.

severinson:
Regarding this PR, we're just waiting to hear back on the bug report you submitted for now?

theAntiYeti:
I merged a copy (+ some changes). I'm leaving this branch open for so the OpenQL people can have a look at the logs if they so wish.

PR #829, opened 2021-12-13
Armada server: improved error handling and messages

severinson:
This PR improves error handling and messages in `internal/armada/repository` and `internal/armada/server/handlers`. The end goal of this line of work is:

1. Making error messages clearer to aid in debugging errors and crashes.
2. Making it possible to extract detailed error information programatically, so that the server can return detailed error messages to the user in case of an error.

This PR does so through the following changes:

1. Functions/methods return error messages of the form `fmt.Errorf("[function_name] error doing x: %w", err)`. Note the `%w` verb, which wraps an error for use with `errors.As`.
2. Some repository functions now return special error types, e.g., `ErrQueueNotFound`.
3. gRPC handlers use `errors.As` to check what the underlying error is (e.g., if it is of type `ErrQueueNotFound`) to set the gRPC return code correctly.

By including the function name in the error message and wrapping errors as they travel up the call stack, and logging the error once it reaches the top-level (typically a gRPC handler), we can recover the sequence of function calls that led to the error. By wrapping errors using the `%w` verb, we can use `errors.As` to recover the source of the error programatically. See the following link for more info about this: https://go.dev/blog/go1.13-errors

 This PR also adds several comments (docstrings, explanations, and todos).


internal/armada/server/lease_test.go
```diff
@@ -199,15 +209,15 @@ func (repo *mockJobRepository) ReturnLease(clusterId string, jobId string) (retu
 	return nil, nil
 }
 
-func (repo *mockJobRepository) DeleteJobs(jobs []*api.Job) map[*api.Job]error {
+func (repo *mockJobRepository) DeleteJobs(jobs []*api.Job) (map[*api.Job]error, error) {
```
theAntiYeti, line 212: Was this changed to reflect the interface of DeleteJobs changing?

internal/armada/server/job_validation.go
```diff
@@ -11,7 +11,7 @@ func validateJobsCanBeScheduled(jobs []*api.Job, allClusterSchedulingInfo map[st
 	activeClusterSchedulingInfo := scheduling.FilterActiveClusterSchedulingInfoReports(allClusterSchedulingInfo)
 	for i, job := range jobs {
 		if !scheduling.MatchSchedulingRequirementsOnAnyCluster(job, activeClusterSchedulingInfo) {
-			return fmt.Errorf("job with index %d is not schedulable on any cluster", i)
+			return fmt.Errorf("[validateJobsCanBeScheduled] the %d-th job (job ID %s) can't be scheduled on any cluster", i, job.Id)
```
theAntiYeti, line 14: I like this

internal/armada/server/authorization.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	return strings.Join(reasons, ", ")
+}
+
+// checkReprioritizePerms checks if the principal embedded in the context has permission
+// to reprioritize all of the given jobs. This may require permissions for several queues.
+// If the principal has sufficient permissions, nil is returned. Otherwise an error is returned.
+func (server *SubmitServer) checkReprioritizePerms(ctx context.Context, jobs []*api.Job) error {
+
+	// Set of queues that has at least 1 of the given jobs
```
theAntiYeti, line 45: Comment not a tiny bit gratuitous? I'd get a second opinion but comments should be more about why you're doing something rather than what. The code itself is fairly clear

internal/armada/server/authorization.go
```diff
@@ -0,0 +1,158 @@
...
+	var e *repository.ErrQueueNotFound
+
+	// TODO Checking permissions shouldn't have side side effects.
+	// Hence, this function shouldn't automatically create queues.
+	// Further, we should consider removing the AutoCreateQueues option entirely.
+	// Since it leads to surprising behavior (e.g., creating a queue if the name is misspelled).
+	// Creating queues should always be an explicit decision.
+	// The less surprising behavior is to return ErrQueueNotFound (perhaps wrapped).
+	if errors.As(err, &e) && attemptToCreate && server.queueManagementConfig.AutoCreateQueues {
+		// TODO Is this correct? Shouldn't the relevant permission be permissions.CreateQueue?
```
theAntiYeti, line 98: Good question

internal/armada/repository/usage.go
```diff
@@ -75,43 +76,58 @@ func (r *RedisUsageRepository) GetClusterLeasedReports() (map[string]*api.Cluste
 func (r *RedisUsageRepository) GetClusterPriority(clusterId string) (map[string]float64, error) {
 	result, err := r.db.HGetAll(clusterPrioritiesPrefix + clusterId).Result()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("[RedisUsageRepository.GetClusterPriority] error reading from database: %s", err)
 	}
-	return toFloat64Map(result)
```
theAntiYeti, line 80: What's the story behind toFloat64Map returning an error as well? Did it always do this and we just ignored the error? Is this a thing that can commonly fail?

internal/armada/server/lease_test.go
severinson: Yes

internal/armada/server/job_validation.go
severinson: :)

internal/armada/repository/usage.go
severinson: It parses strings to floats, which can fail. I think that before we just didn't wrap it, i.e., the caller had to handle the underlying error directly.

internal/armada/server/authorization.go
severinson: I agree. I've updated the comment to say `If the principal has insufficient permissions, an error of type ErrNoPermission is returned.`.

internal/armada/server/authorization.go
JamesMurkin: This is the correct permission - for current usage. Not say it isn't a bit of an abuse of the roles.

However the use case is quite contrived and should probably be removed as stated in the above TODO comment - or be made more explicit. (i.e AutoCreateQueue permission, but that is pretty messy)

JamesMurkin: We'd need to migrate existing users off this first

internal/armada/server/job_validation.go
JamesMurkin: Nitpicks - This will create 1-th, 2-th and we typically use `job id or job Id`

internal/armada/server/submit.go
```diff
@@ -58,33 +57,44 @@ func NewSubmitServer(
...
-	queue, e := server.queueRepository.GetQueue(req.Name)
-	if e == repository.ErrQueueNotFound {
-		return nil, status.Errorf(codes.NotFound, "Queue %q not found", req.Name)
-
-	} else if e != nil {
-		return nil, status.Errorf(codes.Unavailable, "Could not load queue %q: %s", req.Name, e.Error())
+	queue, err := server.queueRepository.GetQueue(req.Name)
+	var e *repository.ErrQueueNotFound
+	if errors.As(err, &e) {
+		return nil, status.Errorf(codes.NotFound, "[GetQueue] error: %s", err)
```
JamesMurkin, line 84: I assume we've check this + other similar that the resulting error is similarly readable and useful

internal/armada/server/submit.go
```diff
@@ -219,137 +245,169 @@ func (server *SubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmitRe
...
+			return result, status.Errorf(codes.Unavailable, "[cancelJobsBySetAndQueue] error checking permissions: %s", err)
+		}
+		cancelledIds = append(cancelledIds, result.CancelledIds...)
 
-			if util.CloseToDeadline(ctx, time.Second*1) {
-				return &api.CancellationResult{CancelledIds: cancelledIds}, status.Errorf(
-					codes.DeadlineExceeded,
-					"not all jobs were cancelled: took too long")
-			}
+		// TODO I think the right way to do this is to include a timeout with the call to Redis
```
JamesMurkin, line 331: I think we need to update to a newer version of go-redis with this supported, however that indeed would be better.

internal/armada/server/submit.go
```diff
@@ -387,97 +454,46 @@ func (server *SubmitServer) reprioritizeJobs(jobIds []string, newPriority float6
...
 	if request.Queue == "" {
-		return nil, fmt.Errorf("queue is not specified")
+		return nil, fmt.Errorf("[createJobs] queue not specified")
 	}
 
 	for i, item := range request.JobRequestItems {
 
 		if item.PodSpec != nil && len(item.PodSpecs) > 0 {
-			return nil, fmt.Errorf("job with index %v has both pod spec and pod spec list specified", i)
+			return nil, fmt.Errorf("[createJobs] jobs must specify either a pod spec. or pod spec. list, but the %d-th job of job set %s specifies both", i, request.JobSetId)
```
JamesMurkin, line 487: Maybe we should say jobs must specify either `podSpec` or `podSpecs`. I find it confusing to read currently

JamesMurkin (dismissed):

internal/armada/server/job_validation.go
severinson: I updated the message to be `fmt.Errorf("[validateJobsCanBeScheduled] job %d out of %d with Id %s can't be scheduled on any cluster", i, len(jobs), job.Id)`.

internal/armada/repository/event.go
```diff
@@ -87,29 +88,27 @@ func (repo *RedisEventRepository) ReadEvents(queue, jobSetId string, lastId stri
...
 	messages := make([]*api.EventStreamMessage, 0)
 	for _, m := range cmd[0].Messages {
 		data := m.Values[dataKey]
 		msg := &api.EventMessage{}
 		bytes := []byte(data.(string))
-		e = proto.Unmarshal(bytes, msg)
-		if e != nil {
-			return nil, e
+		err = proto.Unmarshal(bytes, msg)
+		if err != nil { // TODO Should we log an error and keep going instead?
```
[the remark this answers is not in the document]
severinson: I see. Let's keep the current behavior then, i.e., propagate the error up the stack.

internal/armada/repository/event_nats.go
```diff
@@ -76,13 +77,14 @@ func (p *RedisEventProcessor) handleBatch(batch []*eventstream.Message) error {
...
 	if err != nil {
-		return fmt.Errorf("error while reporting events to event store for %d events: %v", len(events), err)
+		return fmt.Errorf("[handleBatch] error reporting %d events: %w", len(events), err)
 	}
 
 	for _, msg := range batch {
 		err = msg.Ack()
 		if err != nil {
-			log.Errorf("error while acknowledging event: %v", err)
+			err = fmt.Errorf("[handleBatch] error acknowledging event: %s", err)
```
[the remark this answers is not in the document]
severinson: Yes. Errors wrapped with `%w` effectively become part of the API since they can be extracted programmatically. Hence, I've avoided using `%w` to wrap errors coming from libraries, since those could change, which would change the API. Since this is internal code, I don't think it matters much, but that's at least the motivation :)

internal/armada/repository/job.go
```diff
@@ -69,6 +87,8 @@ func NewRedisJobRepository(
 	return &RedisJobRepository{db: db, retentionPolicy: retentionPolicy}
 }
 
+// TODO DuplicateDetected should be remove in favor of setting the error to
```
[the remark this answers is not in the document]
severinson: I see your point. I'll leave this for now. Agreed about the refactoring.

internal/armada/repository/job.go
```diff
@@ -525,87 +627,123 @@ func (repo *RedisJobRepository) updateJobs(ids []string, mutator func([]*api.Job
...
+	}
 
-func updateJobAndPriority(db redis.Cmdable, job *api.Job, newPriority float64, jobData *[]byte) *redis.Cmd {
-	return updateJobAndPriorityScript.Run(db,
-		[]string{jobQueuePrefix + job.Queue, jobObjectPrefix + job.Id},
-		job.Id, newPriority, *jobData)
+	// TODO I think we should return a more informative error message instead of logging a warning
+	// and returning a Redis error. I leave it for now since tests depend on getting a redis.TxFailedErr
+	log.Warnf("[RedisJobRepository.updateJobBatchWithRetry]: Redis Transaction failed after retrying, giving up (job ids %s)", strings.Join(ids, ", "))
+	return nil, redis.TxFailedErr
```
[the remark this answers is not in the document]
severinson: Yea, I think the batching system can be cleaned up significantly.

internal/armada/repository/job.go
```diff
@@ -873,6 +1036,7 @@ func addJob(db redis.Cmdable, job *api.Job, jobData *[]byte) *redis.Cmd {
 		job.Id, job.Priority, *jobData, job.ClientId)
 }
 
+// TODO This script will create the queue it it doesn't already exist.
```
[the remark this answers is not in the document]
severinson: I've elaborated a bit on this and made it a regular comment (i.e., removed the todo).

internal/armada/scheduling/lease.go
```diff
@@ -124,23 +131,27 @@ func calculateQueueSchedulingLimits(
 	return schedulingInfo
 }
 
+// TODO Do we need to log error messages here?
```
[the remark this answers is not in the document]
severinson: I updated the todo to reflect this.

internal/armada/server/submit.go
severinson: This should be ok. It'll be even better in a subsequent PR.

internal/armada/server/submit.go
severinson: Agreed. I changed it.

internal/armada/server/authorization.go
severinson: All right

JamesMurkin (approved):

PR #830, opened 2021-12-13
Add basic roadmap as a start

stackedsax:
I'd love some more eyes/thoughts on this before we merge it in.  I threw some high-level aspirational bullet points that I saw floating around and put some dates on them, but it's all completely arbitrary.

I'd love:
- Any more big-ticket items that people have in their minds
- Any suggestions as to how to more intelligently order these items into releases
- Input on the dates that I've arbitrarily set here

Thanks!

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -1,0 +1,15 @@
+# Armada Roadmap
+
+## v0.4 (Planned on Jul 1st, 2022)
+
+- First-class observability for admins
+- Admin tooling
+- SDN integration
+
+## v0.3 (Planned on May 1st, 2022)
+
+- Smart job prioritisation
+- First-class observability
+- New tech compatibility
+- Performance tuning / scaling / reliability
+- User tooling


severinson:
Is monthly point releases the right model for Armada? Since we control the release pipeline, we could make releases at whatever frequency/granularity we want.

Do you have more information about the features you've listed? As stated, they're not defined well enough to determine if they're done or not. Further, what's the reason for putting specific dates on a given feature? Are other parts of the business making decisions based on the promises we make?

stackedsax:
@severinson Are you suggesting monthly releases or did you think I was?  Rest assured, I don't have any underlying theory or logic to the dates I assigned here.  I was merely taking a convention that I've seen elsewhere and applying it arbitrarily to the short list of roadmap items I know we have on our minds.

We could take out the dates entirely for now, or perhaps change the months to quarters?  It might also be good to separate these items out into more distinct releases -- I'm open to all suggestions.

> Do you have more information about the features you've listed?

If you want to break these items down into smaller-but-still-roadmap-level-big items, by all means.  If you just want to add some more descriptions to each of these, that would be great, too.

I imagine we have more ideas than just this list, but this was all I found rummaging around Confluence.

andrewpcone (approved):
The log output should be sanitized, but that is not related to this PR, so we'll deal with it later

rankindaveyg (approved):

rankindaveyg (approved):

JamesMurkin:
Happy if we have something like this, but these are now out of date and the deadline has passed

PR #831, opened 2021-12-14
Fix cwe117

theAntiYeti:
Branched off from the previous branch to keep it alive for the CodeQL people to view logs, etc.
This is the "correct" solution, according to the [CodeQL codebase](https://github.com/github/codeql-go/blob/main/ql/src/Security/CWE-117/LogInjectionGood.go), so I intend to mark this as a false positive and merge in.

### DIFF
--- a/pkg/client/auth/oidc/device.go
+++ b/pkg/client/auth/oidc/device.go
@@ -146,7 +146,8 @@ func makeErrorForHTTPResponse(resp *http.Response) error {
 	if err != nil {
 		return err
 	}
-	safeURL := safeString(resp.Request.URL.String())
+	safeURL := strings.Replace(resp.Request.URL.String(), "\n", "", -1)
+	safeURL = strings.Replace(safeURL, "\r", "", -1)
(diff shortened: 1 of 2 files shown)


pkg/client/auth/oidc/device.go
```diff
@@ -145,5 +145,8 @@ func makeErrorForHTTPResponse(resp *http.Response) error {
 	if err != nil {
 		return err
 	}
-	return fmt.Errorf("%s %s returned HTTP %s; \n\n %#q", resp.Request.Method, resp.Request.URL, resp.Status, bodyBytes)
+	url := resp.Request.URL.String()
+	safeURL := strings.Replace(url, "\n", "", -1)
+	safeURL = strings.Replace(safeURL, "\r", "", -1)
+	return fmt.Errorf("%s %s returned HTTP %s; \n\n %#q", resp.Request.Method, safeURL, resp.Status, bodyBytes)
 }
```
robertdavidsmith, line 152: Suggest making a sanitizeUserInput function under https://github.com/G-Research/armada/tree/master/internal/common somewhere

theAntiYeti: of course

robertdavidsmith (approved):

PR #832, opened 2021-12-15
Revert "[issue-702] Add unit tests for queue based API (armada server)"

severinson:
Reverts G-Research/armada#789

After discussing with @robertdavidsmith and @carlocamurri , I suggest rolling back this PR. The reason being that the way of implementing API handlers introduced with PR 789 adds significant complexity, thus making it harder to understand the code and ensuring that we're implementing the API correctly. 

Further, we currently have two ways of implementing API handlers (the one used originally and the one introduced by PR 789, which is only used for parts of the API). Rolling back PR 789 makes the API handlers more consistent.


internal/armada/server/submit.go
```diff
@@ -59,41 +58,97 @@ func NewSubmitServer(
...
+	if e := checkPermission(server.permissions, ctx, permissions.DeleteQueue); e != nil {
+		return nil, e
+	}
+
+	active, e := server.jobRepository.GetQueueActiveJobSets(request.Name)
+	if e != nil {
+		return nil, status.Errorf(codes.InvalidArgument, e.Error())
+	}
+	if len(active) > 0 {
+		return nil, status.Errorf(codes.FailedPrecondition, "Queue is not empty.")
```
theAntiYeti, line 143: is there no way to put the name of the queue here?

severinson: I'm sure there is, but it may be cleaner to do that in a separate PR. So that this PR is only about reverting a prior change. For example, the following one.
https://github.com/G-Research/armada/pull/829

theAntiYeti (approved):
LGTM

PR #838, opened 2021-12-17
Resource limits validation

theAntiYeti:
Add validation for when jobsets submit resource requests/limits which are too low (usually by forgetting units)

This is done by adding a server minimum to the config list and validating that the resource fields are all submitted above said minimums. This was done to give more flexibility for future usecases and also because the first suggestion (validate people haven't forgotten units) was inflexible, could only be verified client-side (before numbers were parsed into resource.Quantity objects) and was potentially prone to bad maintenance (regex validating strings more prone to failing correct cases, missing incorrect cases, etc)


internal/common/validation/podspec.go
```diff
@@ -33,14 +33,36 @@ func ValidatePodSpec(spec *v1.PodSpec, maxAllowedSize uint) error {
...
+func validateContainerResource(
+	resourceSpec v1.ResourceList,
+	minJobResources v1.ResourceList,
+	container_name string,
+	requestOrLimit string,
+) error {
+	for rc, containerRsc := range resourceSpec {
+		serverRsc, nonEmpty := minJobResources[rc]
+		if nonEmpty && containerRsc.Value() < serverRsc.Value() {
+			return fmt.Errorf("%s %s (%s) on container %v below server minimum (%s)", requestOrLimit, rc, &containerRsc, container_name, &serverRsc)
```
severinson, line 60:
```suggestion
			return fmt.Errorf("[validateContainerResource] containers must request at least %s of resource %s, but container %q only requests %s", serverRsc.Value(), rc, container_name, containerRsc.Value())
```

severinson: I suggest updating the other error messages introduced by this PR similarly.

internal/common/validation/podspec.go
```diff
@@ -8,7 +8,7 @@ import (
 	"k8s.io/component-helpers/scheduling/corev1/nodeaffinity"
 )
 
-func ValidatePodSpec(spec *v1.PodSpec, maxAllowedSize uint) error {
+func ValidatePodSpec(spec *v1.PodSpec, maxAllowedSize uint, minJobResources v1.ResourceList) error {
```
severinson, line 11: Same comment as above. Should we pass in a larger object that contains all values needed to do the validation?

go.sum
```diff
@@ -1402,27 +1418,44 @@ honnef.co/go/tools v0.0.1-2020.1.3/go.mod h1:X/FiERA/W4tHapMX5mGpAtMSVEeEUOyHaw9
...
+k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b h1:wxEMGetGMur3J1xuGLQY7GEQYg9bZxKn3tKo5k/eYcs=
+k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b/go.mod h1:jPW/WVKK9YHAvNhRxK0md/EJ228hCsBRufyofKtW8HA=
 rsc.io/binaryregexp v0.2.0/go.mod h1:qTv7/COck+e2FymRvadv62gMdZztPaShugOCi3I+8D8=
 rsc.io/quote/v3 v3.1.0/go.mod h1:yEA65RcK8LyAZtP9Kv3t0HmxON59tX3rD+tICJqUlj0=
 rsc.io/sampler v1.3.0/go.mod h1:T1hPZKmBbMNahiBKFy5HrXp6adAjACjK9JXDnKaTXpA=
+sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6 h1:fD1pz4yfdADVNfFmcP2aBEtudwUQ1AlLnRBALr33v3s=
+sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6/go.mod h1:p4QtZmO4uMYipTQNzagwnNoseA6OxSUutVw05NhYDRs=
 sigs.k8s.io/structured-merge-diff/v4 v4.0.2/go.mod h1:bJZC9H9iH24zzfZ/41RGcq60oK1F7G282QMXDPYydCw=
 sigs.k8s.io/structured-merge-diff/v4 v4.1.2 h1:Hr/htKFmJEbtMgS/UD0N+gtgctAqz81t3nu+sPzynno=
 sigs.k8s.io/structured-merge-diff/v4 v4.1.2/go.mod h1:j/nl6xW8vLS49O8YvXW1ocPhZawJtm+Yrr7PPRQ0Vg4=
```
severinson, line 1461: This PR updates a lot of dependencies. What's going on?

severinson (changes requested):

internal/armadactl/app_test.go
```diff
@@ -229,10 +229,10 @@ jobs:
               - ls
             resources:
               limits:
-                memory: 1Mi
+                memory: 100Mi
                 cpu: 1
               requests:
-                memory: 1Mi
+                memory: 100Mi
```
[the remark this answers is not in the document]
theAntiYeti: I guess the default on development machines is irrelevant and I could completely remove the limit here. I'll just set it to 1Mi to leave it in the config (tacitly remind people that it's a possibility to change)

internal/armada/server/submit.go
```diff
@@ -432,7 +432,7 @@ func (server *SubmitServer) createJobs(request *api.JobSubmitRequest, owner stri
 
 		for j, podSpec := range item.GetAllPodSpecs() {
 			server.applyDefaultsToPodSpec(podSpec)
-			e := validation.ValidatePodSpec(podSpec, server.schedulingConfig.MaxPodSpecSizeBytes)
+			e := validation.ValidatePodSpec(podSpec, server.schedulingConfig.MaxPodSpecSizeBytes, server.schedulingConfig.MinJobResources)
```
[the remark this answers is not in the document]
theAntiYeti: Changing the tests will be a pain but I see what you mean here

go.sum
theAntiYeti: I'll see if I can get rid of these

internal/common/validation/podspec.go
severinson: Doesn't value return the amount of resource (e.g., 1000 if 1kb was requested)? Since these errors will be returned to users, I think we should make it as clear as possible what went wrong and what they should do to fix it. I think showing internal details of how we store limits etc. may confuse users, who shouldn't need to know those things.

theAntiYeti: ah, passing &serverRsc will actually call the .String() function and will print 1Ki rather than 1000. I'm unconvinced printing .Value() is a good idea when we'll definitely have limits/reqs in the Gi, maybe even Ti

go.sum
theAntiYeti: Yeah I think I know why this happened, I imported a package directly that I then ended up not using directly (the resource library). Have reverted

internal/common/validation/podspec.go
severinson: Ah. In that case I agree. Passing `&serverRsc` would be much better.

go.sum
severinson: Cool :)

internal/common/validation/podspec.go
theAntiYeti: On review, I completely disagree with your format. That looks nothing like any of the errors raised by ValidatePodSpec or podspec.go

theAntiYeti: To clarify, in all the errors we raise; none of them have an error type in square brackets a the start, and all of them respond in a very indicative format "<RESOURCE> X has done something" or "<RESOURCE> Y is something> .. this is why it's wrong". Your response is imperative/subjunctive in that it declares the ideal state of the world rather than starting with the current state of the world. I hope this makes sense

severinson (approved):

severinson (approved):
lgtm

PR #839, opened 2021-12-20
Add template for JobId in Job spec

theAntiYeti:
This change allows a job spec annotation or label to contain {JobId}, which gets templated out with the actual JobId on submission.

### DIFF
--- a/internal/armada/server/submit.go
+++ b/internal/armada/server/submit.go
@@ -456,6 +456,7 @@ func (server *SubmitServer) checkQueuePermission(
 
 // brought outside for mocking.
 var NewULID = util.NewULID
+var now = time.Now
 
 func (server *SubmitServer) createJobs(request *api.JobSubmitRequest, owner string, ownershipGroups []string) ([]*api.Job, error) {
 	jobs := make([]*api.Job, 0, len(request.JobRequestItems))
@@ -523,7 +524,7 @@ func (server *SubmitServer) createJobs(request *api.JobSubmitRequest, owner stri
 
 			PodSpec:                  item.PodSpec,
 			PodSpecs:                 item.PodSpecs,
-			Created:                  time.Now(),
+			Created:                  now(), // Replaced with now for mocking unit test
 			Owner:                    owner,
 			QueueOwnershipUserGroups: ownershipGroups,
 		}
--- a/internal/armada/server/submit.go
+++ b/internal/armada/server/submit.go
@@ -455,7 +455,7 @@ func (server *SubmitServer) checkQueuePermission(
 }
 
 // brought outside for mocking.
-var NewULID = util.NewULID
+var newULID = util.NewULID
 var now = time.Now
 
 func (server *SubmitServer) createJobs(request *api.JobSubmitRequest, owner string, ownershipGroups []string) ([]*api.Job, error) {
@@ -505,7 +505,7 @@ func (server *SubmitServer) createJobs(request *api.JobSubmitRequest, owner stri
 			}
 		}
 
-		jobId := NewULID()
+		jobId := newULID()
 		j := &api.Job{
 			Id:       jobId,
 			ClientId: item.ClientId,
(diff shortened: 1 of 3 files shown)


internal/armada/server/submit.go
```diff
@@ -454,6 +454,10 @@ func (server *SubmitServer) checkQueuePermission(
 	return nil, groups
 }
 
+// brought outside for mocking.
+var NewULID = util.NewULID
```
samclark, line 458: inconsistent exports

internal/armada/server/submit.go
```diff
@@ -529,6 +534,16 @@ func (server *SubmitServer) createJobs(request *api.JobSubmitRequest, owner stri
 	return jobs, nil
 }
 
+func enrichText(text map[string]string, jobId string) map[string]string {
+	returnText := util.DeepCopy(text)
```
samclark, line 538: do we need to deep copy? can we not just replace in-place?

internal/armada/server/submit.go
```diff
@@ -529,6 +534,16 @@ func (server *SubmitServer) createJobs(request *api.JobSubmitRequest, owner stri
...
+
+	for key, value := range returnText {
+		returnText[key] = strings.Replace(value, "{JobId}", jobId, -1)
```
samclark, line 541: although unlikely there is no way to escape `{JobId}`. I think we should provide a way to escape this value incase users want to use it literally.

internal/armada/server/submit_test.go
```diff
@@ -727,3 +727,114 @@ func withSubmitServerAndRepos(action func(s *SubmitServer, jobRepo repository.Jo
...
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	owner := "test"
```
samclark, line 829: why have a separate variable for `owner` here?

internal/armada/server/submit.go
JamesMurkin, line 541: Why not use ReplaceAll?

internal/armada/server/submit.go
JamesMurkin, line 458: I don't love this style. Feels like hack and messes up the code for the sake of testing.

Considering createJobs could be extracted out of submit entirely. Possibly it could be some struct should handle createJobs, it can then be configured with a uuidFunc and timeFunc. 

At the very least, it shouldn't be exported

internal/armada/server/submit_test.go
```diff
@@ -727,3 +727,114 @@ func withSubmitServerAndRepos(action func(s *SubmitServer, jobRepo repository.Jo
...
+	ownershipGroups := make([]string, 0)
+	withSubmitServer(func(s *SubmitServer, events repository.EventRepository) {
+		output, err := s.createJobs(request, owner, ownershipGroups)
+		fmt.Println(err)
```
JamesMurkin, line 833: Should we not test that there is NoError rather than just printing?

theAntiYeti: Oh that's a massive catch - I only put that in to debug, wasn't going to test that at all but this makes more sense

internal/armada/server/submit.go
theAntiYeti: fixed the exporting. Going to leave this like this for today at least and maybe come back to it tomorrow

internal/armada/server/submit.go
theAntiYeti: we probably could. Something about making non mutative functions generally feels better to me but I guess this isn't that important or used again?

internal/armada/server/submit.go
theAntiYeti: I don't believe {} are allowed in labels or annotations anyway are they?

samclark: I've not tested but I suspect almost anything will be allowed in annotation values.

theAntiYeti: I'm still not entirely convinced this isn't just a YAGNI, if you have any concrete example of when someone would want to have `{JobId}` in their annotation then I'll add it but it seems more than feasible they could use `{Job_Id}` or some other variation instead. What is the use case for having the escaped string there anyway?

samclark: Annotations by their nature are very flexible and it's just good practice as the platform to not get in the way. Escaping the magic token surely can't add much complexity to the code and providing a way to escape it will be expected and intuitive to users.

samclark (approved):
LGTM

PR #840, opened 2021-12-26
[feature/issues/816] Create queue using configuration files

steffnova:
[Problem]

A more configurable way of creating queues is required that allows an
explicit way of giving queue permissions. Current gives all permissions
to queue's user and group owners.

Users should be able to assign list of permissions per user or group.

[Solution]

Add queue configuration file written in yaml:

```yaml
apiVersion: armadaproject.io/v1
kind: Queue
name: queue-a
permissions:
- subjects:
  - name: group
    kind: Group
  verbs:
  - cancel
  - reprioritize
  - watch
priorityFactor: 3.0
resourceLimits:
  cpu: 1.0
  memory: 1.0
```

Name, priorityFactor and resourceLimits are existing queue configuration
options. Permissions is a new way of defining granular permissions. Each
permission has subjects and verbs. Each subject is defined by name and it's
kind (allowed values: Group and User). Verbs define which actions a subject
can execute. Supported verbs: "submit", "cancel", "reprioritize", "watch"

Add new command `create -f <file_path>`. It creates a new resource (only
queue is supported for now) using a configuration file. Command has
-dry-run option, where file validation is performed only, without resource
creation.

Add new Queue structure that can be translated from and into a api.Queue
structure. This allows backward compatibility with existing queues and
translation of queue's groups and owners to new permissions abastraction.

Add strict validation of queue attributes for: resourceLimits, priorityFactor
and permissions.

Update all armada server handlers to use queue's permissions to authorize
queue specific request.

Add example queue configuration file

Update quick-start documentation with an option to create queues with
configuration files.

[Tests]

Add unit test for new codebase and update existing tests.
Manual testings

Close #840


severinson:
I started looking through this. I have a few questions and I think it'll be more efficient to discuss than to post them as review comments. How about a call sometime next week?

internal/armadactl/queue.go
```diff
@@ -4,22 +4,42 @@ import (
...
+		return fmt.Errorf("[armadactl.CreateQueue] error creating queue %s: %s", queue.Name, err)
+	}
+	fmt.Fprintf(a.Out, "Created queue %s\n", queue.Name)
+	return nil
+}
+
+func (a *App) CreateResource(fileName string, dryRun bool) error {
+	var resource client.Resource
+	if err := util.BindJsonOrYaml(fileName, &resource); err != nil {
+		return fmt.Errorf("file %s error :%s", fileName, err)
```
severinson, line 24:
```suggestion
		return fmt.Errorf("file %s error: %s", fileName, err)
```

severinson: Will this produce an error message that is nicely formatted and easily understood by the user? If not, I think we should change it. Same comment for the other error messages in this function.

pkg/client/queue/priority_factor.go
```diff
@@ -0,0 +1,48 @@
+package queue
+
+import (
+	"encoding/json"
+	"fmt"
+	"math/rand"
+	"reflect"
+)
+
+type PriorityFactor float64
```
severinson, line 10: It seems to me that this adds unnecessary complexity. Why do we need to introduce a custom type for a priority factor instead of using `float64`?

pkg/client/queue/queue.go
```diff
@@ -0,0 +1,103 @@
+package queue
```
severinson, line 1: Why do we need a second queue type in addition to the one in the `api` package? Adding new types, files, and code carries a significant cost in terms of system complexity and ease of understanding.

severinson (commented):
I've added a few questions and comments. There are several things I think warrant some discussion.

steffnova:
> I started looking through this. I have a few questions and I think it'll be more efficient to discuss than to post them as review comments. How about a call sometime next week?

@severinson sure. I won't be available on 6th and 7th of January.

pkg/client/queue/priority_factor.go
steffnova: Float64 means that any float64 value is a valid priority factor value, which isn't the case. Priority factor has to be greater than or equal to 1.0. ([reference](https://github.com/G-Research/armada/blob/2d63f9dd154f3462c1f426b28954fdf5619840cf/internal/armada/server/submit.go#L126-L131))

docs/quickstart/queue-a.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: armadaproject.io/v1
+kind: Queue
+name: queue-a
+permissions:
```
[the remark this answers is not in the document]
steffnova: @severinson it was mine concern as well. I've had a discussion with @robertdavidsmith and @JamesMurkin and for current requirements it was concluded that this should suffice.

pkg/client/queue/priority_factor.go
severinson: Sure, but that could be checked when validating the queue struct itself. Unless I'm mistaken, creating a new queue with `Queue{}` would result in a queue for which the priority factor is `0` (the default value for a `float64`), so we need to validate the queue struct regardless of if the priority factor is a `float64` or a custom type.

docs/quickstart/queue-a.yaml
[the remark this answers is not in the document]
severinson: All right. I'm still concerned since the file format becomes part of Armada's API. If other systems, or worse, customers (although I don't think we have customers that can create queues?) become dependent on it, changing it may be difficult. If we're merging this as a temporary measure, I suggest adding comments to the related `.proto` and `.go` files saying something like `This API is not considered stable and may be changed without warning at any time.`

pkg/client/queue/priority_factor.go
steffnova: You can always instantiate invalid data structure even in standard library, Go language doesn't prevent you from improper usage.

What this does ensure, is that float64 value cannot be assigned to PriorityFactor as they are not the same type. To Generate PriorityFactor from float64 `NewPriorityFactor` should be used instead.

pkg/client/queue/queue.go
steffnova: Here is my reasoning process:
1. api package is auto generated gRPC package. It defines request and response types used in gRPC protocol. It also doesn't feel right of adding methods to autogenerated types. Armada server business logic (repository abstraction for example) is coupled to autogenerated types. This is an odd design by itself, as API directly exposes and is coupled to public available data structures.  Any changes to API indirectly changes the repo layer and vice-versa.
2. Introducing new package decouples public api from business and repo layers.
3. Any business logic should depend on this `queue` package as it defines a valid source of truth for Queue data. If used properly, any queue from this package can be considered valid (valid values for priority factor, resource limits etc...)
4. The changes in this pull request redefine how queue data structure looks like. User owners and  Group owners have been removed and have been replaced with permissions, subjects and verbs. The new type ensures transformation between two version of Queue. These transformation are required for gRPC request and from reading data from repositories.

pkg/client/queue/queue.go
```diff
@@ -0,0 +1,103 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	return false
+}
+
+type Queues []Queue
```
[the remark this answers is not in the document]
steffnova: This is a very common Go pattern. Especially if there is a need to have []T (t is a arbitrary type) as a method receiver.
In this case converting []queue.Queue to []*api.Queue.

docs/quickstart/queue-a.yaml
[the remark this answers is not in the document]
steffnova: The main feature introduced here is adding ability to create any Armada resource through `armadactl create -f file_path` command. Old command for creating queues stays the same, and all existing functionalities have unchanged behaviour. It expands current owners system to be more granular when needed. Regarding stability I'm not sure how it should be treated. @robertdavidsmith @JamesMurkin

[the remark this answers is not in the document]
severinson: Just saying it's a nice habit to add a comment to parts of the API that we don't consider to be stable.

severinson (approved):

PR #841, opened 2021-12-29
Consider cordoned resource when calculating fair share

theAntiYeti:
When calculating the share of capacity to allocate queues, the total resource to allocate doesn't consider the resource taken up by jobs running on cordoned nodes but the resource allocated to queues does. This leads to a situation where a queue with can have more resource running on cordoned nodes than the server physically believes allocable, leading to new jobs not being scheduled even when the cluster has plenty of spare resource on schedulable nodes.

This PR aims to add the resource taken on cordoned nodes to the total allocable resource.


internal/executor/utilisation/cluster_utilisation.go
```diff
@@ -165,24 +172,63 @@ func (clusterUtilisationService *ClusterUtilisationService) GetAllNodeGroupAlloc
...
+			usage.Add(containerResource)
+		}
+	}
+	fmt.Printf("%v\n", usage)
+	return usage
+}
+
+// castAsComputeResources casts a ResourceList type as common.ComputeResources.
+// Both are aliases for map[string]resource.Quantity.
+// This feels dirty, there must be a way to directly cast them (remove this line or replace this function)
```
severinson, line 221: Perhaps just remove the one in common? If the right abstraction exists in the Kubernetes API, we may not need to create our own version. I suggest adding TODO to the start of the sentence and putting a suggested fix into the comment.

pkg/api/usage.pb.go
```diff
@@ -747,6 +758,30 @@ func (m *NodeTypeUsageReport) MarshalToSizedBuffer(dAtA []byte) (int, error) {
 	_ = i
 	var l int
 	_ = l
+	if len(m.CordonedUsage) > 0 {
```
severinson, line 761: This could be cleaned up. Perhaps split into a separate function :)

internal/executor/utilisation/cluster_utilisation.go
theAntiYeti: As far as I can tell you, the reason we do this is because in go you can't extend structs outside the package they're defined in.

We have a good number of methods we put on top of common.ComputeResources, for instance termwise, mutative addition and subtraction (https://github.com/G-Research/armada/blob/master/internal/common/resource.go).

Maybe the question we should be asking is why we don't make these normal functions (maybe make them pure functions rather than modifying the subject as well) on map[string]resource.Quantity? and then we *should* be able to cast that directly to v1.ResourceList. That would be cleaner to my understanding of golang wanting methods only to be for implementing interfaces (which we don't actually do in the case of common.ComputeResources).

internal/common/util/cluster_usage_report.go
```diff
@@ -5,6 +5,8 @@ import (
 	"github.com/G-Research/armada/pkg/api"
 )
 
+// GetClusterCapaicty returns the total capacity on all nodes on a cluster,
```
d80tb7, line 8: nit: spelling

internal/executor/utilisation/cluster_utilisation.go
```diff
@@ -165,24 +172,63 @@ func (clusterUtilisationService *ClusterUtilisationService) GetAllNodeGroupAlloc
 		return []*NodeGroupAllocationInfo{}, err
 	}
 
+	batchPods, err := clusterUtilisationService.clusterContext.GetBatchPods()
+	if err != nil {
+		// Failing this step will just mean cordoned utilisation won't be accounted for
+		// This isn't a critical failure so we log and continue
+		log.Errorf("Error fetching batch pods to calculate cordond node utilisation: %v", err)
```
d80tb7, line 179: should we explicitly set `batchPods` to the empty list here so that we're not taking a value from a function that has errored?

internal/executor/utilisation/cluster_utilisation.go
```diff
@@ -165,24 +172,63 @@ func (clusterUtilisationService *ClusterUtilisationService) GetAllNodeGroupAlloc
...
+	cordonedNodes := util.FilterNodes(nodes, func(node *v1.Node) bool { return node.Spec.Unschedulable })
+	podsOnNodes := GetPodsOnNodes(pods, cordonedNodes)
+	usage := common.ComputeResources{}
+	for _, pod := range podsOnNodes {
+		for _, container := range pod.Spec.Containers {
+			containerResource := castAsComputeResources(container.Resources.Limits) // Not 100% on whether this should be Requests or Limits
+			usage.Add(containerResource)
+		}
+	}
+	fmt.Printf("%v\n", usage)
```
d80tb7, line 215: did you mean to leave this in?

internal/executor/utilisation/cluster_utilisation.go
d80tb7: So I think this is just `common.ComputeResources.FromResourceList` isn't it?  You can probably delete this function and just use that.

As an aside, what you're doing here isn't really a cast, but rather a conversion.

internal/executor/utilisation/cluster_utilisation.go
```diff
@@ -165,24 +172,63 @@ func (clusterUtilisationService *ClusterUtilisationService) GetAllNodeGroupAlloc
...
+// to pods running on cordoned nodes. We need this information in calculating queue fair shares when
+// significant resource is running on cordoned nodes.
+func (clusterUtilisationService *ClusterUtilisationService) getCordonedResource(nodes []*v1.Node, pods []*v1.Pod) common.ComputeResources {
+	cordonedNodes := util.FilterNodes(nodes, func(node *v1.Node) bool { return node.Spec.Unschedulable })
+	podsOnNodes := GetPodsOnNodes(pods, cordonedNodes)
+	usage := common.ComputeResources{}
+	for _, pod := range podsOnNodes {
+		for _, container := range pod.Spec.Containers {
+			containerResource := castAsComputeResources(container.Resources.Limits) // Not 100% on whether this should be Requests or Limits
+			usage.Add(containerResource)
```
d80tb7, line 212: so I don't know nearly enough about k8s here, but there' no risk of overwriting keys here is there?

d80tb7:
I notice we have no tests for this new functionality.  Is there any way we can add some?

internal/executor/utilisation/cluster_utilisation.go
theAntiYeti: I'm fairly certain there is not, especially with the change "castAsComputeResources->FromResourceList" which explicitly copies containers.

I'm not entirely sure what you're thinking would override keys or what keys.

theAntiYeti:
@d80tb7 I'd be very surprised if any of these functions could be unit tested nicely. I'll look into it but this will definitely be a pain.

internal/executor/utilisation/cluster_utilisation.go
d80tb7: so  `usage` here is essentially a map of `string` to `Resource.Quantity`.  I was wondering if calling `Add` on the map would do the correct thing in terms of adding keys together rather than replacing previously added keys with the latest one. I checked the code and it does do the correct thing so I think we're good.

d80tb7:
so both `cluster_utilisation_test.go` and (to a lesser extent) `job_utilisation_reporter_test.go` look to have suitable places where tests can be defined.  If you're finding it difficult to add tests then let me know and I'll have a look.  If we find places in the codebase that are hard to test, then we should deal with this as it generally indicates an issue.

internal/executor/node/node_group.go
```diff
@@ -100,6 +101,10 @@ func (kubernetesNodeInfoService *KubernetesNodeInfoService) GetAllAvailableProce
 	return util2.FilterNodes(allNodes, kubernetesNodeInfoService.isAvailableProcessingNode), nil
 }
 
+func (kubernetesNodeInfoService *KubernetesNodeInfoService) GetAllProcessingNodes() ([]*v1.Node, error) {
```
JamesMurkin, line 104: Possibly this should be "GetAllNodes"

Processing nodes implies we filter it in someway. 
 - i.e We'd include all with taints that are in kubernetesNodeInfoService.toleratedTaints

internal/executor/utilisation/cluster_utilisation.go
```diff
@@ -165,24 +172,52 @@ func (clusterUtilisationService *ClusterUtilisationService) GetAllNodeGroupAlloc
 		return []*NodeGroupAllocationInfo{}, err
 	}
 
+	batchPods, err := clusterUtilisationService.clusterContext.GetBatchPods()
+	if err != nil {
+		// Failing this step will just mean cordoned utilisation won't be accounted for
+		// This isn't a critical failure so we log and continue
+		log.Errorf("Error fetching batch pods to calculate cordond node utilisation: %v", err)
+		batchPods = []*v1.Pod{}
```
JamesMurkin, line 180: Probably simpler to just return error in the case that call fails. If it fails it is probably indicative of a wider problem.

If you have a fair chunk of the cluster cordoned, it could be quite wrong to not report cordoned. Simpler to just have it all succeed or all fail, rather than have partial failures.

JamesMurkin:
Probably we should update https://github.com/G-Research/armada/blob/master/internal/executor/metrics/pod_metrics/cluster_context.go#L219 to reflect cordoned + allocatable

theAntiYeti:
> Probably we should update https://github.com/G-Research/armada/blob/master/internal/executor/metrics/pod_metrics/cluster_context.go#L219 to reflect cordoned + allocatable

WDYM?

d80tb7 (commented):
lgtm

d80tb7 (approved):

PR #844, opened 2022-01-07
adding interceptors for embedding ids in gRPC requests

severinson:
This PR adds interceptors for embedding ids in gRPC requests in a manner that is compatible with, e.g., OpenTelemetry. The ids are embedded in the gRPC metadata, which is part of the request context.


d80tb7 (approved):
lgtm

PR #846, opened 2022-01-10
Adding generic error types and interceptors for automatically setting return codes and adding a request id to error messages

severinson:
This PR adds a new package `internal/common/armadaerrors` with:

- Generic error types for "not found", "already exists", and "invalid argument" errors.
- gRPC interceptors that look for these errors and set the gRPC return code accordingly. These interceptors also insert a request id into the returned error message if one is present.

This functionality is not yet used by any part of Armada.


internal/common/armadaerrors/errors_test.go
```diff
@@ -0,0 +1,155 @@
...
+				t.Fatalf("expected %v but got %v", tc.want, got)
+			}
+		})
+	}
+}
+
+func TestUnaryServerInterceptor(t *testing.T) {
+	ctx := context.Background()
+	ctx = metadata.NewIncomingContext(ctx, metadata.New(map[string]string{}))
+	// ctx = requestid.AddToIncomingContext(ctx, "123")
```
JamesMurkin, line 46: Commented line, can we remove?

internal/common/armadaerrors/errors_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+
+	"github.com/G-Research/armada/internal/common/requestid"
+)
+
+func TestCodeFromError(t *testing.T) {
+	tests := map[string]struct {
```
JamesMurkin, line 19: Does this actually test where we skip if the error is already a status error?

internal/common/armadaerrors/errors_test.go
```diff
@@ -0,0 +1,155 @@
...
+		"pkg.Error => ErrInvalidArgument": {errors.WithMessage(&ErrInvalidArgument{}, "foo"), codes.InvalidArgument},
+		"pkg.Error":                       {errors.New("foo"), codes.Unknown},
+		"nil":                             {nil, codes.OK},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			got := CodeFromError(tc.err)
+			if got != tc.want {
+				t.Fatalf("expected %v but got %v", tc.want, got)
```
JamesMurkin, line 37: Style 

Most of the project would use:
`assert.Equal(t, tc.want, got, "expected %v but got %v", tc.want, got)`

Any reason we are swapping from that?

internal/common/armadaerrors/errors_test.go
severinson: Yes

internal/common/armadaerrors/errors_test.go
severinson: We do now :)

internal/common/armadaerrors/errors_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"strings"
+	"testing"
+
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
+	"github.com/pkg/errors"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+	"gotest.tools/v3/assert"
```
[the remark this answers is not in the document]
severinson: Right. Fixed.

JamesMurkin (approved):

PR #858, opened 2022-01-27
Find armadactl config in the same folder was kubectl

JamesMurkin:
We recently changed the way we find the homedir when looking for `.armadactl.yaml` to the user home dir

The issue with this is it can be quite different to the way kubectl finds its configuration

Swapping back to what we used to use, as this is much more consistent with kubectl
 - This change is pretty much a noop for linux

### DIFF
--- a/pkg/client/command_line.go
+++ b/pkg/client/command_line.go
@@ -8,6 +8,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"gopkg.in/yaml.v2"
+	"k8s.io/client-go/util/homedir"
 
 	"github.com/G-Research/armada/internal/fileutils"
 )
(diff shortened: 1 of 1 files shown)


pkg/client/command_line.go
```diff
@@ -77,9 +78,9 @@ func LoadCommandlineArgsFromConfigFile(cfgFile string) error {
 		}
 		viper.SetConfigFile(cfgFile)
 	} else {
-		homeDir, err := os.UserHomeDir()
-		if err != nil {
-			return fmt.Errorf("[LoadCommandlineArgsFromConfigFile] error getting user home directory: %s", err)
+		homeDir, err := homedir.Dir()
+		if homeDir == "" {
```
carlocamurri, line 82:
```suggestion
		if err != nil {
```

carlocamurri (approved):

PR #860, opened 2022-01-28
Make deserializing Service/Ingress types more strict

JamesMurkin:
Currently if you submit an Service/Ingress type which is an int but invalid, it just uses the first type

I.e if you say I want ServiceType with value 100, you get NodePort (as NodePort is the first ServiceType)
 - Even though value 100 is invalid, as the only options are 0 and 1

This will prevent users relying on invalid numbers that may at some point become used and also it is probably just good practice to not support invalid values and default them
 - Now the user will get an api error back immediately


pkg/api/deserialize.go
```diff
@@ -9,6 +9,10 @@ func (x *IngressType) UnmarshalJSON(data []byte) error {
 	var s int32
 	e := json.Unmarshal(data, &s)
 	if e == nil {
+		_, present := IngressType_name[s]
```
kannon92, line 12: I'll add locks just in case.

pkg/api/deserialize.go
```diff
@@ -29,6 +33,10 @@ func (x *ServiceType) UnmarshalJSON(data []byte) error {
 	var s int32
 	e := json.Unmarshal(data, &s)
 	if e == nil {
+		_, present := ServiceType_name[s]
```
kannon92, line 36: Same here.

jimbobby5:

kannon92:
@JamesMurkin and I decided to get this one merged.  So I took it over and the review is active now.

pkg/api/deserialize_test.go
```diff
@@ -7,6 +7,62 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestMarshalJsonIngressType(t *testing.T) {
+	tests := map[string]IngressType{
```
ClifHouck, line 11: It looks like all these tests are running a set of test cases. Would you be willing to restructure these so they use subtests? They offer some benefits, see:  https://go.dev/blog/subtests . Example in our codebase:  https://github.com/G-Research/armada/blob/5c78c6bec2802738852f434328cde183f199a2a3/internal/lookout/repository/utils_test.go#L262

kannon92: Sure thing!

pkg/api/deserialize_test.go
```diff
@@ -2,21 +2,114 @@ package api
...
+
+	for _, input := range testInputsStrings {
+		t.Run(fmt.Sprintf("TestDeserializeJobStateString(%s)", input),
+			func(t *testing.T) {
+				assert.NoError(t, testDeserializeJobStateString(t, input))
+			})
+	}
 
 	assert.NoError(t, testDeserializeJobStateInt(t, 0))
 	assert.NoError(t, testDeserializeJobStateInt(t, 1))
```
kannon92, line 94: Remove this code.

ClifHouck (approved):
LGTM

PR #866, opened 2022-02-07
updated macos x instructions

### LINKED ISSUE #865: Quickstart instructions for Mac OS X don't work
Two problems:

- The syntax `brew cask install <x>` has been replaced with `brew install --cask <x>`, but the instructions use the former, so they don't wokr.
- You don't actually get the `docker` CLI executable unless you open `/Applications/Docker.app`, so that needs to be included explicitly

andrewpcone:
Closes #865


robertdavidsmith (changes requested):
Hi Andrew,

Thanks for the PR.

The change to quickstart.md looks great, thanks. I'm unsure about the change to kind-config-executor.yaml though. Was this committed by mistake?

Thanks,

Rob

PR #888, opened 2022-03-03
Switch to standalone cobra-cli dependency

liggitt:
Switches to use the extracted cobra-cli command since it is being removed from the github.com/spf13/cobra module (xref https://github.com/spf13/cobra/issues/1597)

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -222,7 +222,7 @@ You can use the cobra cli to add new commands, the below will describe how to ad
 Get cobra cli tool:
 
 ```
-go get -u github.com/spf13/cobra/cobra
+go get -u github.com/spf13/cobra-cli
 ```
 
 Change to the directory of the command-line tool you are working on:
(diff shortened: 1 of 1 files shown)


suprjinx:
afaict, "cobra-cli" is a bootstrapping module for initializing a new cobra/command line app. Ie, we wouldn't use this as a replacement for cobra.

suprjinx:
I do notice that the goal of the cobra PR is to remove viper dependencies -- but we in fact import viper separately anyway so there won't be any reduction in the dependency list.

suprjinx (approved):
LGTM assuming the updated instructions work

kannon92:
Hello @liggitt! Sorry we missed this PR.  

Are you open to fixing the merge conflicts that have appeared?  i think the change looks good but it needs some minor love.

liggitt:
actually, it looks like the cobra cli reference was dropped in https://github.com/G-Research/armada/pull/1692/files#diff-f1ea07149d825024a80119f7f8d068cd154c8780a250b03c6544c68f5bd5261fL179, so this PR isn't needed any more

PR #931, opened 2022-04-28
Log job submit API deduplication

severinson:
This PR adds job submission deduplication to the log submit API. It also adds a key-value store backed by Postgres, which is used for the deduplication.

### DIFF
--- a/e2e/pulsar_test/pulsar_test.go
+++ b/e2e/pulsar_test/pulsar_test.go
@@ -135,6 +135,12 @@ func TestDedup(t *testing.T) {
 			return nil
 		}
 
+		numEventsExpected := len(res.JobResponseItems) * 6
+		_, err = receiveJobSetSequences(ctx, consumer, armadaQueueName, req.JobSetId, numEventsExpected, 10*time.Second)
+		if err != nil {
+			return err
+		}
+
 		// The second time, all jobs should be rejected.
 		req = createJobSubmitRequestWithClientId(numJobs, clientId)
 		ctxWithTimeout, _ = context.WithTimeout(context.Background(), time.Second)
@@ -145,6 +151,12 @@ func TestDedup(t *testing.T) {
 
 		if ok := assert.Equal(t, 0, len(res.JobResponseItems)); !ok {
 			return nil
+		}
+
+		numEventsExpected = len(res.JobResponseItems) * 6
+		_, err = receiveJobSetSequences(ctx, consumer, armadaQueueName, req.JobSetId, numEventsExpected, 10*time.Second)
+		if err != nil {
+			return err
 		}
 
 		// Check for partial rejection
@@ -159,6 +171,12 @@ func TestDedup(t *testing.T) {
 
 		if ok := assert.Equal(t, numJobs, len(res.JobResponseItems)); !ok {
 			return nil
+		}
+
+		numEventsExpected = len(res.JobResponseItems) * 6
+		_, err = receiveJobSetSequences(ctx, consumer, armadaQueueName, req.JobSetId, numEventsExpected, 10*time.Second)
+		if err != nil {
+			return err
 		}
 
 		return nil
--- a/e2e/pulsar_test/pulsar_test.go
+++ b/e2e/pulsar_test/pulsar_test.go
@@ -124,7 +124,7 @@ func TestDedup(t *testing.T) {
 		clientId := uuid.New().String()
 		originalJobIds := make([]string, numJobs)
 
-		// The first time, all jobs should be submitted.
+		// The first time, all jobs should be submitted as-is.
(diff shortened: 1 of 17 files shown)


internal/armada/server.go
```diff
@@ -196,6 +198,27 @@ func Serve(config *configuration.ArmadaConfig, healthChecks *health.MultiChecker
 		}
 		submitServerToRegister = pulsarSubmitServer
 
+		// If postgres details were provided, enable deduplication.
+		log.Infof("DedupTable: %s\n", config.Pulsar.DedupTable)
+		log.Infof("Postgres: %+v\n", config.Pulsar.Postgres)
```
d80tb7, line 203: be careful here else you might end up logging postgres passwords.

internal/armada/server/submit_to_log.go
```diff
@@ -158,21 +170,74 @@ func (srv *PulsarSubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmi
...
 
+// removeDuplicateSubmissions filters out job submit request items with previously seen client ids.
+// Client ids are namespaced by queue.
+func (srv *PulsarSubmitServer) removeDuplicateSubmissions(ctx context.Context, queue string, items []*api.JobSubmitRequestItem) ([]*api.JobSubmitRequestItem, error) {
+
+	// For storage efficiency, we store hashes instead of user-provided strings.
+	// For computational efficiency, we create a lookup table to avoid computing the same hash twice.
+	// Client ids are namespaced by queue. Hence, we hash the client id together with the queue.
+	jobSubmitRequestItems := make([]*api.JobSubmitRequestItem, 0, len(items))
+	combinedHashData := make([]byte, 40)
```
d80tb7, line 198: is 40 bytes enough here?

internal/armada/server.go
severinson: You're right; removed.

internal/armada/server/submit_to_log.go
severinson: Yes, the compiler guarantees that each hash is exactly 20 bytes.

d80tb7: ALso if you're just worried about size in memory then string interning would probably be better.  If you're worried about the size on disk in the db then hashes work

internal/pgkeyvalue/pgkeyvalue.go
```diff
@@ -0,0 +1,198 @@
...
+
+// AddKey is equivalent to Add(ctx, key, nil).
+func (c *PGKeyValueStore) AddKey(ctx context.Context, key string) (bool, error) {
+	return c.Add(ctx, key, nil)
+}
+
+func (c *PGKeyValueStore) createTable(ctx context.Context) error {
+	var pgErr *pgconn.PgError
+	_, err := c.db.Exec(ctx, fmt.Sprintf("create table %s (key text primary key, value bytea, inserted timestamp not null);", c.tableName))
+	if errors.As(err, &pgErr) && pgErr.Code == "42P07" { // Relation already exists (i.e., someone else just created it), which is fine.
```
d80tb7, line 81: I think pgx has some constants for errors

internal/pgkeyvalue/pgkeyvalue.go
```diff
@@ -0,0 +1,198 @@
...
+		})
+	}
+	if tableName == "" {
+		return nil, errors.WithStack(&armadaerrors.ErrInvalidArgument{
+			Name:    "TableName",
+			Value:   tableName,
+			Message: "TableName must be non-empty",
+		})
+	}
+	cache, err := simplelru.NewLRU(100, nil)
```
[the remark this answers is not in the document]
severinson: Agreed; changed to be a parameter and set to one million.

internal/pgkeyvalue/pgkeyvalue.go
severinson: Nice. I've updated the code to use them.

e2e/pulsar_test/pulsar_test.go
```diff
@@ -118,6 +118,85 @@ func TestSubmitJobs(t *testing.T) {
...
+			originalJobIds[i] = res.JobResponseItems[i].GetJobId()
+		}
+
+		numEventsExpected := numJobs * 6
+		_, err = receiveJobSetSequences(ctx, consumer, armadaQueueName, req.JobSetId, numEventsExpected, 10*time.Second)
+		if err != nil {
+			return err
+		}
+
+		// The second time, all jobs should be rejected.
```
d80tb7, line 149: nit- they don't get rejected just the deduplicated ids are returned

d80tb7 (approved):

d80tb7 (approved):

PR #950, opened 2022-05-05
Better Logging In LookoutIngester

d80tb7:
I've added a periodic logger in the pulsar receive code (similar to that in `submit_from_log`) so that we can see that we're actually receiving pulsar messages.

### DIFF
--- a/internal/lookoutingester/instructions/instructions_test.go
+++ b/internal/lookoutingester/instructions/instructions_test.go
@@ -347,14 +347,10 @@ func TestReprioritised(t *testing.T) {
 
 func TestInvalidEvent(t *testing.T) {
 
-	// This event is invalid as it references a job that doesn't exist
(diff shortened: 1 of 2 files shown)


internal/lookoutingester/instructions/instructions_test.go
```diff
@@ -347,14 +347,10 @@ func TestReprioritised(t *testing.T) {
 
 func TestInvalidEvent(t *testing.T) {
 
-	// This event is invalid as it references a job that doesn't exist
-	nonExistingJob, _ := armadaevents.ProtoUuidFromUlidString("01f3j0g1md6qx7z5qb148qnh4r")
```
d80tb7, line 351: We stopped processing `JobRunLeased` messages a while back which broke the purpose of this test.  It should be ignoring the event because its invalid , instead it was now ignoring the event because we don't event parse it.

I've changed it to use a `JobRunRunning` event, which we do care about.

severinson (approved):

severinson (approved):

PR #952, opened 2022-05-06
Store Jobs in JobSet by Queue:JobSet Pair: add storage

carlocamurri:
We currently store a mapping from JobSet -> set of JobIds in that JobSet in Redis in Armada.

This however means that if multiple Queues have the same JobSet name, then Jobs from different Queues will be stored under the same key in Redis.

As such we want to make sure that the key is a Queue:JobSet pair instead. This would make JobCancellation easier, as in order to cancel all Jobs in a JobSet, we currently have to load all Job ids in a queue and filter out those in the specified JobSet.

This is the first step of the migration to avoid breaking existing setups, which allows for storing Job ids in JobSets using both methods.

### DIFF
--- a/internal/armada/repository/job.go
+++ b/internal/armada/repository/job.go
@@ -205,6 +205,9 @@ func (repo *RedisJobRepository) DeleteJobs(jobs []*api.Job) (map[*api.Job]error,
 		deletionResult.deleteJobSetIndexResult = pipe.SRem(jobSetPrefix+job.JobSetId, job.Id)
 		deletionResult.deleteJobRetriesResult = pipe.Del(jobRetriesPrefix + job.Id)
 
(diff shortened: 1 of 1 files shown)


severinson:
Wouldn't it be better to hash the queue and job set name together, i.e., to use as the key
key = sha1(sha1(queue name) + sha1(job set name))?

It's safer (since we're won't have issues like a + ab = aa + b without needing to disallow particular characters), it means keys will be shorter (a fixed 160 bits if using sha1), and it improves string comparison performance (since keys won't have long common prefixes).

We use this method job deduplication:
https://github.com/G-Research/armada/blob/master/internal/armada/server/submit_to_log.go#L103

JamesMurkin (approved):

PR #963, opened 2022-05-10
Fix release builds

JamesMurkin:
- Use machine image. Building with docker in docker seems to be painful. This makes it consistent with other build steps
 -  Use gox v1.0.1, as its a version we've always used and is known to work as we expect
 - Fix store_artifacts, as it doesn't support wildcard
 - Fix dotnet release, as it doesn't need any external dependencies so this just keeps it simple + fixes it failing to get external dependencies


severinson:
The makefile looks for kubectl on the path, so you'd need to remove that check too.
https://github.com/G-Research/armada/blob/master/makefile#L8

severinson (approved):

PR #972, opened 2022-05-12
Python client

andrewpcone:
https://github.com/G-Research/armada/issues/956


severinson:
I have some suggestions:

- Create a new make target to build the python client. Write a new script for it instead of using the existing one.
- Use a python docker image as the base for the docker image to build the python client. That way you don't have to apt install python.
- Using pip to install the grpc tooling is fine. To make it possible to build behind a corporate proxy, you can load in environment variables from the host to set the pip repository (assuming this is possible).

client/python/client.py
```diff
@@ -0,0 +1,284 @@
...
+    # private static ApiJobSubmitRequest CreateJobRequest(string jobSet)
+    def job_submit_request_items_for_test(self, queue, job_set_id):
+        pod = core_v1.PodSpec(
+            volumes=[
+                core_v1.Volume(
+                    name="root-dir",
+                    volumeSource=
+                        core_v1.VolumeSource(flexVolume=
+                            core_v1.FlexVolumeSource(
+                                driver="gr/cifs",
```
severinson, line 193: I think this is a GR-specific component that shouldn't be in the public repo.

client/python/client.py
```diff
@@ -0,0 +1,284 @@
...
+                if e.code() == grpc.StatusCode.CANCELLED:
+                    pass
+                # process cancelled status
+                elif e.code() == grpc.StatusCode.UNAVAILABLE and 'Connection reset by peer' in e.details():
+                    pass
+                # process unavailable status
+                else:
+                    raise
+
+        th = threading.Thread(target=thread_func)
```
severinson, line 170: Using functools.partial may be cleaner here. Concurrent.futures may be better than using threads.

client/python/client.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def reprioritize_jobs(self, new_priority, job_ids=None, job_set_id=None, queue=None):
+        request = submit_pb2.JobReprioritizeRequest(
+            job_ids=job_ids, job_set_id=job_set_id, queue=queue,
+            new_priority=new_priority
+        )
+        response = self.submit_stub.ReprioritizeJobs(request)
+        return response
+
+    def create_queue(self, name, **queue_params):
```
severinson, line 127: Is this the way to do it in Python? Why are we wrapping the auto-generated code like this? Couldn't we just call into the auto-generated code directly?

client/python/client.py
```diff
@@ -0,0 +1,284 @@
...
+                raise Exception("need username and password for Basic auth")
+            self.username = username
+            self.password = password
+        elif self.method == AuthMethod.OpenId:
+            if not oidc_token:
+                raise Exception("need oidc_token for OpenId auth")
+        elif self.method == AuthMethod.Kerberos:
+            pass
+
+class GrpcBasicAuth(grpc.AuthMetadataPlugin):
```
kannon92, line 49: I think it would be nice to explain this a bit more.  We went into the depths of GRPC docs to understand this so we should add a comment and maybe a link to where we discovered this.

client/python/client.py
```diff
@@ -0,0 +1,284 @@
...
+            .decode('ascii')
+        callback(
+            (('authorization', f'basic {b64encoded_auth}'),),
+            None
+        )
+
+class ArmadaClient:
+    def __init__(self, host: str, port: int,
+        auth_data: AuthData = AuthData(),
+        testing: bool = False):
```
kannon92, line 66: This testing flag is really about turning off ssl?  We should probably make this be a more descriptive name.

client/python/client.py
```diff
@@ -0,0 +1,284 @@
...
+            pass
+
+        self.submit_stub = submit_pb2_grpc.SubmitStub(self.channel)
+        self.event_stub = event_pb2_grpc.EventStub(self.channel)
+        self.usage_stub = usage_pb2_grpc.UsageStub(self.channel)
+
+    def get_job_events_stream(self, queue, job_set_id, from_message_id=None, watch=False):
+        jsr = event_pb2.JobSetRequest(queue=queue, from_message_id=from_message_id, watch=False)
+        jse = self.event_stub.GetJobSetEvents(queue, job_set_id, jsr)
+        print(jse)
```
kannon92, line 101: Debugging print leftover.

client/python/client.py
```diff
@@ -0,0 +1,284 @@
...
+
+from armada.client import event_pb2
+from armada.client import event_pb2_grpc
+from armada.client import queue_pb2
+from armada.client import queue_pb2_grpc
+from armada.client import usage_pb2
+from armada.client import usage_pb2_grpc
+from armada.client import submit_pb2
+from armada.client import submit_pb2_grpc
+from k8s.io.api.core.v1 import generated_pb2 as core_v1
```
kannon92, line 18: I probably would avoid using internal k8 protobufs api in this code especially if we are using it to create pods/containers.  For unit/e2e testing, you could probably just use simple pod definitions without the k8 api.

kannon92: This is what I get for only looking at your client code.  It looks like our proto objects depend on this so the k8 is needed.  We don't need to reference it in client as I think the code generation of pods is verbose.  

But we can't avoid using internal k8 protobuf.

client/python/client.py
andrewpcone: I got that from the C# example. With what should it be replaced?

client/python/client.py
andrewpcone: Done in c45a48d86baef3d1271e85d5d1d1709402c191d8

client/python/client.py
andrewpcone: 6da8377

client/python/client.py
andrewpcone: a37346d

client/python/client.py
andrewpcone: I don't understand what you're proposing here. I need those imports in order to construct the proto msg that gets sent to armada, do I not?

client/python/armada/client/event_pb2_grpc.py
```diff
@@ -0,0 +1,166 @@
+# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
+"""Client and server classes corresponding to protobuf-defined services."""
+import grpc
+
+from armada.client import event_pb2 as armada_dot_client_dot_event__pb2
```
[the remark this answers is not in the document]
rankindaveyg: Okay, duh, i see the line above about it being generated code. Nothing to see here, move along...

client/python/client.py
kannon92: Yep.  Ignore this.

build/python-client/Dockerfile
```diff
@@ -0,0 +1,40 @@
+ARG PYTHON_VERSION=3.10.4
+ARG GOGO_PROTOBUF_VERSION=1.3.2
```
severinson, line 2: Gogo-proto is a system for generating go code; you don't need it for Python.

client/python/client.py
```diff
@@ -0,0 +1,284 @@
...
+
+class ArmadaClient:
+    def __init__(self, host: str, port: int,
+        auth_data: AuthData = AuthData(),
+        disable_ssl: bool = False):
+
+        self.host = host
+        self.port = port
+        self.disable_ssl = disable_ssl
+        self.executor = ThreadPoolExecutor(max_workers=1)
```
severinson, line 75: I'd suggest making max_workers an argument of __init__ to allow the caller to determine the number of threads. You could default to the number of logical cores of the system.

client/python/client.py
severinson: All right.

client/python/client.py
```diff
@@ -0,0 +1,284 @@
...
+                if e.code() == grpc.StatusCode.CANCELLED:
+                    pass
+                # process cancelled status
+                elif e.code() == grpc.StatusCode.UNAVAILABLE and 'Connection reset by peer' in e.details():
+                    pass
+                # process unavailable status
+                else:
+                    raise
+
+        self.executor.submit(event_counter)
```
severinson, line 174: Using functools.partial is more explicit than passing variables in via nonlocal.

client/python/client.py
```diff
@@ -0,0 +1,284 @@
...
+        return event_stream
+
+    def unwatch_events(self, event_stream):
+        event_stream.cancel()
+
+class ArmadaClientTest:
```
kannon92, line 180: Can you look into refactor this so test isn't part of the client package?  

I would like to see a tests directory and you should be able to run this via pytest or unittest.  

You can look at my branch for ideas.

client/python/client.py
kannon92: @severinson Can probably correct me but this looks like to me like it is generating a static definition for a pod.  You can probably just use really simple pod definitions or reference an existing one in examples.

kannon92 (changes requested):
In general, I'd like to see a few things.  

1) python typically uses a requirements.txt and a requirements-dev.txt so you can list out your dependencies for a local build (grpc, grpcio-tools).  
2) We should set up formatting, linting and testing for this client package.
3) The packaging should generate a wheel file so clients can use this outside.
4) We should use a more descriptive name for the client package.  The way you structured this I think the package name would be client which can be confusing if someone uses this outside of armada.  `armada_client` could be more descriptive.

Please take a look at fork and see if you integrate these changes.  But I think this is necessary for approval.  


https://github.com/kannon92/armada/tree/python_client_poetry/client/python

client/python/client.py
andrewpcone: 05d334102472f1228cd6cff646df15ad168fb05f

build/proto/Dockerfile
```diff
@@ -76,4 +76,10 @@ COPY --from=builder /usr/local/bin /usr/local/bin
 COPY --from=builder /go/bin /usr/local/bin
 COPY --from=builder /proto /proto
 
+# Install python3 and grpcio-tools
+# TODO maybe install all this manually (ie, not from package management)
+RUN apt-get update \
```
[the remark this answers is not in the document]
andrewpcone: Yes, building `grpcio-tools` takes forever, and it is no longer needed in that Dockerfile, so I removed it in 9409da2.

Instead, please enjoy the same thing taking forever using the new Dockerfile. I recommend building with `make python-client` from the root armada directory.

PR #973, opened 2022-05-12
Postgres Backed Event API

d80tb7:
**Events API backed by Postgres**

There are two main components added here:

**1. Event Ingester**
The role of the Event Ingester is to read messages from Pulsar and insert them into the postgres db.  It is implemented as an async pipeline consisting of the following go routines:

1. Read Messages from pulsar. Each message has a sequence Number (`SeqNo`) which is an int64 giving the position of the message in the stream.
2. Convert each message into an event to be inserted into the database
3. Batch the messages so that we have a batch of either x (default 10K) messages or all the messages from the last  y (default 500ms) millis (whichever occurs first).
4. Insert the batch into the database.  Note that we insert both the events themselves and also populate a `SeqNo` table which records the latest `SeqNo` available for each jobset.
5. Publish the SeqNos we just inserted onto pulsar.  This lets the events api know that new data is available.
6. Ack the messages on pulsar.

There are a couple of optimisations  in the above:
- We compress each event message.  Messages under x bytes (default 1024) will be compressed with `zlib.NoCompression`
- In the database we don't store the jobset and queue with each message as these are potentially long strings and therefore have a high overhead cost.  Instead we have a `Jobset` mapping table which we use to map (queue, jobset) to an int64.  This mapping is only valid internally for the event api and should not escape to other components.

**2. Event API**
This is responsible for serving the events in a compatible manner with the existing api.  It works as follows.

1.  At startup, the api reads the `SeqNo` table to get the latest sequence numbers available for each jobset.  It then subscribes to the seqNo updates topic (the one published to by the ingester) to get new seqNos.  Importantly, it reads this topic from *before* it read the db which means it shouldn't miss any updates.
2. When a request for events comes in the api will first map the (jobset, queue) to the internal int64 identifier.  It then makes a call to the subscription manager to request a channel of events.
3. The subscription manager first determines if we need to catch up (i.e. we have a `SeqNo` for that jobset which is greater than the seqNo the user has asked for). If so it makes db requests as fast as it can until we have caught up.
4. Once we have caught up, the subscription manager will check each second to see if we have new events available.  This is a fast operation as it's an in memory-check using the data we've susbcribed to from pulsar.
5. If there are new events available, the subscription manager will request new events from the db.  For efficiency state we batch these requests such that several jobset requests can be in the same db query.
6. All Events are returned to the api.  Before we can send them to the user we need to a) decompress the events and b) convert from internal events to api events.

To maintain compatibility with the the existing system I have done the following:
- Serving from postgres is disabled by default and can be enabled by a feature flag.
- Even when serving from postgres is enabled, we still may have to server events from redis.  This is because clients can ask for events from a specific seqNo and the postgres seqNos are not cmpatible with the redis ones.  To deal with this, I first check if the seqNo requested is compatible with postgres and if not I send it to redis.  Once the feature has been enabled for a couple of weeks, we should be able to turn off the redis fallback.

**Current Status**
There is still a bit more to do on this, but I wanted to raise the PR now as the code is in the right shape and basic functionality is working.  Still to do is:

- Retry logic for database/pulsar failures
- Docker image for event ingester
- Helm chart for event ingester
- More unit tests for the event ingester
- Make the e2e tests work using the postgres event api


internal/armada/server.go
```diff
@@ -286,9 +292,36 @@ func Serve(config *configuration.ArmadaConfig, healthChecks *health.MultiChecker
...
+	var eventApi *serving.EventApi = nil
+	// Setup new events api if enabled
+	if config.EventApi.Enabled {
+
+		// Set up eventDb
+		pool, err := postgres.OpenPgxPool(config.Postgres)
+		eventDb := eventdb.NewEventDb(pool)
+
+		// Setup pulsar
+		pulsarClient, err := pulsarutils.NewPulsarClient(&config.Pulsar)
```
severinson, line 304: Re-use the existing Pulsar client.

internal/eventapi/serving/subscription_manager.go
```diff
@@ -0,0 +1,248 @@
...
+	Unsubscribe(subscriptionId int64)
+}
+
+type internalSubscription struct {
+	subscriptionId int64
+	jobset         int64
+	channel        chan []*model.EventRow
+}
+
+type DefaultSubscriptionManager struct {
```
severinson, line 27: If you don't want to expose the default implementation, I suggest calling it "subscriptionManager" and creating a method "New" with return type "SubscriptionManager" (i.e., the interface type).

severinson:
I've added some comments in-line. I also have the following high-level comments and questions.

Ingester
- I think the idea for the ingester is right (i.e., insert batches into pg in an idempotent manner).
- Do you have logic to seek to the correct place in each partition based on the sequence ids stored in pg? Also for when partitions are rebalanced between consumers?
- Is it necessary to publish the seqnos on Pulsar? We could also use a polling model (e.g., look for new messages at most once per second) or a model based on gRPC calls to indicate updates. If we do need to publish seqnos on Pulsar I'd guggest moving them to another package than armadaevents.
- Instead of going to the db to get the int64 associated with each (queue, jobset), I'd suggest using the hash sha1(sha1(queue) + sha1(jobset)). That'll reduce the risk of collisions and it could save us a network round trip in some cases.
- Could we achieve the same performance without the need for async? If Pulsar queries messages in the background, we don't need to implement async ourselves. We need to wait for the pg transaction before moving on so that part must be sync regardless.

API
- Does there need to be coordination with the ingester? Perhaps we could keep active pairs (queue, jobset) in-memory and every second get all new events for those topics. Whenever there's a new active pair (queue, jobset) (due to an api call requesting events) we'd query the db for all events associated with that pair and then continue as usual. We regularly clean up pairs (queue, jobset).
- Alternatively, could we leverage ephemeral topics in Pulsar? For each combination (queue, jobset) we create a new ephemeral topic and republish all events to that topic. Later, we close the consumer, which causes the topic to be garbage collected. In this case we'd use the in-memory storage of the brokers instead of the server to store events clients are actively listening for.

d80tb7:
So to answer the questions:

> Do you have logic to seek to the correct place in each partition based on the sequence ids stored in pg? Also for when partitions are rebalanced between consumers?

I don't think we need to seek here as Pulsar will guarantee that we pick off from where we left off last time (including if we get a handover).  There is an edge case whereby we have put a message into the db but not acked it yet but in that case we will exclude the message when we come to write it to the database.

> Is it necessary to publish the seqnos on Pulsar

It's not necessary, but I think it's a nicer architectural solution than polling the database as it reduces database load and means we get updates faster. Grpc would be more complex (the api could poll the ingester but then it would have to know about the ingester which right now it doesn't) and we already have pulsar available. The one caveat here might be if we find out that it's possible for us to put a message into the db but not publish the update.  I don't think this is the case as we don't ack the message until the update is published, and when the message is reprocessed it will not be inserted to the db (as it already exists) but we *will* resend the update to say the seqNo is available.

> Instead of going to the db to get the int64 associated with each (queue, jobset), I'd suggest using the hash sha1(sha1(queue) + sha1(jobset)). That'll reduce the risk of collisions and it could save us a network round trip in some cases.
There is no risk of collision here. The int64s are assigned atomically by the database so we know they will be unique.  The advantage of using an int64 is that it adds 8 bytes of overhead per message rather than 40and also that it means we have an (efficient) int primary key in the db rather than a (less efficient) bytea key.  The network round trip should hopefully not be too bad as cache the mappings in memory and so the common case should be a single lookup when we first encounter the jobset.  All subsequent mappings for that jobset will be memory-local.

> Could we achieve the same performance without the need for async? If Pulsar queries messages in the background, we don't need to implement async ourselves. We need to wait for the pg transaction before moving on so that part must be sync regardless.
There is an argument that I've got more goroutines than we actually need here- for example it's unlikely that acking the messages is going to be a bottleneck, but I think you need at least two goroutines as you want to have the next batch of messages ready for inserting as soon as you have finished persisting the current batch.  To answer the specific point you raised we don't need to wait for the transaction to finish before we can start unmarshalling, validating and compressing the next set of messages.

> Does there need to be coordination with the ingester? Perhaps we could keep active pairs (queue, jobset) in-memory and every second get all new events for those topics. Whenever there's a new active pair (queue, jobset) (due to an api call requesting events) we'd query the db for all events associated with that pair and then continue as usual. We regularly clean up pairs (queue, jobset).

Sorry I'm not sure what you're suggesting here?  Are you saying that every second we should poll the db for new events for all jobsets that we know we have subscriptions to.  It's not a bad idea and would be definitely be more efficient in the case that we had many people subscribing to the same jobset.  The tricky bit would be synchronising the initial fetch with the periodic fetches but that sounds like a solvable issue.

>Alternatively, could we leverage ephemeral topics in Pulsar? For each combination (queue, jobset) we create a new ephemeral topic and republish all events to that topic. Later, we close the consumer, which causes the topic to be garbage collected. In this case we'd use the in-memory storage of the brokers instead of the server to store events clients are actively listening for.

This I did think about and I think it would be a better long term strategy.  I think this is worth exploring once they have message routing working as the main issue right now is that we would end up subscribing to thousands (or even 10s thousands) of topics and that's where pulsar struggled last time.

internal/eventapi/serving/subscription_manager.go
d80tb7: So it's nothing to do with not wanting to hide the internals (all the internal stuff is module private- anyway).  The reason for the interface type is purely for test purposes so that we can easily inject a mock/dummy subscription manager that is much simpler than this.

internal/eventapi/serving/subscription_manager.go
```diff
@@ -0,0 +1,248 @@
...
+	catchupChannel    chan *model.EventRequest // A channel for queries to be made in a low latency manner at the cost of overall throughput
+	subscriptionIndex int64
+	subscriptionsById map[int64]*internalSubscription
+	subscriptionMutex sync.RWMutex
+	clock             clock.Clock
+}
+
+// EventDbRO is a Simplified view of EventsDb which helps testing
+type EventDbRO interface {
+	GetEvents(requests []*model.EventRequest, limit int) ([]*model.EventResponse, error)
```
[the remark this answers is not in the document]
d80tb7: no it doesn't- let me fix

internal/pulsarutils/async.go
```diff
@@ -11,10 +11,25 @@ import (
...
 
 	"github.com/G-Research/armada/internal/common/logging"
-	"github.com/G-Research/armada/internal/lookoutingester/model"
-	"github.com/G-Research/armada/pkg/armadaevents"
 )
 
+// ConsumerMessageId wraps a pulsar message id  and an identifier for the consumer which originally received the
+// corresponding message.  This exists because we need to track which messages came from which consumers so that
+// we can ACK them on the correct consumer.
+type ConsumerMessageId struct {
```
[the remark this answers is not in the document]
d80tb7: So we can have concurrency from multiple pulsar readers inside a single process.  I'm not convinced we absolutely need it and I currently default this concurrency to 1, but it wasn't too hard to add and I think it may come in useful one day.

severinson (approved):

PR #974, opened 2022-05-13
Always ack messages, run services in errgroups

severinson:
This pr updates Pulsar subscribers to always ack messages.

It also updates the server to run concurrent services in errgroups.
This allows errors to be propagated between services so that all services can be shut down gracefully in case of an error.
Further, if any service fails, we can shut down the app so that it can be restarted by k8s.

### DIFF
--- a/cmd/armada/main.go
+++ b/cmd/armada/main.go
@@ -1,20 +1,25 @@
 package main
 
 import (
+	"context"
+	"fmt"
 	"net/http"
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/G-Research/armada/internal/armada"
 	"github.com/G-Research/armada/internal/armada/configuration"
 	"github.com/G-Research/armada/internal/common"
 	gateway "github.com/G-Research/armada/internal/common/grpc"
 	"github.com/G-Research/armada/internal/common/health"
+	"github.com/G-Research/armada/internal/common/logging"
 	"github.com/G-Research/armada/pkg/api"
 )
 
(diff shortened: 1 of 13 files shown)


internal/armada/server_test.go
```diff
@@ -177,7 +181,8 @@ func withRunningServer(action func(client api.SubmitClient, leaseClient api.Aggr
 	if err != nil {
 		panic(err)
 	}
-	defer minidb.Close()
+	// minidb.Close hangs indefinitely (likely due to a bug in miniredis)
+	// defer minidb.Close()
```
JamesMurkin, line 185: Is this a good idea? Did you try newer versions of miniredis or do they have breaking api changes that makes upgrading non-trivial?

internal/armada/server.go
```diff
@@ -53,12 +67,36 @@ func Serve(config *configuration.ArmadaConfig, healthChecks *health.MultiChecker
 	authServices := auth.ConfigureAuth(config.Auth)
 	grpcServer := grpcCommon.CreateGrpcServer(authServices)
 
-	// Allows for registering functions to be run periodically in the background
+	// Shut down grpcServer if the context is cancelled.
+	// Give the server 5 seconds to shut down gracefully.
+	services = append(services, func() error {
+		<-ctx.Done()
+		go func() {
+			time.Sleep(5 * time.Second)
```
JamesMurkin, line 75: Not something to change. However we should probably review the graceful shutdown times and make sure they are sensible and consistent (and configurable?)

JamesMurkin (approved):

internal/armada/server_test.go
severinson: I don't think it matters much for us. Since the server is cleaned up when the process exits.

internal/armada/server.go
severinson: I agree.

cmd/armada/main.go
```diff
@@ -63,19 +82,22 @@ func main() {
...
+	// Start Armada server
+	g.Go(func() error {
+		return armada.Serve(ctx, &config, healthChecks)
+	})
 
-		// TODO Can't be sure this will run, since there are calls to panic/os.Exit in sub-calls
-		shutdown()
+	// Assume the server is ready if there are no errors within 10 seconds.
+	go func() {
+		time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
severinson: We didn't have one before because we assumed that services were ready immediately after being started. This is more conservative. I agree, but I'm not yet sure how to implement it.

PR #975, opened 2022-05-16
Automatically populate queue/jobSetId if not provided for job cancellation

severinson:
This change is necessary for compatibility with the legacy submit API.

### DIFF
--- a/internal/armada/server/submit_to_log.go
+++ b/internal/armada/server/submit_to_log.go
@@ -254,6 +254,12 @@ func (srv *PulsarSubmitServer) CancelJobs(ctx context.Context, req *api.JobCance
 		if err != nil {
 			return nil, err
 		}
+		if len(jobs) == 0 {
+			return nil, &armadaerrors.ErrNotFound{
+				Type:  "job",
+				Value: req.JobId,
+			}
+		}
 		if len(jobs) != 1 { // Internal error; should never happen.
 			return nil, fmt.Errorf("expected 1 job result, but got %v", jobs)
 		}
--- a/internal/armada/server/submit_to_log.go
+++ b/internal/armada/server/submit_to_log.go
@@ -245,6 +245,48 @@ func (srv *PulsarSubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmi
 }
 
 func (srv *PulsarSubmitServer) CancelJobs(ctx context.Context, req *api.JobCancelRequest) (*api.CancellationResult, error) {
+
+	// For requests specifying a job id, we need to check which queue the job is part of
+	// (see comments below).
+	// This must be done before checking authroization, since the auth check expects a queue.
+	if req.JobId != "" {
+		jobs, err := srv.SubmitServer.jobRepository.GetJobsByIds([]string{req.JobId})
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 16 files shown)


internal/armada/server/submit_to_log.go
```diff
@@ -245,6 +245,48 @@ func (srv *PulsarSubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmi
 }
 
 func (srv *PulsarSubmitServer) CancelJobs(ctx context.Context, req *api.JobCancelRequest) (*api.CancellationResult, error) {
+
+	// For requests specifying a job id, we need to check which queue the job is part of
+	// (see comments below).
+	// This must be done before checking authroization, since the auth check expects a queue.
```
d80tb7, line 251: [minor] `authroization`

internal/armada/server/submit_to_log.go
```diff
@@ -245,6 +245,48 @@ func (srv *PulsarSubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmi
...
+		}
+		if len(jobs) != 1 { // Internal error; should never happen.
+			return nil, fmt.Errorf("expected 1 job result, but got %v", jobs)
+		}
+		queue := jobs[0].Job.GetQueue()
+		jobSetId := jobs[0].Job.GetJobSetId()
+
+		// We allow clients to submit requests only containing a job id.
+		// For these requests, we need to populate the queue and jobSetId fields.
+		if req.Queue == "" {
```
d80tb7, line 265: shouldn't we always fill in the queue here?  Otherwise a malicious user could fill in the queue to whatever they want and they would effectively bypass our check.

internal/armada/server/submit_to_log.go
```diff
@@ -245,6 +245,48 @@ func (srv *PulsarSubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmi
...
+			req.Queue = queue
+		}
+		if req.JobSetId == "" {
```
d80tb7, line 268: Again- I'd fill this in based on the job.  We don't want weird behaviour down the line because the user has passed in  inconsistent job and jobsetIds

internal/armada/server/submit_to_log.go
severinson: Fixed, thanks.

internal/armada/server/submit_to_log.go
```diff
@@ -245,6 +245,48 @@ func (srv *PulsarSubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmi
...
+
+	// For requests specifying a job id, we need to check which queue the job is part of
+	// (see comments below).
+	// This must be done before checking authroization, since the auth check expects a queue.
+	if req.JobId != "" {
+		jobs, err := srv.SubmitServer.jobRepository.GetJobsByIds([]string{req.JobId})
+		if err != nil {
+			return nil, err
+		}
+		if len(jobs) != 1 { // Internal error; should never happen.
```
d80tb7, line 257: This would happen if the user sends us some crazy job id.  It might be worth distinguishing between jobNotFound (which is probably their fault) and multiple jobs found (which is definitely our fault).

severinson: I agree. Fixed.

d80tb7 (approved):

PR #976, opened 2022-05-16
New scheduler spec

severinson:
This PR adds a docstring describing a new proposed scheduler.


devlounge (approved):
lgtm

docs/scheduler.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Elasticity and dynamic resource creation
+
+Armada supports workflows that require a dynamically changing amount of resources (e.g., [Apache Spark](https://spark.apache.org/) clusters and AirFlow jobs) using the following two approaches:
+
+- The workflow is submitted to a higher-level scheduler (e.g., Apache AirFlow) that dynamically submits jobs to Armada, monitors their status, and makes decisions on which jobs to submit next.
+- An "operator" job is submitted to Armada (e.g., corresponding to a Spark cluster manager) that has the power to request resources from Armada (e.g., to run additional pods) by submitting further Armada jobs. The operator job may do so either via the Armada job submission API or via a special set of Kubernets API endpoints that translate resource creation commands into Armada job submissions.
```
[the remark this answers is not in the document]
d80tb7: We should probably also say explicitly that this mechanism can be used to release resource.

docs/scheduler.md
```diff
@@ -0,0 +1,152 @@
...
+2. Cluster assignment
+3. Node assignment
+
+The first two stages take place in the server and the third in the executor. We explain the first two stages here and the third stage in a later section. At any one time, each job is in one of the following three stages:
+
+1. Queued
+2. Pending
+3. Running
+
+Each job that is submitted is part of one of the `N` queues managed by the scheduler and is in the queued stage. Jobs within each queue are ordered by priority but there is no ordering between jobs in different queues. Jobs are moved from the queued stage into the pending stage one at a time and jobs in the pending stage are assigned to worker clusters on a first-in-first-out basis. Hence, all jobs in the pending stage are ordered and the global prioritisation stage consists of deciding in what order jobs should be moved from the queued to pending stage. Within each queue, jobs are moved from the queued to pending stage strictly in order of priority, i.e., for each queue, the next job to be moved into the pending stage is always the highest priority job in that queue.
```
d80tb7, line 46: "There is no ordering between jobs in different queues". Surely we need to say something about relative queue priority?

d80tb7: Actually you talk about it below, but I'd mention it here if only to say it's discussed in detail below.

docs/scheduler.md
```diff
@@ -0,0 +1,152 @@
...
+
+Armada jobs are created by submitting a job specification to the Armada server, either via client libraries or via the `armadactl` command-line-utility. In this way, users can submit jobs to Armada directly. However, users may also submit workflows to higher-level schedulers, such as [Apache AirFlow](https://airflow.apache.org/), which in turn submit jobs to Armada.
+
+### Distributed jobs
+
+Distributed jobs are represented as [PodGroups](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/kep/42-podgroup-coscheduling/README.md), i.e., groups of cooperating pods, and the pods that make up a `PodGroup` are scheduled jointly using a gang-scheduling algorithm. Similar to to [kube-batch](https://github.com/kubernetes-sigs/kube-batch) and [Volcano](https://volcano.sh/en/docs/podgroup/), the gang-scheduling algorithm is implemented as a [custom Kubernetes scheduler](https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/).
```
d80tb7, line 21: Do we have anything to say about pre-empting distributed jobs.

docs/scheduler.md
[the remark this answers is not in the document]
severinson: I clarified it a bit. I agree, I've updated the text.

docs/scheduler.md
severinson: I've updated the text.

docs/scheduler.md
severinson: I added a comment on this.

severinson:
Here's an overview of the various job types I've seen.

d80tb7:
- Topology awareness
- If all the ML jobs can be fit into the PodGroup model, or the job/indexed job model
- Speaking to the quants that run these jobs to understand their use case
- Prioritising these things
- Updated submit API and events

severinson:
Dominant resource fairness (DRF) is likely a better policy than max-min fairness.
DRF is introduced in https://cs.stanford.edu/~matei/papers/2011/nsdi_drf.pdf
Volcano and a similar scheduler used by Nvidia both use DRF.
Volcano: https://github.com/volcano-sh/volcano/blob/master/docs/design/drf.md
Nvidia: https://www.youtube.com/watch?v=yWOcsNv1tKo

severinson:
We could consider using hierarchical queues to more effectively balance resource usage between teams.
There's a related notion in network engineering we could draw inspiration from.
https://en.wikipedia.org/wiki/Token_bucket

severinson:
Consider automatically reducing the weight of queues that arr inactive for extended periods of time.

severinson:
We should have a notion of job urgency in the spec. Can be implemented as follows.

For each job, the user specifies one of the following priority groups:
1. Run now; my life depends on it
2. Schedule as usual
3. Only run if there are spare resources

Jobs in group 2 and 3 are only scheduled if there are no jobs in group 1.
Jobs in group 3 are only scheduled if there are no jobs in group 2.
Jobs in group 3 must be immediately preemptible.

We may want to a much more granular approach (e.g., an arbitrarily large number of groups). But the above would be a good starting point.

severinson:
Currently, we have no text on how we decide in which order jobs are preempted. We should have some notion of how expensive it is to preempt a given job.

PR #979, opened 2022-05-17
Timeout read call in Armada C# client

d80tb7:
In #953 we attempted to add a timeout when reading form the stream in the `watch` call.  This is because various network issues can cause the stream to hang, in which case we can't distinguish between "the stream is fine and there's just no messages" and "the stream has hung and you're never getting messages".  Unfortunately this caused issues with resources not being cleaned up so we rolled it back in #977.  This PR aims to add back the timeout.  

Basically the strategy is as follows:

* Add an extension method to `Task`- `TimeoutAfter(Timespan timeout)`  which will fail the task with a `TimeoutException` if the task does not complete within the specified timeout.
* Change the `ReadAynsc()` in watch  to `ReadAsync().withTimeout()`
* Remove the need for us to call `Reader.EndOfSteam()`, instead relying on `ReadAsync()` to return `null` to let us know that the end of stream has been reached.  By removing  `Reader.EndOfSteam()` we ensure that we only have a single call (`ReadAsync()`) to worry about timing out.

The above behaviour should ensure that we will throw a `TimeoutException` if we receive no messages for 10 minutes (whether that is because there are genuinely no messages or because of bad network).  As in #953 this will cause us to restart the connection to the server and thus (hopefully) fix any network weirdness that may have occurred.


client/DotNet/Armada.Client/TaskExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <param name="timeout">Timespan after which the task will fail with a TimeoutException</param>
+        /// <typeparam name="TResult">Result type returned by the task</typeparam>
+        /// <returns>Modified Task that will </returns>
+        public static async Task<TResult> TimeoutAfter<TResult>(this Task<TResult> task, TimeSpan timeout) {
+
+            using (var timeoutCancellationTokenSource = new CancellationTokenSource()) {
+                var completedTask = await Task.WhenAny(task, Task.Delay(timeout, timeoutCancellationTokenSource.Token));
+                if (completedTask == task) {
+                    timeoutCancellationTokenSource.Cancel();
+                    return await task;  // thi si needed to propagate exceptions
```
JamesMurkin, line 22:
```suggestion
                    return await task;  // this is needed to propagate exceptions
```

client/DotNet/Armada.Client/TaskExtensions.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace GResearch.Armada.Client
+{
+    public static class TaskExtensions
+    {
+        /// <summary>
+        /// Extension method to let a Task timeout with a TimeoutException iof it does not complete before the given timeout
```
JamesMurkin, line 10:
```suggestion
        /// Extension method to let a Task timeout with a TimeoutException if it does not complete before the given timeout
```

JamesMurkin (dismissed):

samclark (approved):
LGTM

PR #993, opened 2022-05-23
Reduce default lease expiry timeout

severinson:


d80tb7 (approved):

JamesMurkin:
This will be quite a risky configuration for prod, probably if anything 15 minutes is too short. 

Any interruption to the executor - server communication would mean all jobs will get lost. That is a very short time and c have many causes. On large clusters, just restarting the executor pod can take 2-3 minutes to catch up on pod state.

severinson:
> This will be quite a risky configuration for prod, probably if anything 15 minutes is too short.
> 
> Any interruption to the executor - server communication would mean all jobs will get lost(and killed on reconnection). That is a very short time and could have many causes. On large clusters, just restarting the executor pod can take 2-3 minutes to catch up on pod state.

In that case, let's hold off on merging this and implement logic that closes the lease on job X if we've received lease renewal messages from this executor in the last Y minutes, but we've received no lease renewal messages that include job X in the last Z minutes.

PR #995, opened 2022-05-23
Fix executor deleting events

JamesMurkin:
Currently the executor fails to delete events due to insufficient perms. However the log message doesn't make it clear why

 - Fix the log to display error properly when failing to delete pod events
 - Update clusterrole to allow deleting events


internal/executor/context/cluster_context.go
```diff
@@ -322,7 +322,7 @@ func (c *KubernetesClusterContext) ProcessPodsToDelete() {
 			if err == nil {
 				deletePodEventsErr := c.deletePodEvents(podToDelete)
 				if deletePodEventsErr != nil {
-					log.Errorf("Failed to delete pod events for pod %s/%s because %s", podToDelete.Name, podToDelete.Namespace, err)
+					log.Errorf("Failed to delete pod events for pod %s/%s because %s", podToDelete.Name, podToDelete.Namespace, deletePodEventsErr)
```
severinson, line 325:
```suggestion
					log.WithError(deletePodEventsErr).Errorf("Failed to delete pod events for pod %s/%s", podToDelete.Name, podToDelete.Namespace)
```

severinson: Shouldn't it be namespace before name, i.e., namespace/pod_name?

severinson: I think we should conform to the k8s way of doing it.

severinson (approved):

PR #997, opened 2022-05-23
Prevent concurrent access to map

JamesMurkin:
In Go it is not safe to read and write to a map at the same time
 - If this ever happens on separate threads, it should be protected as per the doc

This specific area has been seen to panic on occasion (higher load has higher chance)

Adding appropriate mutex to prevent panic


d80tb7 (changes requested):
does `registerIssue` suffer from the same issue.

I'd be tempted to say we should use a `sync.map` here to be a bit safer although:

a) I accept that you still might need a mutex in certain places
b) the sync.map inteface is a bit unfortunate in that it's not he same as a mpa interface

So I quite understand if you don't actually want to use sync.map here.

severinson:
Based on the docs of sync.Map, it seems we won't gain much benefit from using it over a regular map protected by a mutex? If so, I'd say it's preferable to use a regular map with mutex.
https://pkg.go.dev/sync#Map

severinson:
Based on the docs of sync.Map, it seems we won't gain much benefit from using it over a regular map protected by a mutex? If so, I'd say it's preferable to use a regular map with mutex.
https://pkg.go.dev/sync#Map

d80tb7 (approved):

PR #1002, opened 2022-05-24
[WIP] Handle new jobRunErrors Reasons

d80tb7:


internal/common/eventutil/eventutil.go
```diff
@@ -935,10 +935,10 @@ func EventSequenceFromApiEvent(msg *api.EventMessage) (sequence *armadaevents.Ev
 }
 
 // Id used for messages for which we can't use the kubernetesId.
-const LEGACY_RUN_ID = "00000000000000000000000000"
+const LEGACY_RUN_ID = "00000000-0000-0000-0000-000000000000"
```
d80tb7, line 943: I've changed this because previously `legacyJobRunId` was creating a `ulid` which is inconsistent with all the valid runIds which are `uuids`.  This lead to strange behaviour when you tried to do e.g. string comparisions.

This change makes every representation of `LEGACY_RUN_ID ` a uuid and is thus consistent with other runIds.

severinson (approved):

severinson (approved):

PR #1004, opened 2022-05-25
Better show errors when Pulsar Send fails

d80tb7:
I've moved all the pulsar sending code in `submit_to_log` into a common function that does async sends.
I've also made sure we log out the full error (including stack) of the errors we get when sending from this function.

### DIFF
--- a/internal/armada/server/submit_to_log.go
+++ b/internal/armada/server/submit_to_log.go
@@ -11,6 +11,7 @@ import (
 	"github.com/gogo/protobuf/types"
 	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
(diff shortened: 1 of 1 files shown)


internal/armada/server/submit_to_log.go
```diff
@@ -238,20 +233,11 @@ func (srv *PulsarSubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmi
...
-		}
-		pulsarrequestid.AddToMessage(msg, requestId)
-		_, err = srv.Producer.Send(ctx, msg)
-		if err != nil {
-			return nil, status.Errorf(codes.Internal, "failed to send message")
-		}
+	err = srv.publishToPulsar(ctx, []*armadaevents.EventSequence{sequence})
+
+	if err != nil {
+		log.Errorf("Error sending message to pulsar: %+v", err)
```
severinson, line 239: log.WithError(err).Error("failed send to Pulsar")

severinson (approved):

severinson (approved):

PR #1006, opened 2022-05-25
Set a max message size for Pulsar messages

d80tb7:
The pulsar go client defaults the maximum message size to 128KB.  The pulsar server defaults to 5MB.  We were having issues messages > 128Kb so I've made this configurable in the armada config and defaulted it to 4MB which is still comfortably lower than the server limit.

### DIFF
--- a/config/armada/config.yaml
+++ b/config/armada/config.yaml
@@ -53,3 +53,8 @@ eventRetention:
   retentionDuration: 336h # Specified as a Go duration
 metrics:
   refreshInterval: 10s
+pulsar:
+  enabled: false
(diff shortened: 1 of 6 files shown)


config/armada/config.yaml
```diff
@@ -53,3 +53,8 @@ eventRetention:
   retentionDuration: 336h # Specified as a Go duration
 metrics:
   refreshInterval: 10s
+pulsar:
```
d80tb7, line 56: we weren't defaulting any pulsar properties before.  This is a minimal set.

JamesMurkin (approved):

PR #1015, opened 2022-05-26
Limit executor object creation rate

severinson:
This PR makes the following changes to the executor:

- Introduces a QPS and burst limit shared by all Kubernetes API clients.
- Introduces a service that monitors the health of etcd by scraping its metrics.
- Stops requesting jobs when ectd is close to full (a soft limit).
- Stops pod creation when etcd is even closer to full (a hard limit).


internal/common/armadaerrors/errors.go
```diff
@@ -43,97 +43,6 @@ type ErrNoPermission struct {
 	Message string
 }
 
-// retryablePostgresErrors represents set of postgres errors that can be retried. Fundamentally these are all
-//issues with postgres itself, with the network or with authentication
-var retryablePostgresErrors = map[string]bool{
```
severinson, line 48: This code it just moved within the same file.

JamesMurkin (approved):

PR #1018, opened 2022-05-29
WIP: Finessing the eventapi

d80tb7:
- Migration job for eventdb
- Cleanup job for eventdb
- Removed caching (queue, jobset) -> ids
- Added retry logic for db operations


jimbobby5:

PR #1020, opened 2022-05-30
Add streaming gRPC service to transfer jobs from the server to the executor

severinson:
Currently, we send jobs to the executor in batches using unary gRPC calls. If such a call fails due to network error, those jobs are in limbo until the leases expire, which may be a long time.  The issue is that, for unary gRPC calls, the server has no way of knowing if sending a response back succeeded or not.

This PR adds a way of sending jobs over a bidirectional gRPC stream, where jobs are from the server to the executor and ids of received jobs are sent back to the server. Only acked jobs are marked as leased by the server.


internal/executor/service/job_lease.go
```diff
@@ -81,38 +84,148 @@ func (jobLeaseService *JobLeaseService) RequestJobLeases(availableResource *comm
 		Queues:     leasedQueueReports,
 	}
 
-	leaseRequest := api.LeaseRequest{
+	// Setup a bidirectional gRPC stream.
+	// The server sends jobs over this stream.
+	// The executor sends back acks to indicate which jobs were successfully received.
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
JamesMurkin, line 90: 10 seconds here is unfortunately quite short for some cases. I think it should be larger for now until scheduling is changed on the server side
 - It used to be 30s

internal/armada/repository/job.go
```diff
@@ -66,6 +66,7 @@ type JobRepository interface {
 	GetQueueJobIds(queueName string) ([]string, error)
 	RenewLease(clusterId string, jobIds []string) (renewed []string, e error)
 	ExpireLeases(queue string, deadline time.Time) (expired []*api.Job, e error)
+	ExpireLeasesById(jobIds []string, deadline time.Time) (expired []*api.Job, e error)
```
JamesMurkin, line 69: Is this still needed?

pkg/api/queue.proto
```diff
@@ -49,6 +49,23 @@ message LeaseRequest {
     repeated NodeInfo nodes = 7 [(gogoproto.nullable) = false];
 }
 
+// For the bidirectional streaming job lease request service.
+// For the first message, populate all fields except SubmittedJobs, which should be empty.
+// For subsequent messages, these fields may be left empty, in which case the last non-zero value received is used.
+// By streaming back job ids, the server knows which jobs were received in case of an outage.
+message StreamingLeaseRequest {
+    // Each cluster has a unique name associated with it.
+    string cluster_id = 1;
```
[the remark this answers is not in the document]
severinson: Fixed

internal/armada/repository/job.go
severinson: No, but I think it's nicer.

JamesMurkin (approved):

JamesMurkin (approved):

PR #1025, opened 2022-05-31
Make EtcdHealthMonitor handle some instances down + metrics

JamesMurkin:
EtcdHealthMonitor should handle some instances being down without disrupting submitting to the cluster

Etcd is often run as a cluster of 3 nodes

During maintenance it isn't unexpected one of these nodes would be down, so we shouldn't fail health checks just because one of the instances is down
 - I have added `minimumAvailable` to configure how many have to be available to consider there to be enough instances to perform the health check
 - We no longer break all scrapes etcd health because one failed

Also added metrics to allow seeing the current state Armada thinks etcd is in
 - Allows alerting if Armada thinks etcd is down (and hence submission will be blocked)
 - Allows debugging

Some other minor cleanup + adding tests


internal/executor/configuration/types.go
```diff
@@ -36,24 +36,29 @@ type KubernetesConfiguration struct {
...
-	MinimumPodAge                       time.Duration
-	StuckTerminatingPodExpiry           time.Duration
-	FailedPodExpiry                     time.Duration
-	MinimumJobSize                      common.ComputeResources
-	PodDefaults                         *PodDefaults
-	PendingPodChecks                    *podchecks.Checks
+	MetricUrls                      []string
+	FractionOfStorageInUseSoftLimit float64
+	FractionOfStorageInUseHardLimit float64
+	MinimumAvailable                int
```
severinson, line 61: I'd add a comment to explain what MinimumAvailable is.

internal/executor/etcd/healthmonitor.go
```diff
@@ -0,0 +1,306 @@
+package etcd
```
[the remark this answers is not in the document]
severinson: I think having the health monitor be part of the executor is good.

internal/executor/node/node_group_test.go
```diff
@@ -102,6 +104,17 @@ func TestFilterAvailableProcessingNodes(t *testing.T) {
 	assert.True(t, result, 1)
 }
 
+func TestMultiError(t *testing.T) {
+	var result *multierror.Error
+
+	result = multierror.Append(result, fmt.Errorf("Error1"))
+	result = multierror.Append(result, fmt.Errorf("Error2"))
+
+	fmt.Println("1" + result.Error())
```
carlocamurri, line 113: Remove this

carlocamurri (dismissed):

carlocamurri (approved):

PR #1032, opened 2022-05-31
Fix panic + improve panic handling

JamesMurkin:
The recovery interceptor must be the first interceptor

So that it can handle panics in the endpoint code + all subsequent interceptor code (i.e it wraps all the code that can panic)

Currently it is not the first interceptor, so if any interceptor that was added before it panics, the server will crash


kannon92 (changes requested):
I got bit by this earlier today, I think.  

Would it be possible to add tests for Errors?

PR #1033, opened 2022-05-31
Move linting and unit tests to Github Actions

jayofdoom:
This PR migrates our unit tests and linting to github actions. It uses some of the prior art from PR #675 as a guide, but was created greenfield otherwise.

The testing matrix is equivalent to our matrix in CircleCI -- we're testing both against go 1.16 and go 1.18.


jayofdoom:
Successful test passes can be seen in my fork:
https://github.com/jayofdoom/armada/actions/runs/2423754952
https://github.com/jayofdoom/armada/actions/runs/2423754957

.github/workflows/go.yml
```diff
@@ -0,0 +1,47 @@
+on: [push]
```
kannon92, line 1: You can add more options to this.  I think push is good.  

I've seen people have pull_request as one option also.  

I'm not sure if you need to change it though.  I've seen people use pretty elaborate rules but we aren't there yet.

kannon92 (approved):

.github/workflows/go.yml
jayofdoom: I have intentionally not pursued things like this yet, including e.g. only running the TS linter if TS files are changed... right now I'm trying to stick to replicating functionality provided by CircleCI.

jayofdoom:
Whoever merges this will have to mash buttons to change the required jobs for merge to remove the circleci check_and_unit_test_* required ones, and instead require the workflow to pass.

.github/workflows/go.yml
kannon92: What I was wondering is for github actions, will they run when a PR is merged into main?  

It says push but is a merge considered a push or a pull_request closed?

I'm not familar with circle CI so I don't know if circle CI has sensible defaults.

jayofdoom: That's a solid question, I'll check that behavior.

jayofdoom: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push

A merged pull request will trigger the push webhook. So it'll run on push.

(Just saying this out loud) -- We'll need to be more delicate about what we run jobs on once we add releasing to the github actions.

PR #1048, opened 2022-06-06
E2E Tests for the Event API

d80tb7:
All E2E test for the event api now pass


jimbobby5:

PR #1053, opened 2022-06-07
Update all k8s.io/* dependencies to 0.24.1

jayofdoom:
Dependabot updates for these were failing independently; upon further research they appear to be integrated and need to be updated atomically. This is an attempt to get those tests passing.


kannon92:
I ran into a lot of issues when I was on python client when I accidentally updated this apis.

https://github.com/kubernetes-sigs/controller-tools/issues/643

it looks like maybe an update of go may fix this.

jayofdoom:
This can't be done until Go 1.18 is the minimum supported version in our repo.

PR #1054, opened 2022-06-07
Annotations clickable (https://github.com/G-Research/armada/issues/700)

kannon92:
1) Added pulsar to docs/dev/setup.sh so we can dev test with pulsar.  @carlocamurri mentioned that this is what he uses for testing lookout
2) Combine pulsar script so we can reuse same pulsar priming
3) Add ability for urls to render as clickable links if annotations are a valid link.


internal/lookout/ui/src/components/job-dialog/DetailRow.tsx
```diff
@@ -9,14 +9,19 @@ type DetailRowProps = {
   value: string
   className?: string // Class to be applied to value <span> element
 }
-
+const isUrl = (value: string) => value.startsWith("http")
```
rankindaveyg, line 12: What about using a validation library, like https://www.npmjs.com/package/validator

jimbobby5:
Normally I'd recommend doing these as separate PRs, feels like the dev setup stuff for Pulsar is logically different from a UI change.

kannon92:
> Normally I'd recommend doing these as separate PRs, feels like the dev setup stuff for Pulsar is logically different from a UI change.

Yep.  I'll open them as two separate PRs.  Pulsar was added as default so I first got pulsar working before I started tackling UI change but I'll open them as separate PRs.

PR #1059, opened 2022-06-08
Add testsuite

severinson:


d80tb7:
Proposed Initial set of tests:

Test | Steps | Automatable
-- | -- | --
Submit a single Job that succeeds | <ul><li>Submit Hello World Job1</li><li> Read Events</li><li> Assert Succeeded Event</li></ul> | Yes (easy)
Submit a single Job that fails to pull image |<ul><li>Submit Job With Non-Existant Image</li><li>Read Events</li><li>Assert Job Failed Event</li></ul> | Yes (easy)
Submit a single Job that fails with OOM | <ul><li>Submit Job with xMB memory limit with command line that uses > xMB</li><li>Read Events</li><li>Assert Job Failed</li></ul>  | Yes (easy)
Submit a single job that succeeds close to memory limit | <ul><li>Submit Job with xMB memory limit with command line that uses 90% xMB</li><li>Read Events</li><li>Assert Job Succeeded Event</li></ul>  | Yes (easy)
Submit a job that uses fileshares | <ul><li>Submit Job that reads and writes to file shares</li><li>Read Events</li><li>Assert Job succeeded and files exist.</li></ul>  | Yes (easy)
Submit a duplicate job | <ul><li>Submit Hello World Job with ClientId x</li><li>Submit Job again with same client ID</li><li>Assert that second response indicates duplicate</li><li>Subscribe to Events</li><li>Assert Job Duplicate and Succeeded Messages</li></ul>  | Yes (easy)
Unschedulable Job | <ul><li>Submit Job with invalid taint</li><li>Assert rejected</li></ul>  | Yes (easy)
Batch Submit | <ul><li>Submit 1000 Hello world jobs in a single batch</li><li>Read Events</li><li>Assert Job Succeeded</li></ul>  | Yes (easy)
Performance |<ul><li> Submit 1,000,000 jobs across multiple batches.</li><li>Read Events</li><li>Assert all succeeded in reasonable time frame.</li></ul>  | Yes (hard as limited resources in dev might make this too slow- could use fake executor)
Performance |<ul><li> Submit large number of jobs across may queues such that we generate x thousand messages per second</li><li>.Read Events</li><li>Assert all succeeded in reasonable time frame.</li></li></ul>  | Yes (hard as limited resources in test env might make this too slow- could use fake executor)
Subscribe to middle of stream | <ul><li>Submit Hello World JobRead Events</li><li>Resubscribe to events with some offset</li><li>.Assert Events as expected</li></ul>  | yes (easy)
Cancel a job | <ul><li>Submit Hello World Job with a long sleep</li>Cancel job</li>Read Events</li>Assert Job Cancelled Event | yes (easy)
Cancel lots of jobs | <ul><li>Submit 1,000,000 jobsCancel jobsRead EventsAssert Job Cancelled Event for each job</li></ul>  | yes (easy if perf in dev ok)
Reprioritize Job | <ul><li>Submit Hello World Job with a medium sleep</li><li>Reprioritize job</li><li>Read Events</li><li>Assert Job Reprioritized Event</li></ul>  | Yes (easy- but only if asserting the event is good enough)
Reprioritize lots of jobs |<ul><li> Submit 1,000,000 jobs</li><li>Reprioritize jobs</li><li>Read Events</li><li>Assert Job Reprioritized Event for each job</li></ul>  | yes (easy if perf in dev ok)
Submit Job using python client |<ul><li> Submit Hello World Job using python client</li><li>Read Events</li><li>Assert Succeeded Event</li></ul>  | yes (easy)
Submit job using dotnet client | <ul><li>Submit Hello World Job using dotnet client</li><li>Read Events</li><li>Assert Succeeded Event</li></ul>  | yes (easy)
Lookout | <ul><li>Run above tests</li><li>Assert jobs have shown up in lookout with correct state</li></ul>  | No- Suggestion is to manually check lookout after the above tests and assert that results are as expected.
Binoculars | <ul><li>Submit Hello World Job with some logging</li><li>Go to lookout</li><li>Click on logs and assert you can see the job</li></ul>  | No (needs to be manual test)
Ingress |<ul><li> Submit Hello World Job with ingress that stops job when a http call made to ingress</li><li>Read events and pull out ingress info</li><li>Make call to ingress</li><li>Assert Succeeded Event</li></ul>  | yes (easy/medium)
HA | <ul><li>Submit job every half second for 5 mins<li>Kill one of the armada components (api, lookout ingester, pulsar, redis)</li><li>Subscribe to events</li><li>Assert all jobs succeeded</li></ul>  | Yes (medium/hard)
Postgres failover | <ul><li>Submit jobs<li>Kill postgres master</li><li>See that lookout keeps working</li></ul> | No
Cluster Failover and job timeout | <ul><li>Have two clusters active</li><li>Submit Jobs</li><li>Kill one cluster</li><li>See that lost jobs are rescheduled</li><li>See that new jobs are submitted to good cluster</li></ul>  | Yes (hard)
Queue Permissions | <ul><li>Submit to queue that user is not permissioned to</li><li>Assert that submit is rejected</li></ul>  | Yes (easy)
Namespace permissions | <ul><li>Submit to queue that user is permissioned to but under namespace that they are not permissioned to</li><li>.Subscribe to events<li>See job failed event</li></li></ul>  | yes (easy?)
Long running Subscription | <ul><li>Submit hello world jobs over long time period (hours)Subscribe to events over this period</li><li>Check that all events are as expected</li><li>Repeat for c# and python clients</li></ul> | Yes but hard because dev/staging clusters might not be stable enough?
Queue Ops | <ul><li>Create a queue<li>Check the Queue can be listed</li><li>Submit to the queue</li><li>Check that the job succeeds</li><li>Delete the queue</li><li>Check the queue can no longer be listed</li><li>Submit job to queue</li><li>Check that submit is rejected</li></ul> | yes (easy?)
Podspec too Large | <ul><li>Submit a job with a podspec > 64KB</li><li>See that the job is rejected</li></ul>  | yes (easy)
Too Much ephemeral storage | <ul><li>Submit a job that request storage but writes more data to disk than it asks for</li><li>Check that the job fails </li></ul> | yes (easy)

internal/armadactl/reprioritize.go
```diff
@@ -76,7 +74,7 @@ func (a *App) writeResults(results map[string]string) error {
 	}
 
 	if len(erroredIds) > 0 {
-		return fmt.Errorf("[armadactl.writeResults] error reprioritizing some jobs")
+		return fmt.Errorf("error reprioritizing some jobs")
```
dejanzele, line 77: I think it is always better to use `errors.Errorf` as from the `github.com/pkg/errors` as they include stack traces opposed to `fmt` package.

internal/armadactl/submit.go
```diff
@@ -14,44 +14,39 @@ import (
...
 // Submit a job, represented by a file, to the Armada server.
 // If dry-run is true, the job file is validated but not submitted.
-func (a *App) Submit(path string, dryRun bool) (outerErr error) {
+func (a *App) Submit(path string, dryRun bool) error {
 
 	ok, err := validation.ValidateSubmitFile(path)
 	if !ok {
-		fmt.Fprintf(a.Out, "Invalid jobfile: %s/n", err)
-		return nil
+		return err
```
dejanzele, line 21: I am a big fan of using `errors.Wrap` from `github.com/pkg/errors` to provide additional context to errors.
i.e. `errors.Wrap(err, "error validating submited file")`, it is especially useful if the errors in `validation.ValidateSubmitFile` do not give info on the calling context.

internal/armadactl/submit.go
```diff
@@ -14,44 +14,39 @@ import (
...
-	client.WithConnection(a.Params.ApiConnectionDetails, func(conn *grpc.ClientConn) {
-		submissionClient := api.NewSubmitClient(conn)
+	return client.WithSubmitClient(a.Params.ApiConnectionDetails, func(c api.SubmitClient) error {
 		for _, request := range requests {
-			response, err := client.SubmitJobs(submissionClient, request)
+			response, err := client.SubmitJobs(c, request)
 			if err != nil {
-				outerErr = fmt.Errorf("[armadactl.Submit] error submitting job with request %#v: %s", request, err)
-				return
+				return errors.WithMessagef(err, "error submitting request %#v", request)
```
dejanzele, line 39: Just looked at docs for [WithMessage](https://pkg.go.dev/github.com/pkg/errors#hdr-Adding_context_to_an_error), and honestly I do not understand when to use `Wrap` and when to use `WithMessage`

internal/testsuite/eventwatcher/eventwatcher.go
```diff
@@ -0,0 +1,266 @@
...
+		select {
+		case <-parent.Done():
+			return g.Wait()
+		case <-ctx.Done(): // errgroup cancelled
+			return g.Wait()
+		case msg := <-C:
+			if ingressInfo := msg.GetIngressInfo(); ingressInfo != nil {
+				for _, host := range ingressInfo.IngressAddresses {
+					ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+					defer cancel()
```
dejanzele, line 204: Should the `defer` run at the end of for iteration or at the end of the function call?

internal/testsuite/eventwatcher/eventwatcher.go
```diff
@@ -0,0 +1,266 @@
...
+				}
+			}
+		}
+	}
+}
+
+func getFromIngress(ctx context.Context, host string) error {
+	tokens := strings.Split(host, ".")
+	if len(tokens) == 0 {
+		return fmt.Errorf("malformed hostname: %s", host)
```
dejanzele, line 216: I'd replace `fmt` with `github.com/pkg/errors` package

internal/testsuite/eventwatcher/eventwatcher.go
```diff
@@ -0,0 +1,266 @@
...
+	// which may handle ingress on an arbitrary port.
+	ingressPort := os.Getenv("ARMADA_EXECUTOR_INGRESS_PORT")
+	if ingressPort == "" {
+		ingressPort = "80"
+	}
+
+	// Make a get request to test that the ingress works.
+	// This assumes that whatever the ingress points to responds.
+	httpClient := &http.Client{}
+	httpReq, err := http.NewRequest("GET", url+":"+ingressPort+"/", nil)
```
dejanzele, line 232: I'd rather use `NewRequestWithContext` instead of `NewRequest`.

For the second param, my preference is always in favour of `fmt.Sprintf`

i.e. `fmt.Sprintf("%s:%s/", url, ingressPort).

For the body param, a lot of people suggest to use `http.NoBody` in favour to `nil`

internal/testsuite/eventwatcher/eventwatcher.go
```diff
@@ -0,0 +1,266 @@
...
+		case <-ctx.Done():
+			// When cancelled, return the most recent request err,
+			// or if there have been no errors (context cancelled before the first attempt)
+			// return the context error.
+			if requestErr == nil {
+				return ctx.Err()
+			}
+			return requestErr
+		default:
+			httpRes, err := httpClient.Do(httpReq)
```
dejanzele, line 252: The response body should be closed in a defer.

`defer httpRes.Body.Close()`

internal/testsuite/submitter/submitter.go
```diff
@@ -0,0 +1,115 @@
...
+
+		// Submit jobs.
+		for srv.NumBatches == 0 || numBatchesSent < srv.NumBatches {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-tickerCh:
+				res, err := c.SubmitJobs(ctx, req)
+				if err != nil {
+					return errors.WithStack(errors.WithMessage(err, "error submitting jobs"))
```
dejanzele, line 97: I think this line could be replaced with `errors.Wrap(err, "error submitting jobs")`

internal/armadactl/analyze.go
```diff
@@ -6,30 +6,26 @@ import (
...
 	fmt.Fprintf(a.Out, "Querying queue %s for job set %s\n", queue, jobSetId)
-	client.WithConnection(a.Params.ApiConnectionDetails, func(conn *grpc.ClientConn) {
-		eventsClient := api.NewEventClient(conn)
-
+	return client.WithEventClient(a.Params.ApiConnectionDetails, func(c api.EventClient) error {
 		events := map[string][]*api.Event{}
 		var jobState *domain.WatchContext
 
-		client.WatchJobSet(eventsClient, queue, jobSetId, false, true, context.Background(), func(state *domain.WatchContext, e api.Event) bool {
+		client.WatchJobSet(c, queue, jobSetId, false, true, context.Background(), func(state *domain.WatchContext, e api.Event) bool {
```
kannon92, line 20: Can we use a more descriptive name than c?

pkg/api/testspec.proto
```diff
@@ -0,0 +1,46 @@
+syntax = 'proto3';
```
kannon92, line 1: I don't think we should put this in our public facing api.

pkg/api/testspec.proto
```diff
@@ -0,0 +1,46 @@
...
+import "pkg/api/submit.proto";
+import "pkg/api/event.proto";
+
+option (gogoproto.goproto_stringer_all) = false;
+option (gogoproto.stringer_all) = true;
+
+// Defines a test case for the Armada test suite.
+// Defined as a proto message to enable unmarshalling oneof fields.
+message TestSpec {
+	// Jobs to submit.
```
kannon92, line 16: The whitespace on this proto file is not consistent.

internal/armadactl/submit.go
severinson: Wrap adds both a message and a stack trace.

internal/armadactl/reprioritize.go
severinson: Done

internal/armadactl/submit.go
severinson: I don't think we should add a stack trace here; the stack trace should be added where the error first occurs. This code only annotates the error with some additional info.

internal/testsuite/eventwatcher/eventwatcher.go
```diff
@@ -0,0 +1,266 @@
...
+				}
+			}
+		}
+	}
+}
+
+func isTerminalEvent(msg *api.EventMessage) bool {
+	switch msg.Events.(type) {
+	case *api.EventMessage_Failed:
+		return true
```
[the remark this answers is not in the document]
severinson: Only with an explicit fallthrough statement though?
Just omitting the returns would result in the function returning the wrong thing, no?

internal/testsuite/eventwatcher/eventwatcher.go
severinson: End of the function call.
In fact, it doesn't need to run at all; I've removed the defer to make it cleaner.

internal/testsuite/eventwatcher/eventwatcher.go
severinson: Done

internal/testsuite/eventwatcher/eventwatcher.go
severinson: Done

internal/testsuite/eventwatcher/eventwatcher.go
severinson: Done. We can even close it immediately.

internal/testsuite/submitter/submitter.go
severinson: I think having only errors.WithMessage is better. Since a stack trace should be captured at the point at which the error occurs,

pkg/client/load-test.go
```diff
@@ -124,17 +123,15 @@ func (apiLoadTester ArmadaLoadTester) runSubmission(ctx context.Context, submiss
...
+	go WithSubmitClient(apiLoadTester.apiConnectionDetails, func(client api.SubmitClient) error {
 		defer submissionComplete.Done()
 
-		client := api.NewSubmitClient(connection)
-
 		e := CreateQueue(client, &api.Queue{Name: queue, PriorityFactor: priorityFactor})
 		if status.Code(e) == codes.AlreadyExists {
 			log.Infof("Queue %s already exists so no need to create it.\n", queue)
 		} else if e != nil {
 			log.Errorf("ERROR: Failed to create queue: %s because: %s\n", queue, e)
```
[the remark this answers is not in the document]
severinson: It should. But it may be better to make a separate PR to fix this.
The entire client could do with a once-over...

internal/armadactl/analyze.go
severinson: How about `ec`? The scope of the variable is quite limited.

pkg/api/testspec.proto
severinson: I agree. Unfortunately I recall there being some issues around importing proto files from other locations in the project, which makes it difficult to put this file anywhere else...

pkg/api/testspec.proto
severinson: How so?

cmd/testsuite/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+
+			testFiles, err := filepath.Glob(testFilesPattern)
+			if err != nil {
+				return err
+			}
+
+			// Crate a context that is cancelled on SIGINT/SIGTERM.
+			// Ensures test jobs are cancelled on ctrl-C.
+			ctx, cancel := context.WithCancel(context.Background())
```
d80tb7, line 85: maybe not for now but we should definitely make a utility function for this

cmd/testsuite/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+				}
+			}()
+
+			numSuccesses := 0
+			numFailures := 0
+			start := time.Now()
+			for _, testFile := range testFiles {
+				testStart := time.Now()
+				err := app.TestFile(ctx, testFile)
+				fmt.Printf("\nRuntime: %s\n", time.Since(testStart))
```
d80tb7, line 104: what happens if we get a panic?- I think most test runners will just treat it as a failure and move on to the next test.  Do we want to do that?

makefile
```diff
@@ -303,6 +313,9 @@ tests-e2e-setup: setup-cluster
 		-e ARMADA_APICONNECTION_FORCENOTLS=true \
 		armada-executor --config /e2e/setup/insecure-executor-config.yaml
 
+	# Create test queue if it doesn't already exist
```
kannon92, line 316: Wouldn't this fail the build if this queue exists?

cmd/testsuite/cmd/root.go
severinson: I prefer not having a utility function. I think this is more clear.

internal/testsuite/eventwatcher/eventwatcher.go
```diff
@@ -0,0 +1,266 @@
...
+	numRemaining := len(jobIds)
+	exitedByJobId := make(map[string]bool)
+	for {
+		select {
+		case <-parent.Done():
+			return nil
+		case msg := <-C:
+			if e := msg.GetSubmitted(); e != nil {
+				numActive++
+			} else if e := msg.GetSucceeded(); e != nil {
```
d80tb7, line 148: Do we need to handle the cancelled case here?

internal/testsuite/eventwatcher/eventwatcher.go
```diff
@@ -0,0 +1,266 @@
...
+					g.Go(func() error { return getFromIngress(ctxWithTimeout, host) })
+				}
+			}
+		}
+	}
+}
+
+func getFromIngress(ctx context.Context, host string) error {
+	tokens := strings.Split(host, ".")
+	if len(tokens) == 0 {
```
d80tb7, line 214: I'm not sure I follow this- way does taking the last token here yield a valid host?

pkg/api/testspec.proto
d80tb7: Maybe I'm looking at it wrong but I agree with Kevin here.  At least t he way github is rendering it, some of the fields seems to be indented an extra tab

d80tb7 (approved):

makefile
severinson: No. Because of the `|| true`.

internal/testsuite/app_test.go
```diff
@@ -0,0 +1,8 @@
+package testsuite
+
+import (
+	"testing"
+)
+
+func TestFoo(t *testing.T) {
```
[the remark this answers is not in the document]
severinson: Deleting it.

internal/testsuite/eventwatcher/eventwatcher.go
severinson: You're right; this is wrong. Fixing it.

Issue #702, opened 2021-09-17
Ensure armada permission checking is fully unit tested


robertdavidsmith:
It would be great if someone could fully unit test the permission checks in https://github.com/G-Research/armada/blob/master/internal/armada/server/submit.go 

This has some coverage already (https://github.com/G-Research/armada/blob/master/internal/armada/server/submit_test.go), but not everything is covered. Given this code is security critical, it should have 100% unit test coverage.

andrewpcone:
When `SubmitJobs` causes a queue to be created, that queue gets the groups of present principal, including the group `"everyone"`. As a result, anyone with the `permissions.SubmitJobs` can submit jobs to that queue.

I've committed code that excludes the `"everyone"` group from queue ownership.

Moreover, I am skeptical that the creating principal's groups should automatically have privileges on the new queue. The only thing that we know for sure is that the user themselves should have all privileges. I think it's reasonable (nay optimal) to require the user to grant other users/groups by subsequent API calls.

Thoughts @severinson ?

PR #1067, opened 2022-06-11
Permissions testing

andrewpcone:
This pull request addresses #702

### DIFF
--- a/internal/armada/permissions/permissions.go
+++ b/internal/armada/permissions/permissions.go
@@ -18,6 +18,5 @@ const (
 	ReprioritizeAnyJobs                       = "reprioritize_any_jobs"
 	WatchEvents                               = "watch_events"
 	WatchAllEvents                            = "watch_all_events"
-
-	ExecuteJobs = "execute_jobs"
+	ExecuteJobs                               = "execute_jobs"
 )
--- a/internal/common/auth/authorization/permissions.go
+++ b/internal/common/auth/authorization/permissions.go
@@ -38,9 +38,9 @@ func NewPrincipalPermissionChecker(
 // has that permission.
 func (checker *PrincipalPermissionChecker) UserHasPermission(ctx context.Context, perm permission.Permission) bool {
 	principal := GetPrincipal(ctx)
-	return hasPermission(perm, checker.permissionScopeMap, func(scope string) bool { return principal.HasScope(scope) }) ||
-		hasPermission(perm, checker.permissionGroupMap, func(group string) bool { return principal.IsInGroup(group) }) ||
-		hasPermission(perm, checker.permissionClaimMap, func(claim string) bool { return principal.HasClaim(claim) })
+	return hasPermission(perm, checker.permissionScopeMap, principal.HasScope) ||
+		hasPermission(perm, checker.permissionGroupMap, principal.IsInGroup) ||
+		hasPermission(perm, checker.permissionClaimMap, principal.HasClaim)
 }
 
 // UserOwns check if obj is owned by the principal contained in the context,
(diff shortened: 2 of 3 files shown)


kannon92 (dismissed):
Looks good to me.  I would like a person who is more familar with Armada permissions to look this over though.

robertdavidsmith:
Hi,

Thanks a lot for the PR.

Please can you update the title and description to fully cover what the PR does, as it's a bit more than just adding tests.

e.g. there are some very welcome improvements to error messages, and also some special casing of group "everyone" that wasn't there before.

Thanks,

Rob

robertdavidsmith:
Please can you add unit tests for the removal of "everyone", including the case where "everyone" is not present (which looks like it will fail currently).

I agree adding the groups to queue perms is a bit funny, I believe GR has an internal use case that needs this. Someone such as @JamesMurkin may know more. I would love to remove auto queue creation eventually.

Thanks,

Rob

JamesMurkin:
Hi,

I think we all agree that we want to drop auto queue creation, sadly we have an internal use case that relies on it. We'll have to migrate them off before we can remove it.

I will initiate this internally however I can't give a timeline quite yet, so we'll have to keep it.

internal/armada/server/submit.go
```diff
@@ -270,10 +271,15 @@ func (server *SubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmitRe
 
 	q, err := server.getQueueOrCreate(ctx, req.Queue)
 	if err != nil {
-		return nil, err
+		code := codes.Unknown
+		if e, ok := status.FromError(err); ok {
+			code = e.Code()
+		}
+
+		return nil, status.Errorf(code, "[SubmitJobs] couldn't get/make queue: %s", err)
```
JamesMurkin, line 279: I think we have moved away from adding the function name into the error messages

internal/armada/server/submit.go
```diff
@@ -270,10 +271,15 @@ func (server *SubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmitRe
 
 	q, err := server.getQueueOrCreate(ctx, req.Queue)
 	if err != nil {
-		return nil, err
+		code := codes.Unknown
+		if e, ok := status.FromError(err); ok {
+			code = e.Code()
```
JamesMurkin, line 276: Maybe one for a separate PR but we now have something that magically sets the return code base on error type:
https://github.com/G-Research/armada/blob/master/internal/common/armadaerrors/errors.go#L251

I think the plan was to move to using the appropriate error types rather than setting codes explicitly

internal/armada/server/submit.go
```diff
@@ -689,35 +695,41 @@ func (server *SubmitServer) checkReprioritizePerms(ctx context.Context, jobs []*
...
+		}
+		if !server.permissions.UserHasPermission(ctx, permissions.SubmitAnyJobs) {
+			return nil, status.Errorf(codes.PermissionDenied, "Queue %s not found; won't create because user lacks SubmitAnyJobs permission", queueName)
 		}
 
 		principal := authorization.GetPrincipal(ctx)
-
+		groupNames := principal.GetGroupNames()
+		everyone_index := slices.Index(groupNames, "everyone")
+		groupNames = append(groupNames[:everyone_index], groupNames[everyone_index+1:]...)
```
JamesMurkin, line 717: Several nit picks here:
 - Make make remove string from slice a separate function so it can be tested (there are some util libraries in common
  - Use the "everyone" constant https://github.com/G-Research/armada/blob/master/internal/common/auth/authorization/common.go#L23
   - Have a comment saying why we want to remove this

kannon92 (approved):

andrewpcone:
@dejanzele OK to merge?

PR #1070, opened 2022-06-13
Improve utilisation data collection

JamesMurkin:
When there are bad nodes, sometimes our utilisation gathering gets stuck
 - Now skip collection of metrics for NotReady nodes
 - Add a timeout of 15s so don't hang forever. Not all "bad" nodes are NotReady

### DIFF
--- a/internal/executor/context/cluster_context.go
+++ b/internal/executor/context/cluster_context.go
@@ -1,7 +1,7 @@
 package context
 
 import (
-	ctx "context"
+	"context"
 	"encoding/json"
 	"fmt"
 	"time"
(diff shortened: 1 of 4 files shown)


internal/executor/context/cluster_context.go
```diff
@@ -1,7 +1,7 @@
 package context
```
JamesMurkin, line 1: Rename package to clustercontext

JamesMurkin: This causes lots of changes, will do this as a separate PR

carlocamurri (approved):

PR #1073, opened 2022-06-13
Fix queueReports being sent to server

JamesMurkin:
A queue report contains:
 - resource allocated
 - resource used
 - count of pods per phase

However currently if there are no running pods for a queue, we don't create a queue report for that queue

That means the count of pods per phase is incorrect when there are no running pods
 - As there may be pods in Pending, Succeeded, Failed

Now we calculate a QueueReport based on if any pods exist for a given queue - regardless of their phase


internal/executor/utilisation/cluster_utilisation.go
```diff
@@ -278,30 +278,17 @@ func getAllPodsRequiringResourceOnProcessingNodes(allPods []*v1.Pod, processingN
 }
 
 func (clusterUtilisationService *ClusterUtilisationService) createReportsOfQueueUsages(pods []*v1.Pod) []*api.QueueReport {
```
JamesMurkin, line 280: Write test

carlocamurri (approved):

PR #1076, opened 2022-06-14
Enable configuration of fatal submission errors when creating pods

carlocamurri:
Currently there are a some hard-coded rules for when we want to instantly fail a job instead of retrying when a pod creation operation on an executor cluster fails.

This PR retains the same logic, whilst enabling configuration for when to instantly fail jobs when they match specified regex:
* Always instantly fail a job if we get a Kubernetes `APIError` with a `StatusReasonInvalid` or `StatusReasonForbidden`
* Instantly fail a job if the error string matches any one of the specified regexes in the configuration, in `kubernetes.fatalPodSubmissionErrors`
* Retry every other error


docs/dev/kind.yaml
```diff
@@ -1,5 +1,14 @@
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
+kubeadmConfigPatches:
```
kannon92, line 3: Could you comment on why you added this to dev run?

carlocamurri: This is to enable validating/mutating webhooks on the local kind cluster, in case you might want to test using them. Can remove from this PR if it feels out of scope

kannon92: Could you leave that comment in the PR?

internal/executor/job/submit.go
```diff
@@ -218,17 +207,21 @@ func exposesPorts(job *api.Job, podSpec *v1.PodSpec) bool {
...
+			status.Reason == metav1.StatusReasonForbidden {
+			return false
+		}
 	}
 
-	//This message shows it was rejected by an admission webhook.
-	// By default admission webhooks blocking results in a 500 so we can't use the status code as we could confuse it with Kubernetes outage
-	if strings.Contains(status.Message, admissionWebhookValidationFailureMessage) {
-		return true
+	for _, errorMessage := range allocationService.fatalPodSubmissionErrors {
```
JamesMurkin, line 219: Please make the logic the same as before

JamesMurkin (approved):

Issue #1027, opened 2022-05-31
Create documentation for the python client


rankindaveyg:
Needs instructions for both how to use it, and how to do development.

jayofdoom:
Kevin and I broke this down into smaller steps:
- Add type hints and parameter documentation to pydoc strings at top of each method
- Implement an MVP of sphinx (?) generating the API docs (ugly is OK)
- Make it pretty/put it in the existing marketing site -- might want to ask where the best place is for it
- Have a small set of examples on how to use the client
- Ensure someone unfamiliar with python is able to use dev docs to install/develop on client

jayofdoom:
> Add type hints and parameter documentation to pydoc strings at top of each method
> Implement an MVP of sphinx (?) generating the API docs (ugly is OK)
> Make it pretty/put it in the existing marketing site -- might want to ask where the best place is for it

All completed; the last two items are left as low hanging fruit for the incoming interns.

PR #1084, opened 2022-06-15
Python client: Improve docstrings, generate docs

jayofdoom:
- Improves docstrings and makes them compliant with RST format
  expected by sphinx.
- Adds configuration and automation for generating jekyll-markdown
  files from docstrings (for use with gh-pages)
- Adds documenation on how to use the generation.

The docstrings in the code can still use significant improvement
as our understanding of the API and python types involved improve,
however this is better than what we have today and provides a basis
for future improvement.

First half of changes needed for #1027.


client/python/armada_client/client.py
```diff
@@ -89,36 +123,72 @@ def reprioritize_jobs(
         return response
 
     def create_queue(self, name: str, **queue_params):
-        """Implementation of CreateQueue rpc function"""
+        """Create the queue by name.
+
+        TODO(JayF) what are queue params?
```
kannon92, line 128: Its a proto object that represents the Queue object.  

Its a nested object.

https://github.com/G-Research/armada/blob/master/pkg/api/submit.proto#L89

kannon92 (dismissed):

client/python/armada_client/client.py
jayofdoom: That's the type in general terms, yes, but it's not the python type of the object. That's what I really need to make these strings more accurate with typing. 

I did mean to remove these TODOs though, I'll do that now.

kannon92: ACK.

jayofdoom: Fixed via amended commit.

kannon92 (approved):

severinson (approved):

PR #1086, opened 2022-06-16
Add python client docs to docs/, run in CI

jayofdoom:
- Adds a tox target, which runs in CI, which only passes if the
  generated version of the python armada client docs are the same
  as the version committed.
- Adds a new top level page to armadaproject.io (will be added to
  menu in separate PR once this merges) containing basic
  information about libraries provided for armada usage. This will
  be improved significantly over time as we develop examples for
  our clients and generally make them more friendly.
- Initial commit of generated python armada client docs.

This is part 2 of issue #1027


kannon92 (dismissed):

kannon92 (approved):

richscott (approved):

.github/workflows/python.yml
```diff
@@ -7,53 +7,25 @@ on:
...
-      #----------------------------------------------
-      # install your root project, if required
-      #----------------------------------------------
-      - name: Install library
-        working-directory: client/python
-
-        run: poetry install --no-interaction
-      #----------------------------------------------
+          python-version: '3.9'
+      - run: pip install tox
```
severinson, line 20: I'd suggest always using a specific version.
To avoid new versions potentially breaking the build.

docs/libraries.md
```diff
@@ -0,0 +1,30 @@
+# Libraries 
+
+## Overview
+Client libraries for Armada are under active development. This document is intended to document what clients exist and how to use them.
```
severinson, line 4:
```suggestion
This document is an overview of Armada client libraries.
```

docs/libraries.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Python
+Armada provides python client bindings providing access to the RPC API.
```
severinson, line 7:
```suggestion
Here, we give an overview of the Armada Python client. 
```

.github/workflows/python.yml
jayofdoom: I can do that if you feel strongly; but `tox` is a tool used for coordinating CI runs moreso than something we actually depend on. I've been using it for nearly a decade and have never seen a release that broke things (without loud messages about deprecation in advance). For that reason, I generally prefer leaving it unpinned (so the deprecation warnings are surfaced when we still have months to fix them).

severinson: All right. In that case I agree.

severinson (approved):

PR #1087, opened 2022-06-16
Search Events api addition

kannon92:
Add a new client api call to search events for finished jobs.

Modify unit tests so they don't fail if armada is running.  

Adding unit tests for search events.


client/python/tests/unit/test_search_for_job_complete.py
```diff
@@ -0,0 +1,72 @@
...
+def test_cancelled_event():
+    event_stream_message = event_pb2.EventStreamMessage()
+    event_message = event_pb2.EventMessage()
+    event_stream_message.id = 'test'
+
+    cancelled_event = event_pb2.JobCancelledEvent(job_id="id")
+    event_message.cancelled.CopyFrom(cancelled_event)
+    event_stream_message.message.CopyFrom(event_message)
+
+    events = [event_stream_message]
```
jayofdoom, line 55: This is going to be memory+compute intensive to make a list out of these. Is there a reason you can't pass this (I'm assuming it's a form of generator?) directly into search_for_job_complete?

client/python/armada_client/client.py
```diff
@@ -183,6 +183,47 @@ def get_queue_info(self, name: str):
...
+def search_for_job_complete(event, job_name: str, job_id: str) -> List[str]:
+    """Search the event stream to see if your job has finished running
+
+    :param event: a gRPC event stream
+    :param job_name: The name of your armada job
+    :param job_id: The name of the job id that armada assigns to it
+    :return: A list of [state, message]
+    """
+
+    tuple_return = ["temp", "temp2"]
```
jayofdoom, line 195: this is wrongish... a tuple in python is `('item1', 'item2')` (for N items)

Is there a reason you didn't want it to look like:
```
tuple_return = (
    "queued",
    f"Armada job:id {job_name}:{job_id} is queued"
)
```

Or alternatively, leave it a list (but don't pre-add items you don't need) and add the stuff as you do now, and rename the var.

client/python/tests/unit/test_search_for_job_complete.py
```diff
@@ -0,0 +1,72 @@
...
+    event_stream_message.id = 'test'
+
+    successful_event = event_pb2.JobSucceededEvent(job_id="id")
+    event_message.succeeded.CopyFrom(successful_event)
+    event_stream_message.message.CopyFrom(event_message)
+
+    events = [event_stream_message]
+    job_complete = search_for_job_complete(events, job_name="test", job_id="id")
+    assert job_complete[0] == "successful"
+    assert job_complete[1] == f"Armada job:id test:id succeeded"
```
jayofdoom, line 16: This does not need the `f` before as it has nothing to fill inside.

client/python/tests/unit/test_search_for_job_complete.py
```diff
@@ -0,0 +1,72 @@
...
+    event_stream_message.id = 'test'
+
+    failed_event = event_pb2.JobFailedEvent(job_id="id", reason="test fail")
+    event_message.failed.CopyFrom(failed_event)
+    event_stream_message.message.CopyFrom(event_message)
+
+    events = [event_stream_message]
+    job_complete = search_for_job_complete(events, job_name="test", job_id="id")
+    assert job_complete[0] == "failed"
+    assert job_complete[1] == f"Armada job:id test:id failed\nfailed with reason test fail"
```
jayofdoom, line 30: This does not need the `f` before as it has nothing to fill inside.

client/python/tests/unit/test_search_for_job_complete.py
```diff
@@ -0,0 +1,72 @@
...
+    event_stream_message.id = 'test'
+
+    cancelling_event = event_pb2.JobCancellingEvent(job_id="id")
+    event_message.cancelling.CopyFrom(cancelling_event)
+    event_stream_message.message.CopyFrom(event_message)
+
+    events = [event_stream_message]
+    job_complete = search_for_job_complete(events, job_name="test", job_id="id")
+    assert job_complete[0] == "cancelling"
+    assert job_complete[1] == f"Armada job:id test:id cancelling"
```
jayofdoom, line 44: This does not need the `f` before as it has nothing to fill inside.

client/python/tests/unit/test_search_for_job_complete.py
```diff
@@ -0,0 +1,72 @@
...
+    event_stream_message.id = 'test'
+
+    cancelled_event = event_pb2.JobCancelledEvent(job_id="id")
+    event_message.cancelled.CopyFrom(cancelled_event)
+    event_stream_message.message.CopyFrom(event_message)
+
+    events = [event_stream_message]
+    job_complete = search_for_job_complete(events, job_name="test", job_id="id")
+    assert job_complete[0] == "cancelled"
+    assert job_complete[1] == f"Armada job:id test:id cancelled"
```
jayofdoom, line 58: This does not need the `f` before as it has nothing to fill inside.

client/python/tests/unit/test_search_for_job_complete.py
```diff
@@ -0,0 +1,72 @@
...
+    event_stream_message.id = 'test'
+
+    terminated_event = event_pb2.JobTerminatedEvent(job_id="id")
+    event_message.terminated.CopyFrom(terminated_event)
+    event_stream_message.message.CopyFrom(event_message)
+
+    events = [event_stream_message]
+    job_complete = search_for_job_complete(events, job_name="test", job_id="id")
+    assert job_complete[0] == "terminated"
+    assert job_complete[1] == f"Armada job:id test:id terminated"
```
jayofdoom, line 72: This does not need the `f` before as it has nothing to fill inside.

jayofdoom (commented):
Overall OK, some room for improvement as noted. I'm surprised the unnecessary f-strings in the tests are passing lint.

client/python/tests/unit/test_search_for_job_complete.py
```diff
@@ -11,6 +11,9 @@ def test_succeded_event():
     event_message.succeeded.CopyFrom(successful_event)
     event_stream_message.message.CopyFrom(event_message)
 
+    # We are constructing a EventMessage and creating a list out of it
+    # This is to mock a grpc stream.
+    # IRL this would be given to us from a grpc call
```
jayofdoom, line 16:
```suggestion
    # In a real use case, this is a generator, not a list
```

jayofdoom (approved):

client/python/armada_client/client.py
```diff
@@ -183,6 +183,46 @@ def get_queue_info(self, name: str):
...
+            break
+        if element.message.cancelling.job_id == job_id:
+            job_state = "cancelling"
+            job_message = f"Armada {job_name}:{job_id} cancelling"
+            break
+        if element.message.cancelled.job_id == job_id:
+            job_state = "cancelled"
+            job_message = f"Armada {job_name}:{job_id} cancelled"
+            break
+        if element.message.terminated.job_id == job_id:
```
d80tb7, line 218: terminated and cancelling aren't actually terminal states.

client/python/armada_client/client.py
```diff
@@ -183,6 +183,46 @@ def get_queue_info(self, name: str):
         return response
 
 
+def search_for_job_complete(event, job_name: str, job_id: str) -> Tuple[str, str]:
```
d80tb7, line 186: I understand why we have this but I'm not a massive fan of it.  This is potentially going to fetch millions of events from the server and filter them all out client side, just to read a single event.

I think long term we should make the lookout api into a public api that stores the state of the world and allows clients discover such facts with a much more efficient query.

severinson:
Like Chris, I have some concerns about this. It may be better that clients like the AirFlow operator subscribe to all events and continually update its internal state than it getting lots of events to look at a single job.

kannon92:
I'm not sure I follow.  Updating internal state for airflow?

I was thinking it may be useful to group airflow jobs by a single Job set so the events should be small.  It's also nice to cancel them in a group.  

Should we add a new rpc call that fetches the job status per id?  I think that would clean this code up

client/python/armada_client/client.py
d80tb7: so terminated indicates that the pod was terminated by k8s, but doesn't preclude it being rerun.  You should always get a subsequent `Succeeded`, `Cancelled` or `Failed` message.

d80tb7:
@kannon92  here's a bit of background information that might be helpful.

- Providing an endpoint to let you know the current state of your jobs was what the precursor to armada did.  The problem with this was that clients would continuously poll to determine if their jobs had finished, which posed serious scalability problems as the number of jobs (and thus the number of poll requests) rose.
- The solution to this was the events endpoint that you're using here.  In this case users could subscribe to events about their jobs and gent notifications only when a new event is available. This gives a good efficiency win for the (common) case in which people just want to wait until all their jobs have finished, but is less good for answering state-of-the-world type questions as it's up to the user to parse the entire event stream and reconstruct the state themselves.

My (personal) view on this is that we should migrate the api currently used by lookout to be a public api that could answer such  SOW questions and be used by clients (python, go, c# armadactl).  Examples of such queries would be:
- What are the states of all jobs in jobset x
- What yaml was submitted for Job y
- How long has job z been running for

The caveat here would be that this api should not be used for applications that want to be continuiously updated.  Basically if you're polling this api for updates, you're doing it wrong adn you should be using the event api that's more scalable.

As to how this relates to this PR, I guess the question is wat we're trying to achieve here.  If the answer is that this is something the airflow operator needs to know if all the jobs have finished, then the answer is probably to subscribe to the events.  If this is something we think end users will use as a one off call then by all means add it, but the caveat is that it will be very inefficient and that using it on large jobs sets could cause issues.

kannon92:
So the airflow operator PR is https://github.com/G-Research/armada/pull/1092

Basically airflow needs to know when a single job finishes so that it can start the next job in a dag.

I don't really need the Jobset functionality but following an individual job.  Once that job finishes airflow marks that task as complete and schedules a new task.

Can we maybe add JobId to the GetJobSetEvents? For airflow users we would only show filtered events?

Airflow isn't really designed for streaming or subscription.  Their suggestion is not use it for event based apis.

d80tb7:
Ok so this presents a problem.

As it stands this is going to be very inefficient

1) You're subscribing to to a stream of events and then throwing most of them away.  
2) You're going to be subscribing one stream of events per job.  
3) You're going to be polling so you're doing 1) and 2) many ties for each job.

This is largely the model we used before Armada, but we found it didn't scale. Consider Armada managing > 100K cores with one job per core.  This means that you'll have 100K separate users trying to poll you to ask if their job is done yet.

I don't know enough about Airflow to understand what the right solution is here. Maybe we should have a meeting and work out how to proceed.

kannon92:
> Ok so this presents a problem.
> 
> As it stands this is going to be very inefficient
> 
> 1. You're subscribing to to a stream of events and then throwing most of them away.
> 2. You're going to be subscribing one stream of events per job.
> 3. You're going to be polling so you're doing 1) and 2) many ties for each job.
> 
> This is largely the model we used before Armada, but we found it didn't scale. Consider Armada managing > 100K cores with one job per core. This means that you'll have 100K separate users trying to poll you to ask if their job is done yet.
> 
> I don't know enough about Airflow to understand what the right solution is here. Maybe we should have a meeting and work out how to proceed.



> Ok so this presents a problem.
> 
> As it stands this is going to be very inefficient
> 
> 1. You're subscribing to to a stream of events and then throwing most of them away.
> 2. You're going to be subscribing one stream of events per job.
> 3. You're going to be polling so you're doing 1) and 2) many ties for each job.
> 
> This is largely the model we used before Armada, but we found it didn't scale. Consider Armada managing > 100K cores with one job per core. This means that you'll have 100K separate users trying to poll you to ask if their job is done yet.
> 
> I don't know enough about Airflow to understand what the right solution is here. Maybe we should have a meeting and work out how to proceed.

Let's have that meeting.  I'm demoing airflow to Michell and Adam Pitt our airflow work.  I'll bring up that this isn't ready for prod yet but I'd like to get a path forward.

kannon92:
We had a good discussion and we decided to move this function out of client for now.  I will go ahead and close this PR.

PR #1088, opened 2022-06-17
Adding e2e tests

kannon92:
Closes https://github.com/G-Research/armada/issues/1071

Enable Integration Tests in make-e2e-tests for python client


client/python/tests/integration/test_no_auth.py
```diff
@@ -9,86 +9,95 @@
...
     assert not queue.active_job_sets
-    no_auth_client.delete_queue(name=queue_name)
+
+
+def test_submit_job_and_cancel_by_id():
+    job_set_name = f"set-{uuid.uuid1()}"
+    jobs = no_auth_client.submit_jobs(
+        queue=queue_name, job_set_id=job_set_name, job_request_items=submit_sleep_job()
+    )
+    sleep(5)
```
jayofdoom, line 62: Is there anything we can do to avoid sleeping here? These will be run in sequence which means we're adding at least 20 seconds to test runtime. Any reason we can't poll to see when whatever we're waiting on (the job to actually be submitted?) to be complete?

kannon92: The problem is that we have to sleep before we start polling.  When you submit a job, it may take some time for events to pick it up.

I have to deal with this in my job service where a job polling fails until the first event.

jayofdoom (dismissed):

kannon92:
I am going to look into how I can remove the sleeps as I do think that could be a potential issue in a CI.

build/python-client/Dockerfile
```diff
@@ -6,4 +6,14 @@ RUN mkdir /proto
 
 RUN pip install grpcio grpcio-tools poetry
 
+WORKDIR /code
```
kannon92, line 9: Once you get this PR synced with main, you should refactor this to no longer use poetry.  

I think `pip install .` should do the trick to get this code installed in the docker container.

build/python-client/Dockerfile
```diff
@@ -6,4 +6,14 @@ RUN mkdir /proto
 
 RUN pip install grpcio grpcio-tools poetry
```
kannon92, line 7: Also should probably remove poetry.

client/python/tests/integration/test_no_auth.py
```diff
@@ -1,113 +1,162 @@
...
-    no_auth_client.cancel_jobs(queue=queue_name, job_set_id=job_set_name)
-    no_auth_client.delete_queue(name=queue_name)
 
+    return [no_auth_client.create_job_request_item(priority=0, pod_spec=pod)]
 
-def test_get_queue():
-    queue_name = f"queue-{uuid.uuid1()}"
-    no_auth_client.create_queue(name=queue_name, priority_factor=200)
 
+def wait_for(queue=None, job_set_id=None):
```
kannon92, line 64: I think you should refactor this into two functions.  

A good approach to follow is SRP (Single Responsibility Principle).  A function should do one thing.

Sharpz7: Okay that's fine by me :)

build/python-client/Dockerfile
Sharpz7: I have done this sync

build/python-client/Dockerfile
```diff
@@ -4,6 +4,11 @@ FROM --platform=x86_64 python:${PYTHON_VERSION}-buster
 
 RUN mkdir /proto
 
-RUN pip install grpcio grpcio-tools mypy-protobuf
+# Project Install
+COPY client/python/pyproject.toml /code/pyproject.toml
+RUN pip install /code\[test\] --use-feature=in-tree-build
```
kannon92, line 9: I don't really understand this.

Sharpz7: This is what I had to do to fix the code last night in CI 😅. Would you want comments written about specifically why each thing is done or is discussing it in a meeting better?

rankindaveyg: Comments will live forever! :)

Sharpz7: On it :))

client/python/tests/integration/test_no_auth.py
kannon92: Once you address this, I think it looks good to me!

kannon92: On a slack message, we decided to leave this as is.

kannon92 (commented):
Since I am the original author, I can't approve.  It LGTM.

build/python-client/Dockerfile
```diff
@@ -4,6 +4,19 @@ FROM --platform=x86_64 python:${PYTHON_VERSION}-buster
...
+COPY client/python/pyproject.toml /code/pyproject.toml
+
+# Note that --use-feature=in-tree-build is needed until pip 21.3
+# (https://github.com/libAtoms/QUIP/issues/345)
+#
+# To make sure that the workdir is not changed (This might break something in the future),
+# We pip install from /code
+#
+# Then to install the `test` optional dependencies, we pip install from /code[test]
+# See https://stackoverflow.com/questions/46775346/what-do-square-brackets-mean-in-pip-install
```
jayofdoom, line 16: nit: SO links as doc links are weird. The explanation was probably OK without the link, I just generally prefer using primary sources. (Do not spend time fixing this though :D)

client/python/tests/integration/test_no_auth.py
```diff
@@ -1,113 +1,167 @@
...
-    job_set_name = f"set-{uuid.uuid1()}"
-    no_auth_client.create_queue(name=queue_name, priority_factor=200)
-    no_auth_client.submit_jobs(
-        queue=queue_name, job_set_id=job_set_name, job_request_items=submit_sleep_job()
-    )
-    # Jobs can must be finished before deleting queue
-    no_auth_client.cancel_jobs(queue=queue_name, job_set_id=job_set_name)
-    no_auth_client.delete_queue(name=queue_name)
-    time.sleep(1)
+    if server_ssl == "true":
```
jayofdoom, line 21: nit: Would be nice if we made server_ssl a real bool instead of this.

```
>>> import distutils
>>> distutils.util.strtobool("falsey")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/distutils/util.py", line 319, in strtobool
    raise ValueError("invalid truth value %r" % (val,))
ValueError: invalid truth value 'falsey'
>>> distutils.util.strtobool("truthy")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/distutils/util.py", line 319, in strtobool
    raise ValueError("invalid truth value %r" % (val,))
ValueError: invalid truth value 'truthy'
>>> distutils.util.strtobool("true")
1
>>> distutils.util.strtobool("0")
0
>>> distutils.util.strtobool("false")
0
>>> distutils.util.strtobool("1")
1
```

client/python/tests/integration/test_no_auth.py
```diff
@@ -1,113 +1,167 @@
...
 
+no_auth_client = client()
 
-def test_submit_job_and_cancel_by_queue_job_id():
-    queue_name = f"queue-{uuid.uuid1()}"
-    job_set_name = f"set-{uuid.uuid1()}"
-    no_auth_client.create_queue(name=queue_name, priority_factor=200)
-    no_auth_client.submit_jobs(
-        queue=queue_name, job_set_id=job_set_name, job_request_items=submit_sleep_job()
+queue_name = f"queue-{uuid.uuid1()}"
```
jayofdoom, line 37: should this (and line 35) be in a pytest setup method and/or fixture?

client/python/tests/integration/test_no_auth.py
jayofdoom: Can we at least guard against both values being set, if that's an invalid usage?

makefile
```diff
@@ -402,6 +403,8 @@ tests-e2e: build-armadactl build-docker-no-lookout tests-e2e-setup
 	$(GO_TEST_CMD) go test -v ./e2e/armadactl_test/... -count=1 2>&1 | tee test_reports/e2e_armadactl.txt
 	$(GO_TEST_CMD) go test -v ./e2e/basic_test/... -count=1 2>&1 | tee test_reports/e2e_basic.txt
 	$(GO_TEST_CMD) go test -v ./e2e/pulsar_test/... -count=1 2>&1 | tee test_reports/e2e_pulsar.txt
+	docker run -v${PWD}/client/python:/code -e ARMADA_SERVER=server -e ARMADA_PORT=50051 --entrypoint python3 --network=kind armada-python-client-builder:latest -m pytest -v -s /code/tests/integration/test_no_auth.py
```
jayofdoom, line 406: If these aren't also added to `tests-e2e-no-setup`, they will not run in CI. In fact, they are not currently running in the circleci job triggered by the PR and so haven't been tested yet AFAICT.

jayofdoom (changes requested):

build/python-client/Dockerfile
Sharpz7: I did this because I had never heard of this before, and I found barely any documentaition on it. Good to know tho :))

makefile
jayofdoom: It's clearly in tests-e2e-no-setup above. I have no idea why I couldn't find it before Kevin pointed me at it.

jayofdoom (dismissed):
I was wrong about CI; this LGTM. We may want someone longer-tenured on the project to :+1: adding things to the integration tests suite before it merges...

client/python/tests/integration/test_no_auth.py
Sharpz7: This has been handled  :))

client/python/tests/integration/test_no_auth.py
Sharpz7: I don't think that is an invalid usage, I think its fine for both to be set.

client/python/tests/integration/test_no_auth.py
Sharpz7: Enums can solve this, I shall do that.

client/python/tests/integration/test_no_auth.py
Sharpz7: I do not think so, as they are needed across all the functions as globs. I can put them in a better place in the file though.

build/python-client/Dockerfile
Sharpz7: This has been completed

client/python/tests/integration/test_no_auth.py
Sharpz7: After re-writing this code with fixtures, I have re-written it to be dependent on just the job-set=id being present, as the queue var needs to be present in either case.

jayofdoom (dismissed):

jayofdoom (approved):

jayofdoom (approved):

PR #1092, opened 2022-06-17
Python airflow integration

kannon92:
Airflow Integration

https://github.com/G-Research/armada/issues/1063


kannon92:
https://github.com/G-Research/armada/issues/1137

third_party/airflow/tox.ini
```diff
@@ -0,0 +1,73 @@
...
+deps = 
+    poetry
+whitelist_externals = 
+    poetry 
+    bash
+commands =
+    poetry install -v
+    poetry run pytest tests/unit
+    # This executes the dag files in examples but really only checks for imports and python errors
+    bash -c "find examples/ -maxdepth 1 -type f | xargs poetry run python3"
```
jayofdoom, line 18: nit: if you added a `-name *.py`, you might save someone an annoying problem in the future :)

jayofdoom (commented):
Almost all my comments are mere suggestions or nits -- I see nothing major here except maybe that the `poetry.lock` file looks wrong, but I suspect we just shouldn't commit it.

docs/python_airflow_operator.md
```diff
@@ -0,0 +1,197 @@
+---
```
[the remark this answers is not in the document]
kannon92: Yea I like a followup.  I need to write an actual user doc on how to use airflow and our operator.  So i will come back to this once we are happy with everything.

third_party/airflow/armada/operators/armada.py
```diff
@@ -0,0 +1,97 @@
...
+            raise AirflowException("Armada has issues submitting job")
+        job_state, job_message = search_for_job_complete(
+            job_service_client=self.job_service,
+            queue=self.queue,
+            job_set_id=self.job_set_id,
+            airflow_task_name=self.name,
+            job_id=job_id,
+        )
+        armada_logger.info(
+            f"Armada Job finished with {job_state} and message: {job_message}"
```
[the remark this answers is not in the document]
kannon92: Interesting!  Thank you for that.  So fstrings execute but interpolation doesn't?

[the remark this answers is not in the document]
jayofdoom: I'll note this depends on the logging library, but in any I've looked at in python, it checks the log level first and will not interpolate the fields unless it's actually going to log it. This can (sadly) have a real performance difference on python -- it probably doesn't matter for the client, but it's good to know and a good practice to get into (using the log libraries' built-in interpolation.

If you're just using an f-string, it's going to have the fields inserted before the library is even called.

third_party/airflow/poetry.lock
```diff
@@ -0,0 +1,2943 @@
+[[package]]
```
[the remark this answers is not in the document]
jayofdoom: I think that's more of an opinion as to how to package things; if we include it for the client we should be consistent here -- but are you 100% certain the lockfile committed here is sane? It seemed to have a lot of unrelated packages listed in the extras section.

third_party/airflow/tests/unit/armada_client_mock.py
```diff
@@ -0,0 +1,35 @@
+from google.protobuf import empty_pb2
```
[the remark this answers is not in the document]
kannon92: I think I'll leave it for now then.

third_party/airflow/examples/bad_armada.py
```diff
@@ -0,0 +1,114 @@
+from airflow import DAG
```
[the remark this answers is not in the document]
jayofdoom: Literally just a docstring with inputs/outputs and a one line sentence of what's going on -- that's better than the nothing we have now, and makes room for someone to improve it later if they lack understanding.

.github/workflows/airflow-operator.yml
```diff
@@ -0,0 +1,65 @@
...
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.9'
+      - run: pip install tox poetry
+      # Generate the proto files for python, required for later steps
+      - run: make python
+      - run: |
+          poetry install -v
+          poetry build
+        working-directory: client/python
```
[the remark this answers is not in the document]
jayofdoom: I'm still confused then, because if it's installed into a poetry-specific venv, why does installing it make a difference the later tox runs (it's structured such that it's a dependency for them)?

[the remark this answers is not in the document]
jayofdoom: Talked to @kannon92 about this in slack:
- We need to create a wheel file for the python armada client for the airflow operator to consume.
- Tomorrow, we'll see if we can get that wheel by just running `poetry build` (without `poetry install`).
- Also a comment should be added about why we're doing this build here.

This makes a lot more sense now. Thanks for the chat in slack to give me understanding.

third_party/airflow/docs/source/python_airflow_operator.rst
```diff
@@ -0,0 +1,28 @@
+:orphan:
+
+armada package
```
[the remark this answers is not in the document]
kannon92: Added a little tidbit.

third_party/airflow/docs/source/conf.py
```diff
@@ -0,0 +1,56 @@
...
+import os
+import sys
+sys.path.insert(0, os.path.abspath('../..'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'python-armadaairflowoperator'
+copyright = '2022 Armada Project'
+author = 'kevin@gr-oss.io'
```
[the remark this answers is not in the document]
kannon92: Done.

third_party/airflow/armada/operators/utils.py
```diff
@@ -0,0 +1,113 @@
...
+                                    can be in job_id_not_found
+                                    before we decide it was a invalid job
+    :return: A tuple of state, message
+    """
+    start_time = time.time()
+    # Overwrite time_out_for_failure by environment variable for configuration
+    armada_time_out_env = os.getenv("ARMADA_AIRFLOW_TIME_OUT_JOB_ID")
+    if armada_time_out_env:
+        time_out_for_failure = int(armada_time_out_env)
+    while True:
```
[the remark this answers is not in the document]
kannon92: This method should block as long it takes for an Armada Job to finish.  

Even with job service, that is the design of this function to keep polling the jobservice until we get a terminal event.

jayofdoom (dismissed):
Thank you, those docstrings are exactly what I was looking for. Thanks!

jayofdoom (dismissed):

third_party/airflow/armada/operators/utils.py
```diff
@@ -0,0 +1,115 @@
...
+                job_set_id=job_set_id,
+                job_service_client=job_service_client,
+            )
+        else:
+            job_status_return = job_status_callable(
+                queue=queue, job_id=job_id, job_set_id=job_set_id
+            )
+
+        if job_status_return.state == jobservice_pb2.JobServiceResponse.SUCCEEDED:
+            job_state = "succeeded"
```
Sharpz7, line 86: For the same reason I am using Enums for the job_types in the events PR, do we want these strings to be what is returned, rather than an object that could be better tracked and understood? Also stops issues like typos etc breaking if statements

third_party/airflow/examples/bad_armada.py
```diff
@@ -0,0 +1,136 @@
...
+                    },
+                ),
+            )
+        ],
+    )
+
+    return [submit_pb2.JobSubmitRequestItem(priority=1, pod_spec=pod)]
+
+
+def submit_bad_job():
```
Sharpz7, line 50: I am not sure how termperary this file is, but having a lot of repeatable code is not ideal for an example. Maybe make creating a dummy pod spec a function with args that could mkae ifakail or not fail?

third_party/airflow/examples/hello_armada.py
```diff
@@ -0,0 +1,95 @@
...
+                    },
+                ),
+            )
+        ],
+    )
+
+    return [submit_pb2.JobSubmitRequestItem(priority=1, pod_spec=pod)]
+
+
+"""
```
Sharpz7, line 51: I am not sure how important this is to people, but because of [PEP 257](https://peps.python.org/pep-0257/), most IDE's are going to include these docstrings as part of the files module, rather than comments for the individual places where they are used.

This will have limited scope, but it could be annoying if an IDE decides to show all these docstrings when the user only expects to see the files/module docstring.

third_party/airflow/tox.ini
```diff
@@ -0,0 +1,58 @@
+[tox]
```
Sharpz7, line 1: Does this file mean we will need to run tox commands from two places now just to get a CI pass? 

If that is the case, do we maybe want a makefile that cn auto-run the commands for us? Just for a bit of ease of use.

kannon92: Makefiles aren’t really a common tool for python developers.  I would like to approach build systems as being native like to their language.  Also I hope some day we can maybe migrate clients to separate repos so I don’t want the go makefile controlling the build of our python code

third_party/airflow/examples/hello_armada.py
kannon92: @jayofdoom suggested docs in examples as a way to provide documentation so someone could understand how to write their own examples.  

Are you suggesting we use normal comments and not do strings.

third_party/airflow/armada/operators/utils.py
kannon92: Good idea!

third_party/airflow/examples/hello_armada.py
Sharpz7: Yeah exactly. Just suggesting to use normal comments

third_party/airflow/tox.ini
kannon92: > Does this file mean we will need to run tox commands from two places now just to get a CI pass?
> 
> If that is the case, do we maybe want a makefile that can auto-run the commands for us? Just for a bit of ease of use.

I think this is more of a bandaid.  Our goal should be that the airflow operator uses a published python client but we haven't picked up deployment of that yet.

third_party/airflow/examples/hello_armada.py
jayofdoom: Eh, I don't think we should reduce the value of the comments because they show up in an ide.

third_party/airflow/tox.ini
jayofdoom: It's totally reasonable to have something in the makefile to run both. It can be done as a follow on pr as desired.

jayofdoom (approved):

PR #1108, opened 2022-06-22
setting to use local certs, more platform independence

### LINKED ISSUE #1107: Makefile cert situation is messy and not platform independent
Our makefile assumes the CA certs are all in `/etc/ssl/certs/ca-certificates.crt`, which is true in Debian/Ubuntu, but not on AmazonLinux (where they are in `/etc/ssl/certs/ca-bundle.crt`) or Mac OS (where they live in some keychain thing that you can't easily access).

To remedy this, I propose generalizing the makefile to look in a few standard places, and also adding a simple variable that toggles whether or not to use the host system certs. I think the default should just be to use whatever is in the docker instances.

andrewpcone:
Hopefully closes #1107


makefile
```diff
@@ -108,6 +125,37 @@ export SHELLOPTS:=$(if $(SHELLOPTS),$(SHELLOPTS):)pipefail:errexit
 gobuildlinux = go build -ldflags="-s -w"
 gobuild = go build
 
+ifdef USE_SYSTEM_CERTS
```
kannon92, line 128: A nitpick but this doesn't need to be in an ifdef.  You could remove this out of ifdef and just put the calls to this target in the ifdef.

kannon92 (dismissed):
I was able to build the dotnet client with this change.  Thank you Andrew!

I have a small nitpick but it looks good to me.

kannon92 (approved):

richscott (approved):

dejanzele (approved):

Issue #1037, opened 2022-06-01
Migrate from CircleCI to Github Actions


richscott:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-48) by [Unito](https://www.unito.io)

jayofdoom:
Unit tests and linting merged, but not required for merging -- we're evaluating it versus Circle CI first.

Integration tests have performance issues that I believe are caused by a smaller runner being used in GHA (this is not configurable; you use their static-sized runners or provide your own). There's a discussion here about how to move forward: https://gr-oss.slack.com/archives/C03F1DS85AQ/p1654807364006189 -- it boils down to three options:
- Investigate hosting our GHA runners
- Revert all test migrations to GHA
- Continue to use a split setup, with unit tests and linting running alongside CodeQL in github actions, with integration tests running in circle ci. This is the option I think gives us the best mix of functionality and quick runtime.

jayofdoom:
I have a PoC running tests using paid-for github runners in ~15-17 minutes. https://github.com/jayofdoom/armada/runs/7733643008

There are a handful of things needed before this is ready for landing:
- Do some matrix testing across multiple different instance sizes; be sure we're at the best size
- Ensure in the case where we cannot spawn a runner, we still run integration tests on a free runner (if possible)
- Ensure *all* items supported by circleci (including releases / uploads of docker images) are supported in our GHA

stackedsax:
@jayofdoom thanks for doing this.  I've never checked out the paid-for runners -- do we just pay more and maybe get that time down under 10mins?

jayofdoom:
I'm not convinced that vertical scaling is possible without actual work to parallelize the tests explicitly. I'm trying to explore those possibilities before doing the cleanup work around this to make it mergeable. (Which will include making sure we're at the point where we have all the performance we can pay for without excess)

For instance, I started something testing before I stepped away this afternoon that is using docker's new build system which might be faster outright, but if it isn't I'm convinced it could be made to be faster with a few changes to our dockerfiles.

jayofdoom:
https://github.com/jayofdoom/armada/runs/7736074257 looks like I just gained quite a bit of time just by setting an environment variable 👍

jayofdoom:
One #1156 is merged, we'll have the following remaining steps:
- Implement releases in Github Actions
- Disable run of jobs in CircleCI

richscott:
[...]

PR #1110, opened 2022-06-22
Fix when GHAs run

jayofdoom:
Aiming to make it run on pushes of all branches (except master, which
do not need to run tests after merge -- this will change when releasing
is moved to GHA).

Should also fix issue where PRs from forks don't have the job runners
show up in the list at the bottom of the pull request.

Related: Issue #1037

### DIFF
--- a/.github/workflows/not-python.yml
+++ b/.github/workflows/not-python.yml
@@ -11,8 +11,8 @@ on:
       - '.github/workflows/python.yml'
       - 'docs/python_armada_client.md'
   pull_request:
-    branches:
-      - master
+    branches-ignore:
+      - gh-pages
     paths-ignore:
       - 'client/python/**'
       - 'build/python-client/**'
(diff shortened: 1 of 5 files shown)


jayofdoom:
The only opportunity to DRY up, AFAICT, would be using yaml anchors to make us only list the paths once per workflow. That adds a bit of complexity (someone reading this file has to grok yaml anchors) for very minimal benefit. I would go the route of DRYing it up if I could do it across github workflows.

If you feel strongly about it; I can use the yaml anchors but for small config snippets like this, I find the readability of being explicit worth the tradeoff.

.github/workflows/not-python.yml
```diff
@@ -2,6 +2,17 @@ name: Run Python Linting and Unit Tests
...
+      - master
+    paths-ignore:
+      - 'client/python/**'
+      - 'build/python-client/**'
+      - 'pkg/api/*.proto'
+      - '.github/workflows/python.yml'
+      - 'docs/python_armada_client.md'
+  pull_request:
+    branches:
+      - master
```
jayofdoom, line 15: this needs to be updated to branches-ignore, like in the go.yml, and I missed this several places

PR #1113, opened 2022-06-23
Fixes to test suite

severinson:


internal/testsuite/eventwatcher/eventwatcher.go
```diff
@@ -210,11 +209,10 @@ func GetFromIngresses(parent context.Context, C chan *api.EventMessage) error {
 }
 
 func getFromIngress(ctx context.Context, host string) error {
-	tokens := strings.Split(host, ".")
-	if len(tokens) == 0 {
-		return errors.Errorf("malformed hostname: %s", host)
+	ingressUrl := os.Getenv("ARMADA_EXECUTOR_INGRESS_URL")
```
kannon92, line 212: It seems like you are missing the case of a bad hostname from host parameter.  

Is that intentional?

severinson: What do you mean?

severinson: I think that's fine.

kannon92: Kk.  I didn't see that this was in test suite.  Yea I think that's fine.

kannon92 (approved):

d80tb7 (approved):

PR #1114, opened 2022-06-23
Fix issue with proto files generation w/bad perms

jayofdoom:
Before this change, running `make proto` on linux would leave you
with files inside proto dir which were owned by root and/or set
to have no write permissions whatsoever. This prevented a dev
from being able to run `make proto` successively without cleanup.

This also attempts to cleanup some of the code by:
- maintaining windows compatability by factoring out -u args for
  windows machines
- using a single environment variable to set user options for docker

This also fixes a single instance where make was resolving the docker
command as `docker -u : ` without properly inserting uid/gid.

Co-Authored-By: Kevin Hannon <kannon1992@gmail.com>

### DIFF
--- a/makefile
+++ b/makefile
@@ -41,6 +41,14 @@ else
 	DOCKER_NET = --network=host
 endif
 
+ifeq ($(DOCKER_RUN_AS_USER),)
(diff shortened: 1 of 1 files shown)


richscott (approved):

jayofdoom:
I have this bookmarked, will make this mergable some day in the future.

PR #1120, opened 2022-06-24
add healthcheck in testsuite for armada server

dejanzele:


pkg/client/connection.go
```diff
@@ -17,6 +22,7 @@ import (
 
 type ApiConnectionDetails struct {
 	ArmadaUrl                   string
+	ArmadaRestUrl               string
```
kannon92, line 25: Is it possible to just construct the ArmadaRestUrl from ArmadaUrl?  

I don't love a new config value that requires our client users to put both a RestURL and a grpc URL.

severinson: No. They may be completely different.

cmd/testsuite/cmd/test.go
```diff
@@ -0,0 +1,147 @@
...
+	cmd.Flags().String("tests", "", "Test file pattern, e.g., './testcases/*.yaml'.")
+	cmd.Flags().String("junit", "", "Write a JUnit test report to this path.")
+
+	return cmd
+}
+
+func testCmdRunE(app *testsuite.App) func(cmd *cobra.Command, args []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		if err := app.Params.ApiConnectionDetails.ArmadaRestServerHealthcheck(); err != nil {
+			return errors.Wrap(err, "armada server health check failed")
```
severinson, line 41:
```suggestion
			return errors.WithMessage(err, "Armada server health check failed")
```

pkg/client/connection.go
kannon92: So how does someone understand what to put for ArmadaRestUrl?  

This is a config that our users of armadactl and other client libs use so I don't understand how they would know that the RestURL that grpc is generating could be very different from the connection details they would use for the internal grpc server.

cmd/testsuite/cmd/test.go
```diff
@@ -0,0 +1,147 @@
...
+		// If junitPath is set, write a JUnit report.
+		testSuite.Time = fmt.Sprint(time.Since(start))
+		testSuites := &junit.Testsuites{
+			Name: testFilesPattern,
+			Time: testSuite.Time,
+		}
+		testSuites.AddSuite(*testSuite)
+		if junitPath != "" {
+			if err := createJUnitReport(junitPath, testSuites); err != nil {
+				return errors.Wrap(err, "error creating junit report")
```
severinson, line 93:
```suggestion
				return errors.WithMessage(err, "error creating junit report")
```

pkg/client/connection.go
```diff
@@ -97,5 +103,25 @@ func transportCredentials(config *ApiConnectionDetails) grpc.DialOption {
...
 		return grpc.WithTransportCredentials(credentials.NewClientTLSFromCert(nil, ""))
 	}
-	return grpc.WithInsecure()
+	return grpc.WithTransportCredentials(insecure.NewCredentials())
+}
+
+func (a *ApiConnectionDetails) ArmadaRestServerHealthcheck() error {
+	url := a.ArmadaRestUrl
+	if url == "" {
+		return errors.New("armada server rest api url not provided")
```
severinson, line 112: I prefer using capitalised Armada everywhere. Since it's a name I think that's more correct.

pkg/client/connection.go
dejanzele: @kannon92 grpc and rest could quite possibly be on completely different (sub)domains, where the REST server exposes "safe" endpoints and could be made public, and the gRPC server exposes "unsafe/admin" endpoints, and could be made private.

PR #1121, opened 2022-06-24
# 1119 update quickstart to remove manual steps and highlight install…

rankindaveyg:
… scripts

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -5,7 +5,7 @@ The purpose of this guide is to install a minimal local Armada deployment for te
 ## Pre-requisites
 
 - Git
-- Docker (Docker Desktop recommended for local development)
+- Docker (Docker Desktop recommended for local development on Windows/OSX)
 - Helm v3.5+
 - Kind v0.11.1+
 - Kubectl
@@ -24,7 +24,7 @@ All commands are intended to be run from the root of the repository.
 
 Armada is a resource intensive application due to the need to run multiple Kubernetes clusters - for a local installation you will need at least 16GB of RAM available.
 
-### One-click setup
+### One-click Setup
 
 Run the following script in order to provision an Armada cluster locally:
 ```bash
@@ -40,7 +40,7 @@ Run the following script if you want to destroy your local Armada cluster:
 ## Usage
 Create queues, submit some jobs and monitor progress:
 
-### Queue creation
+### Queue Creation
 ```bash
 ./armadactl create queue queue-a --priorityFactor 1
 ./armadactl create queue queue-b --priorityFactor 2
@@ -58,7 +58,8 @@ For more control, queues can be created via `armadactl create`, which allows for
 ./armadactl create -f ./docs/quickstart/queue-b.yaml
 ```
 
-### Job submission
+
+### Job Submission
 ```
 ./armadactl submit ./docs/quickstart/job-queue-a.yaml
 ./armadactl submit ./docs/quickstart/job-queue-b.yaml
@@ -76,7 +77,7 @@ Watch individual queues:
 Log in to the Grafana dashboard at [http://localhost:30001](http://localhost:30001) using the default credentials of `admin` / `prom-operator`.
 Navigate to the Armada Overview dashboard to get a view of jobs progressing through the system.
 
-Try submitting lots of jobs and see queues build and get processed:
+Try submitting lots of jobs and see queues get built and processed:
 
 ```bash
 for i in {1..50}
@@ -107,12 +108,30 @@ Nov  4 11:44:17 | Queued:   0, Leased:   0, Pending:   0, Running:   1, Succeede
 Nov  4 11:44:26 | Queued:   0, Leased:   0, Pending:   0, Running:   0, Succeeded:   2, Failed:   0, Cancelled:   0 | event: *api.JobSucceededEvent, job id: 01drv3mey2mzmayf50631tzp9m
 ```
 
+
+### Grafana Configuration
+
+Run the following commands to setup Grafana in your environment:
+
+```bash
+curl -X POST -i http://admin:prom-operator@localhost:30001/api/datasources -H "Content-Type: application/json" -d '{"name":"cluster-0","type":"prometheus","url":"http://'$EXECUTOR_0_IP':30001","access":"proxy","basicAuth":false}'
+curl -X POST -i http://admin:prom-operator@localhost:30001/api/datasources -H "Content-Type: application/json" -d '{"name":"cluster-1","type":"prometheus","url":"http://'$EXECUTOR_1_IP':30001","access":"proxy","basicAuth":false}'
+curl -X POST -i http://admin:prom-operator@localhost:30001/api/dashboards/import --data-binary @./docs/quickstart/grafana-armada-dashboard.json -H "Content-Type: application/json"
+```
+
 Grafana:
 
 ![Armada Grafana dashboard](./quickstart/grafana-screenshot.png "Armada Grafana dashboard")
 
-Note that the jobs in this demo simply run the `sleep` command so do not consume much resource.
+Note that the jobs in this demo simply run the `sleep` command so do not consume many resources.
 
-Lookout:
+### Lookout Configuration
+
+Armada Lookout UI can be configured by doing the following:
+
+```bash
+kubectl port-forward svc/armada-lookout 8080:8080 
+```
+Then access it by opening [http://localhost:8080](http://localhost:8080) in your browser.
 
 ![Lookout UI](./quickstart/lookout.png "Lookout UI")
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -26,15 +26,11 @@ Armada is a resource intensive application due to the need to run multiple Kuber
 
 ### One-click Setup
 
-Run the following script in order to provision an Armada cluster locally:
-```bash
-./docs/local/setup.sh
-```
+To install Armada and all its dependencies you can use this script: 
+https://github.com/G-Research/armada/blob/master/docs/local/setup.sh
 
-Run the following script if you want to destroy your local Armada cluster:
-```bash
-./docs/local/destroy.sh
-```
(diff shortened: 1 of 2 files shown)


docs/quickstart.md
```diff
@@ -81,147 +36,6 @@ Run the following script if you want to destroy your local Armada cluster:
 ./docs/local/destroy.sh
 ```
 
-### Manual setup
```
jayofdoom, line 84: Since these have been tested to work, can they be moved to another document for manual setup, for folks (like me) who wouldn't be very comfortable running a bash script to configure an open source project.

jayofdoom (changes requested):
One comment inline, and this appears to contain a kubectl binary that shouldn't be here.

docs/quickstart.md
```diff
@@ -81,147 +36,6 @@ Run the following script if you want to destroy your local Armada cluster:
...
-helm install lookout gresearch/armada-lookout -f docs/quickstart/lookout-values.yaml
-```
-
-You can view the UI by running the following:
-```bash
-kubectl port-forward svc/armada-lookout 8080:8080
-```
-You will be able to view Lookout at `http://localhost:8080`.
-
-#### Grafana configuration
```
dejanzele, line 177: This block should also stay, even tho you get the same info from the install script

dejanzele (changes requested):

docs/quickstart.md
kannon92: Yea, think of the script just installing what you need to a k8 cluster.  

You still need to run these commands in a separate terminal in order to actually view the lookout UI.  

Same for Grafana/Prometheus

docs/quickstart.md
```diff
@@ -5,54 +5,11 @@ The purpose of this guide is to install a minimal local Armada deployment for te
 ## Pre-requisites
 
 - Git
-- Docker
+- Docker (Docker Desktop recommended for local development)
```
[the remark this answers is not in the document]
jayofdoom: I don't use docker desktop on linux, and wouldn't, because it adds a VM layer that hurts performance and needs separate management (e.g. allocating it more ram/cpus as needed) -- in fact the install guide dictates you configure KVM before doing this, which makes it a nonstarter on systems where folks are running another hypervisor software (like vmware/virtualbox).

I do agree that Docker Desktop makes the most sense on Mac and Windows, though.

docs/quickstart.md
rankindaveyg: Can't you just run the script step by step in that case? Most projects I see use 'pip install foo' or 'gem install bar'. I'd like to get out of the business of maintaining both versions.

kannon92: https://github.com/G-Research/armada/discussions/1109

I like to follow similar patterns in the open source community.  I have not found any example that tells someone to run a bash script.   I think we should actually scrap the script and have code blocks that people can copy.  I would also prefer us to clean up this installation process so it is possible to do in a few steps rather than 10-20.

In the CNCF projects, I see a common pattern is to use kubectl to install a project onto an existing cluster.   

It may seem weird but it is common practice to have people copy code blocks into their terminal to run for these quick starts.  Look at the airflow documentation in my discussion.  

I have maybe a 10-15% confidence that this script will actually work on people's machines.  That is a big part why I wrote the discussion because I think we should adopt the changes I proposed in that discussion before we update the quickstart.  

So far, Pulsar seems to be a big issue on most of our machines so this script will probably not work for most of our users.  I think its related to the fickleness of Kind/Minikube as a local method for deploying Kubernetes.  Deploying via helm charts is typically reserved for a production installation and the charts don't seem to work well with local installation.  

Maybe we should have a discussion about this before we update the quickstart?

kannon92: In terms of maintaining , I agree.  But no developer on our team or in GR uses the quickstart guide as a means of developing.  So it will always be out of date.  

I would prefer that we migrate our quickstart guide to be focused on a developer persona with using docker-compose which is a common way that we develop our code.  We also keep this up to date.  My hope is to also use docker-compose environment in our CI as a way to keep our dev section up to date.  If not, I find this to be the most reliable way of setting up Armada for development so I think it would be kept update.   

I would like to use a production ready K8 cluster in our quickstart guide because I think if we release this as is, we will spend a lot of time debugging issues that people have on various machines.  We already have a series of issues about how our original quickstart guide doesn't work on certain machines.    

This script currently doesn't work on mac M1, AWS EC2, or Rich's linux box.  It may work on other teams machines if they have enough memory.  But I think this script will cause more issues than it is worth IMO.

rankindaveyg: This script just worked fine on my Mac M1, including the Pulsar bits. How about we take this as a cleanup on the page and enhance it further when we have a K8 cluster for new users to try out? I think this is still something of a win, if not the end state.

dejanzele: @kannon92 @richscott @jayofdoom @rankindaveyg 
A lot of CNCF projects can only be installed in Kubernetes and cannot possibly work outside of k8s. (Argo, Istio, Cert Manager...)

Armada has a hard dependency on Kubernetes, where the executors need a k8s cluster in which to schedule a job, but rest of components can be run outside of k8s. Even tho some components can run outside of k8s, it is not their natural habitat. Natural habitat for Armada and its components is inside a k8s cluster.

My proposal is we redo the Quickstart and move towards using the Armada Bundle helm chart to install a fully functional Armada cluster.

Guide could have a prerequisite for a k8s cluster, and if the user does not have one, we provide docs to setup one (minikube or kind), and after that we just tell them now use `helm install armada-bundle gresearch/armada-bundle`.

WDYT?

kannon92: Wow.  I haven't got this script working on either of my machines..

kannon92: I think that is good but the main point of using Armada is to provide multi cluster scheduling.  

Will this be handled by a helm chart bundle?  Or will that install the executor onto the same cluster as the server?

dejanzele: The control plane/server/main cluster part has the most components, you need cert manager, prometheus, nginx controller... where the executor cluster only needs one thing, the executor.

Also, Quickstart guides for popular tools assume you have a k8s cluster and dive straight into installing using kubectl, helm, kustomize...

* ArgoCD - https://argo-cd.readthedocs.io/en/stable/getting_started/
* Cert Manager - https://cert-manager.io/docs/installation/
* Istio (they provide a bash script but all whom I knew used Istio, absolutely avoided the bash script and went straight to Helm chart) - https://istio.io/latest/docs/setup/getting-started/
* NGINX Controller (it has nuances between different k8s flavours, so it has examples for varios k8s deployments) - https://kubernetes.github.io/ingress-nginx/deploy/

rankindaveyg: Holy moly, we are creeping the scope of what is just a rearranging of the furniture here. I'd like to propose that we accept this work as-is as a cleanup of the website and move discussions of the future of the developer experience into a new discussion.

richscott: It sucks that K8S (and K8S lookalikes such as Kind/minikube/k3s) is such a buggy mess to get fully installed and running correctly. I'm seriously drawn towards just telling potential Armada users "first, get a running K8S cluster" (they may have one in-house already, or we suggest using AWS EKS or Google's managed clusters), then we provide a Helm chart to install Armada. 

But if it runs at least reasonably well using Kind as the cluster manager/k8s-simulator on Docker Desktop for Mac, can/should we declare that as the supported dev-desktop configuration for now?

kannon92: Our point is that most OSS do not tell you to run a script from their webpage.  If they do, you use something like `pip install` or `helm install` which is verified.  I think that running a script without knowing what's inside is a security issue in most organizations.  Its security 101 to not run an executable from someone you don't know sent to you via email.  I think the same thing should apply on any website.  
 
 The other issue is that this script doesn't really provide a stable environment.  So we "fix" the issue on the webpage but we will keep fixing support issues for people who try to run this as an evaluation.  And it makes us look sloppy and we don't get any people adopting our software.

docs/quickstart.md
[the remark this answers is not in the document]
richscott: Also note that Docker Desktop has built-in K8S support (at least, on macOS) - turn it on via a checkbox in the config panel, then restart Docker Desktop - this may run more reliably than Kind-managed clusters over Docker.

docs/quickstart.md
richscott: For what it's worth, I just noticed that the Volcano project essentially expects you to have a running K8S cluster as a prequisite:  https://volcano.sh/en/docs/installation/  - then guide to installing Volcano via `kubectl` with YAML spec files, or using `helm` with their public charts.

richscott (approved):

kannon92 (approved):

docs/quickstart.md
```diff
@@ -81,147 +36,6 @@ Run the following script if you want to destroy your local Armada cluster:
...
-You will be able to view Lookout at `http://localhost:8080`.
-
-#### Grafana configuration
-
-```bash
-curl -X POST -i http://admin:prom-operator@localhost:30001/api/datasources -H "Content-Type: application/json" -d '{"name":"cluster-0","type":"prometheus","url":"http://'$EXECUTOR_0_IP':30001","access":"proxy","basicAuth":false}'
-curl -X POST -i http://admin:prom-operator@localhost:30001/api/datasources -H "Content-Type: application/json" -d '{"name":"cluster-1","type":"prometheus","url":"http://'$EXECUTOR_1_IP':30001","access":"proxy","basicAuth":false}'
-curl -X POST -i http://admin:prom-operator@localhost:30001/api/dashboards/import --data-binary @./docs/quickstart/grafana-armada-dashboard.json -H "Content-Type: application/json"
-```
-### CLI installation
```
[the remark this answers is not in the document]
rankindaveyg: Done, pls resolve if you feel it looks good.

jayofdoom (approved):

kannon92 (approved):

PR #1125, opened 2022-06-28
Expose client gRPC keepalive settings

severinson:


d80tb7 (approved):

pkg/client/connection.go
```diff
@@ -51,21 +61,24 @@ func CreateApiConnectionWithCallOptions(
...
 	streamInterceptors := grpc.WithChainStreamInterceptor(grpc_retry.StreamClientInterceptor(retryOpts...))
 	dialOpts := append(additionalDialOptions,
 		defaultCallOptions,
-		keepAliveOptions,
 		unuaryInterceptors,
 		streamInterceptors,
 		transportCredentials(config),
 	)
+	// gRPC keepalive options.
+	if config.GrpcKeepAliveTime > 0 || config.GrpcKeepAliveTimeout > 0 {
```
d80tb7, line 73: is this correct? What happens if `config.GrpcKeepAliveTime` = 0 and  `config.GrpcKeepAliveTime` = 1.  Should it be `&&`?

pkg/client/connection.go
```diff
@@ -20,8 +20,18 @@ import (
 )
 
 type ApiConnectionDetails struct {
-	ArmadaUrl                   string
-	ArmadaRestUrl               string
+	ArmadaUrl     string
+	ArmadaRestUrl string
+	// After a duration of this time, if the client doesn't see any activity it
+	// pings the server to see if the transport is still alive.
+	// If set below 10s, a minimum value of 10s is used instead.
```
d80tb7, line 27: Is the `max(time, 10s)` logic implemented?  I can't see it here

severinson: It's implemented inside the gRPC keepalive logic.

pkg/client/connection.go
severinson: I didn't want to be opinionated here on which settings to allow. I prefer to put in known-good default values later once we've rolled out a server that supports frequent keepalive everywhere.

PR #1127, opened 2022-06-28
Python client release issue 1072

kannon92:
PR for updating Author for Python Client and instructions for a manually upload of the package.  

We should automate this so I will leave it up to team if we want to merge this as is.  Or use it as inspiration for an automated story.  

Change the author name to our PyPI service account and add a README for manually uploading release artifacts.


jayofdoom (approved):

client/python/pyproject.toml
```diff
@@ -1,8 +1,8 @@
 [tool.poetry]
 name = "armada_client"
-version = "0.1.0"
+version = "0.1.0-dev"
```
JamesMurkin, line 3: Should this be `0.1.0`  instead of `0.1.0-dev`

PR #1153, opened 2022-06-30
Move unwatch py client

### LINKED ISSUE #1145: Move unwatch event inside Armada Client as static method
https://github.com/G-Research/armada/blob/master/client/python/armada_client/client.py#L187

```py
def unwatch_events(event_stream) -> None:
    """Closes gRPC event streams
    Closes the provided event_stream.queue
    :param event_stream: a gRPC event stream
    :return: nothing
    """
    event_stream.cancel()
```

I think it makes sense to put this function inside the main class as a static method so it is not left on its own.

Sharpz7:
- Moved client inside main client class
- Added `pylintrc` for other developers. Ignores gRPC stubs so that IDE's with pylin installed don't go haywire.

Closes #1145


kannon92 (changes requested):
When changing files you will have to generate the docs and commit them.  

The github action is yelling about the docs diff is different from what we generate.

kannon92 (dismissed):

client/python/pylintrc
```diff
@@ -0,0 +1,4 @@
+[MASTER]
```
jayofdoom, line 1: If we want to run pylint, we need to run it in CI. I am -1 to adding this file just for individual developer use.

jayofdoom (changes requested):

client/python/pylintrc
jayofdoom: I have no objection whatsoever if you want to push a GHA change to add pylint (look at `.github/workflows/python.yml`) -- just make sure that it's configured to work with `black` instead of against it :)

jayofdoom (approved):

client/python/pylintrc
Sharpz7: I think a better solution for now is me just adding these changes to my personal pylint settings. I might change my mind later ;))

kannon92 (approved):

richscott (approved):

richscott (approved):

kannon92 (approved):

dejanzele (approved):

PR #1156, opened 2022-07-01
Armada integration/e2e tests in Github Runners

jayofdoom:
Originally this used AWS hosted runners. Now, instead of that, we use runners from buildjet.com, simplifying configuration and removing basically all security concerns. Now this works for incoming PRs to our org as well, awesome.

Now the only issue is that a malicious PR could misuse runners provided by us until a reviewer cancelled the jobs. This is a very minor security issue and is in line with other risks G-R github org takes (e.g. `ilgpu` is using GPU runners with a very similar set of limitations).

Issue #1037


.github/workflows/go.yml
```diff
@@ -51,3 +51,33 @@ jobs:
         with:
           name: junit.xml
           path: test_reports/junit.xml
+  go-integration-tests:
```
jayofdoom, line 54: remove this from this file

scripts/pulsar.sh
```diff
@@ -2,12 +2,12 @@
 # pulsar-admin errors if the Pulsar server hasn't started up yet.
 sleep 10 
 
-docker exec -it pulsar bin/pulsar-admin tenants create armada
-docker exec -it pulsar bin/pulsar-admin namespaces create armada/armada
-docker exec -it pulsar bin/pulsar-admin topics delete-partitioned-topic persistent://armada/armada/events -f || true
-docker exec -it pulsar bin/pulsar-admin topics create-partitioned-topic persistent://armada/armada/events -p 2
+docker exec -i pulsar bin/pulsar-admin tenants create armada
```
jayofdoom, line 5: bad merge/rebase, revert this

.github/workflows/go-integration.yml
```diff
@@ -0,0 +1,89 @@
+name: Go End-to-End
+
+on:
+  push:
+    branches-ignore:
```
jayofdoom, line 5: These on: values are good for testing; make them look like the ones in go.yaml for merge

.github/workflows/go-integration.yml
```diff
@@ -0,0 +1,89 @@
...
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION }}
+      - name: Start EC2 runner
+        id: start-ec2-runner
+        uses: machulav/ec2-github-runner@v2
+        with:
+          mode: start
+          github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
+          ec2-image-id: ami-0d3d3ae42ef82e511
```
kannon92, line 38: Should these be secrets also?

jayofdoom: Explicitly not imo, for a handful of reasons:
1) it's not actually secret
2) it's easier to test/change/upgrade images using PRs (e.g. you can PR to test a new image)
3) troubleshooting job failures is painful if you're not certain what base image is being used

.github/workflows/go-setup/action.yml
```diff
@@ -4,10 +4,13 @@ description: "Sets up go environment and installs deps for go testing of armada"
 runs:
   using: "composite"
   steps:
+    # By setting cache:true, it enables caching of go modules, as long as the
```
jayofdoom, line 7: I think all the changes to this file are just ... leftover from how old this branch is. They can be reverted.

kannon92 (dismissed):

jayofdoom:
If statement clearly worked; this will need secrets setup in gr/armada repo for this to pass tests now.

.github/workflows/go-integration.yml
```diff
@@ -0,0 +1,91 @@
...
+            INTEGRATION_ENABLED: true
+        - run: make junit-report
+        - name: Upload junit report
+          uses: actions/upload-artifact@v2.2.4
+          with:
+            name: junit.xml
+            path: test_reports/junit.xml
+  stop-runner:
+      name: Stop self-hosted runner
+      if: always()
```
kannon92, line 71: Why would you always stop?

kannon92 (approved):

.github/workflows/go-integration.yml
jayofdoom: See if the comments I just added makes it clearer :)

ljubon:
Hey @jayofdoom, I've put some comments regarding the current workflow - but in general, as i understand you want to run some EC2 instances, do your tests, and shut down 🤔. I wanted to share with you something that @pavlovic-ivan did and its called [ephemeral-github-runner](https://github.com/pavlovic-ivan/ephemeral-github-runner)  and it's pretty cool and might solve some problems which you also targeting with this PR. Think about it if it's not too complex or out of scope for this PR. Anyway, we are both here to help and support you.
If the team agrees that the current workflow should be implemented, I would advise to challenge current security as sharing AWS credentials should be compact and clean 🔒

jayofdoom:
This needs secrets added to g-research/armada in order to run and pass tests and as a prereq to merge.

.github/workflows/go.yml
```diff
@@ -63,3 +63,23 @@ jobs:
...
+        - uses: actions/checkout@v3
+        - uses: ./.github/workflows/go-setup
+        - run: make build-ci
+        - run: make tests-e2e-setup
+        - run: make tests-e2e-no-setup
+          env:
+            INTEGRATION_ENABLED: true
+        - run: make junit-report
+        - name: Upload junit report
+          uses: actions/upload-artifact@v2.2.4
```
severinson, line 82: Instead of just uploading an artefact, can we have something that understands junit reports? So that we can have a nice dashboard showing which tests passed and failed, how long they took etc.
Maybe something like this: https://github.com/marketplace/actions/junit-report-action

dave-gantenbein: +1 for a dashboard!!

kannon92: This sounds good but I'd vote for a new story on it.  It'd be nice to get this merged in.

kannon92 (approved):

.github/workflows/go.yml
jayofdoom: That sounds useful but I agree it's a separate task and I'd like this merged.

severinson (approved):

.github/workflows/go.yml
severinson: Ok. Let's merge it.

severinson (approved):

PR #1164, opened 2022-07-06
Stop sending JobTerminatedEvents + no longer delete finished pods

JamesMurkin:
We now only delete pods that are still running when the lease cannot be renewed
 - There is no need to delete finished pods as they are already finished. Also deleting the pods means we lose the logs etc

We no longer send JobTerminatedEvent on lease being revoked
 - This seemed to cause more confusion than benefit it added.
 - Also there are some edge cases where it could be sent unexpectedly

### DIFF
--- a/internal/executor/service/job_manager.go
+++ b/internal/executor/service/job_manager.go
@@ -46,8 +46,17 @@ func (m *JobManager) ManageJobLeases() {
 	for _, chunk := range chunkedJobs {
 		failedJobs, err := m.jobLeaseService.RenewJobLeases(chunk)
 		if err == nil && len(failedJobs) > 0 {
-			m.reportTerminated(extractPods(failedJobs))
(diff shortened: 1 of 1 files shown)


internal/executor/service/job_manager.go
```diff
@@ -46,8 +46,17 @@ func (m *JobManager) ManageJobLeases() {
 	for _, chunk := range chunkedJobs {
 		failedJobs, err := m.jobLeaseService.RenewJobLeases(chunk)
 		if err == nil && len(failedJobs) > 0 {
-			m.reportTerminated(extractPods(failedJobs))
-			m.jobContext.DeleteJobs(failedJobs)
+			// This happens in case of lease being revoked - normally due to cancellation
```
kannon92, line 49: Can you add some tests in job_manager for this logic?

Issue #889, opened 2022-03-11
Fixing broken URL's on armada website pages


dans77777:
I found out some url's gives 404 error and are not working properly, I will try to fix it for better user / developer experience

jayofdoom:
Made a sheet listing multiple duplicate/out of date pages. Will need significant cleanup even before we deal with the CI... https://docs.google.com/spreadsheets/d/1LgQT6jEaL33LcglNl6rxEHVP_gu3m5BNdZEmN0DjV_M/edit?usp=sharing

jayofdoom:
Status | URL | Source link text
-- | -- | --
-1 Invalid URL | http://localhost:30001/ | http://localhost:30001
-1 Invalid URL | http://localhost:8080/ | http://localhost:8080
403 Forbidden | https://docs.github.com/articles/github-community-guidelines/ | GitHub Community Guidelines [301 from https://help.github.com/articles/github-community-guidelines/]
404 Not Found | https://armadaproject.io/helm/server.md | server Helm docs
404 Not Found | https://armadaproject.io/helm/executor.md | executor Helm docs
404 Not Found | https://armadaproject.io/quickstart/grafana-screenshot.png | img/src
404 Not Found | https://armadaproject.io/quickstart/lookout.png | img/src
404 Not Found | https://armadaproject.io/cmd/armadactl/cmd/submit.go | examples
404 Not Found | https://armadaproject.io/assets/img/kit/pro/ill/index_bg.svg | style: .index-page .section-hero
404 Not Found | https://armadaproject.io/pkg/api/submit.proto | definition
404 Not Found | https://armadaproject.io/pkg/api/event.proto | event.proto
404 Not Found | https://armadaproject.io/pkg/api/queue.proto | queue.proto
404 Not Found | https://armadaproject.io/pkg/api/usage.proto | usage.proto
404 Not Found | https://armadaproject.io/pkg/api/api.swagger.json | here
404 Not Found | https://armadaproject.io/internal/armada/permissions/permissions.go | here
404 Not Found | https://armadaproject.io/pkg/client/queue/permission_verb.go | here

Status	URL	Source link text
-1 Invalid URL	http://localhost:30001/	[http://localhost:30001](https://armadaproject.io/quickstart)
-1 Invalid URL	http://localhost:8080/	[http://localhost:8080](https://armadaproject.io/quickstart)
[...]

PR #1172, opened 2022-07-06
Fix 404 errors in api.md

jayofdoom:
The API documentation points to paths relative to github that only
work from github. Instead, we make them fully qualified paths. This
has a side effect of making links from docs/ in forks point to the
upstream official documentation.

Additionally, the helm chart documentation has been consolidated into
a single file to permit it to be integrated into armadaproject.io.
This was required to get rid of all 404 errors in /api since it links
to the helm charts.

In the future, we may want to consider setting up a helm collection
in gh-pages that can be populated via a directory, but given the helm
docs hadn't been touched in almost 2 years, the simpler solution seemed
wiser.

Issue: #889

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -27,7 +27,8 @@ jobs:
       - name: Copy site files
         run: |
           cp master/docs/*.md gh-pages/
-          cp -r master/docs/quickstart gh-pages/_pages
+          rm gh-pages/docs-readme.md
+          cp -r master/docs/quickstart gh-pages/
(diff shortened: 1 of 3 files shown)


jayofdoom:
If we'd like, I can add a separate commit to gh-pages branch to add the now-published helm docs to the menu bar if we'd like. I don't have good context on how useful or quality those docs are given they haven't been touched in a while, but I think it's probably OK for now just to be published and linked to from other pages.

richscott (approved):

kannon92 (approved):

dejanzele (approved):

PR #1176, opened 2022-07-07
increase grpc message limit to 32mb

dejanzele:


kannon92 (changes requested):
Since we are changing this, I think making this a configurable value is probably the best move right now.

PR #1179, opened 2022-07-07
[small][Readme File]

kindyluv:


makefile
```diff
@@ -5,7 +5,7 @@ PWD := $(shell pwd)
 # Using 'where' on Windows and 'which' on Unix-like systems, respectively
 # We do not check for 'date', since it's a cmdlet on Windows, which do not show up with where
 # (:= assignment is necessary to force immediate evaluation of expression)
-EXECUTABLES = git docker kubectl
```
kannon92, line 8: I think we want to keep kubectl in our makefile.  It is used in this file for non client work so we should still require it.

client/python/tests/test.py
```diff
@@ -0,0 +1,117 @@
...
+from armada_client.k8s.io.api.core.v1 import generated_pb2 as core_v1
+from armada_client.k8s.io.apimachinery.pkg.api.resource import (
+    generated_pb2 as api_resource,
+)
+
+# The python GRPC library requires authentication
+#  data to be provided as an AuthMetadataPlugin.
+# The username/password are colon-delimted and base64 encoded as per RFC 2617
+
+class GrpcBasicAuth(grpc.AuthMetadataPlugin):
```
kannon92, line 17: Why was this added as part of the README changes?

client/python/README.md
```diff
@@ -1,13 +1,61 @@
...
+3) Install [poetry](https://python-poetry.org/) on your local machine by following the instructions on [python poetry documentation](https://python-poetry.org/docs/). 
+
+4) Fork the G-Research Armada repository to your personal repository.
+
+5) Clone the forked armada repository on your personal git repository by running git clone [Armada Repo](https://github.com/G-Research/armada.git) and navigate into the new directory Armada project by running cd armada.
+
+On Mac Machine
+-
+
+1) Open your favorite shell, for example, [Homebrew](https://brew.sh/) or [MacPorts](https://www.macports.org/).
```
kannon92, line 28: Do you mean package manager?

client/python/README.md
```diff
@@ -1,13 +1,61 @@
...
+
+2) Install [Git](https://git-scm.com/) by running `$ brew install git` or `$ sudo port install git` or follow the instructions [here](https://www.atlassian.com/git/tutorials/install-git)
+
+3) Install [Python](https://docs.python-guide.org/starting/install3/osx/)
```
kannon92, line 32: I probably would suggest pyenv or something that allows more than one python version.  We also use python 3.9 in our CI so we should tell them to install python 3.9.  I don't think we support 3.10 yet.

client/python/README.md
```diff
@@ -1,13 +1,61 @@
...
+
+4) Install [poetry](https://python-poetry.org/) on your local machine by following the instructions on [python poetry documentation](https://python-poetry.org/docs/).
+
+5) Fork the G-Research Armada repository to your personal repository.
 
+6) Clone the forked armada repository on your personal git repository by running git clone [Armada Repo](https://github.com/G-Research/armada.git) and navigate into the new directory Armada project by running cd armada.
 
 
+On Windows Machine
```
kannon92, line 41: I wonder if we need WSL to be enabled for windows?  You have descriptions for installing on native windows.

kannon92 (changes requested):
README looks good.  I have a few minor fixes in the review comments. 

I think this PR should only include the README.md changes, so please remove kind, tests, your changes to makefile and the proto directory.

PR #1182, opened 2022-07-08
Fix deleted pods not getting reported done 

JamesMurkin:
Currently if a pod is unexpectedly deleted
 - We send a JobFailedEvent

This isn't sufficient as this cases the lease to expire and the job to be reran which is confusing.

We now also correctly report done for the job, so it doesn't get reran


internal/executor/service/job_manager_test.go
```diff
@@ -10,15 +10,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
-	"github.com/G-Research/armada/internal/common/util"
+	commonUtil "github.com/G-Research/armada/internal/common/util"
 	podchecksConfig "github.com/G-Research/armada/internal/executor/configuration/podchecks"
 	"github.com/G-Research/armada/internal/executor/context"
+	context2 "github.com/G-Research/armada/internal/executor/context/fake"
```
carlocamurri, line 16: Is `context2` appropriate here? Would it be better to call it `fakecluster` or `fakecontext`?

PR #1184, opened 2022-07-08
Jobservice events

kannon92:
Solve https://github.com/G-Research/armada/issues/1138

Add a redis cache for JobService.


cmd/jobservice/main.go
```diff
@@ -0,0 +1,54 @@
...
+	"golang.org/x/sync/errgroup"
+
+	"github.com/G-Research/armada/internal/common"
+	"github.com/G-Research/armada/internal/jobservice"
+	"github.com/G-Research/armada/internal/jobservice/configuration"
+)
+
+const CustomConfigLocation string = "config"
+
+func init() {
```
severinson, line 21: I suggest looking at how the testsuite is implemented. I think that's a cleaner approach. It doesn't require having an init function and so on.

internal/jobservice/configuration/types.go
```diff
@@ -0,0 +1,20 @@
+package configuration
+
+import (
+	"github.com/go-redis/redis"
+
+	grpcconfig "github.com/G-Research/armada/internal/common/grpc/configuration"
+	"github.com/G-Research/armada/pkg/client"
+)
+
+type JobServiceConfiguration struct {
```
severinson, line 10: I think we can re-use the config we already use for the Armada client? The test suite also takes this approach.

cmd/jobservice/main.go
```diff
@@ -0,0 +1,54 @@
...
+	pflag.StringSlice(CustomConfigLocation, []string{}, "Fully qualified path to application configuration file (for multiple config files repeat this arg or separate paths with commas)")
+	pflag.Parse()
+}
+
+func main() {
+	common.ConfigureLogging()
+	common.BindCommandlineArguments()
+	g, ctx := errgroup.WithContext(context.Background())
```
severinson, line 29: Why both an errgroup and a waitgroup? I suggest using only errgroups.

cmd/jobservice/main.go
```diff
@@ -0,0 +1,54 @@
...
+		case <-ctx.Done():
+			return nil
+		case sig := <-stopSignal:
+			wg.Done()
+			shutdown()
+			return fmt.Errorf("received signal %v", sig)
+		}
+	})
+
+	wg.Wait()
```
severinson, line 52: This doesn't look right. Starting the service in an errgroup and waiting on the waitgroup.

internal/jobservice/application.go
```diff
@@ -0,0 +1,51 @@
...
+	defer func() {
+		if err := db.Close(); err != nil {
+			log.WithError(err).Error("failed to close Redis client")
+		}
+	}()
+
+	grpcServer := grpcCommon.CreateGrpcServer(config.Grpc.KeepaliveParams, config.Grpc.KeepaliveEnforcementPolicy, []authorization.AuthService{&authorization.AnonymousAuthService{}})
+	log.Info("JobService service listening on ", config.GrpcPort)
+
+	redisJobRepository := repository.NewRedisJobServiceRepository(db)
```
severinson, line 35: Is Redis a good fit for this? Why do we need a RedisJobServiceRepository?

severinson: I'd probably use postgres instead of Redis to store the cache state on disk. And then also store the state of all jobs watched in-memory in the application itself as a map.

internal/jobservice/eventstojobs/event_job_response.go
```diff
@@ -0,0 +1,36 @@
+package eventstojobs
+
+import (
+	"fmt"
+
+	"github.com/G-Research/armada/pkg/api"
+	"github.com/G-Research/armada/pkg/api/jobservice"
+)
+
+func EventsToJobResponse(message api.EventMessage) (*jobservice.JobServiceResponse, error) {
```
severinson, line 10: Why do we need this? Are not just interested in tracking the current state of each job?

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,72 @@
...
+	jobServiceRepository repository.JobServiceRepository) *EventsToJobService {
+	return &EventsToJobService{
+		queue:                queue,
+		jobsetid:             jobsetid,
+		jobid:                jobid,
+		jobServiceConfig:     jobServiceConfig,
+		jobServiceRepository: jobServiceRepository}
+}
+
+func (eventToJobService *EventsToJobService) SubscribeToJobSetId(context context.Context) error {
```
severinson, line 39: I'd use the approach used in the test suite to watch events.

internal/jobservice/eventstojobs/event_job_response.go
kannon92: In short, the job response only cares about terminal events and running.  So we only return those but there are 14-15 other events that can happen so we use this to translate events into our jobresponse.

internal/jobservice/configuration/types.go
kannon92: Nice find!  I attempted to do this.  Please take a look.

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,129 @@
...
+			Watch:          true,
+			FromMessageId:  fromMessageId,
+			ErrorIfMissing: false,
+		})
+		if err != nil {
+			log.Error(err)
+			return err
+		}
+		for {
+			msg, err := stream.Recv()
```
richscott, line 88: I sort of wish that protoc generated code such that GetJobSetEvents() returned a Go channel that we could select upon, instead of doing repeated Recv() calls. (This is gogo's protoc? I wonder if other, newer protoc's would generate more Go-chan-oriented code?)

kannon92: Yea I think we need to do some research into go and protoc anyway.  We can add this to that one issue.

config/jobservice/config.yaml
```diff
@@ -0,0 +1,21 @@
+grpcPort: 60003
+httpPort: 8090
+SubscribeJobSetTime: 10
```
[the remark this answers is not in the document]
kannon92: Fixed.

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (eventToJobService *EventsToJobService) SubscribeToJobSetId(context context.Context) error {
+
+	return client.WithEventClient(&eventToJobService.jobServiceConfig.ApiConnection, func(c api.EventClient) error {
+		jobIdMap, err := eventToJobService.StreamCommon(c, context)
+		for key, element := range jobIdMap {
+			log.Infof("key %s element: %s", key, element.State)
+			e := eventToJobService.jobServiceRepository.UpdateJobServiceDb(key, element)
+			if e != nil {
+				panic(e)
```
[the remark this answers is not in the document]
kannon92: @severinson and @d80tb7 What are your thoughts on error handling in this code base? 

I changed this to return an error but I don't really know what pattern to follow.

internal/jobservice/eventstojobs/event_job_response.go
```diff
@@ -0,0 +1,56 @@
+package eventstojobs
+
+import (
+	"fmt"
+
+	"github.com/G-Research/armada/pkg/api"
+	"github.com/G-Research/armada/pkg/api/jobservice"
```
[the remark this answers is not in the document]
jayofdoom: I have the opposite preference here -- I don't like aliased package names in golang unless they are required to ensure separated namespaces. I think that makes @kannon92 the tie-breaking vote :laughing:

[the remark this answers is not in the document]
kannon92: Since I changed it already, I vote for js.

internal/jobservice/eventstojobs/event_job_response.go
```diff
@@ -0,0 +1,56 @@
...
+	switch message.Events.(type) {
+	case *api.EventMessage_Submitted, *api.EventMessage_DuplicateFound, *api.EventMessage_Running, *api.EventMessage_Failed, *api.EventMessage_Succeeded, *api.EventMessage_Cancelled:
+		return true
+	default:
+		return false
+	}
+}
+func IsEventTerminal(message api.EventMessage) bool {
+	switch message.Events.(type) {
+	case *api.EventMessage_Cancelled, *api.EventMessage_Succeeded, *api.EventMessage_Failed:
```
d80tb7, line 39: I think DuplicateFound is technically a terminal state

internal/jobservice/server/server.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+		return jobResponse, err
+	}
+	response, err := s.jobRepository.GetJobStatus(opts.JobId)
+	if err != nil {
+		log.Fatal(err)
+		return nil, err
+	}
+	var subscribeToJobSet = eventstojobs.IsStateTerminal(response.State)
+	if !subscribeToJobSet {
```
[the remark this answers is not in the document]
d80tb7: oh- I see- you return earlier in the function if that's the case- I stil think you have a race condition though where two people subscribe at the same time.  I think you need a lock.

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,129 @@
...
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			// This allows us to subscribe for x amount of time
+			case <-ctxTimeout.Done():
+				log.Info("Hit a timeout")
+				return nil
+			default:
+			}
+			if !IsEventAJobResponse(*msg.Message) {
```
d80tb7, line 104: wouldn't this be simpler as just 
```
if IsEventAJobResponse{
  ...
}
```
rather than declaring an empty if and then using the else for the logic

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,129 @@
...
+			if err != nil {
+				log.Error(err)
+				return err
+			}
+			fromMessageId = msg.GetId()
+			currentJobId := api.JobIdFromApiEvent(msg.Message)
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			// This allows us to subscribe for x amount of time
```
d80tb7, line 98: why do we want to subscribe for some time?

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,129 @@
...
+				if eventJobErr != nil {
+					// This can mean that the event type reported from server is unknown to the client
+					log.Error(eventJobErr)
+				}
+				terminalEventClientId := false
+				if eventToJobService.jobid == currentJobId {
+					terminalEventClientId = IsEventTerminal(*msg.Message)
+				}
+				val, ok := jobIdMap[currentJobId]
+				if ok && val.State != jobStatus.State {
```
d80tb7, line 116: why do we have an if here when we do the same on both conditions?

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,129 @@
...
+func NewEventsToJobService(
+	queue string,
+	jobsetid string,
+	jobid string,
+	jobServiceConfig *configuration.JobServiceConfiguration,
+	jobServiceRepository repository.JobServiceRepository) *EventsToJobService {
+	return &EventsToJobService{
+		queue:                queue,
+		jobsetid:             jobsetid,
+		jobid:                jobid,
```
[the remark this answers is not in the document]
kannon92: We can discuss.

In general, I found that I can't really subscribe from a client grpc call in the way that you would prefer.  If you return a rpc call, it will cancel the go routine once the call returns.  

So I built a subscribe timer that subscribes on a job-set for x amount of time and it will return if you have a terminal state for your job id.

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,129 @@
...
+	jobIdMap := make(map[string]*js.JobServiceResponse)
+	var fromMessageId string
+	// I found that GRPC will not allow you to run something in background and return a value back to caller.
+	// GRPC will cancel the context once your request returns.
+	// So we are going to introduce a timer for how long to subscribe to event.
+	// This will allow the rpc call to listen for all events in a given job-set
+	// But we will return after SubscribeJobSetTime s
+	ctxTimeout, cancel := context.WithTimeout(ctx, time.Duration(eventToJobService.jobServiceConfig.SubscribeJobSetTime)*time.Second)
+	defer cancel()
+	err := client.WithEventClient(&eventToJobService.jobServiceConfig.ApiConnection, func(c api.EventClient) error {
```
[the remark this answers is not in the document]
kannon92: Good call.

kannon92:
Decided to do this as a new PR.  I addressed the comments of this PR in https://github.com/G-Research/armada/pull/1259 and followed up with Chris offline to discuss the implementation.

Issue #1177, opened 2022-07-07
build fails on linux if TESTS_IN_DOCKER isn't set


andrewpcone:
`make proto` should work out of the box. Right now it doesn't, because `TESTS_IN_DOCKER` isn't set, which makes `GO_TEST_CMD` empty, which causes `swagger` and other commands to run on the host, rather than in docker. If the user's `PATH` doesn't include `~/go/bin` or other `$GOPATH/bin`, then `swagger` fails.

I propose that we build exclusively in docker by using `$GO_CMD` in front of all the `make proto` commands instead of relying on the system go installation.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-57) by [Unito](https://www.unito.io)

andrewpcone:
alternatively we can have the makefile call, for example  `` `go env GOPATH`/swagger `` or similar

kannon92:
Could someone reproduce this? Not sure if it’s fixed or not.

Sharpz7:
Since we now use mage, this isn't needed. Closing

PR #1185, opened 2022-07-08
GO_CMD for dockerized build

andrewpcone:
Addresses #1177


makefile
```diff
@@ -54,6 +54,7 @@ endif
 # For now just take the first one
 DOCKER_GOPATH_TOKS := $(subst :, ,$(DOCKER_GOPATH:v%=%))
 DOCKER_GOPATH_DIR = $(word 1,$(DOCKER_GOPATH_TOKS))
+@echo docker gopath: ${DOCKER_GOPATH_DIR}
```
kannon92, line 57: I'm unfamilar with this syntax in make.  What does this do?

andrewpcone:
oops that's debug code sorry

makefile
richscott: If it's not expected to be needed in the future, I'd suggest removing it.

kannon92 (approved):

andrewpcone:
This isn't needed if go is added to the path of the host box, so I'm going to get rid of it

PR #1189, opened 2022-07-08
Added new wrappers to client

### LINKED ISSUE #1188: Extra Python client wrappers
`submit_pb2.JobSubmitRequest` and `submit_pb2.JobSubmitRequestItem` are not currently wrapped into `client.py`, unlike all other `submit_pb2` methods.

This does create issues with how weird the arg objects are, but kwargs can be used to solve this, exactly the same as how `**query_params` are used for the queue wrapper.

Sharpz7:
I tried to document the methods as best as I could. Please suggest any changes. I have also tested that these changes work by running a modified version of `test_basic_auth` with and without the changes.

Closes #1188


client/python/armada_client/client.py
```diff
@@ -194,3 +195,34 @@ def unwatch_events(event_stream) -> None:
...
         event_stream.cancel()
+
+    def create_job_request(
+        self, queue: str, job_set_id: str, job_request_items: List, **job_params
+    ):
+        """Create a job request.
+
+        :param queue: The name of the queue
+        :param job_set_id: The name of the job set (a grouping of jobs)
+        :param job_params: Job Object
```
kannon92, line 206: I'm confused by job_request_items and **job_params.  

Should we have both?

Sharpz7: Yes, because job_request_items are a list of `submit_pb2.JobSubmitRequestItem` where as `job_params` covers all other possible kwargs. Can you think of a better way to word it? 

I'll let you respond, but you have made me realise a docstring param is missing so I will add that in when we have finished this discussion.

client/python/armada_client/client.py
```diff
@@ -194,3 +195,34 @@ def unwatch_events(event_stream) -> None:
...
+        )
+
+    def create_job_request_item(
+        self, pod_spec: core_v1.PodSpec, priority: int = 1, **job_item_params
+    ):
+        """Create a job request.
+
+        :param priority: The priority of the job
+        :param pod_spec: The k8s pod spec of the job
+        :param job_item_params: Job Item Object
```
jayofdoom, line 223: I'm a bit confused about why you're using `**` throughout here... seems like we shouldn't have a variable number of arguments? If it's supposed to be a list of JobItems, the docstring should indicate that.

I'm really curious why the construct you used in the `create_job_request` method won't work here as well.

client/python/docs/README.md
```diff
@@ -10,7 +10,7 @@ Usage
 Easy way:
 - Ensure all protobufs files needed for the client are generated by running
   `make python` from the repository root.
-- `tox -edocs` will create a valid virtual environment and use it to generate
+- `tox -e docs` will create a valid virtual environment and use it to generate
```
jayofdoom, line 13: FWIW this works with or without the space, but I'm OK adding the space

client/python/armada_client/client.py
jayofdoom: Heh, I see Kevin's review/comment now. We need to make sure the docstring makes this clear since that makes two of us who didn't understand what was happening :)

Sharpz7: The first method has an argument that is actually a list of what I am putting in, where as in the 2nd method, there are other args that I have not stated, as they are present, but their objects are more complicated. See here:

```py
message JobSubmitRequestItem {
    double priority = 1;
    string namespace = 3;
    string client_id = 8;
    map<string, string> labels = 4;
    map<string, string> annotations = 5;
    map<string, string> required_node_labels = 6 [deprecated = true]; // Use PodSpec.NodeSelector instead
    k8s.io.api.core.v1.PodSpec pod_spec = 2 [deprecated = true]; // Use PodSpecs instead
    repeated k8s.io.api.core.v1.PodSpec pod_specs = 7;
    repeated IngressConfig ingress = 9;
    repeated ServiceConfig services = 10;
}
```

Kevin mentioned this to me in a DM in slack, that is why I kept it that way, but in theory, yes I could add all the args for this instead of using kwargs. I would then ask why the same hasn't been done for queues, as it uses `**query_params` but it too could be completly typed up arg by arg - I assumed there was a reason for not doing so.

https://files.mcaq.me/056kb.jpg

If you give the co-ahead, I can add "Rich" arguments to all the queue methods and the new methods I created ;))

client/python/docs/README.md
Sharpz7: Never new that! Good to know ;))

kannon92 (approved):

client/python/armada_client/client.py
jayofdoom: I mean, I think the core issue here is that the API is awkward, probably because (as you said in another issue) gRPC+python just works a little awkward together. Maybe having better examples will clear it up, but there's no reason to hold up merge of this for that reason.

jayofdoom (approved):

dejanzele (approved):

PR #1191, opened 2022-07-08
Example python files

### LINKED ISSUE #1146: Write python example scripts and add Documentaition
Create 4 example python scripts:

- queue.py covering queue code
- jobs.py covering basic jobs code
- monitor.py covering how to monitor jobs and jobsets
- priority.py covering how to change the priority of a job.

Also create a README within an examples folder with more infomation, like what is done [here](https://github.com/zenml-io/zenml/tree/main/examples)

Sharpz7:
Closes #1146


client/python/.env
```diff
@@ -0,0 +1,2 @@
+HOST=armada.dev.armadaproject.io
```
kannon92, line 1: We should not commit this URL into the repo.  The url is open for all to use.

I think you should just add a placeholder like this: https://stackoverflow.com/a/43664611

Sharpz7: I forgot to add this to the gitignore... My bad

client/python/examples/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+```bash
+export ARMADA_SERVER=localhost
+export ARMADA_PORT=443
+export DISABLE_SSL=true
+```
+
+You can then simply run one of the examples
+
+```bash
```
kannon92, line 30: Are you missing an explanation of how to create a python virtual environment and how to install client to an environment?

kannon92: True but if any tries to copy your code in this PR it will fail without any notice of why that is the case?

Sharpz7: I get what you mean, I would assume that if a user is checking the example README they would already have it installed, but I understand that this is not always the case. I will link to https://github.com/G-Research/armada/blob/master/client/python/README.md directly as a guide for installing?

client/python/examples/README.md
```diff
@@ -0,0 +1,185 @@
...
+#### Walkthrough
+
+```py
+def watch_job_set(client: ArmadaClient, queue: str, job_set_id):
+```
+
+Watches all changes on a job_set. It attempts to connect multiple times instead of sleeping like in `general.py`. It also demonstrates using the clients `unmarshal_event_response()` method to better access the event object.
+
+```py
+def watch_queue(client, queue):
```
kannon92, line 125: We should ask if this would be useful before we put in examples?

client/python/examples/async_logging.py
```diff
@@ -0,0 +1,210 @@
...
+            print(f"Queue {queue} already exists")
+            client.update_queue(name=queue, priority_factor=1)
+        else:
+            raise e
+
+    # Time for the watcher to pick it up
+    time.sleep(1)
+
+    # Some different commands for logging to detect
+    client.update_queue(name=queue, priority_factor=2)
```
kannon92, line 151: This seems to make the reprioritize_jobs does nothing.  You update queue with 2 and then reprioritize with 2 again

Sharpz7: Oh. I thought that queues had priority and then job_sets had their own seperate priority?

client/python/examples/README.md
Sharpz7: Maybe discuss this on Wednesday?

kannon92: Sure but I don't know if any of us have a good answer for you.  I'd ask the GR Armada team about this.

client/python/examples/async_logging.py
kannon92: Yea you know this better than I.  You are right.

client/python/examples/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+Example of using the armada client to create a queue, jobset and job,
+then watch for the job to succeed or fail.
+
+#### Walkthrough
+
+```py
+def create_dummy_job(client: ArmadaClient):
+```
+Armada uses kubernetes to control jobs, so we need to define a podspec that are job will use.
```
kannon92, line 46:
```suggestion
Armada uses Kubernetes to control jobs, so we need to define a podspec that our job will use.
```

client/python/examples/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+```bash
+python3 simple.py
+```
+
+## Example Files
+
+### simple.py
+
+Example of using the armada client to create a queue, jobset and job,
```
kannon92, line 38:
```suggestion
Example of using the Armada client to create a queue, jobset and job,
```

client/python/examples/async_logging.py
```diff
@@ -0,0 +1,210 @@
...
+                    },
+                ),
+            )
+        ],
+    )
+
+    return [client.create_job_request_item(priority=1, pod_spec=pod)]
+
+
+def watch_job_set(client: ArmadaClient, queue: str, job_set_id):
```
kannon92, line 50: I don't love the infinite function here but I don't know if you need to change.

client/python/examples/general.py
```diff
@@ -0,0 +1,173 @@
...
+    Will automatically return if the job fails or the event stream is closed.
+    """
+
+    # Contains all the possible message types
+    for event in event_stream:
+
+        event = client.unmarshal_event_response(event)
+
+        if event.message.job_id == job_id:
+            if event.type == EventType.failed:
```
kannon92, line 64: There are a few other events you are missing.  (Cancelled, DuplicatedFound). You can look at the events code in go to see what we consider terminal.

client/python/examples/general.py
```diff
@@ -0,0 +1,173 @@
...
+        print("Job submitted")
+
+    elif not test:
+        print("Failed")
+
+    # Close the event stream
+    client.unwatch_events(event_stream)
+
+
+def creating_queues_example(client, queue):
```
kannon92, line 105: Should we also do a more complicated queue with all the parameters?  It's be nice to explain the other queues and what they mean in code.

client/python/examples/simple.py
```diff
@@ -0,0 +1,90 @@
+"""
+Example of using the armada client to create a queue, jobset and job,
```
kannon92, line 2:
```suggestion
Example of using the Armada client to create a queue, jobset and job,
```

kannon92 (commented):
I think I'd like to see a bit more in queue examples and there are a few typos.

client/python/examples/general.py
Sharpz7: I think doing this in a seperate file definitely makes sense. I think I would need a hand understanding the service and ingress configs though.

client/python/examples/async_logging.py
Sharpz7: Would putting a timer on it be effective? Like maybe 30 seconds - a minute?

client/python/examples/general.py
kannon92: Same!

client/python/examples/async_logging.py
kannon92: We will see what others think.  A timer would be good but I may not be necessary for this

client/python/examples/README.md
kannon92: I talked with @d80tb7 and I think we should remove watch_queue everywhere.

Sharpz7: Okay. I think its just used in `async_logging` so I will remove that shortly

client/python/examples/async_logging.py
Sharpz7: The infinte function will be removed and replaced with something else.

Sharpz7:
This is getting towards being ready now, but I have a few concerns:

- I think `wait_for_job_event` should be intergrated into the python client
- I think there should be some file in the client called `dummy.py` for creating dummy job_specs

The reasons for these suggestions are I feel examples (Especially Python code examples) should have as little bloat in them as physically possible.

client/python/examples/README.md
```diff
@@ -0,0 +1,263 @@
...
+
+Example of using the Armada client to create a queue, jobset and job,
+then watch for the job to succeed or fail.
+
+#### Walkthrough
+
+```py
+def create_dummy_job(client: ArmadaClient):
+```
+Armada uses Kubernetes to control jobs, so we need to define a podspec that are job will use.
```
kannon92, line 50:
```suggestion
Armada uses Kubernetes to control jobs, so we need to define a podspec that our jobs will use.
```

client/python/examples/README.md
```diff
@@ -0,0 +1,265 @@
+# Python Client Examples
+
+Currently there are three example files:
```
kannon92, line 3:
```suggestion
Currently there are five example files:
```

client/python/examples/README.md
```diff
@@ -0,0 +1,265 @@
...
+client.cancel_jobs(job_id=job_id)
+```
+
+To cancel a specific job, its job_id must be read from the `client.submit_jobs` response, and then it can be cancelled.
+
+```py
+client.cancel_jobs(queue=queue, job_set_id=job_set_id1)
+```
+
+You can also cancel all jobs on a job_set with its queue and job_set id.
```
kannon92, line 162: Show an example here also of that?  I think cancelling by jobset is the more common use case imo.

client/python/examples/README.md
```diff
@@ -0,0 +1,265 @@
...
+
+Currently there are three example files:
+
+- [simple.py](#simple.py)
+- [general.py](#general.py)
+- [queues.py](#queuespy)
+- [cancelling.py](#cancellingpy)
+- [monitor.py](#monitor.py)
+
+There are also some codeblocks with useful code snippets:
```
kannon92, line 11: I’d make this reflect what you trying to show with auth ie this section could reflect examples with more complicated auth

kannon92 (changes requested):
Looks good.  Few minor things and a request to add cancel by jobset

client/python/examples/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+> All queues need a priority factor to be created, which is used to determine the order in which jobs are run.
+> ```py
+> client.create_queue(name=queue, priority_factor=1)
+> ```
+
+
+### general.py
+
+[Link to File](https://github.com/G-Research/armada/tree/master/client/python/examples/general.py)
```
kannon92, line 68: These links don't work if you try to view the markdown.  Maybe you should just use a relative link and not the branch name?

Sharpz7: They won't work until the code is merged, but I can use relative links now

kannon92 (dismissed):

client/python/examples/simple.py
```diff
@@ -0,0 +1,90 @@
+"""
+Example of using the Armada client to create a queue, jobset and job,
+then watch for the job to succeed or fail.
```
jayofdoom, line 3: This is the sort of information that should be above, where you're talking about what examples we have. IMO

client/python/examples/README.md
```diff
@@ -0,0 +1,246 @@
+# Python Client Examples
+
+Currently there are five example files:
+
+- [simple.py](#simplepy)
```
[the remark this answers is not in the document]
jayofdoom: Yeah, you have the info I want below, but I think you should just make the actual-primary-name you refer to the examples as descriptive.

jayofdoom (commented):
I presume the examples are still being linted since it's in the right folder, if not we should make sure of that.

My comments are mostly just suggestions as to how to make this more friendly; feel free to take them or leave them or just take the spirit of them and do something different :)

client/python/examples/README.md
```diff
@@ -0,0 +1,246 @@
+# Python Client Examples
+
+Currently there are five examples:
+
+- [Intro to using the Python Client](#simplepy)
```
jayofdoom, line 5: Since you still call these "simple.py" and similar below, consider adding those names here
e.g.
```suggestion
- [Intro to using the Python Client - simple.py](#simplepy)
```

jayofdoom (approved):

kannon92 (dismissed):

jayofdoom (approved):

kannon92 (approved):

richscott (approved):
These are excellent - this will be a great welcome to Python devs who want to start playing with Armada.

PR #1202, opened 2022-07-11
Started to add comments.

Sharpz7:
Closes https://github.com/G-Research/armada/issues/1200


Sharpz7:
It seems like a there are a lot of tests to change that the proto files do not change... Should I close this issue for now? @jayofdoom

pkg/api/event.proto
```diff
@@ -239,13 +259,15 @@ message EventMessage {
...
 
+// The cause for a job to error.
 enum Cause {
     Error = 0;
     Evicted = 1;
     OOM = 2;
     DeadlineExceeded = 3;
 }
 
+// A Docker containers status.
```
kannon92, line 270: Not sure if this is true.  This is the container status and Docker is one container runtime for kubernetes.  

I would just leave this as A container status.

PR #1203, opened 2022-07-11
Custom pyclient event response

Sharpz7:
Closes https://github.com/G-Research/armada/issues/1198


Sharpz7:
I am going to redesign the Objects so that there is an Event Object, and a Message Object that will only accept parameters. This will make documenting it a lot cleaner

client/python/armada_client/event.py
```diff
@@ -0,0 +1,102 @@
...
+
+import google.protobuf.timestamp_pb2 as timestamp_pb2
+
+from armada_client.armada import event_pb2, queue_pb2
+from armada_client.k8s.io.apimachinery.pkg.api.resource import (
+    generated_pb2 as resource_pb2,
+)
+
+
+class EventType(enum.Enum):
```
kannon92, line 13: Can you add some unit tests for this class?

kannon92: I'd like to see accessing members of these events and aiming to test all the enums.

kannon92 (changes requested):

client/python/armada_client/client.py
```diff
@@ -59,6 +61,16 @@ def get_job_events_stream(
         )
         return self.event_stub.GetJobSetEvents(jsr)
 
+    def unmarshal_event_response(self, event: event_pb2.EventStreamMessage) -> Event:
```
kannon92, line 64: This should be a static function?  It doesn't seem to depend on self for anything.

client/python/armada_client/event.py
```diff
@@ -0,0 +1,100 @@
...
```
jayofdoom, line 13: Is this a duplication of types that are in the protobuf? Is this code likely to break as Armada's API evolves?

client/python/armada_client/event.py
```diff
@@ -0,0 +1,100 @@
...
+    reprioritizing = "reprioritizing"
+    updated = "updated"
+
+
+@dataclass
+class EventMessage:
+    """
+    EventMessage is the message type for the event stream.
+
+    Based on event_pb2.EventMessage
```
jayofdoom, line 43: IMO This is likely unacceptable; we cannot create manually curated files out of code that is generated from proto files. We cannot expect every commit to change an Armada API to require manual python code changes. 

Have you looked into if we can modify *how* the code generation happens for python, and perhaps create these dynamically?

jayofdoom (changes requested):
I believe it's unacceptable to manually create types that have their source of truth in a proto file. 

I agree it makes the interface better, but it's unlikely to be long-term maintainable unless we make these types generate automatically from the protobuf definitions.

client/python/armada_client/event.py
Sharpz7: Yeah. In its current form it would break, but I can add another enum for "unknown", which won't break any of its type access.

client/python/armada_client/event.py
Sharpz7: There are pretty much 0 options for generating the code differently, without major changes to how the proto stuff is layed out.

The problem is the fact that there's one proto message for each event type, and then G-Research have used the "one of" function, which is not very well supported in python, or golang. If you check the golang event.go file, it's virtually layed out the same as what I have done.

Sharpz7: https://github.com/Sharpz7/armada

kannon92: Actually, the golang event_util.go is a manual function they added to make working with events easier:

https://github.com/G-Research/armada/blob/master/pkg/api/event_util.go

Sharpz7: So casting this class onto the grpc class doesn't actually change anything about the object, so this will never break anything. It's purely here to make working with the class easier, and have everything written in a nicer fashion. 

If this is okay, I will add a docstring explaining this.

jayofdoom: We talked about this in slack. I have a better understanding of the struggles involved around this.

The idea moving forward is:
- This is fine, but not awesome. There is no awesome solution.
- We will document what code (outside of *.proto files) needs to be touched when an API is updated. -- this can be in a follow up / or just have an issue filed to do this later

Adam may, at his choice, spend some time in a timebox to generate this in CI -- but we think it'll be difficult to implement this in an consistently unbroken way.

Sharpz7:
Depending on what features are kept / removed based on https://github.com/G-Research/armada/pull/1239, more docstrings might be required to explain the subtle differences between how the python client handles these proto objects vs Golang or C#

client/python/docs/source/conf.py
```diff
@@ -12,27 +12,28 @@
 #
 import os
 import sys
-sys.path.insert(0, os.path.abspath('../..'))
+
+sys.path.insert(0, os.path.abspath("../.."))
```
jayofdoom, line 16: Please revert all changes in this file since they are noops.

Sharpz7: Coolio :)

jayofdoom (dismissed):

kannon92 (changes requested):
I'd like to see a unit test around event.py. 

Looks like it is possible to test __repr__:  

https://stackoverflow.com/questions/32569732/correct-way-of-unit-testing-repr-with-dict

kannon92 (approved):

client/python/tests/unit/test_event.py
```diff
@@ -0,0 +1,60 @@
...
+        return self.fake_name
+
+
+class FakeEventStreamMessage:
+    def __init__(self, name):
+        self.message = FakeEventMessage(name)
+        self.id = 1
+
+
+@pytest.mark.parametrize(
```
jayofdoom, line 28: image

jayofdoom (approved):

PR #1214, opened 2022-07-13
add test to e2e testsuite for gpu node

dejanzele:


testsuite/testcases/submit_gpu_1x1.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - priority: 0
+    namespace: personal-anonymous
+    podSpec:
+      containers:
+        - name: cuda-container
+          image: nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda10.2
+          resources:
+            limits:
+              nvidia.com/gpu: 1
+      nodeSelector:
```
JamesMurkin, line 14: Are these selectors + tolerations generic? I'm not sure how we exactly intend to use these testcases

Same question in general, if users don't have gpus won't this always fail? (Same for any tests we use this on)

testsuite/testcases/submit_gpu_1x1.yaml
```diff
@@ -0,0 +1,24 @@
...
+batchSize: 1
+queue: e2e-test-queue
+jobs:
+  - priority: 0
+    namespace: personal-anonymous
+    podSpec:
+      containers:
+        - name: cuda-container
+          image: nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda10.2
+          resources:
```
JamesMurkin, line 11: I think general good practice would be to specify all resources (cpu, memory) as we do in all other tests

JamesMurkin (approved):

PR #1216, opened 2022-07-13
Send JobLeaseReturned event server side

JamesMurkin:
On max lease retries we current get events like:

JobPendingEvent
JobUnableToScheduleEvent
JobFailedEvent
JobLeaseReturnedEvent

This is bad because:
 - The lease was really returned before the job was failed
 - It is hard to process by downstream systems (such as Lookout) which can't work out what state the job is currently in

I have now moved the JobLeaseReturned event to server side so events now look like:

JobPendingEvent
JobUnableToScheduleEvent
JobLeaseReturnedEvent
JobFailedEvent

Which now reflects the real order these events occurred and also fixes Lookout ingestion. It also means the executor no longer needs to worry about this issue


JamesMurkin:
A point for discussion
 - We could equally just sent the event first on the executor side
 
 The reasons I didn't do this:
  - Prevent issues with clock drift between server and executor causing confusion to lookout event ingestion
  - Makes the executor simpler

However I'm not strongly opinionated here and happy to swap it.

carlocamurri (approved):

PR #1219, opened 2022-07-13
C# / .NET Armada client updates for .NET 6.0

richscott:
This is a basic, rough first attempt at updating the .NET client code to be buildable on .NET 6, update the NSWag version and invocation to the latest version, to generate updated C# method bindings, and fix any miscellaneous breakage in the existing unit tests for the C# client.


client/DotNet/Armada.Client.Test/Tests.cs
```diff
@@ -1,64 +1,75 @@
...
 using RichardSzalay.MockHttp;
 
 namespace GResearch.Armada.Client.Test
 {
     public class Tests
     {
+        public static DateTimeOffset now = (DateTimeOffset)DateTime.UtcNow;
+        public static string testQueue = "test-" + now.ToUnixTimeSeconds().ToString();
+
+        public static string testServer = "https://armada.dev.armadaproject.io/api";
```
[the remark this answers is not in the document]
richscott: I'll make `testServer` default to an empty string, and fetch it from the environment.

kannon92:
Should we have some targets in the makefile that build this?  And should we add this to the CI?

client/DotNet6/Armada.Client.Test/Tests.cs
```diff
@@ -0,0 +1,156 @@
...
+using NUnit.Framework;
+using RichardSzalay.MockHttp;
+
+namespace GResearch.Armada.Client.Test
+{
+    public class Tests
+    {
+        public static DateTimeOffset now = (DateTimeOffset)DateTime.UtcNow;
+
+        // ARMADA_SERVER should be defined in this process's environment, of the
```
kannon92, line 14: I'm trying to test this but I'm unsure what to put here for my local dev armada server.  

Maybe we should have a little README.md that can explain how to set up tests and build?

kannon92: I've tried http://localhost:8080/api in my ARMADA_SERVER env variable but that doesn't seem to work on the local ci..

jayofdoom: You have to point it at the installed version of armada for testing; the URL isn't here :)

kannon92: Yea but I can create a dev armada server and test against that?  But I don't know what the URL is.  

I don't think we should point this to an isntalled version of Armada for a CI test?  It should be whatever test-e2e-setup creates?

PR #1225, opened 2022-07-14
add gzip support to server and testsuite client

dejanzele:


pkg/client/connection.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"google.golang.org/grpc/encoding/gzip"
+
```
severinson, line 10: Spacing mistake?

severinson (approved):

PR #1237, opened 2022-07-19
Add support for benchmark and statistics

dejanzele:


pkg/api/jobid.go
```diff
@@ -43,3 +45,47 @@ func JobIdFromApiEvent(msg *EventMessage) string {
 	}
 	return ""
 }
+
+func CreatedFromApiEvent(msg *EventMessage) time.Time {
```
severinson, line 49: Don't need this anymore, right?

severinson (approved):

PR #1239, opened 2022-07-19
Auto Generate server event types

### LINKED ISSUE #1238: Auto-generate python client event types
Automatically create a `typings.py` file that will correctly identify all the possible type options for events and create an Enum to represent them.

Sharpz7:
Closes #1238


kannon92 (changes requested):
Two things:

1) I think we should move this to a tox target in our python_client.  I don't think this should be in scripts.  Let's contain this code to armada_client.
2) I'd like to have some tests around this code.

Sharpz7:
> Two things:
> 
> 1. I think we should move this to a tox target in our python_client.  I don't think this should be in scripts.  Let's contain this code to armada_client.
> 2. I'd like to have some tests around this code.

I feel that having generated code, that will have to be imported into other modules, does maybe belong in the script? That is where we generate the python stubs, really this is just another stub. Or would we rather the process was run `make python` and run a tox command as well?

I will make some tests tonight, that makes sense to me :))

kannon92:
> I feel that having generated code, that will have to be imported into other modules, does maybe belong in the script? That is where we generate the python stubs, really this is just another stub. Or would we rather the process was run `make python` and run a tox command as well?

I wasn't very clear.  I think its okay to generate these typings in the script but I'd like to move the python code to client/python.

Sharpz7:
To document the message event type, we should use a Union typing and then autogenerate all the types. This new "type" can be stored as a variable and then used in `event.py`

client/python/tests/integration/test_typing_gen.py
```diff
@@ -0,0 +1,27 @@
+import os
```
kannon92, line 1: I don't really think you need an integration test here.  I think the unit test sufficies.

kannon92: I see why you did this.  You want the writing of a file done as an integration test.  I like your approach but I don't think we should include this as an integration test.  

I would propose a refactor where we generate the typing file in memory and then you test that.  The prod code can write this buffer to disk and that shouldn't be tested.  

This is a pattern I like to follow with interacting with external resources.  

This is overkill with the code that you have but it is a good practice to learn.  For example, you do not want to create a database in order to unit test code.  So you can build a simple mock and tests your logic and you trust the DB library to be tested enough.

kannon92 (changes requested):

client/python/tests/integration/test_typing_gen.py
Sharpz7: Sorry this is not 100% clear, do you want me to remove this file, or refactor it so that the file is only stored in memory and that is asserted?

kannon92: Both.  

1) Remove the integration test file.  
2) Add a unit test that generates the file in memory and asserts it matches what you expect.
3) Production code should generate the file and then you write that out to dist.

Sharpz7: Cool. On it :))

scripts/build-python-client.sh
```diff
@@ -30,4 +30,8 @@ sed -i 's/from github.com/from armada_client.github.com/g' client/python/armada_
 sed -i 's/from google.api/from armada_client.google.api/g' client/python/armada_client/armada/*.py
 sed -i 's/from google.api/from armada_client.google.api/g' client/python/armada_client/google/api/*.py
 
-find client/python/armada_client/ -name '*.py' | xargs sed -i 's/from k8s.io/from armada_client.k8s.io/g'
\ No newline at end of file
+find client/python/armada_client/ -name '*.py' | xargs sed -i 's/from k8s.io/from armada_client.k8s.io/g'
+
+# Generate better docs for the client
+export PYTHONPATH=${PWD}/client/python
+python3.8 ${PWD}/client/python/armada_client/gen/event_typings.py
```
kannon92, line 37:
```suggestion
python3 ${PWD}/client/python/armada_client/gen/event_typings.py
```

Minor suggestion.  If we update the docker file, it would be nice to not have to update this file.  I think python3 should work instead.

client/python/tests/unit/test_gen.py
```diff
@@ -0,0 +1,149 @@
...
+
+def test_union_var():
+
+    assert get_all_job_event_classes() == expected_events
+
+
+def test_file_gen():
+
+    import_text, states_text, union_text = gen_file(
+        expected_event_states, expected_events
```
kannon92, line 144:
```suggestion
        get_event_states(), get_all_job_event_classes()
```

I wouldn't use the expected values as parameters.

Sharpz7: But then its not a unit test, because it relies on other functions?

kannon92: I say it is.  You are testing the python code in isolation.  That is a unit test in my eyes.  

I view an integration test is one which you need other components to test against.  An integration test usually includes spawning a database or creating an armada test.  Testing against kubernetes etc.  

If I can test the code in isolation from all other pieces, that is what I would define as a unit test.

Sharpz7: Alrighty I shall implement it as you have suggested :))

kannon92 (approved):

jayofdoom (approved):

PR #1243, opened 2022-07-19
Fixed merge issues

Sharpz7:
Closes https://github.com/G-Research/armada/issues/1242

### DIFF
--- a/client/python/pyproject.toml
+++ b/client/python/pyproject.toml
@@ -3,7 +3,7 @@ name = "armada_client"
 version = "0.1.0-dev"
 description = "Armada gRPC API python client"
 readme = "README.md"
-requires-python = ">=3.7,<3.9"
+requires-python = ">=3.7,<3.9.13"
 dependencies = ["grpcio~=1.46.3", "grpcio-tools~=1.46.3", "mypy-protobuf~=3.2.0"]
 license = { text = "Apache Software License" }
 authors = [{ name = "G-Research Open Source Software", email = "armada@armadaproject.io" }]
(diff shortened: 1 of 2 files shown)


kannon92 (changes requested):
Looks good to me minus the requires-python change.

client/python/pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "armada_client"
 version = "0.1.0-dev"
 description = "Armada gRPC API python client"
 readme = "README.md"
-requires-python = ">=3.7,<3.9"
+requires-python = ">=3.7,<3.9.13"
```
jayofdoom, line 6: Is there a reason this is using a minor version? If the desired behavior is py 3.9 support, IMO we should just make it `<3.10` so that we pick up py 3.9 minor releases for free.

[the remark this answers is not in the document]
Sharpz7: Ah. Looks like I shall need to install another version of python

Sharpz7: Resolved by kevins comment

richscott (approved):

client/python/tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 isolated_build = true
 envlist =
-    format
+    format-code
```
kannon92, line 4: I think its bad practive to auto format code in a commit.  

We have format because we want the committer to run the formatter and commit his changes.  

It gets nasty if you have git auto format because then it should also commit its changes as a separate commit.  So let's avoid that.

Sharpz7: Oops this was not meant to be commited. Thanks for spotting :)

jayofdoom (approved):

Issue #1227, opened 2022-07-14
Integrate python dependencies into dependabot


kannon92:
We should use dependabot to keep our python packages up to date.

AgataGans:
working on it

jayofdoom:
Closed via #1251

Thank you for your contribution!

PR #1251, opened 2022-07-21
Enabled Dependabot for Python packages (#1227)

AgataGans:
* Added Dependabot integration for pip

#1227

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,3 +25,17 @@ updates:
     # Check for updates once a week
     schedule:
       interval: "daily"
+
+  # Enable version updates for pip -- client/python dir
+  - package-ecosystem: "pip"
+    directory: "./client/python"
+    # Check the pip registry for updates every day (weekdays)
+    schedule:
+      interval: "daily"
+
+  # Enable version updates for pip -- third_party/airflow dir
+  - package-ecosystem: "pip"
+    directory: "./third_party/airflow"
+    # Check the pip registry for updates every day (weekdays)
+    schedule:
+      interval: "daily"


.github/dependabot.yml
```diff
@@ -25,3 +25,10 @@ updates:
     # Check for updates once a week
     schedule:
       interval: "daily"
+
+  # Enable version updates for pip
+  - package-ecosystem: "pip"
+    directory: "/"
```
jayofdoom, line 31: We have two python modules in this repo:
- one under `client/python`
- one under `third_party/airflow`

IMBW, but I believe this will have to be pointed at the module dirs directly (where the pyproject.toml is) in order for this to be effective.

jayofdoom (commented):
Thank you for contributing, there's a comment that'll need resolution.

Where'd you find us? Is there any general interest in Armada beyond just this contribution?

.github/dependabot.yml
AgataGans: I made the required changes. Unfortunately, it's still impossible to add multiple dirs into the same package ecosystem.
Here's what I found on this: 
https://github.com/dependabot/dependabot-core/issues/2824
https://github.com/iZettle/example-integration/pull/7

However, I think that my version was correct - it was should have been checking all python packages in the whole root dir. At least accrording to what I've read so far. IMBW, too, of course. It's something for me to learn more about, for sure!

---
I was looking for some open source projects I could contribute to for the first time and to practice using GitHub, and this way I found Armada.
I'm actually trying to learn more about Armada itself rn. : )

jayofdoom (approved):

.github/dependabot.yml
jayofdoom: I think the revision is better; I prefer being explicit about where to check. 

Thanks for coming by! If you have any questions, feel free to pop them into github discussions. Otherwise, if you continue to be interested in contributing, please just pick up an untouched issue and comment on it. Those of us working full time on Armada work right out of github issues, so that should be up to date.

PR #1257, opened 2022-07-22
Testing libraries.md fix

Sharpz7:
Closes https://github.com/G-Research/armada/issues/1147


docs/libraries.md
```diff
@@ -1,24 +1,26 @@
...
+Here, we give an overview of the Armada Python client.
 
 ### Installation
 Prerequisites for installation include a working Python installation and poetry. You may wish to install the client into a virutalenv.
 
 Our python client bindings can be installed directly from the Armada repository.
  1. Checkout the [armada git repository](https://github.com/g-research/armada)
  2. From the root of the repository, run `make python` to generated needed gRPC code.
- 3. Install the client with poetry: `cd client/python && poetry install -v` 
+ 3. Install the client with poetry: `cd client/python && poetry install -v`
```
kannon92, line 15: I don't think you actually changed this line but this calls out that we need to update Libraries.md with the latest building changes.  

@jayofdoom do you think we should just link to README.md or contributor.md in our docs so we don't need to maintain both?

kannon92 (changes requested):
Your changes for adding return types looks good to me.  

there are some small issues that I think you probably are fixing.  

1) It looks like your doc generation is broken and you committed a bad doc.  
2) I'd like to fix the Libraries installation directions in here or at least file a ticket to clean that up in a followup.  I'm open to either option.

docs/libraries.md
Sharpz7: I would agree that just linking to README.md and removing this stuff would be wise. I will hold off for now though.

jayofdoom: Yeah. Literally consider all the existing stuff in the libraries.md boilerplate that can/should be replaced with internal links and/or improved information.

Sharpz7: Cool. I shall make those changes :))

docs/libraries.md
```diff
@@ -1,24 +1,26 @@
...
 
 ### Usage
 Once installed, the client can be imported with `import armada_client.client`.
 
 ### API Documentation
-For full documentation of our python module API, please see our [autogenerated python API docs](/python_armada_client)
+For full documentation of our python module API, please see our [autogenerated python API docs](./python_armada_client.md)
```
jayofdoom, line 13: This link is going to be broken once generated. If you access `https://armadaproject.io/api.md` -- or really any md file directly -- it gets downloaded.

This needs to be a fully qualified link, including armadaproject.io.  Please ensure you've reviewed https://github.com/G-Research/armada/blob/master/docs/docs-readme.md and https://github.com/G-Research/armada/blob/gh-pages/README.txt

docs/python_armada_client.md
```diff
@@ -11,6 +11,9 @@ title: armada_client package
 
 Armada Python GRPC Client
 
+For the api definitions:
+[https://github.com/G-Research/armada/blob/master/docs/api.md](https://github.com/G-Research/armada/blob/master/docs/api.md)
```
jayofdoom, line 15: this should instead link to armadaproject.io/api -- we have tended to prioritize making the website links better than the github-based links

jayofdoom (changes requested):

docs/libraries.md
Sharpz7: My apoligies, I should have checked these files. Thanks for passing that info on :))

kannon92 (approved):

jayofdoom (approved):

Issue #1142, opened 2022-06-29
Update minimum supported Go version from 1.16


jayofdoom:
This is a tracking issue to track various issues we're having regarding continued support of Go 1.16. It may not be the time to make the change now, but we should ensure we're documenting the tradeoffs we're making.

- We are locked into our current versions of `k8s.io/component-helpers`, `k8s.io/apimachinery` and `k8s.io/client-go`, newer versions do not support Go 1.16.
  - Related PRs: #1097 #1096 #1095 #1135
- Go 1.16 is losing security support when Go 1.19 is released. This is scheduled for August 2022 (https://tip.golang.org/doc/go1.19)

jayofdoom:
We've preliminarily agreed to move to Go 1.18.3.

jayofdoom:
As part of this, a nice to have would be some documentation around:
- Where and why we pin versions for CI (we will pin to a patch version)
- How we perform go version upgrades (for minor or patch versions)

jayofdoom:
So, I'm going to write a small doc on changing golang versions, but the reality is this is mostly trivial to push a button on when G-R is ready.

Originally, I was thinking we'd publish separate artifacts -- one for old (1.16) go  and one for newer (1.18) go -- but that doesn't make sense for a compiled application like it does for python.

The output of this issue then, before I table it for later is:
- a PR upgrading to Go 1.18 when we're ready, that includes a small doc (or really, probably just comments in the makefile) about how/where to change the go version we build against
- an email to stakeholders running this in production, letting them know that PR exists, and that we're ready to merge it at will

This work should be complete by EOD tomorrow as long as there's no controversy about how to do this.

kannon92:
https://github.com/G-Research/armada/pull/1258 closes this issue.

PR #1258, opened 2022-07-22
Use go 1.18 to build artifacts

jayofdoom:
This should not merge until we are ready for all armada consumers to use go 1.18.

Issue #1142


kannon92 (commented):
Do you need to change the version in go.sum or go.mod?

jayofdoom:
Nice catch; yes I do. Probably should run `go mod tidy`, too. 

Gonna be honest; I didn't expect this change to pass tests so I didn't jump through all the needed hoops yet :laughing:

jayofdoom:
Once https://github.com/G-Research/armada/pull/1276 merges, this rebased on top of it should pass tests.

.circleci/config.yml
```diff
@@ -106,10 +106,10 @@ commands:
           git push -q origin HEAD
 
 jobs:
-  checks_and_unit_tests-ubuntu2004-go116: &checksAndUnitTests
+  checks_and_unit_tests-ubuntu2204-go118:
     machine:
       docker_layer_caching: true
-      image: ubuntu-2004:202104-01
+      image: ubuntu-2204:current
```
kannon92, line 112: This seems like a large change?  Should we pin ubuntu to a version?

kannon92 (dismissed):

.circleci/config.yml
jayofdoom: This is the way 1.18 tests are running already -- it'd be pretty stunning to see ubuntu break the most basic of functionality we use in a security update release, and I'd rather get the security updates. I understand why some folks may not wanna do that though.

jayofdoom:
This should not merge without explicit approval and timing from GR.

.circleci/config.yml
kannon92: Yea for this it is probably fine.

d80tb7:
LGTM- let's coordinate with @JamesMurkin to switch the ci over to requiring 1.18 to pass

jayofdoom:
The work has already been done on the public side; if there's anything I can do to help with other CI systems please just ask.

PR #1259, opened 2022-07-22
Poc in memory

kannon92:
Closes https://github.com/G-Research/armada/issues/1138

POC for an in memory cache for the Job Service.

Still working on all the logic we want for this service but operating on a local map.


internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,95 @@
...
+		jobServiceRepository: jobServiceRepository}
+}
+
+func (eventToJobService *EventsToJobService) SubscribeToJobSetId(context context.Context) error {
+
+	return eventToJobService.StreamCommon(&eventToJobService.jobServiceConfig.ApiConnection, context)
+}
+func (eventToJobService *EventsToJobService) StreamCommon(clientConnect *client.ApiConnectionDetails, ctx context.Context) error {
+	var fromMessageId string
+	// This corresponds to the amount of time since the job-set has been updated.
```
d80tb7, line 43: I'm not sure this is what we want.  Rather than unsubscribing when no events have come in for a job set, we want to unsubscribe when nobody has *requested a status update* on that jobset after x hours

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,95 @@
...
+	// If no new events come from the job-set, then the timer allows automatic unsubscribing.
+	duration := time.Duration(eventToJobService.jobServiceConfig.SubscribeJobSetTime) * time.Second
+	timeOut := time.NewTimer(duration)
+	err := client.WithEventClient(clientConnect, func(c api.EventClient) error {
```
d80tb7, line 47: I think we need some error handling here to reconnect to the armada event server if necessary.

internal/jobservice/repository/job_service_in_memory.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func (inMem *InMemoryJobServiceRepository) IsJobSetAlreadySubscribed(jobSetId string) bool {
+	inMem.jobStatus.subscribeLock.Lock()
+	defer inMem.jobStatus.subscribeLock.Unlock()
+	_, ok := inMem.jobStatus.subscribeMap[jobSetId]
+	if ok {
+		return true
+	}
+	inMem.jobStatus.subscribeMap[jobSetId] = &jobSetId
```
d80tb7, line 67: I think this could be problematic because:
- we're marking that we are subscribed before we actually subscribe.  E.g. calling this method will cause us to become subscribed even though no subscription has acually ben made
- I wouldn't expect a function called `IsJobSetAlreadySubscribed` to mutate state.

internal/jobservice/repository/job_service_in_memory.go
```diff
@@ -0,0 +1,113 @@
...
+	return false
+}
+
+func (inMem *InMemoryJobServiceRepository) UnSubscribeJobSet(jobSetId string) error {
+	inMem.jobStatus.jobLock.Lock()
+	defer inMem.jobStatus.jobLock.Unlock()
+	_, ok := inMem.jobStatus.subscribeMap[jobSetId]
+	if !ok {
+		return fmt.Errorf("JobSetId %s already unsubscribed", jobSetId)
```
d80tb7, line 76: Is this an error?   I'd argue it should just be a no-op

internal/jobservice/repository/job_service_in_memory.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	delete(inMem.jobStatus.subscribeMap, jobSetId)
+	log.Infof("JobSetId %s unsubscribed", jobSetId)
+	return nil
+}
+
+// This is a very slow function until we get a database.
```
d80tb7, line 83: If you wanted to make this faster I think you just need to keep a map of `jobsetId -> Seq[JobId]` although I understand if you don't want to optimise this now.

internal/jobservice/server/server.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func NewJobService(config *configuration.JobServiceConfiguration, inMemoryService repository.JobServiceRepository) *JobServiceServer {
+	return &JobServiceServer{jobServiceConfig: config, jobRepository: inMemoryService}
+}
+
+func (s *JobServiceServer) GetJobStatus(ctx context.Context, opts *js.JobServiceRequest) (*js.JobServiceResponse, error) {
+	response, err := s.jobRepository.GetJobStatus(opts.JobId)
+	if err != nil {
+		log.Fatal(err)
```
d80tb7, line 27: not sure this should be fatal!

internal/jobservice/server/server.go
```diff
@@ -0,0 +1,37 @@
...
+	jobServiceConfig *configuration.JobServiceConfiguration
+	jobRepository    repository.JobServiceRepository
+}
+
+func NewJobService(config *configuration.JobServiceConfiguration, inMemoryService repository.JobServiceRepository) *JobServiceServer {
+	return &JobServiceServer{jobServiceConfig: config, jobRepository: inMemoryService}
+}
+
+func (s *JobServiceServer) GetJobStatus(ctx context.Context, opts *js.JobServiceRequest) (*js.JobServiceResponse, error) {
+	response, err := s.jobRepository.GetJobStatus(opts.JobId)
```
d80tb7, line 25: I'd  get the status after the subscription

internal/jobservice/eventstojobs/event_job_response.go
```diff
@@ -0,0 +1,56 @@
...
+		return &js.JobServiceResponse{State: js.JobServiceResponse_RUNNING}, nil
+	case *api.EventMessage_Failed:
+		return &js.JobServiceResponse{State: js.JobServiceResponse_FAILED, Error: message.GetFailed().Reason}, nil
+	case *api.EventMessage_Succeeded:
+		return &js.JobServiceResponse{State: js.JobServiceResponse_SUCCEEDED}, nil
+	case *api.EventMessage_Cancelled:
+		return &js.JobServiceResponse{State: js.JobServiceResponse_CANCELLED}, nil
+	}
+
+	return nil, fmt.Errorf("unknown event type")
```
d80tb7, line 26: why are we returing errros here? Isn't it expected that there will be events which we want to ignore?  I'd suggest just silently ignoring any events we don;'t care about.

d80tb7: To be more precise- I'd get rid of `IsEventAJobResponse` and just call this method regardless which will return nil if it's not an event we care about.

kannon92: Good idea!  Done.

internal/jobservice/eventstojobs/eventstojobs.go
kannon92: Any code I could use as an example in Armada?

internal/jobservice/repository/job_service_in_memory.go
kannon92: Yea I'll hold on this because I think this will be performant once I add a databse.

kannon92:
Followed up with chris:

1) We will unsubscribe if the client hasn't requested a status after x amount of time.
2) Implement logic to handle cancelling subsubscription once job-set is unsubscribed.

internal/jobservice/application.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil
+	})
+	g.Go(func() error {
+		ticker := time.NewTicker(time.Duration(config.SubscribeJobSetTime) * time.Second)
+		for range ticker.C {
+			for _, value := range inMemoryJobService.GetSubscribedJobSets() {
+				log.Infof("Subscribed job sets : %s", value)
+				if inMemoryJobService.CheckToUnSubscribe(value, config.SubscribeJobSetTime) {
+					inMemoryJobService.DeleteJobsInJobSet(value)
+					inMemoryJobService.UnSubscribeJobSet(value)
```
jayofdoom, line 70: This should be done in reverse, right? To avoid a race condition... seems like it'd be possible for a jobset to update after deletion but before unsubscribe.

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,104 @@
...
+		jobServiceConfig:     jobServiceConfig,
+		jobServiceRepository: jobServiceRepository,
+	}
+}
+
+func (eventToJobService *EventsToJobService) SubscribeToJobSetId(context context.Context) error {
+
+	return eventToJobService.StreamCommon(&eventToJobService.jobServiceConfig.ApiConnection, context)
+}
+func (eventToJobService *EventsToJobService) StreamCommon(clientConnect *client.ApiConnectionDetails, ctx context.Context) error {
```
jayofdoom, line 43: nit: please add whitespace between these functions

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,104 @@
...
+	var fromMessageId string
+	conn, connErr := client.CreateApiConnection(clientConnect)
+	eventToJobService.jobServiceRepository.SubscribeJobSet(eventToJobService.jobsetid)
+	if connErr != nil {
+		log.Warnf("Connection Issues with EventClient %v", connErr)
+	}
+	defer conn.Close()
```
jayofdoom, line 50: This is potentially bugged in the broken case; you will not always have a conn to Close in the case where connErr is set.

internal/jobservice/repository/job_service_in_memory.go
```diff
@@ -0,0 +1,156 @@
...
+	jobStatus        *JobStatus
+	jobServiceConfig *configuration.JobServiceConfiguration
+}
+
+func NewInMemoryJobServiceRepository(jobMap *JobStatus, config *configuration.JobServiceConfiguration) *InMemoryJobServiceRepository {
+	return &InMemoryJobServiceRepository{jobStatus: jobMap, jobServiceConfig: config}
+}
+
+func (inMem *InMemoryJobServiceRepository) GetJobStatus(jobId string) (*js.JobServiceResponse, error) {
+	inMem.jobStatus.jobLock.RLock()
```
jayofdoom, line 35: You could consider using sync/atomic.Value to avoid having to do the locking yourself. I'm not 100% sure it's a perfect fit though.  

https://pkg.go.dev/sync/atomic#Value

internal/jobservice/repository/job_table.go
```diff
@@ -0,0 +1,19 @@
+package repository
```
jayofdoom, line 1: I would suggest putting only the interface in the repository package, and locate the implementations outside.

internal/jobservice/application.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// Setup an errgroup that cancels on any job failing or there being no active jobs.
+	g, _ := errgroup.WithContext(ctx)
+
+	grpcServer := grpcCommon.CreateGrpcServer(config.Grpc.KeepaliveParams, config.Grpc.KeepaliveEnforcementPolicy, []authorization.AuthService{&authorization.AnonymousAuthService{}})
+
+	inMemoryMap := make(map[string]*repository.JobTable)
+	subscribedJobSets := make(map[string]*repository.SubscribeTable)
+	jobStatusMap := repository.NewJobStatus(inMemoryMap, subscribedJobSets)
+	inMemoryJobService := repository.NewInMemoryJobServiceRepository(jobStatusMap, config)
```
jayofdoom, line 44: If the idea is to create and use a JobServiceRepository interface, we need to ensure these are typed to the interface, and not a specific implementation. This issue sorta exists throughout, but this seemed the most sensible place to comment it.

internal/jobservice/repository/job_service_in_memory.go
```diff
@@ -0,0 +1,156 @@
...
+	log.Infof("Deleting jobs in job-set %s", jobSetId)
+	for key, val := range inMem.jobStatus.jobMap {
+		if val.jobSetId == jobSetId {
+			delete(inMem.jobStatus.jobMap, key)
+		}
+	}
+	return nil
+}
+
+func (inMem *InMemoryJobServiceRepository) PrintAllItems() {
```
jayofdoom, line 135: nit: I believe this should be a private method, since it's not in the interface

third_party/airflow/armada/operators/utils.py
```diff
@@ -94,6 +94,8 @@ def search_for_job_complete(
             job_status_return = job_status_callable(
                 queue=queue, job_id=job_id, job_set_id=job_set_id
             )
+
+        time.sleep(3)
```
jayofdoom, line 98: :question: 

If this is needed, it should be commented and an issue created to remove the need for a sleep in the future

kannon92: Is it big deal to add a sleep so we aren’t polling the api every nanosecond?  

It’s prob leftover debugging but it spams the grpc logging bcuz every rpm call gets logged

internal/jobservice/repository/job_service_in_memory.go
kannon92: So lowercase is private?  I know Capital function is exported

third_party/airflow/armada/operators/utils.py
jayofdoom: Oh, that's fair. I generally comment any `sleep()` calls in code though because they tend to catch the eye

internal/jobservice/repository/job_service_in_memory.go
kannon92: I think I need this bcuz maps by default aren’t threadsafe.  I think go throws a panic if they are accessed concurrently.

internal/jobservice/repository/job_service_in_memory.go
jayofdoom: Yeah, that's my understanding :+1:

cmd/jobservice/cmd/run.go
```diff
@@ -0,0 +1,50 @@
...
+func runCmd(app *jobservice.App) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "test",
+		Short: "Test an Armada deployment by submitting jobs and watching for expected events.",
+		RunE:  runCmdE(app),
+	}
+
+	return cmd
+}
+func runCmdE(app *jobservice.App) func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
kannon92: Not sure.  I copied this code from TestSuite.  I think the naming corresponds for what function this gets called from Cobra.

internal/jobservice/repository/job_service_in_memory.go
jayofdoom: I am pretty sure this is exactly the case that the `sync/atomic` package is designed for. I've only used it for inc/dec integers across threads though, not higher order objects.

Either way, I wanted to make sure you were aware this existed in the stdlib; there may be good reasons to not use it (including but not limited to 'this code is already written and I didn't :D).

cmd/jobservice/cmd/run.go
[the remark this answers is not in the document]
kannon92: RunCmdE is a cobra function function that says it will run but it takes an error.  So kinda redundant

internal/jobservice/repository/job_service_in_memory.go
jayofdoom: I think you're right, this is a bad fit on further analysis. `sync/atomic` only validates that a write can happen before the read; but it does nothing to ensure multiple writers are not updating simultaneously. It seems like if anything, you'd want to use it *in addition to* more locking.

kannon92: Yea there is a sync map that I could potentially use but it seemed more complicated than I needed.

internal/jobservice/application.go
kannon92: Actually UnSubscribeJobSet does deletion so I removed.

internal/jobservice/repository/job_service_in_memory.go
kannon92: Done

internal/jobservice/repository/job_table.go
kannon92: See comment below.  I tried to do this but I think that the exporting gets tricky and I'm unsure if we really need a repostiryo package for different implementations.

internal/jobservice/application.go
d80tb7: Personally I don't think we need an interface here.  This component is pretty small and focused so I think it;'s better to be explicit rather than trying to abstract things away

internal/jobservice/eventstojobs/eventstojobs.go
d80tb7: you also close the connection on line 62- I'm not sure if it's legal to close an already closed connection, but it seems a bit wrong to close it twice either way.

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -0,0 +1,107 @@
...
+func (eventToJobService *EventsToJobService) SubscribeToJobSetId(context context.Context) error {
+
+	return eventToJobService.streamCommon(&eventToJobService.jobServiceConfig.ApiConnection, context)
+}
+
+func (eventToJobService *EventsToJobService) streamCommon(clientConnect *client.ApiConnectionDetails, ctx context.Context) error {
+	var fromMessageId string
+	conn, connErr := client.CreateApiConnection(clientConnect)
+	eventToJobService.jobServiceRepository.SubscribeJobSet(eventToJobService.jobsetid)
+	if connErr != nil {
```
d80tb7, line 50: if `connErr` is not nil and you couldn't create a connection is there anything useful you can do here?  I'm pretty sure all the goroutines below will fail in that case so you'd be much better just bailing here without creating them

internal/jobservice/repository/job_service_in_memory.go
```diff
@@ -0,0 +1,176 @@
...
+
+// Update in memory JobStatus Map with jobId and our JobTable
+func (inMem *InMemoryJobServiceRepository) UpdateJobServiceDb(jobId string, jobTable *JobTable) error {
+	log.Infof("Updating JobId %s with State %s", jobId, jobTable.jobResponse.State)
+	inMem.jobStatus.jobLock.Lock()
+	defer inMem.jobStatus.jobLock.Unlock()
+	jobResponse, ok := inMem.jobStatus.jobMap[jobId]
+	if ok && jobResponse.jobResponse.State == jobTable.jobResponse.State {
+	} else {
+
```
d80tb7, line 59: (nit)- you can remove this blank line

internal/jobservice/repository/job_service_in_memory.go
```diff
@@ -0,0 +1,176 @@
...
+	return ok
+}
+
+// Mark our JobSet as being subscribed
+// SubscribeTable contains JobSet and time when it was created.
+func (inMem *InMemoryJobServiceRepository) SubscribeJobSet(jobSetId string) {
+	inMem.jobStatus.subscribeLock.Lock()
+	defer inMem.jobStatus.subscribeLock.Unlock()
+	_, ok := inMem.jobStatus.subscribeMap[jobSetId]
+	if ok {
```
d80tb7, line 85: i think you can simplify this to just:
```
if !ok {
    inMem.jobStatus.subscribeMap[jobSetId] = NewSubscribeTable(jobSetId)
}
```

internal/jobservice/repository/job_service_in_memory.go
```diff
@@ -0,0 +1,176 @@
...
+func (inMem *InMemoryJobServiceRepository) UnSubscribeJobSet(jobSetId string) {
+	inMem.jobStatus.subscribeLock.RLock()
+	defer inMem.jobStatus.subscribeLock.RUnlock()
+	_, ok := inMem.jobStatus.subscribeMap[jobSetId]
+	if !ok {
+		log.Infof("JobSetId %s already unsubscribed", jobSetId)
+		return
+	}
+	delete(inMem.jobStatus.subscribeMap, jobSetId)
+	log.Infof("JobSetId %s unsubscribed", jobSetId)
```
d80tb7, line 103: personally I'd put the log line at the end of the function

internal/jobservice/repository/job_service_in_memory.go
```diff
@@ -0,0 +1,176 @@
...
+
+// Checks JobSet table to make determine if we should unsubscribe from JobSet
+// configTimeWithoutUpdates is a configurable value that is read from the config
+// We allow unsubscribing if the jobset hasn't been updated in configTime
+func (inMem *InMemoryJobServiceRepository) CheckToUnSubscribe(jobSetId string, configTime int64) bool {
+	if !inMem.IsJobSetSubscribed(jobSetId) {
+		return false
+	}
+	currentTime := time.Now().Unix()
+	for _, val := range inMem.jobStatus.subscribeMap {
```
d80tb7, line 115: do you need a lock here as you're iterating over the map which isn't thread safe?

internal/jobservice/repository/job_service_in_memory.go
```diff
@@ -0,0 +1,176 @@
...
+	inMem.jobStatus.jobLock.RLock()
+	defer inMem.jobStatus.jobLock.RUnlock()
+	for key, value := range inMem.jobStatus.jobMap {
+		log.Infof("JobKey: %s Queue %s JobId %s JobSet %s State %s TimeStamp %d", key, value.queue, value.jobId, value.jobSetId, value.jobResponse.State, value.timeStamp)
+	}
+}
+
+// Get a list of SubscribedJobSets
+func (inMem *InMemoryJobServiceRepository) GetSubscribedJobSets() []string {
+	var returnJobSets []string
```
d80tb7, line 165: does this need a lock?

internal/jobservice/repository/job_service_in_memory.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// Update JobSet Map with time that a Job in that JobSet was requested
+func (inMem *InMemoryJobServiceRepository) UpdateJobSetTime(jobSetId string) error {
+	inMem.jobStatus.subscribeLock.Lock()
+	defer inMem.jobStatus.subscribeLock.Unlock()
+
+	_, ok := inMem.jobStatus.subscribeMap[jobSetId]
+	if ok {
+		inMem.jobStatus.subscribeMap[jobSetId] = NewSubscribeTable(jobSetId)
```
d80tb7, line 132: This is a bit confusing- the function is called `UpdateJobSetTime` but it looks like we replace *all* the existing data associated with the jobset in the subscriptionMap.

d80tb7: Actually I'm really confused here- the `subscribeMap` is keyed by `jobSetId` but the object it contains has a `jobId`?

internal/jobservice/repository/subscribe_table.go
```diff
@@ -0,0 +1,12 @@
+package repository
+
+import "time"
+
+type SubscribeTable struct {
+	subscribedJobSet     string
+	lastRequestTimeStamp int64
+}
+
+func NewSubscribeTable(jobSetId string) *SubscribeTable {
```
d80tb7, line 10: so generally `jobSet` isn't the primary key, rather `(queue, jobset`) is the primary key.  I'm not sure if your concatting the two somewhere and calling the result `jobset`, but even if so I'd probably recommend keeping them separate.

Note that this comment applies to a bunch of places where only `JobSet` is used as the PK.

internal/jobservice/eventstojobs/eventstojobs.go
kannon92: Yea you are right.

internal/jobservice/repository/subscribe_table.go
kannon92: Good call out!  I was thinking job-set can be the PK but I will fix this.

internal/jobservice/application.go
kannon92: I went ahead and removed it.

internal/jobservice/repository/job_service_in_memory.go
kannon92: yep!

internal/jobservice/repository/job_service_in_memory.go
kannon92: Good call.

internal/jobservice/repository/subscribe_table.go
kannon92: Alright I fixed this now too.

internal/jobservice/eventstojobs/eventstojobs.go
kannon92: I returned an error here.  I think we probably shouldn't panic?

internal/jobservice/application.go
```diff
@@ -0,0 +1,89 @@
...
+				log.Infof("Subscribed job sets : %s", value)
+				if inMemoryJobService.CheckToUnSubscribe(value, config.SubscribeJobSetTime) {
+					inMemoryJobService.UnSubscribeJobSet(value)
+				}
+			}
+		}
+		return nil
+	})
+	g.Go(func() error {
+		defer log.Println("Stopping server.")
```
d80tb7, line 76: should this be `log.Info` rather than `log.Println`

internal/jobservice/application.go
```diff
@@ -0,0 +1,89 @@
...
+
+		log.Info("JobService service listening on ", config.GrpcPort)
+		if err := grpcServer.Serve(lis); err != nil {
+			log.Fatalf("failed to serve: %v", err)
```
d80tb7, line 80: the return under here is misleading as (I think)  `log.Fatal` causes a panic.

d80tb7 (approved):

internal/jobservice/application.go
kannon92: Will resolve in a followup!

internal/jobservice/application.go
kannon92: Can do!

PR #1269, opened 2022-07-25
Do not rely on system-provided golang for python jobs

jayofdoom:
This is a required step to ensure jobs will be able to pass in the
future against golang 1.18.

Issue #1142


kannon92 (approved):
Nice find

Issue #1030, opened 2022-05-31
Add native gRPC support to the dotnet client


dave-gantenbein:
dotnet supports generating grpc clients from proto files, update the client to use this approach.

Tagged as a stability issue b/c we have a large number of issues with the REST client that could be mitigated w/grpc.

ARMADA-1098

d80tb7:
@andrewpcone If you look at https://github.com/G-Research/armada/tree/f/chrisma/dotnetclient I've got a bit of code you might be able to use as a start for this.  The basic strategy is:

- Create a new Solution under `client/DotNet.gRPC/Armada.Client.Grpc` with the intention of releasing the grpc client as a completely separate library.  Once this client is mature we can decomission the old rest client.
- Modify `build/proto/Dockerfile` such that the Docker image it creates is capable of generating dotnet code.  The only real change here is to download `grpc_csharp_plugin` from nuget and make it available to protoc.
- Modify `scripts/proto.sh` to generate dotnet code under `client/DotNet.gRPC/Armada.Client.Grpc/generated`

Note that the new `Armada.Client.Grpc` solution basically has nothing in it when you check out of git.  This is intentional and the idea is that a developer must first run `make proto` to  add the generated code to the project.  

Things that are not done/well-resolved
- Whether the client should consist of anything other than the generated code.  My feeling here is that it shouldn't because the generated code is pretty much enough, but others may disagree.  @severinson or @JamesMurkin may have opinions here.
- Any sort of tests.  If we just go with the just generated code we probably don't need unit tests. It would be nice to have an e2e test, however.
- An example of how to use the client.  We could possible create an examples project inside the `Armada.Client.Grpc` solution for this.
- Documentation.
- If you look at the generated code, I generate classes for our gogoproto, k8s.io etc dependencies.  We need these classes to be available to users because when they submit an armada job they are going to have to be able to create for example a k8s.io podspec.  What this means, however is that our `Armada.Client.Grpc` library is going to be publishing classes in a  `k8s.io` namespace which seems like bad etiquette.  It might be worth finding some grpc/dotnet expert and seeing what the solutionf for this should be.

dave-gantenbein:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Armada-62c744dc3b819100132ca032/issues/G-Research/armada/1030?planning-poker) with ZenHub @dejanzele @jayofdoom @kannon92 @richscott

kannon92:
[...]

PR #1277, opened 2022-07-26
WIP of GRPC C# client (#1030)

richscott:
This is heavily based upon strategy ideas and suggestions as outlined by Chris Martin in
    https://github.com/G-Research/armada/issues/1030

A new, second Dockerfile for protoc processing on arm64 processor hosts has been introduced (these changes were developed on a Macbook M1).

The prototype .sln and .csproj are directly taken from the branch mentioned in the above GH issue. The proto.sh has been enhanced with just the .NET-related chunks from that branch version. All .cs files, etc, are unmodified from their generation by the 'proto' make target.


kannon92:
I think the consensus was to not commit GRPC generated code into the repo.  Can you remove the generated files from the repo?

richscott:
> I think the consensus was to not commit GRPC generated code into the repo. Can you remove the generated files from the repo?

Sure, I can, but - Chris Martin (@d80tb7) requested that I directly include the generated `.cs` bits, and that we put out a package build of this "as-is" for the other C# devs, then evolve it in subsequent PRs based on their feedback & contributions.

kannon92:
> > I think the consensus was to not commit GRPC generated code into the repo. Can you remove the generated files from the repo?
> 
> Sure, I can, but - Chris Martin (@d80tb7) requested that I directly include the generated `.cs` bits, and that we put out a package build of this "as-is" for the other C# devs, then evolve it in subsequent PRs based on their feedback & contributions.

I think there is confusion about this.  

In his issue, he mentioned this:

>
> Note that the new Armada.Client.Grpc solution basically has nothing in it when you check out of git. This is intentional and > the idea is that a developer must first run make proto to add the generated code to the project.

I read this as the Armada.Client.Grpc should not include the generated files because the developer can generate them. 

I would like all GRPC clients to have the following behavior:

a) We track in git the things that aren't autogenerated by protoc.  
 - This should include your client implementation of the Submit and Event classes.  
 - Unit Testing
 - Solution/VcProj files
b) All generated files are generated as part of a CI and then they are uploaded to the package repo.  So the package should/needs these packages to build. 

This is currently the exact pattern we have for the python client and the airflow operator.  Note, the golang code and the dotnet rest client don't follow this behavior but I'd like to address that in the future.

d80tb7:
Yes- I'm pretty anti checking in generated code here.  Reasons as follows:
- You end up with huge generated files that cause e.g. github issues when they pop up on code reviews
- You end up with two sources of truth (e.g. the generated code is out of sync with the proto file) which we then have to check in ci
- You end up with generated code relating to someone else's project! (in this case we'll have a load of generated code for all of the k8s api) which is just crazy.

If this was a normal dotnet project we'd just check int he protos and then use dotnet build to build it all on the fly. I think we should do the same here.

richscott:
I'm fine with not committing the generated .cs logic, but the repo appears to have committed swagger-generated .json files - shouldn't we do the same there?

d80tb7:
Personally I'd hold off removing the swagger generated code.  In an ideal world I think we'd remove it, but it's not super high priority and if we do the grpc client right it should become a thing of the past in the near future anyway.

build/proto/Dockerfile.arm64
```diff
@@ -0,0 +1,100 @@
+# Dockerfile for building .proto files.
```
kannon92, line 1: I think you could use setup-proto and that should allow you to clean this DOCKERFILE up a lot.

makefile
```diff
@@ -418,6 +425,9 @@ junit-report:
 	$(GO_TEST_CMD) bash -c "cat test_reports/*.txt | go-junit-report > test_reports/junit.xml"
 
 setup-proto: download
+	# Work around a "permission denied" error on macOS, when the following 'rm -rf' attempts to
```
kannon92, line 428: https://github.com/G-Research/armada/issues/1279 is the issue that is tracking these permission errors.  I'd like to fix this at the container level rather than sprinkling around these chmod everywhere.

jayofdoom: I mean, I agree this is kinda awful, but the bug itself is awful too. IMO, we shouldn't block this on #1279 given it represents literally a week of Rich's work.

client/DotNet.gRPC/Armada.Client.Grpc/Armada.Client.Grpc.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>netstandard2.0</TargetFramework>
```
jayofdoom, line 4: TIL about .netstandard target, this is cool

build/proto/Dockerfile.arm64
jayofdoom: Agree but am also OK with a ticketed follow-up to DRY it up, given folks are strongly desiring having this client out

scripts/build-dotnet-client.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+# This script is intended to be run under the docker container at $ARMADADIR/build/proto/
+
+export PATH=$PATH:/dotnet
+export DOTNET_CLI_TELEMETRY_OPTOUT=1
+
+export TYPES=Mgoogle/protobuf/any.proto=github.com/gogo/protobuf/types,\
```
jayofdoom, line 7: nit: maybe nice to have a comment or link to some docs about what this syntax means?

scripts/proto.sh
```diff
@@ -71,3 +71,4 @@ protoc \
 --grpc-gateway_out=logtostderr=true,$TYPES:. \
 --swagger_out=logtostderr=true,$TYPES,allow_merge=true,simple_operation_ids=true,json_names_for_fields=true,merge_file_name=./pkg/api/binoculars/api:. \
 pkg/api/binoculars/binoculars.proto \
+
```
jayofdoom, line 74: nit: appears to be an unintentional change, but it does no harm

jayofdoom (commented):
looks alright to me, but I haven't tested it personally and I don't know much C#

build/proto/Dockerfile
```diff
@@ -32,23 +35,61 @@ ENV GOPRIVATE=$GOPRIVATE
...
+# Install the csharp grpc plugin
+RUN curl -sSL https://www.nuget.org/api/v2/package/Grpc.Tools/2.46.3 --output grpctools.zip && \
+    unzip -q -d grpctools grpctools.zip && \
+    mv grpctools/tools/linux_x64/grpc_csharp_plugin /usr/local/bin/grpc_csharp_plugin && \
+    chmod +x /usr/local/bin/grpc_csharp_plugin
+
+# Remove after I copy proto files
 ARG GOGO_PROTOBUF_VERSION
-RUN GO111MODULE=on go get github.com/gogo/protobuf/protoc-gen-gogofaster@v${GOGO_PROTOBUF_VERSION}
+RUN GO111MODULE=on go get github.com/gogo/protobuf/protoc-gen-gogofaster@v${GOGO_PROTOBUF_VERSION} && \
```
kannon92, line 56: Let's talk on monday and figure out a way to avoid adding all this back in.  

https://github.com/G-Research/armada/pull/1083/files

I merged this PR to try and clean these docker files up so we don't need to manage the version of K8 files in docker files.  I really want a single place where we pull these docker files and not replicate it again and again for every proto generation.

makefile
richscott: The chmod hack has been removed in commit f263b353 - I have borrowed the solution that @jayofdoom proposed in #1279 and it works consistently well now.

scripts/build-dotnet-client.sh
richscott: I agree! I actually spent 90 minutes researching in many places, plus cloning the protobuf and gogo repos, looking at the source code, and cannot find any mention anywhere describing the semantics of this syntax. They do have a few example invocations that use this multiple-pathing-whatever syntax, but no precise description exists. Probably someone would have to fire up a debugger and single step through a protoc run to find out how it parses this.

scripts/proto.sh
richscott: I think this might have been auto-added my editor (Vim with several plugins) or IDE (VS Code) - I wonder if it didn't like the final line of the file ending with a `\` char?

jayofdoom: Dropping the `\` on line 73 would be a better change than this one, looking  more closely. But also :shrug:

makefile
kannon92: https://app.circleci.com/pipelines/github/G-Research/armada/5799/workflows/4011e134-dcc0-4222-aa8d-7816b845d671/jobs/13592/parallel-runs/0/steps/0-104

You got a pretty serious fail in the integration tests.

jayofdoom: aha! This is how it broke in my PR too, I remember now.

richscott:
```
mkdir -p .build/server
docker run --rm -u 1001:1002 -v /home/circleci/go/src/github.com/G-Research/armada:/go/src/armada -w /go/src/armada --network=host -e GOPROXY -e GOPRIVATE -e INTEGRATION_ENABLED=true -e CGO_ENABLED=0 -e GOOS=linux -e GARCH=amd64 -v /home/circleci/.go_workspace:/go golang:1.16-buster go build -ldflags="-s -w" -o ./.build/server/server cmd/armada/main.go
cp -a ./config/armada ./.build/server/config
docker build  -t armada -f ./build/armada/Dockerfile ./.build/server/
failed to initialize build cache at /.cache/go-build: mkdir /.cache: permission denied
make: *** [makefile:225: build-docker-server] Error 1
```
What/where is this /.cache thing created?

kannon92: This code is what is failing:  

https://github.com/G-Research/armada/blob/master/makefile#L210

I guess its something to do with the docker cache?

kannon92: Looks like this may work but unsure what it does to build performance:

https://github.com/openshift/release/issues/9748

build/proto/Dockerfile.arm64
d80tb7: +1 for tidying this up afterwards.  I'd like to get us publishing the generated classes asap

scripts/build-dotnet-client.sh
d80tb7: Do we need this? I *thought* (but admit I don't know for sure)  that this stuff was go specific- i.e. it tells the protoc compiler to generate idiomatic go structs for stuff like Timestamps and Int Objects rather than generating the naïve code which is a bit hard to work with.  Since we're not generating go code here- I'd hope we wouldn't need this?

d80tb7: Also I think this feature might be documented [here](https://developers.google.com/protocol-buffers/docs/reference/go-generated).  Specifically:
> The Go import path may be specified on the command line when invoking the compiler, by passing one or more M${PROTO_FILE}=${GO_IMPORT_PATH} flags. Example usage:

richscott: @d80tb7 thanks for the reference - I've added a link to that in the script

richscott: Also, of course I forgot - TYPES doesn't need to be in build-dotnet-client.sh, it's now excised.

scripts/proto.sh
richscott: The better change is in commit 58af24ea

makefile
richscott: Builds now appear to be working, both locally on my Macbook and Linux systems, and in GH Actions; please take a look, re-review, etc etc.

kannon92: Looks like there are still some issues on integration tests.

build/proto/Dockerfile
kannon92: We are going to file this as a separate issue to cleanup later.

makefile
kannon92: We fixed this with a separate PR.

kannon92 (approved):

PR #1278, opened 2022-07-27
Add job logger

dejanzele:


internal/testsuite/app.go
```diff
@@ -254,9 +270,12 @@ func (a *App) Test(ctx context.Context, testSpec *api.TestSpec, asserters ...fun
 
 	// benchmark
 	report := eventBenchmark.NewTestCaseBenchmarkReport(testSpec.GetName())
-	report.Print(a.Out)
+	//report.Print(a.Out)
```
dejanzele, line 273: This is for my testing, will revert it once this PR gets in a good shape

internal/testsuite/app.go
```diff
@@ -232,6 +233,21 @@ func (a *App) Test(ctx context.Context, testSpec *api.TestSpec, asserters ...fun
 	watcher.Out = a.Out
 	g.Go(func() error { return watcher.Run(ctx) })
 
+	// TODO: handle better
+	kubectxs := []string{
```
dejanzele, line 237: I do not fully understand how and where will we run the testsuite, out-of-cluster or in-cluster, and how will we solve connectivity with Executor clusters

internal/testsuite/joblogger/service.go
```diff
@@ -0,0 +1,128 @@
...
+
+	g.Go(func() error {
+		return srv.runScraper(ctx)
+	})
+
+	return g.Wait()
+}
+
+func getKubeConfigPath() *string {
+	var kubeconfig *string
```
dejanzele, line 99: This will be refactored once we agree on how and where will we run testsuite

go.mod
```diff
@@ -75,9 +75,9 @@ require (
 	google.golang.org/genproto v0.0.0-20211208223120-3a66f561d7aa
 	google.golang.org/grpc v1.43.0
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.22.4
```
kannon92, line 78: We had a lot of issues with updating this.  I believe we have to upgrade to go 1.18 in order to actually get the tests to run.  

We have dependabot PR that are trying to update this but they keep failing tests.  

I'd ask that if you need to update this you should do it in a separate PR.

dejanzele: Actually I need to import a new library kubernetes/client-go, and it updated a lot of dependencies, need to figure out an appropriate version

kannon92: I think that was already present in the go.mod.

internal/testsuite/app.go
```diff
@@ -232,6 +234,21 @@ func (a *App) Test(ctx context.Context, testSpec *api.TestSpec, asserters ...fun
...
+		"arn:aws:eks:us-east-1:235758441054:cluster/dev-armada",
```
kannon92, line 239: Are there any security issue exposing these in a public repo?

dejanzele: I added a comment on this block, this is WIP, I am still not sure how and where will we run testsuite, will it be in-cluster or out-of-cluster, this definitely will be removed in future commits to this PR

pkg/client/connection.go
```diff
@@ -23,6 +23,9 @@ import (
 type ApiConnectionDetails struct {
 	ArmadaUrl     string
 	ArmadaRestUrl string
+	// Specify executor clusters configured for Armada server
+	// Used for monitoring of Armada Jobs
```
severinson, line 27:
```suggestion
	// URLs of executors.
	// Used by the test suite to download logs from pods running tests.
```

severinson (approved):

PR #1288, opened 2022-07-28
enhance executor logs with more resource support

dejanzele:


dejanzele:
Example output from the GROSS EKS GPU cluter:

```
INFO[2022-07-28T15:02:29Z] Requesting new jobs with free resource cpu: 10795m, memory 73833Mi, ephemeral-storage: 273375Mi, nvidia.com/gpu: 1. Received 0 new jobs.
```

PR #1297, opened 2022-08-01
Added missing tests

Sharpz7:
Closes https://github.com/G-Research/armada/issues/1268

### DIFF
--- a/client/python/tests/unit/server_mock.py
+++ b/client/python/tests/unit/server_mock.py
@@ -29,10 +29,8 @@ class SubmitService(submit_pb2_grpc.SubmitServicer):
 
 
 class EventService(event_pb2_grpc.EventServicer):
-    def Watch(self, request, context):
-        return event_pb2.EventMessage()
-
     def GetJobSetEvents(self, request, context):
+
         events = [event_pb2.EventStreamMessage()]
 
         for event in events:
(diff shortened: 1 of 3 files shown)


client/python/tests/unit/server_mock.py
```diff
@@ -31,3 +31,9 @@ def UpdateQueue(self, request, context):
 class EventService(event_pb2_grpc.EventServicer):
     def Watch(self, request, context):
```
kannon92, line 32: Since this is depreciated, can you just remove the Watch function?

richscott (approved):

kannon92 (changes requested):
I think we should remove Watch from EventService.  After that, this LGTM.

client/python/tests/unit/test_event.py
```diff
@@ -58,3 +89,43 @@ def test_event_class(name, event_type):
...
+        "ingress_info",
+        "reprioritizing",
+        "updated",
+    ],
+)
+def test_unmarshal_event_response(name):
+    test_event = FakeEventStreamMessage(name)
+    test_event_from_class = Event(test_event)
+
+    test_event_from_method = tester.unmarshal_event_response(test_event)
```
kannon92, line 122: This leaves a bad taste in my mouth.  I really don't think we need all this setup to test a static function that takes an EventMessage.  

We aren't testing any RPC calls here so I'd like to see a way to avoid having to create mock grpc servers to test this.

kannon92: Maybe you can move this to the event class?

Sharpz7: Yeah I was not a huge fan of this either, but I did not experiment with trying to use the ArmadaClass Directly. I will see if I can get this to work and go from there.

kannon92 (approved):

richscott (approved):

PR #1305, opened 2022-08-02
Jobservice sqllite

### LINKED ISSUE #1265: [Airflow Operator] Job Service Cache should use sqllite as the persistence database
In this issue, we should address that the jobservice needs a persistence datastore.  From our design doc, we are going to use Sqllite.  

1) All writes should be done in batches from the in memory database.

kannon92:
Closes #1265 

Add SQLlite backend to our job service cache


richscott (commented):
Just a single style coment, and questions about possible locking.

internal/jobservice/repository/sql_job_service.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// Save our in memory map to Database and delete from in memory map.
+// Is failing to persist a fatal error?  I'd think so..
+func (s *SQLJobService) PersistDataToDatabase() error {
+	log.Info("Saving Data to Database")
+	s.jobStatus.jobLock.RLock()
+	defer s.jobStatus.jobLock.RUnlock()
+	for key, value := range s.jobStatus.jobMap {
+		stmt, err := s.db.Prepare("INSERT INTO jobservice VALUES (?, ?, ?, ?, ?, ?)")
```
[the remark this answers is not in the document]
kannon92: Good call!

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -61,7 +60,7 @@ func (eventToJobService *EventsToJobService) streamCommon(clientConnect *client.
 		// This will log an error to the jobservice log saying that the connection was used.
 		ticker := time.NewTicker(time.Duration(eventToJobService.jobServiceConfig.SubscribeJobSetTime) * time.Second)
 		for range ticker.C {
-			if !eventToJobService.jobServiceRepository.IsJobSetSubscribed(queueJobSetKey) {
+			if !eventToJobService.jobServiceRepository.IsJobSetSubscribed(eventToJobService.queue, eventToJobService.jobsetid) {
```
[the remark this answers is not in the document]
kannon92: Yep! Good call

internal/jobservice/repository/sql_job_service.go
```diff
@@ -0,0 +1,267 @@
...
+	// If we were able to delete from DB than we assume all is good
+	return rowsAffected, errDelete
+}
+
+func (s *SQLJobService) deleteJobsFromInMemoryMap(queue string, jobSet string) (int64, error) {
+	var rowsAffected int64 = 0
+	for _, val := range s.jobStatus.jobMap {
+		if val.queue == queue && val.jobSetId == jobSet {
+			rowsAffected++
+			delete(s.jobStatus.jobMap, val.jobId)
```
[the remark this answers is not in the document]
kannon92: yea, good call

internal/jobservice/application.go
```diff
@@ -41,19 +42,26 @@ func (a *App) StartUp(ctx context.Context) error {
 	inMemoryMap := make(map[string]*repository.JobTable)
 	subscribedJobSets := make(map[string]*repository.SubscribeTable)
 	jobStatusMap := repository.NewJobStatus(inMemoryMap, subscribedJobSets)
-	inMemoryJobService := repository.NewInMemoryJobServiceRepository(jobStatusMap, config)
-	jobService := server.NewJobService(config, *inMemoryJobService)
+
+	db, err := sql.Open("sqlite", "jobservice.db")
```
jayofdoom, line 46: this path should likely be configurable

internal/jobservice/repository/job_table.go
```diff
@@ -8,15 +8,15 @@ import (
...
+	queue       string
+	jobSetId    string
+	jobId       string
+	jobResponse js.JobServiceResponse
+	timeStamp   int64
 }
 
 // Construct a JobTable and adds a timestamp when it was created.
 func NewJobTable(queue string, jobSetId string, jobId string, jobResponse js.JobServiceResponse) *JobTable {
 	// Construct a key of (queue, jobSetId) since that uniqutely determines jobset.
```
jayofdoom, line 20: The code that did this is removed but the comment has not been removed.

internal/jobservice/repository/sql_job_service.go
```diff
@@ -0,0 +1,267 @@
...
+
+// Get the JobStatus given the jodId
+// If a job is not in the map, we return JOB_ID_NOT_FOUND
+// This should not be an error.
+func (s *SQLJobService) GetJobStatus(jobId string) (*js.JobServiceResponse, error) {
+	s.jobStatus.jobLock.RLock()
+	jobResponse, ok := s.jobStatus.jobMap[jobId]
+	s.jobStatus.jobLock.RUnlock()
+	if !ok {
+		sqlResponse, errSql := s.GetJobStatusSQL(jobId)
```
jayofdoom, line 75: AFAICT, the jobId is never sanitized against SQL injection at any point. We should ensure these inputs are well validated before feeding them to sqlite (and/or the SQL wrapper functions, depending on how you wanna structure it -- IMO it's best to do it closest to the DB call as possible).

dave-gantenbein: There must be some OSS sanitation library we can leverage for this purpose, right?

jayofdoom: I think there's something in the golang stdlib, actually, but I haven't looked in detail.

jayofdoom: Kevin and I talked about this; using parameterized queries like exist already ensure you can't do sql injection. It's so embedded in the stdlib you get it for borderline-free!

richscott: SQL sanitization techniques are not needed here - the `GetJobStatusSQL()` func uses parameter binding - this robustly thwarts injection attempts at the driver level:
```
func (s *SQLJobService) GetJobStatusSQL(jobId string) (*js.JobServiceResponse, error) {
  row := s.db.QueryRow("SELECT JobResponseState, JobResponseError FROM jobservice WHERE JobId=?", jobId)
  ```
If there's any creation of complex SQL via string concatenation, then yeah, it gets risky very quickly.

Comment (richscott) [reply internal/jobservice/repository/sql_job_service.go:75]: The classic sprintf() technique (it protects nothing) is warned against in https://tip.golang.org/doc/database/sql-injection

File: internal/jobservice/application.go (hunk 4 above)
Comment (kannon92) [reply internal/jobservice/application.go:46]: I agree with you but this is the name of the database for this library.  It is not the path.  This library only takes what you pass as second parameter to be the name of the database.  Unclear of ways around that.  

Now I can make the name of the databse configurable.

Comment (jayofdoom) [reply internal/jobservice/application.go:46]: What needs to be configurable is the place on disk the jobservice.db is created/opened from. That may include being able to change the name.

File: internal/jobservice/repository/sql_job_service.go (hunk 6 above)
Comment (kannon92) [reply internal/jobservice/repository/sql_job_service.go:73]: Thanks all!

File: internal/jobservice/repository/sql_job_service.go [hunk 7]
```diff
@@ -0,0 +1,266 @@
...
+// Open to suggestions on how to make this better
+func (s *SQLJobService) CreateTable() {
+	_, err := s.db.Exec("DROP TABLE IF EXISTS jobservice")
+	if err != nil {
+		panic(err)
+	}
+	_, err = s.db.Exec(`
+CREATE TABLE jobservice (
+Queue TEXT,
+JobSetId TEXT,
```
richscott, line 56: I just thought of this - of course this may be completely "premature optimization", but any chance you might need indexes on JobSetId and/or JobId, if you're doing queries/updates on those, and the cardinality (num rows) of this table gets big?  (If it's less than a few thousand rows, then indexes aren't worth the bother.)

internal/jobservice/application.go
```diff
@@ -41,19 +42,26 @@ func (a *App) StartUp(ctx context.Context) error {
...
+	if err != nil {
+		log.Errorf("Error Opening Sqlite DB from jobservice.db %v", err)
```
d80tb7, line 48: I think you can do `log.withError(err).Err` (at least you can with logrus) which gives you the stacktrace.

d80tb7: also if you have an error here I think you might want to exit?  (or possibly try and instantiate a nnew db).  I don't think trying to run the below code with a broken db will end well!

internal/jobservice/repository/sql_job_service.go
d80tb7: I'm pretty sure we'll want indexes

internal/jobservice/repository/sql_job_service.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+type SubscribedTuple struct {
+	Queue  string
+	JobSet string
+}
+
+// Create a Table from a hard-coded schema.
+// Open to suggestions on how to make this better
+func (s *SQLJobService) CreateTable() {
```
d80tb7, line 48: I'm not sure I understand this?  The purpose of sql lite is to recover state when we restart, however this function drops the table and is called every time on startup?

internal/jobservice/repository/sql_job_service.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// Get the JobStatus given the jodId
+func (s *SQLJobService) GetJobStatus(jobId string) (*js.JobServiceResponse, error) {
+	s.jobStatus.jobLock.RLock()
+	jobResponse, ok := s.jobStatus.jobMap[jobId]
+	s.jobStatus.jobLock.RUnlock()
+	if !ok {
+		sqlResponse, errSql := s.GetJobStatusSQL(jobId)
+		if sqlResponse != nil {
```
d80tb7, line 74: won't this return an error if the job isn't present yet?  Is that correct?

internal/jobservice/repository/sql_job_service.go
```diff
@@ -0,0 +1,266 @@
...
+)`)
+	if err != nil {
+		panic(err)
+	}
+}
+
+// Get the JobStatus given the jodId
+func (s *SQLJobService) GetJobStatus(jobId string) (*js.JobServiceResponse, error) {
+	s.jobStatus.jobLock.RLock()
+	jobResponse, ok := s.jobStatus.jobMap[jobId]
```
d80tb7, line 70: It looks like this code first attempts to serve the status from memory and only then tries to fall back to the db.  Have you done some tests to see what sort of performance can be gained if we just serve everything out of the db?  It would obviously lead to a simpler system.

internal/jobservice/repository/sql_job_service.go
d80tb7: Actually I checked and it won't because  `GetJobStatusSQL` returns (`notFound`, `nil`) in this case.  I think you probably want to change `if sqlResponse != nil` to `if err != nil` on line 74

internal/jobservice/repository/sql_job_service.go
```diff
@@ -0,0 +1,266 @@
...
+	case "CANCELLED":
+		jobProtoResponse.State = js.JobServiceResponse_CANCELLED
+	case "JOB_ID_NOT_FOUND":
+		jobProtoResponse.State = js.JobServiceResponse_JOB_ID_NOT_FOUND
+	}
+	return jobProtoResponse, nil
+}
+
+// Update in memory JobStatus Map with jobId and our JobTable
+func (s *SQLJobService) UpdateJobServiceDb(jobTable *JobTable) {
```
d80tb7, line 115: I would change the name of this function as it's a bit weird to have a function called `UpdateJobServiceDb` that doesn't actually update the DB.

internal/jobservice/repository/sql_job_service.go
```diff
@@ -0,0 +1,266 @@
...
+// Save our in memory map to Database and delete from in memory map.
+func (s *SQLJobService) PersistDataToDatabase() error {
+	log.Info("Saving Data to Database")
+	s.jobStatus.jobLock.Lock()
+	defer s.jobStatus.jobLock.Unlock()
+	stmt, err := s.db.Prepare("INSERT INTO jobservice VALUES (?, ?, ?, ?, ?, ?)")
+	if err != nil {
+		panic(err)
+	}
+	for key, value := range s.jobStatus.jobMap {
```
d80tb7, line 257: I think this will be quite slow because you'll have to persist every job to the database.  Also will this work if the job is already in the DB? In most DBs this would have to be an upsert rather than an insert but I don't know about SQLLite

d80tb7:
Looks good- a couple of general comments:

It's not clear to me how we can use sql lite to recover from restarts.  Ideally I think the following would happen:
  - The service recovers all jobstates from the db
  - The service resubscribes to all the jobsets it was subscribed to before the restart *using the correct offsets*.  

I'm not 100% sold on the mechanism of storing the state first in a map and then putting in a db once every so often.  Fundamentally it means:
 - If the service goes down before the data was persisted, then that dat is lost.
 - We end up with the data stored in multiple places and consistency is hard
The simplest solution would just be to store all data in the database and never cache anything in memory.  This might not be fast enough but we should benchmark it to see if  it is, because if we can get away with it then it's by far the simplest option.
A different option would be to make the db the primary store and then add a caching layer on top.  The way I would do this is to have all event rpc streams write state directly into the db (if this is too slow you should be able to write in batches to make it faster). On the serving side I'd have a map of cached states like you do right now but I'd populate them lazily from the cache.  Every time I wrote a new rpc message into the db I'd invalidate the cache entry for that job.  Hopefully that means that most entries wil be served from cache, but the db remains the source of truth for restarts.

dave-gantenbein:
@d80tb7 makes some good points, but I could also see putting them in the bucket of "further optimizations" for version 0.2 and proceeding with some real life tests so see if anything else breaks. It would be good to start to get some miles on the entire system.

internal/jobservice/repository/sql_job_service.go
kannon92: I'll add some indexes.

internal/jobservice/repository/sql_job_service.go
kannon92: I'll add a unit test for that!

kannon92:
@d80tb7 and I followed up.  We will do the following in this PR:

1) We will remove the in memory job component and just focus on the database being the source of truth.
   a) Reason for this is to simplify the implementation and focus on improvements once we have a real production environment.  Future performance improvements can be adding a cache and/or playing with persistence interval.
2) We will add indexes to queue and job_set_id.  Hopefully job_id iss already indexed but we will verify that.  
3) If we want smarter fault tolerance we would need to have a subscribe sql table.  Right now any jobservice crash/restart will purge our subscribe map and then we have to resubscribe to any events that are still being polled.  This is a future improvement.

internal/jobservice/repository/sql_job_service.go
kannon92: We discussed this and we will keep this as is.  If we want true fault tolerance we need to have a subscribe job set table in addition to the job table.  Future work

internal/jobservice/repository/sql_job_service.go
kannon92: Please take a look!  I marked that JOB_ID must be a unique id and added an index for queue and job-set.

internal/jobservice/repository/sql_job_service.go
kannon92: Function actually does update a database now.

internal/jobservice/repository/sql_job_service.go
d80tb7: I think the index is good- I don't think you need the constraint, rather you should define `JobId` to be the primary key as per https://www.sqlitetutorial.net/sqlite-primary-key/.  That will essentially define it is unique and indexed.

internal/jobservice/repository/sql_job_service.go
```diff
@@ -0,0 +1,220 @@
...
+func (s *SQLJobService) UpdateJobServiceDb(jobTable *JobTable) {
+	stmt, err := s.db.Prepare("INSERT OR REPLACE INTO jobservice VALUES (?, ?, ?, ?, ?, ?)")
+	if err != nil {
+		panic(err)
+	}
+	jobState := jobTable.jobResponse.State.String()
+	_, errExec := stmt.Exec(jobTable.queue, jobTable.jobSetId, jobTable.jobId, jobState, jobTable.jobResponse.Error, jobTable.timeStamp)
+
+	if errExec != nil {
+		panic(errExec)
```
d80tb7, line 112: let's put a todo here to make this a bit more robust

internal/jobservice/repository/sql_job_service.go
```diff
@@ -0,0 +1,220 @@
...
+	defer s.jobSetSubscribe.subscribeLock.Unlock()
+	primaryKey := queue + jobSet
+	_, ok := s.jobSetSubscribe.subscribeMap[primaryKey]
+	if !ok {
+		s.jobSetSubscribe.subscribeMap[primaryKey] = NewSubscribeTable(queue, jobSet)
+	}
+
+}
+
+// UnSubscribe to JobSet and delete all the jobs in the in memory map
```
d80tb7, line 151: thinkt he comment might be out of date as now it deletes the jobs in the db?

d80tb7: Also technically this function doesn't unsubscribe (i.e. stop the goroutine with the subscription).  I think it should be renamed `cleanupUnsubscribedData` or something

internal/jobservice/repository/sql_job_service.go
```diff
@@ -0,0 +1,220 @@
...
+	_, ok := s.jobSetSubscribe.subscribeMap[primaryKey]
+	if ok {
+		s.jobSetSubscribe.subscribeMap[primaryKey] = NewSubscribeTable(queue, jobSet)
+		return nil
+	} else {
+		return fmt.Errorf("queue %s jobSet %s is already unsubscribed", queue, jobSet)
+	}
+}
+
+// Delete Jobs in the map.
```
d80tb7, line 201: comment out of date

d80tb7 (approved):

d80tb7 (approved):

PR #1308, opened 2022-08-02
Compress JobFailedEvent Messages in Redis

d80tb7:
Compress the `Failed` Message in Redis.  This is because this field can get quite large and would benefit from compression.  

Unfortunately this results in a new message showing up in the public api which is not for public consumption (they will never get messages of this type).  Once we redo events from pulsar we can get rid of this message.


internal/armada/repository/event.go
```diff
@@ -25,23 +30,56 @@ type EventRepository interface {
...
+		}), &poolConfig)
+
+	decompressorPool := pool.NewObjectPool(context.Background(), pool.NewPooledObjectFactorySimple(
+		func(context.Context) (interface{}, error) {
+			return compress.NewZlibDecompressor()
+		}), &poolConfig)
+	return &RedisEventRepository{db: db, eventRetention: eventRetention, compressorPool: compressorPool, decompressorPool: decompressorPool}
 }
 
+// ReportEvent reports the event to redis.  Note that this function my modify the supplied message in-place
```
JamesMurkin, line 65:
```suggestion
// ReportEvent reports the event to redis.  Note that this function may modify the supplied message in-place
```

internal/armada/repository/event.go
```diff
@@ -25,23 +30,56 @@ type EventRepository interface {
...
 func (repo *RedisEventRepository) ReportEvent(message *api.EventMessage) error {
 	return repo.ReportEvents([]*api.EventMessage{message})
 }
 
+// ReportEvents reports events to redis.  Note that this function my modify the supplied messages in-place
```
JamesMurkin, line 70:
```suggestion
// ReportEvents reports events to redis.  Note that this function may modify the supplied messages in-place
```

JamesMurkin (approved):

PR #1312, opened 2022-08-03
Allow cancelling by state in Lookout 

JamesMurkin:
Adding new endpoint to Armada API `CancelJobSet`
 - This has a filter, that allows cancelling by state

Added to Lookout to add buttons to select the states you want to cancel
 - Currently we only support Queued or Pending + Running, as Armada has no concept of Pending
 - We will add this in the future, however users need this feature sooner


JamesMurkin:
image

pkg/api/submit.proto
```diff
@@ -173,6 +184,12 @@ service Submit {
             body: "*"
         };
     }
+    rpc CancelJobSet (JobSetCancelRequest) returns (CancellationResult) {
```
kannon92, line 187: Why do we need a new API for this?  

Wouldn't it be better to just add JobSetFilter to the JobCancelRequest and utilize CancelJobs rpc call for this?

JamesMurkin: I think it's something we want to discuss internally. For a while we've wanted to split that endpoint rather than having a combined endpoint (cancel by job and cancel by jobset).

We absolutely could continue to extend the existing and the only reason I didn't, is to move towards what we want in future. 

If we decide we don't want that, I'll swap to extending the existing as you suggested.

kannon92: I like that.  I have been meaing to create a python functions that call the CancelJob via jobset/queue or job since the behavior is different and this saves me that work.  

So our goal would be to have CancelJobs take only a jobId and CancelJobSet to take queue and JobSet and list of states?

We should file that as an issue so we document that this is something we'd like.  It is a good issue for anyone to tackle too!

kannon92:
Since we are adding a new API, we should mark that future GRPC clients need to be updated with this RPC call.  

Python should be updated to use this call.  I think we should get in the habit of calling these out.  I'm not saying you need to make the change now but this is incurring tech debt and we should add this API call to our python client. 

Can you create a github issue to create a corresponding change in the python client that reflects this new API call?

pkg/api/submit.proto
```diff
@@ -62,6 +62,17 @@ message JobCancelRequest {
...
 
+// swagger:model
+message JobSetCancelRequest {
+    string job_set_id = 1;
+    string queue = 2;
+    JobSetFilter filter = 3;
+}
+
+message JobSetFilter {
+    repeated string state = 1;
```
kannon92, line 73: Can you make this a enum instead of a raw string?

kannon92:
I did some testing on this branch.  I found a minor state issue.  

If you open the CancelJobSet Dialog and you have a failed cancellation ie try to cancel Queued jobs when they aren't any, you are not able to select a new checkbox in the dialog.  

Steps to reproduce:

1) Open the JobSet Cancel Dialog
2) Select a known error case (queued jobs when all are running)
3) You should get a error message
4) Try to select Running.

internal/armada/repository/job.go
```diff
@@ -838,36 +839,60 @@ func (repo *RedisJobRepository) GetQueueJobIds(queueName string) ([]string, erro
...
+type JobSetFilter struct {
+	IncludeQueued bool
+	IncludeLeased bool
+}
+
+func (repo *RedisJobRepository) GetJobSetJobIds(queue string, jobSetId string, filter *JobSetFilter) ([]string, error) {
+	var queuedIdsCommandResult *redis.StringSliceCmd
+	var leasedIdsCommandResult *redis.StringSliceCmd
 
 	tx := repo.db.TxPipeline()
```
d80tb7, line 857: Not part of this review specifically- but are you sure this TX is doing anything.  My understanding of Redis Transactions was that you had to pass in a list of keys to `Watch` before you call `Exec` if you want transactional behaviour.

internal/armada/repository/job.go
```diff
@@ -838,36 +839,60 @@ func (repo *RedisJobRepository) GetQueueJobIds(queueName string) ([]string, erro
...
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
 
-	activeIds := util.StringListToSet(append(queuedIds, leasedIds...))
+	activeIdsSet := util.StringListToSet(activeJobIds)
 	activeSetIds := []string{}
 	for _, id := range jobSetIds {
-		if activeIds[id] {
+		if activeIdsSet[id] {
```
d80tb7, line 895: the naming here is a bit confusing. We have `activeIdsSet` (which is a set) and `activeSetIds` (which is logically a set).

internal/armada/server/submit_to_log.go
```diff
@@ -361,6 +362,70 @@ func (srv *PulsarSubmitServer) CancelJobs(ctx context.Context, req *api.JobCance
...
+				CancelJob: &armadaevents.CancelJob{JobId: jobId},
+			},
+		})
+	}
+
+	err = srv.publishToPulsar(ctx, []*armadaevents.EventSequence{sequence})
+
+	if err != nil {
+		log.WithError(err).Error("failed send to Pulsar")
+		return nil, status.Error(codes.Internal, "Failed to send message")
```
d80tb7, line 421: I think we could give a better error message to the user here- I.e "error sending cancel job messages to pulsar" or something.  It'll make our lives easier when we get reports of failure.

pkg/api/submit.proto
JamesMurkin: Yeah that was my thought too, will go over the public API change with Chris today to decide what we want
 - This being an enum was a point I mentioned to him already

JamesMurkin:
So I should have fi

> I did some testing on this branch. I found a minor state issue.
> 
> If you open the CancelJobSet Dialog and you have a failed cancellation ie try to cancel Queued jobs when they aren't any, you are not able to select a new checkbox in the dialog.
> 
> Steps to reproduce:
> 
> 1. Open the JobSet Cancel Dialog
> 2. Select a known error case (queued jobs when all are running)
> 3. You should get a error message
> 4. Try to select Running.

I believe I have fixed this now, you can now freely change the checkboxes on the retry screen

internal/armada/repository/job.go
JamesMurkin: I think that it is doing is trying to get a consistent snapshot of the ids.

A transaction in theory says all these commands will run without anyone running in between them - so that state change change between them.

I believe this is how it work - without having to do watches. However I could be wrong, but that is what it is trying to achieve.

internal/armada/server/submit_to_log.go
JamesMurkin: Agreed, I just used what was already used for cancelling jobs

internal/armada/domain/job_set_filter.go
```diff
@@ -0,0 +1,28 @@
+package domain
+
+type JobState string
```
d80tb7, line 3: so this is a real nit but it's not obvious to me how all this code is better than the more succinct:
```
var validSates = map[string]bool { "queued" : true,  "pending": true, "running": true}

func IsValidFilterState(s string) bool {
	_, ok := validSates[s]
	return ok
}
```

As an aside- Ideally the states would not have mixed case e.g. we'd have "RUNNING" rather then "Running"- if we did that we could normalise the case everywhere and forget about casing issues. I'm guessing that's not possible though because Armada already uses a mixed case elsewhere?

d80tb7 (approved):

internal/armada/repository/job.go
d80tb7: no actually you are right- you only need a watch for other keys

internal/armada/server/submit_from_log.go
```diff
@@ -510,7 +506,11 @@ func (srv *SubmitFromLog) CancelJobSet(ctx context.Context, queueName string, jo
 	} else if err != nil {
 		return true, err
 	}
+	//TODO Set userid correctly so we know who actually cancelled the jobset
```
kannon92, line 509: When I see todo in code, I think two things:

should you do this now while the code is fresh?

or should you file this as a issue so we have it in our backlog?

kannon92:
> > I did some testing on this branch. I found a minor state issue.
> > If you open the CancelJobSet Dialog and you have a failed cancellation ie try to cancel Queued jobs when they aren't any, you are not able to select a new checkbox in the dialog.
> > Steps to reproduce:
> > 
> > 1. Open the JobSet Cancel Dialog
> > 2. Select a known error case (queued jobs when all are running)
> > 3. You should get a error message
> > 4. Try to select Running.
> 
> I believe I have fixed this now, you can now freely change the checkboxes on the retry screen

I’ll test out again today but feel free to merge if you need this

internal/armada/server/submit_from_log.go
JamesMurkin: I didn't add this, I spotted this bug while doing this PR.

We should probably raise an issue, as there is no obvious solve

internal/armada/server/submit_to_log.go
```diff
@@ -361,6 +362,68 @@ func (srv *PulsarSubmitServer) CancelJobs(ctx context.Context, req *api.JobCance
...
+			Message: "JobSetId is empty",
+		}
+	}
+
+	err := validation.ValidateJobSetFilter(req.Filter)
+	if err != nil {
+		return nil, err
+	}
+
+	userId, groups, err := srv.Authorize(ctx, req.Queue, permissions.CancelAnyJobs, queue.PermissionVerbCancel)
```
d80tb7, line 386: this is a nit but I'd always make the authorize statement the very first thing in the function (I understand this means making authorize reject empty queue/jobset).  I'd do this because otherwise there's a risk that someone adds some code above the authorize that needs the user to be verified before exceution.

This is probably out of scope of the PR so shouldn't block merging, but if you agree we can raise a ticket to audit the existing code to make it so.

d80tb7 (approved):

internal/armada/server/submit_to_log.go
JamesMurkin: I agree its probably best practice. We should raise a ticket to go over all uses and make them consistent (auth first)

Issue #1316, opened 2022-08-04
Update Python/Client README.md to explain our release process for python client


kannon92:
With #1072 we are now able to manually run a release github action.  

https://github.com/G-Research/armada/blob/master/client/python/README.md#releasing should be updated to reflect our new release process.  

1) Create a PR that updates the pyproject.toml version number for the python client.
2) Merge that PR into main.
3) Select the Release PyPI action and deploy a new wheel file to PYPI.

kannon92:
<img width="1282" alt="Screen Shot 2022-08-04 at 9 37 18 AM" src="https://user-images.githubusercontent.com/3780425/182861609-41587336-2747-4a7c-aa32-73c7515f6e2e.png">

This is how you select the release action for deployment.

jayofdoom:
Going to knock this out + make a general pass over client/python/README.md for accuracy/layout

PR #1318, opened 2022-08-04
Update and rearrange docs + add info on releasing

jayofdoom:
There were some items in CONTRIBUTING.md that were incomplete, and
some items in README.md that needed to be moved to CONTRIBUTING.md.

Closes: #1316

### DIFF
--- a/client/python/CONTRIBUTING.md
+++ b/client/python/CONTRIBUTING.md
@@ -1,48 +1,106 @@
-How to develop on this python client project 
-=
+# Contributing to the Armada python Client
 <hr />
 
 Armada Project welcomes contributions from the opensource community.
 
-This client is written primarily targeting supported versions of python 3 on Linux, Mac, and WSL on Windows.
+This client is written primarily targeting supported versions of python 3 on Linux, Mac, and WSL on Windows. Please file
+an [issue](https://github.com/g-research/armada/issues/new) if you need support for another platform.
 
-### Setting up your own fork of this repository on Linux, Mac, or WSL OS.
+## First time contribution using a fork
+<hr />
+Armada encourages using the [fork then PR](https://gist.github.com/Chaser324/ce0505fbed06b947d962)
+workflow for contributing. First time contributors can follow the guide below to get started.
+
+### Prerequisites
+- A working knowledge of git and python development.
+- Supported versions of `python`, `pip`, and `tox` are installed.
+- A working docker client, or docker-compatible client available under `docker`.
+- Network access to fetch docker images and go dependencies.
+
+### Fork and clone repository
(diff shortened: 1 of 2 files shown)


jayofdoom:
I likely went overboard on the detail, but I'm not sure. Let me know what you think; I'm happy to prune when+where we think is appropriate.

Part of me wonders if we should remove all of the detailed git usage documentation from contributing. If we're going to have it though, it needs to be more complete than it was before.

client/python/README.md
```diff
@@ -1,57 +1,30 @@
...
+- A working docker client, or docker-compatible client available under `docker`.
+- Network access to fetch docker images and go dependencies.
 
-Log into the Armada-GROSS account on PyPI.
-Generate the API tokens and copy those tokens to ~/.pypirc.
-`poetry build`
-`poetry run twine upload dist/*`
+To generate all needed code, and install the python client:
+1) From the root of the repository, run `make python`
+2) Install the client using `pip install client/python`. It's strongly recommended you do this inside a virtualenv.
```
richscott, line 30: I'd suggest changing "virtualenv" to "virtual environment", maybe even with a ", for example, inside a Docker container" - so nobody thinks `virtualenv` may be a specific tool name, etc.

client/python/CONTRIBUTING.md
```diff
@@ -1,48 +1,106 @@
...
+### Preparing repository for development
+<hr />
+Unlike most python projects, the Armada python client contains a large quantity of generated code. This code must be
+generated in order to compile and develop against the client.
 
-1) On github interface click on the `Fork` button.
-2) Clone your forked of armada repository on your personal git repository by running `git clone github.com:Your_GitHub_Username/armada.git`.
-3) Enter the directory `cd armada`
-4) Add upstream repository `git remote add upstream https://github.com/G-Research/armada.git`.
+From the root of the repository (e.g. `armada/`), run `make python`. This will generate python code needed to build
```
kannon92, line 31:
```suggestion
From the root of the repository, run `make python`. This will generate python code needed to build
```

There are folders in this repo called armada and that is dependent on w/e someone called the repo when they cloned.

kannon92 (dismissed):

client/python/README.md
jayofdoom: `virtualenv` *is* a specific tool name: https://virtualenv.pypa.io/en/latest/ (aka the `venv` module in base python 3)

richscott: Wow! I thought it was a vernacular word - I just expected any Python tool had to have a `py` prefix or some such.

richscott (approved):

kannon92 (approved):

Issue #1319, opened 2022-08-04
protobuf processing redundancy and slowness in Dockerfile, makefile


richscott:
The recent changes to implement GRPC bindings for C#/.Net introduced redundancies in the build/proto/Dockerfile and makefile - the logic to copy proto files in the former should be removed, and all Make targets that do protobuf processing should use the `setup-proto` prerequisite target instead.

Also, the `setup-proto` Make target sequentially invokes many golang containers, each running a single `cp` command - these copy commands should be coalesced into a single container execution, to reduce Docker container spin-up time cost.

kannon92:
Didn't auto close this.  Can close with #1320

PR #1320, opened 2022-08-04
Remove redundant proto file logic and speed up copying (#1319)

richscott:
Remove the redundant proto file copying that I added in the recent dotnet-grpc changes; speed up proto file copying operations in the makefile by running them all in just one container instance, rather than in many sequential ones.

With these changes, running `make dotnet-grpc` goes from taking ~39 secs to ~12 secs on my Macbook Pro; on my AMD 12-core 32GB Linux desktop system, it goes from ~11.5 secs to ~6.8 secs. The generated artifacts are the same.

closes 1319


kannon92 (approved):
Thanks for adding this so quickly!

jayofdoom (approved):
Tested working for me. I saw a speedup as well.

richscott:
> Thanks for adding this so quickly!

Atoning for my earlier sins.  😅

PR #1330, opened 2022-08-05
Made pod_spec optional

Sharpz7:
Closes https://github.com/G-Research/armada/issues/1322


kannon92:
Should you add logic to enforce either a single pod_spec or multiple pod_spec?  Having both as optional means that someone could not specify a pod spec which would be a failure..

Sharpz7:
> Should you add logic to enforce either a single pod_spec or multiple pod_spec? Having both as optional means that someone could not specify a pod spec which would be a failure..

I think that gRPC will probably error itself, but I will test it on Monday while doing the examples PR.

kannon92 (approved):
LGTM.

jayofdoom (dismissed):

jayofdoom (approved):

kannon92 (approved):

PR #1339, opened 2022-08-08
.NET GRPC Client

samclark:
This pull request contains changes to the dotnet grpc client:

- move from `Grpc` to `Grpc.Net.Client` (`Grpc` has been deprecated and is also incompatible with our grpc services)
- generate dotnet code from proto definitions using `dotnet build...` instead of `protoc` commands in bash
- csharp namespaces added to all armada protos (a bit ugly but the only supported way [details]( https://github.com/protocolbuffers/protobuf/issues/6846))
- publish to nuget using `dotnet pack...` and `dotnet nuget push...` via circle ci


JamesMurkin (dismissed):

client/DotNet/ArmadaProject.Io.Client.Test/Tests.cs
```diff
@@ -0,0 +1,113 @@
...
+                Credentials = credentials,
+            };
+
+            return GrpcChannel.ForAddress(address, options);
+        }
+
+        private static JobSubmitRequest CreateSubmitRequest(string queue, string jobSetId)
+        {
+            var flex = new FlexVolumeSource();
+            flex.Driver = "gr/cifs";
```
kannon92, line 75: Can we avoid putting these 
GR specific things in our public repo?  I'd vote for just simple pod specs with no volumes for our repo.

samclark: Yes this makes sense. I've removed the volume from the old and new clients.

JamesMurkin (approved):

JamesMurkin (approved):

PR #1346, opened 2022-08-09
Use docker buildkit to improve build times

jayofdoom:
Docker buildkit builds are more efficient and allows parallelization.
Beyond any initial performance benefit, we could also use newer syntax
in our dockerfiles to make it more parallel, with use of buildkit mode.

Issue: #1037

### DIFF
--- a/makefile
+++ b/makefile
@@ -13,6 +13,12 @@ else
 	K := $(foreach exec,$(EXECUTABLES),$(if $(shell which $(exec)),some string,$(error "No $(exec) in PATH")))
 endif
 
+# Docker buildkit builds work in parallel, lowering build times. Outputs are compatable.
+# Ignored on docker <18.09. This may lead to slower builds and different logs in STDOUT,
+# but the image output is the same.
+# See https://docs.docker.com/develop/develop-images/build_enhancements/ for more info.
+export DOCKER_BUILDKIT = 1
+
 # Get the current date and time (to insert into go build)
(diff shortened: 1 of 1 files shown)


jayofdoom:
14:19 in this PR for CircleCI build_and_integration_tests

19:18 for a run in another PR (https://app.circleci.com/pipelines/github/G-Research/armada/6038/workflows/ecfbd27e-723a-482b-92a7-ec7cb47db0bc/jobs/14237) 

I think that's a clear improvement.

makefile
```diff
@@ -13,6 +13,11 @@ else
 	K := $(foreach exec,$(EXECUTABLES),$(if $(shell which $(exec)),some string,$(error "No $(exec) in PATH")))
 endif
 
+# Docker buildkit builds work in parallel, lowering build times. Outputs are compatable.
+# Requires docker 18.09 to be effective (and is ignored on newer docker)
```
kannon92, line 17: This comment is a little confusing.  So this is enabled past Docker 18.09? And ignored for anything newer?

jayofdoom: Yes. Docker 18.09 added support for docker buildkit mode. Earlier versions will ignore this flag.

There is a noticeable difference in output between the two modes, so the comment is meant to help folks grok why it might look different.

kannon92 (approved):

makefile
jayofdoom: We worked out a better comment in slack.

severinson (approved):

PR #1347, opened 2022-08-09
Batch fetch lookout deduplication ids

d80tb7:
The current job de-duplication logic requires two SQL queries to be made for every job.  If a batch of jobs is submitted then each of these SQL queries is submitted in serial, which can lead the `SubmitJobs` call taking a significant amount of time (we have observed > one second for a batch of 200 jobs).

This PR addresses this by changing the logic so that there are at most two SQL calls per batch.  This is achieved as follows:

- Add a new function `LoadOrStoreBatch` to `PGKeyValueStore` that either loads the original jobIds for the supplied keys, or adds the supplied jobIds if the key is not not present. Because this function takes a batch of jobIds, it needs only to make at most two SQL calls to process the entire batch.
- Modify `PulsarSubmitServer` to call this function and then consider the jobs as duplicates if the originalJobId returned by `LoadOrStoreBatch` does not equal the JobId on the Job.

cc @JamesMurkin @severinson


internal/pgkeyvalue/pgkeyvalue.go
```diff
@@ -164,6 +256,48 @@ func (c *PGKeyValueStore) Get(ctx context.Context, key string) ([]byte, error) {
 	return *value, nil
 }
 
+func (c *PGKeyValueStore) GetBatch(ctx context.Context, keys []string) (map[string][]byte, error) {
```
JamesMurkin, line 259: Is this actually used? 

It isn't obvious if it, but I am only looking via Github so maybe its just non obvious and it actually implements some interface

internal/pgkeyvalue/pgkeyvalue.go
```diff
@@ -86,6 +108,76 @@ func (c *PGKeyValueStore) createTable(ctx context.Context) error {
...
+	}
+
+	if keysFound == len(batch) {
+		return addedByKey, nil
+	}
+
+	err := c.db.BeginTxFunc(ctx, pgx.TxOptions{}, func(tx pgx.Tx) error {
+
+		// See which keys are already in the db
+		sql := fmt.Sprintf("SELECT key, value from %s WHERE key = ANY($1)", c.tableName)
```
JamesMurkin, line 134: Are there any concerns about concurrent calls updating the same ids?

internal/pgkeyvalue/pgkeyvalue.go
d80tb7: No it's not- this was an api call I added while testing an alternative approach and forgot to remove.  Done now.

internal/pgkeyvalue/pgkeyvalue.go
d80tb7: You're right- it's not as bulletproof as it could be- I've modified it to be a single statement which should make it work in all circumstances.

Note that the existing `Add` method suffers from the same issue as I'd just copied the pattern used there.  For now that's not an issue as we don't use it anywhere but we should fix in a follow up PR.

internal/armada/server/submit_to_log.go
```diff
@@ -755,3 +718,67 @@ func (srv *PulsarSubmitServer) publishToPulsar(ctx context.Context, sequences []
...
+	combinedHashFromClientId := make(map[string][20]byte)
+	kvs := make([]*pgkeyvalue.KeyValue, 0, len(apiJobs))
+	for _, apiJob := range apiJobs {
+		if apiJob.ClientId != "" {
+			// Hash the ClientId together with the queue (or get it from the table, if possible).
+			combinedHash, ok := combinedHashFromClientId[apiJob.ClientId]
+			if !ok {
+				clientIdHash := sha1.Sum([]byte(apiJob.ClientId))
+				// Compute the combined hash.
+				for i, b := range clientIdHash {
```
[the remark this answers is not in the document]
d80tb7: So this code wasn't new as such- I just moved its position in the file.  That said- you're right it's not the most intuitive.  I decided to remove all the caching logic as its not the most intuitive and is almost certainly a premature optimisation given that we should be able to do millions of SHA1 hashes per second ([link](https://automationrhapsody.com/md5-sha-1-sha-256-sha-512-speed-performance/)).  Note that what Ive done here actually breaks backward compatibility so there will be a short window where de-duplication won't work, but I think that's fine given that it simplifies things considerably.

JamesMurkin (approved):

PR #1349, opened 2022-08-09
Remove checks and unit tests from circleci

jayofdoom:
These are implemented and working in github actions. Stop paying
to run them in circleci as well.

Issue #1037


jayofdoom:
In order for this to merge the following changes need to be made:
- Ensure CircleCI checks_and_unit_tests-ubuntu2204-go118 is no longer required to pass to merge.
- Ensure .github/workflows/go.yml workflow is required to pass in order to merge.

Additionally, but not technically required for this PR, I'd prefer the following tests be made required before merge:
- "Run Python Linting and Unit Tests" 
- "Run Python Armada Airflow Operator Linting and Unit Tests"

kannon92 (dismissed):

JamesMurkin:
@jayofdoom  Is there any reason we don't have a standard naming for these github actions tests?

To be consistent shouldn't they be called `Run Go Linting and Unit Tests`

kannon92:
> @jayofdoom Is there any reason we don't have a standard naming for these github actions tests?
> 
> To be consistent shouldn't they be called `Run Go Linting and Unit Tests`

I like this!  But to be clear, this isn't a comment on this PR but a general request for our GHA?  

So you'd like to see a consistent name for all the jobs rather than some being the file location and others being the name?

JamesMurkin:
> > @jayofdoom Is there any reason we don't have a standard naming for these github actions tests?
> > To be consistent shouldn't they be called `Run Go Linting and Unit Tests`
> 
> I like this! But to be clear, this isn't a comment on this PR but a general request for our GHA?
> 
> So you'd like to see a consistent name for all the jobs rather than some being the file location and others being the name?

Yeah I'm not actually bothered which format we go with, just them being the same. Although I prefer descriptive over just a file name.

The reason I bring it up now, is so I don't have to change the PR checks twice (before and after they are consistent), considering its probably a one line change.

However I can do it twice if you want time to discuss this first.

kannon92:
> 
> Yeah I'm not actually bothered which format we go with, just them being the same. Although I prefer descriptive over just a file name.
> 
> The reason I bring it up now, is so I don't have to change the PR checks twice (before and after they are consistent), considering its probably a one line change.
> 
> However I can do it twice if you want time to discuss this first.

Jay will be online in a few hours so we can ask when he starts working.  I vote for making the change in one go.  I feel that Jay would agree on descriptive as we have been doing that for the most recent ones actions we have added.

jayofdoom:
I'll find some sane-sounding name to change them all to. Probably better to ape what CodeQL does (e.g. call go.yml "Go" so it'd say "Go / go-lint" and the like.

.github/workflows/ts.yml
```diff
@@ -1,3 +1,5 @@
+name: Lookout
```
kannon92, line 1: Should we make this name be more descriptive?  This is really a javascript unit testing and linting.  Lookout has a backend Go component and a UI.  This action tests javascript only.

jayofdoom: Tell me what to call it and I'll call it that :)

kannon92 (dismissed):
Just a nit about the lookout name but LGTM

.github/workflows/ts.yml
kannon92: After I posted this I looked at what they are named in the actions

Lookout / ts-lint is fine

JamesMurkin:
This looks good now, so to confirm I need to remove:
 - ci/circleci: checks_and_unit_tests-ubuntu2204-go118 
 
 And enforce:
  - Go Tests / lint (1.18)
  - Python Airflow Operator / tox-tests-and-lint (1.18)
  - Python Client / tox-tests-and-lint (1.18)
  
  If so,  can do that in and merge this PR

jayofdoom:
So, in a perfect world we can require workflows instead of jobs; if that's the case we want to require everything in:
- Go Tests
- Lookout
- Python Client
- Python Airflow Operator
- ci/circleci: build_and_integration_tests (hopefully not for long!)

If we can't require by workflow, I need to rework this PR to ensure the dummy workflows have the exact same job names. 

Sorry it's not clear; I can't even see the settings so it's hard for me to grok how to set it all up.

jayofdoom:
We synced up in Slack. Github only allows requiring of jobs, not workflows, so I had to make a round trip to make the job names unique.

With the new names, I think we want the following jobs required:
- go-lint
- go-unit-tests
- ts-lint
- ts-unit-tests
- python-client-tox
- airflow-tox

This should mean only three things running not required:
- circleci integration tests
- python-client-integration-tests
- CodeQL stuff

kannon92 (approved):

PR #1362, opened 2022-08-10
WIP: add golangci lint configuration and workflow

dejanzele:


severinson (dismissed):

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,30 @@
+name: Lint & Annotate
```
kannon92, line 1:
```suggestion
name: Lint & Annotate for Golang
```

Or something like that.  Basically we have multiple languages so we should distinguish in the name that this is for the golang code.

severinson: Maybe "Lint & Annotate (Golang)"

dejanzele: done

Issue #909, opened 2022-04-01
Allow showing full jobset in Lookout jobset table


robertdavidsmith:
Currently, the jobset name gets truncated in the jobset table. We could do a few things to make it better:

Make table wider if job set names are longer, currently it only takes up a fraction of the page
Make column resizable
Make cell show full job set with scroll if a user clicks/double clicks

robertdavidsmith:
Suggest we widen the scope of this ticket to "make columns in lookout resizable and re-orderable"

rankindaveyg:
Ensure this also resolves #711

dave-gantenbein:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Armada-62c744dc3b819100132ca032/issues/G-Research/armada/909?planning-poker) with ZenHub @dejanzele @jayofdoom @kannon92 @richscott

PR #1371, opened 2022-08-11
[Lookout] Improve visibility of large jobsets on the jobset page

JamesMurkin:
- More space dedicated to the jobset column
 - Wrap on long jobset


JamesMurkin:
This will make some headway towards #909 but isn't a fully fix.

I experimented with resizable columns and it sort of works, however if the jobset is so large that you need to resize, you don't have any screen real-estate to resize into

Instead just having better default (more size + wrapping) will probably be sufficient for 90% of issues

samclark (approved):

PR #1374, opened 2022-08-12
Fix ubuntu version in circleCI machine provisioning

theAntiYeti:
This PR updates all circleCI VMs to be Ubuntu 22.04 release machines at the current latest to ensure go compilation uses go 1.18 at all stages (to fix the armadactl release pipeline).

It also changes the couple of machines that were running on `2204:current` to be version pinned on tag `2022.07.1` to ensure stability and control as circleCI updates their machines.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,7 +109,7 @@ jobs:
   build_and_integration_tests:
     machine:
       docker_layer_caching: true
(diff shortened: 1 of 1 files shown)


samclark (approved):
LGTM ... deterministic builds can only be a good thing (not using floating tags)

samclark (approved):

Issue #1388, opened 2022-08-16
JobService panics if directory for SQLite db does not already exist


richscott:
If the directory for the SQLite database that JobService uses (as specified for the key `databasePath` in `config/jobservice/config.yaml`) does not already exist, JobService will immediately panic and exit.  The panic is preceded by a misleading error that SQLite is "out of memory". 

The code should check for the existence of the directory, and if not there, it should attempt to create it. If the creation fails, it should log and exit.

kannon92:
Merged with #1389

PR #1389, opened 2022-08-16
Create dir that SQLite uses, if missing (#1388)

richscott:
If the directory for the SQLite database that JobService uses (as specified for the key `databasePath` in `config/jobservice/config.yaml`) does not already exist, create it. Without this, JobService panics and exits with a misleading message from SQLite stating "out of memory".  If the `mkdir` fails, render an error stating as such, and exit.

### DIFF
--- a/internal/jobservice/application.go
+++ b/internal/jobservice/application.go
@@ -5,6 +5,8 @@ import (
 	"database/sql"
 	"fmt"
 	"net"
+	"os"
+	"path/filepath"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jayofdoom:
Lint needs fixing, otherwise lgtm

richscott:
> Lint needs fixing, otherwise lgtm

I missed the notice - just saw the "All checks have passed" - why didn't it fail the check?

jayofdoom (approved):

dejanzele (approved):

PR #1394, opened 2022-08-17
run golangci to autofix lint issues

kannon92:
I ran golang-ci to auto fix linter issues.  

There are a lot more we need to fix manually but this is a good start to get complicance with our linter.


kannon92:
Yikes.  This wasn’t as smooth as I thought.  Going to close and aim to go about this in a smarter way.

PR #1397, opened 2022-08-18
Fix/npm vulnerability

kindyluv:
Npm Audit Vulnerability


kannon92 (changes requested):
Talking with the team, I don't think we should use overrides.  

You should upgrade the dependencies via npm install.

I'd try and update react 18 and see what happens from there.

kannon92:
Closing this PR as we will not be fixing these npm audit warnings.

PR #1407, opened 2022-08-22
Py create queues request

### LINKED ISSUE #1406: Create method `create_queue_request`
To complete #1215, we ideally want to have the queue_requests pre-created, and just pass a list of queues to the `create_queues` method that will be made as part of #1215.

This would involve creating something like the following...
```py
    def create_queue_request(
        self,
        name: str,
        priority_factor: Optional[float],
        user_owners: Optional[List[str]] = None,
        group_owners: Optional[List[str]] = None,
        resource_limits: Optional[Dict[str, float]] = None,
        permissions: Optional[List[Permissions]] = None,
    ) -> None:
        """
        Create a queue request object.

        :param name: The name of the queue
        :param priority_factor: The priority factor for the queue
        :param user_owners: The user owners for the queue
        :param group_owners: The group owners for the queue
        :param resource_limits: The resource limits for the queue
        :param permissions: The permissions for the queue
        :return: A queue request object.
        """

        permissions = [p.to_grpc() for p in permissions] if permissions else None

        return submit_pb2.Queue(
            name=name,
            priority_factor=priority_factor,
            user_owners=user_owners,
            group_owners=group_owners,
            resource_limits=resource_limits,
            permissions=permissions,
        )
```

And then re-writing all the tests, examples from...
```py
client.create_queue(name=queue, priority_factor=1)
```

To..
```py
queue_req = client.create_queue_request(queue=queue, priority_factor=1)
client.create_queue(queue_req)
```

This is more troublsome for creating single queues, but will look significantly better for building batch queues. A Future PR will cover documenting these in examples.

Sharpz7:
Closes #1406

### DIFF
--- a/client/python/armada_client/client.py
+++ b/client/python/armada_client/client.py
@@ -155,76 +155,7 @@ class ArmadaClient:
         response = self.submit_stub.ReprioritizeJobs(request)
         return response
 
-    def create_queue(
-        self,
-        name: str,
-        priority_factor: Optional[float],
-        user_owners: Optional[List[str]] = None,
-        group_owners: Optional[List[str]] = None,
-        resource_limits: Optional[Dict[str, float]] = None,
-        permissions: Optional[List[Permissions]] = None,
-    ) -> empty_pb2.Empty:
-        """Create the queue by name.
-
-        Uses the CreateQueue RPC to create a queue.
-
-        :param name: The name of the queue
-        :param priority_factor: The priority factor for the queue
-        :param user_owners: The user owners for the queue
-        :param group_owners: The group owners for the queue
-        :param resource_limits: The resource limits for the queue
-        :param permissions: The permissions for the queue
-        :return: A queue object per the Armada api definition.
-        """
-
-        permissions = [p.to_grpc() for p in permissions] if permissions else None
-
-        request = submit_pb2.Queue(
-            name=name,
-            priority_factor=priority_factor,
-            user_owners=user_owners,
-            group_owners=group_owners,
-            resource_limits=resource_limits,
-            permissions=permissions,
-        )
-        response = self.submit_stub.CreateQueue(request)
-        return response
-
-    def update_queue(
-        self,
-        name: str,
-        priority_factor: Optional[float],
-        user_owners: Optional[List[str]] = None,
-        group_owners: Optional[List[str]] = None,
-        resource_limits: Optional[Dict[str, float]] = None,
-        permissions: Optional[List[Permissions]] = None,
-    ) -> None:
-        """Update the queue of name with values in queue_params
-
-        Uses UpdateQueue RPC to update the parameters on the queue.
-
-        :param name: The name of the queue
-        :param priority_factor: The priority factor for the queue
-        :param user_owners: The user owners for the queue
-        :param group_owners: The group owners for the queue
-        :param resource_limits: The resource limits for the queue
-        :param permissions: The permissions for the queue
-        :return: None
-        """
-
-        permissions = [p.to_grpc() for p in permissions] if permissions else None
-
-        request = submit_pb2.Queue(
-            name=name,
-            priority_factor=priority_factor,
-            user_owners=user_owners,
-            group_owners=group_owners,
-            resource_limits=resource_limits,
-            permissions=permissions,
-        )
-        self.submit_stub.UpdateQueue(request)
-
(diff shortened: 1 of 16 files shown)


client/python/armada_client/client.py
```diff
@@ -224,6 +224,26 @@ def update_queue(
         )
         self.submit_stub.UpdateQueue(request)
 
+    def send_create_queue_request(self, queue: submit_pb2.Queue) -> empty_pb2.Empty:
```
kannon92, line 227: This RPC call isn't really consistent with the names of the other calls.  I'd probably just rename this to create_queue

client/python/armada_client/client.py
```diff
@@ -270,6 +290,38 @@ def unwatch_events(event_stream) -> None:
...
 
+    def create_queue_request(
+        self,
+        name: str,
+        priority_factor: Optional[float],
+        user_owners: Optional[List[str]] = None,
+        group_owners: Optional[List[str]] = None,
+        resource_limits: Optional[Dict[str, float]] = None,
+        permissions: Optional[List[Permissions]] = None,
+    ) -> None:
```
kannon92, line 301: This is not correct. Should be Queue?

client/python/armada_client/client.py
```diff
@@ -224,6 +224,26 @@ def update_queue(
...
+
+        :param queues: A list of queues to create.
+        """
+
+        response = self.submit_stub.CreateQueue(queue)
+        return response
+
+    def send_update_queue_request(self, queue: submit_pb2.Queue) -> empty_pb2.Empty:
+        """
+        Uses the CreateQueue RPC to update a queue.
```
kannon92, line 239: Comment should be updated for Update Queue

kannon92 (changes requested):
Just a few naming items, doc string issues.

I also like to see some unit tests and docs on create queue request

client/python/armada_client/client.py
Sharpz7: But we already have a method named `create_queue`. Do we want to replace it? that is a massive breaking change IMO

client/python/armada_client/client.py
Sharpz7: Yeah good spot I will fix this

client/python/armada_client/client.py
Sharpz7: Same with this

client/python/armada_client/client.py
kannon92: What is the end goal?  We want to replace create_queue with this call eventually?

We don't have any users other than airflow client so I'm not that concerned with a breaking change.  We have releases to pypi so someone could use an older release if that was a big deal.

I do like that you are thinking about it but queue creation isn't really a hotspot for our users.

Sharpz7: Well I think the end goal is to remove `create_queue` as well and soley use `create_queues`. But if this is acceptable, I will work on the replacement process now :))

client/python/armada_client/client.py
```diff
@@ -155,74 +155,25 @@ def reprioritize_jobs(
...
         Uses the CreateQueue RPC to create a queue.
 
-        :param name: The name of the queue
-        :param priority_factor: The priority factor for the queue
-        :param user_owners: The user owners for the queue
-        :param group_owners: The group owners for the queue
-        :param resource_limits: The resource limits for the queue
-        :param permissions: The permissions for the queue
-        :return: A queue object per the Armada api definition.
+        :param queues: A list of queues to create.
```
kannon92, line 162: This comment doesn't match the code.

kannon92 (dismissed):
Just one small comment usage fix for create_queue and it LGTM!

client/python/armada_client/client.py
```diff
@@ -155,74 +155,25 @@ def reprioritize_jobs(
...
-        :param group_owners: The group owners for the queue
-        :param resource_limits: The resource limits for the queue
-        :param permissions: The permissions for the queue
-        :return: None
+    def update_queue(self, queue: submit_pb2.Queue) -> empty_pb2.Empty:
         """
+        Uses the UpdateQueue RPC to update a queue.
 
-        permissions = [p.to_grpc() for p in permissions] if permissions else None
+        :param queues: A list of queues to create.
```
kannon92, line 172: Comment doesn't match the code.

jayofdoom (approved):

PR #1408, opened 2022-08-22
Lookout fix URL params conflicting with local storage, enable annotations in URL params

carlocamurri:
Previously, column filters in the Jobs table were set from both local storage and URL parameters. This causes confusion when providing URL parameters if local storage is present already.

This change makes it so the filters are only inferred from the URL, whilst retaining local storage of which columns are displayed along with auto refresh logic.

Note that this change also enables annotation filters to be represented as URL params.


internal/lookout/ui/src/services/JobsQueryParamsService.ts
```diff
@@ -98,13 +80,18 @@ export default class JobsQueryParamsService {
   }
 
   saveState(state: JobsContainerState) {
+    console.log(state.defaultColumns)
```
kannon92, line 83: Debug logging left over?

internal/lookout/ui/src/services/JobsLocalStorageService.ts
```diff
@@ -9,36 +9,44 @@ export type JobsLocalStorageState = {
   annotationColumns?: ColumnSpec<string>[]
 }
 
-function convertToLocalStorageState(loadedData: Record<string, unknown>): JobsLocalStorageState {
+function isColumnArray(data: unknown): boolean {
```
kannon92, line 12: Any objection for adding some unit test for these functions?

internal/lookout/ui/src/services/JobsLocalStorageService.ts
```diff
@@ -9,36 +9,44 @@ export type JobsLocalStorageState = {
...
+  return data !== undefined && Array.isArray(data) && data.every((col: any) => isColumnSpec(col))
+}
+
+function loadedDataIsValid(loadedData: Record<string, unknown>): boolean {
+  return (
+    loadedData.autoRefresh !== undefined &&
+    typeof loadedData.autoRefresh == "boolean" &&
+    isColumnArray(loadedData.defaultColumns) &&
```
kannon92, line 20: Are defaultColumns and annotationColumns always defined?

carlocamurri: No, the null check happens in `isColumnArray`

kannon92 (commented):
I don't really have the context to understand the bug fix.  The code and tests look good to me.  

Could you explain how to test this and I'll see if I can grok the issue and see if this fixes it?

internal/lookout/ui/src/services/JobsLocalStorageService.ts
kannon92: Thank you!

internal/lookout/ui/src/services/JobsLocalStorageService.ts
```diff
@@ -9,36 +9,44 @@ export type JobsLocalStorageState = {
...
 
-function convertToLocalStorageState(loadedData: Record<string, unknown>): JobsLocalStorageState {
+function isColumnArray(data: unknown): boolean {
+  return data !== undefined && Array.isArray(data) && data.every((col: any) => isColumnSpec(col))
+}
+
+function loadedDataIsValid(loadedData: Record<string, unknown>): boolean {
+  return (
+    loadedData.autoRefresh !== undefined &&
+    typeof loadedData.autoRefresh == "boolean" &&
```
JamesMurkin, line 19:
```suggestion
    typeof loadedData.autoRefresh === "boolean" &&
```

JamesMurkin (approved):

PR #1418, opened 2022-08-23
minimal linter without errors

kannon92:
Minimal golang-ci linter that will pass without too many changes.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -17,7 +17,7 @@ linters:
 
 linters-settings:
   lll:
-    line-length: 200
+    line-length: 261
   misspell:
     locale: UK
     ignore-words:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,42 +9,11 @@ linters:
   disable-all: true
   enable:
     - lll
-    - misspell
     - gofumpt
-    - govet
     - deadcode
     - goimports
     - ineffassign
-    - varcheck
-    - errcheck
-    - revive
-    - gosec
-    - gosimple
 
 linters-settings:
   lll:
-    line-length: 150
-  maligned:
-    suggest-new: true
-  misspell:
-    locale: US
-  revive:
-    rules:
-      - name: blank-imports
-      - name: context-as-argument
-      - name: dot-imports
-      - name: error-return
-      - name: error-strings
-      - name: error-naming
-      - name: exported
-      - name: if-return
-      - name: increment-decrement
-      - name: package-comments
-      - name: range
-      - name: receiver-naming
-      - name: indent-error-flow
-      - name: empty-block
-      - name: superfluous-else
-      - name: unused-parameter
-      - name: unreachable-code
-      - name: redefines-builtin-id
+    line-length: 261
--- a/e2e/pulsar_test/pulsar_test.go
+++ b/e2e/pulsar_test/pulsar_test.go
@@ -32,9 +32,8 @@ import (
 )
 
 // Pulsar configuration. Must be manually reconciled with changes to the test setup or Armada.
-const pulsarUrl = "pulsar://localhost:6650"
-
 const (
+	pulsarUrl            = "pulsar://localhost:6650"
 	pulsarTopic          = "persistent://armada/armada/events"
 	pulsarSubscription   = "e2e-test"
 	armadaUrl            = "localhost:50051"
--- a/e2e/pulsar_test/pulsar_test.go
+++ b/e2e/pulsar_test/pulsar_test.go
@@ -33,12 +33,15 @@ import (
 
 // Pulsar configuration. Must be manually reconciled with changes to the test setup or Armada.
 const pulsarUrl = "pulsar://localhost:6650"
-const pulsarTopic = "persistent://armada/armada/events"
-const pulsarSubscription = "e2e-test"
-const armadaUrl = "localhost:50051"
-const armadaQueueName = "e2e-test-queue"
-const armadaUserId = "anonymous"
-const defaultPulsarTimeout = 30 * time.Second
+
+const (
+	pulsarTopic          = "persistent://armada/armada/events"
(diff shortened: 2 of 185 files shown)


.golangci.yml
```diff
@@ -9,42 +9,18 @@ linters:
...
-    - gosec
-    - gosimple
+    - misspell
 
 linters-settings:
   lll:
-    line-length: 150
-  maligned:
-    suggest-new: true
+    line-length: 261
```
jayofdoom, line 20: Is this the final spot we want it? If not, might want to put a note/todo as to the desired final length

e2e/pulsar_test/pulsar_test.go
```diff
@@ -33,12 +33,15 @@ import (
 
 // Pulsar configuration. Must be manually reconciled with changes to the test setup or Armada.
 const pulsarUrl = "pulsar://localhost:6650"
```
jayofdoom, line 35: Why don't we set this const with the others? I think this was a sanity-fail by the lint autofix...

internal/armada/repository/job.go
```diff
@@ -18,15 +18,17 @@ import (
...
+	jobLeasedPrefix    = "Job:Leased:"   // {queue}            - sorted set of jobIds by lease renewal time
+	jobSetPrefix       = "Job:Set:"      // {jobSetId}         - set of jobIds
+	jobClusterMapKey   = "Job:ClusterId" //                    - map jobId -> cluster
+	jobRetriesPrefix   = "Job:Retries:"  // {jobId}            - number of retry attempts
+	jobClientIdPrefix  = "job:ClientId:" // {queue}:{clientId} - corresponding jobId
+	keySeparator       = ":"
+)
 
 // Number of jobs queried from Redis at a time in IterateQueueJobs.
 const queueResourcesBatchSize = 20000
```
jayofdoom, line 34: I'm really confused about when the autofixer chose to combine const sets and when it didn't...

internal/lookoutingester/lookoutdb/insertion.go
```diff
@@ -684,7 +673,9 @@ func conflateJobRunUpdates(updates []*model.UpdateJobRunInstruction) []*model.Up
 
 // filterEventsForCancelledJobs queries the database for any jobs that are in the cancelled state and removes them from the list of
 // instructions.  This is necessary because Armada will generate event stauses even for jobs that have been cancelled
-//  The proper solution here is to make it so once a job is cancelled, no more events are generated for it, but until
+//
```
jayofdoom, line 676: This looks like another mistake with the linter's autofix

internal/lookoutingester/lookoutdb/insertion_test.go
```diff
@@ -17,31 +17,34 @@ import (
...
-
-var baseTime, _ = time.Parse("2006-01-02T15:04:05.000Z", "2022-03-01T15:04:05.000Z")
-var updateTime, _ = time.Parse("2006-01-02T15:04:05.000Z", "2022-03-01T15:04:06.000Z")
-var startTime, _ = time.Parse("2006-01-02T15:04:05.000Z", "2022-03-01T15:04:07.000Z")
-var finishedTime, _ = time.Parse("2006-01-02T15:04:05.000Z", "2022-03-01T15:04:08.000Z")
+const (
+	jobIdString = "01f3j0g1md4qx7z5qb148qnh4r"
+	runIdString = "123e4567-e89b-12d3-a456-426614174000"
+)
+
```
jayofdoom, line 26: looks like it pulled jobIdproto (unused?) but wasn't smart enough to unify the two sets of const here

jayofdoom (commented):
A little concerned as well about the amount of dead code being removed? Are we sure no consumers are patching this downstream in a way that these removals could impact?

.golangci.yml
kannon92: Mostly I found the length that doesn't report any errors.  If I go to 150 I will have to change alot of code.  even 200 has a 10s of files that I should change. Figured could put a upper limit and fail if anything is greater.  We have 1 file that has 260 line length

internal/armada/repository/job.go
kannon92: Looks like it groups by type?

kannon92: Looks like comments turn it off for grouping.

internal/executor/utilisation/pod_utilisation.go
```diff
@@ -130,11 +130,12 @@ func (q *KubeletPodUtilisationService) RefreshUtilisationData() {
...
 /*
- We define an active pod as:
- - Any pod that is not Failed/Succeeded and doesn't have a deletion timestamp within the last inactivePodGracePeriod
-   - This rules out pods that get stuck in terminating for greater than inactivePodGracePeriod
- - Any pod that is Failed/Succeeded within the last inactivePodGracePeriod
-   - The kubelet stops reporting metrics for completed pods, just having a grace period to try catch any last metrics
+We define an active pod as:
+- Any pod that is not Failed/Succeeded and doesn't have a deletion timestamp within the last inactivePodGracePeriod
+  - This rules out pods that get stuck in terminating for greater than inactivePodGracePeriod
+
```
kannon92, line 136: Tried to change this but golangci keeps changing it back.

jayofdoom (approved):
Seems sane. I do renew my comment that someone in GR UK should review this, particularly with an eye for the code being removed. It's hard to imagine we've accumulated this much dead code, and I'd like to make sure it's not being used by downstream tooling.

.golangci.yml
jayofdoom: This works :) I was just thinking, in the future, it might be a low hanging fruit for someone to get this lowered to some point, if we agree on the end point (e.g. 150 characters?). Doesn't really matter now :)

Issue #1144, opened 2022-06-30
Clicking on annotation link in lookout should open a new tab 


kannon92:
It would be better if the links opened in a new tab, rather than replacing lookout.

_Originally posted by @robertdavidsmith in https://github.com/G-Research/armada/issues/701#issuecomment-1171469338_

When a user clicks on a link in annotations, the browser should open a new tab rather than overwriting the existing tab.

dave-gantenbein:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Armada-62c744dc3b819100132ca032/issues/G-Research/armada/1144?planning-poker) with ZenHub @dejanzele @jayofdoom @kannon92 @richscott

kannon92:
We can close with https://github.com/G-Research/armada/pull/1421

PR #1421, opened 2022-08-24
[small][Fixed The link routing issue]

kindyluv:


jayofdoom:
Hey, thanks for this.

Can we get a little more information about what the change in behavior is? The original behavior, how you fixed it, and the new behavior?

Additionally, it appears your change has broken the typescript linting and tests. Those failures will need to be addressed as well before this can be reviewed.

kannon92:
> Hey, thanks for this.
> 
> Can we get a little more information about what the change in behavior is? The original behavior, how you fixed it, and the new behavior?
> 
> Additionally, it appears your change has broken the typescript linting and tests. Those failures will need to be addressed as well before this can be reviewed.

Hey @jayofdoom!

@kindyluv has been working on #1144.  This is an issue focused on when someone clicks an annotation that is a URL, it will open a new tab in your browser.  

I suggested that she create a simple POC using just a dummy react application so we can verify that this works without all the armada nonsense.  

She has created https://github.com/kindyluv/LinkRouting which is the POC for testing this out.  I was able to build and test and it does in fact create a browser in a new window!

jayofdoom:
That's super cool, thanks for letting me know Kevin!

kannon92 (approved):
I tested this and it works!

PR #1423, opened 2022-08-24
Added batch queue example

Sharpz7:
Closes #https://github.com/G-Research/armada/issues/1420

Adds example code for:
- Creating or Updating a batch of queues.
- Cancelling a Job-Set with a filter


kannon92 (approved):

client/python/examples/queues.py
```diff
@@ -9,11 +9,9 @@
 from armada_client.permissions import Permissions, Subject
 
 
-def creating_full_queues_example(client, queue):
+def create_queue_request(client, queue):
```
jayofdoom, line 12: any reason we aren't putting types in here, like we do for the code? (throughout)

jayofdoom (dismissed):

client/python/examples/queues.py
Sharpz7: Python is a dynamic language. I just feel that we do not want to enforce the typing standards we set for Python Armada Client on the people that use it.

client/python/examples/cancelling.py
```diff
@@ -103,9 +126,9 @@ def quick_create_queue(client, queue):
 def workflow():
     """
     Starts a workflow, which includes:
-        - Creating a queue
-        - Creating a jobset
-        - Creating a job
+        - Creating a queue a job
```
jayofdoom, line 129: missing an and? Or something, this doesn't make sense as written

client/python/examples/queues.py
jayofdoom: I can see that, but type hints don't break non-type hinted code. Either way, I'll trust your judgement here.

jayofdoom (commented):
Please fix the one place with the nonsensical comment

kannon92 (approved):

client/python/examples/cancelling.py
Sharpz7: Yeah this needs fixed. On it shortly

kannon92 (approved):

richscott (approved):

PR #1427, opened 2022-08-26
Add Kubernetes native authentication flow

theAntiYeti:


internal/common/auth/configuration/types.go
```diff
@@ -53,3 +54,7 @@ type LDAPConfig struct {
 	GroupSearchBase string
 	CacheExpiry     time.Duration
 }
+
+type KubernetesAuthConfig struct {
```
kannon92, line 58: Could you include a comment here to explain what the KidMappingFileLocation is?

pkg/client/auth/kubernetes/authentication.go
```diff
@@ -0,0 +1,116 @@
...
+			// Make HTTPs Request to TokenRequest endpoint
+			request, err := http.NewRequest("POST", url, bytes.NewBuffer([]byte(data)))
+			if err != nil {
+				return nil, err
+			}
+			request.Header.Add("Authorization", "Bearer "+localKubernetesToken)
+			request.Header.Add("Content-Type", "application/json; charset=utf-8")
+
+			tr := &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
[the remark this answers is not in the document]
kannon92: This looks serious.  CodeQL seems to flag this in a few places.

kannon92 (changes requested):
Thanks for the PR!  I think CodeQL is yelling at us in a few places about the insecureVerify flag.  

I would also like to ask that you add a auth readme that explains how to set this up.  I'd be happy with a simple config file if possible? 

What are valid values for a config with this Auth?

Do we need OIDC setup separately from Kubernetes in order for this to work?  

If you are pressed for time, please file a ticket to document how to use this authentication flow.

theAntiYeti:
This PR is still WIP

internal/common/auth/authorization/kubernetes.go
```diff
@@ -0,0 +1,217 @@
...
+	log "github.com/sirupsen/logrus"
+	authv1 "k8s.io/api/authentication/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	"github.com/G-Research/armada/internal/common/auth/configuration"
+)
+
+// TODO: Should include a cache
```
kannon92, line 23: Comment should be removed since it looks like you added a token cache?

docs/auth.md
```diff
@@ -0,0 +1,65 @@
+# Kubernetes Native Authentication
```
kannon92, line 1: This is beautiful!

internal/common/auth/authorization/kubernetes.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	// Get token time
+	expirationTime, err := parseTime(token)
+	if err != nil {
+		return nil, err
+	}
+
+	if time.Now().After(*expirationTime) {
+		return nil, fmt.Errorf("invalid token, expired")
```
kannon92, line 63: Nit: You don't need Errorf here since you have no variable.

internal/common/auth/authorization/kubernetes.go
```diff
@@ -0,0 +1,206 @@
...
+			Valid: true,
+		},
+		expirationTime.Sub(time.Now()))
+
+	// Return very basic Principal
+	return NewStaticPrincipal(name, []string{name}), nil
+}
+
+func (authService *KubernetesNativeAuthService) getClusterURL(token string) (string, error) {
+	log.Infof("Token: %s", token)
```
kannon92, line 104: Should we log these things?  This seems like a security vulnerability?

pkg/client/auth/kubernetes/token_credentials.go
```diff
@@ -0,0 +1,55 @@
...
+	encodedCa := base64.RawURLEncoding.EncodeToString([]byte(ca))
+	body := fmt.Sprintf(`{"token":"%s", "ca":"%s"}`, jwt, encodedCa)
+	encoded := base64.RawURLEncoding.EncodeToString([]byte(body))
+
+	return map[string]string{
+		"authorization": "KubernetesAuth " + encoded,
+	}, nil
+}
+
+func (c *NativeTokenCredentials) RequireTransportSecurity() bool {
```
kannon92, line 36: Is this a TODO to enable at some point?

pkg/client/auth/kubernetes/token_credentials.go
```diff
@@ -0,0 +1,55 @@
...
+func (c *NativeTokenCredentials) getJWT(source oauth2.TokenSource) (string, error) {
+	t, e := source.Token()
+	if e != nil {
+		return "", e
+	}
+	return t.AccessToken, nil
+}
+
+func getClusterCA() (string, error) {
+	fromFile, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt")
```
kannon92, line 49: I would probably put this as a config.  Will all K8 environments that potential run armada store their ca in this file location?

docs/auth.md
```diff
@@ -0,0 +1,65 @@
...
+The Armada Server must have a ConfigMap in its namespace with the following
+format for entries:
+
+```yaml
+data:
+  "<CLUSTER_KID>": "<EXECUTOR_CLUSTER_CALLBACK_URL>"
+  ...
+```
+
+This ConfigMap may be mounted anywhere on the Server's Pod.
```
kannon92, line 32: This seems to be a implementation detail of how Armada stores ConfigMaps.  

Is it required that a configmap is stored as a volume or can someone create a configmap as a k8 object?

internal/common/auth/authorization/kubernetes.go
kannon92: You do this in quite a few places.  No reason to change it.

kannon92 (commented):
I have some small items to fix.  I don't really feel comfortable giving a Yay or Nay for this PR as I don't understand how GR handles Auth internally.

docs/auth.md
theAntiYeti: :heart:

internal/common/auth/authorization/kubernetes.go
theAntiYeti: I don't believe there's an equivalent to Errorf without formatting. Do let me know if this is wrong

internal/common/auth/authorization/kubernetes.go
theAntiYeti: We indeed should not be logging these things, am in the process of taking them out

pkg/client/auth/kubernetes/token_credentials.go
theAntiYeti: No, you'll find this to be the same in the equivalent object in the oidc package in common.go

I believe in this case transportCredentials actually is just a euphemism for client TLS certs

docs/auth.md
theAntiYeti: wdym create a configmap as a k8s object?
The configmap needs to be mounted as this is what allows the pod to read updates without having to rotate the pod.
There are alternatives to how you get configmap info into a pod but they usually do not rotate except with the pod.

pkg/client/auth/kubernetes/token_credentials.go
theAntiYeti: I believe yes but that doesn't mean you're wrong to want to put it as a config. I'll keep this conversation unresolved for now.

docs/auth.md
kannon92: Didn't know that

theAntiYeti:
This change has now been tested in dev and is now ready for review

docs/auth.md
theAntiYeti: There are four different ways that you can use a ConfigMap to configure a container inside a Pod:

    Inside a container command and args
    Environment variables for a container
    Add a file in read-only volume, for the application to read
    Write code to run inside the Pod that uses the Kubernetes API to read a ConfigMap

Source: https://kubernetes.io/docs/concepts/configuration/configmap/
As I see it 1 and 2 are clearly fixed after first read, 3 updates when the configmap does (though this is controlled by kubernetes and I don't believe is instant, though fast enough for our uses). 4 also clearly updates with the configmap but is the least abstracted way of doing this so we definitely shouldn't be relying on it.

go.mod
```diff
@@ -83,6 +83,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require github.com/patrickmn/go-cache v2.1.0+incompatible
```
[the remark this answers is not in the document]
theAntiYeti: This was discussed with Sam, I think it's been placed here by `go mod tidy` due to the +incompatible flag

As this was done automatically rather than manually we decided to leave it.

internal/common/auth/authorization/kubernetes.go
```diff
@@ -0,0 +1,203 @@
...
+
+	ca, err := base64.RawURLEncoding.DecodeString(uMbody.Ca)
+	if err != nil {
+		return "", "", err
+	}
+
+	return uMbody.Token, string(ca), nil
+}
+
+func parseTime(token string) (*time.Time, error) {
```
d80tb7, line 182: I wouldn't return a pointer here- but rather a struct.

internal/common/auth/authorization/kubernetes.go
```diff
@@ -0,0 +1,203 @@
...
+		return nil, missingCredentials
+	}
+
+	// Get token time
+	expirationTime, err := parseTime(token)
+	if err != nil {
+		return nil, err
+	}
+
+	if time.Now().After(*expirationTime) {
```
d80tb7, line 61: I've put a comment lower down to say that parseTime should return a struct rather than a pointer, but if you decide not to do that then I'd add a nil check here.

internal/common/auth/authorization/kubernetes.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if err := json.Unmarshal(decoded, &unmarshalled); err != nil {
+		return "", err
+	}
+
+	if unmarshalled.Kid == "" {
+		return "", fmt.Errorf("kubernetes serviceaccount token KID must not be empty")
+	}
+
+	url, err := os.ReadFile(authService.KidMappingFileLocation + unmarshalled.Kid)
```
d80tb7, line 121: this is probably overkill, but could someone trick you into reading random files on the filesystem by sending a `kid` of the form `../../path/to/random/file`

pkg/client/auth/kubernetes/token_credentials.go
```diff
@@ -0,0 +1,55 @@
...
+	TokenSource oauth2.TokenSource
+}
+
+func (c *NativeTokenCredentials) GetRequestMetadata(context.Context, ...string) (map[string]string, error) {
+	jwt, err := c.getJWT(c.TokenSource)
+	if err != nil {
+		return nil, err
+	}
+
+	ca, err := getClusterCA()
```
d80tb7, line 22: are we doing this on every request?  Does it make any measurable difference to performance?

d80tb7: I say this because we could obviously cache the encoded ca in memory, if only for a short time.

internal/common/auth/authorization/kubernetes.go
theAntiYeti: I see what you're saying, do you feel there should be some validation checking on the string format of the kid? ensure it doesn't contain slashes of any form?

pkg/client/auth/kubernetes/token_credentials.go
theAntiYeti: I don't believe it will have a major difference on performance, I can look into caching it in memory though

d80tb7 (approved):

PR #1428, opened 2022-08-26
Add log-based scheduler

severinson:
This PR introduces a new scheduler to Armada. This scheduler is backed by postgres and derives its internal state entirely from Pulsar messages, with the exception of node info, which it receives directly from executors.

This is a first step towards replacing the current scheduler. The scheduler added with this PR works, in the sense that it passes testsuite tests, but it doesn't have feature parity with the existing scheduler yet.

Specifically, the new scheduler consists of:

- Postgres database (for storing queued jobs and node info)
- Ingester (responsible for populating postgres)
- API (the set of endpoints on which executors request jobs)
- Scheduling logic (responsible for generating leases)

Of these, the database, ingester and API are fairly mature; I don't expect any major changes to be necessary. The scheduling logic is a prototype; it just assigns jobs to executors in a round-robin fashion.


internal/armada/server.go
```diff
@@ -330,18 +340,137 @@ func Serve(ctx context.Context, config *configuration.ArmadaConfig, healthChecks
...
+
+		// Scheduler jobs ingester.
+		schedulerIngester := &eventscheduler.Ingester{
+			PulsarClient: pulsarClient,
+			ConsumerOptions: pulsar.ConsumerOptions{
+				Topic:            config.Pulsar.JobsetEventsTopic,
+				SubscriptionName: "pulsar-scheduler-ingester",
+				Type:             pulsar.KeyShared,
+			},
+			MaxWriteInterval: 10 * time.Second,
```
d80tb7, line 381: this seems quite a long time- why is this?

internal/armada/server.go
```diff
@@ -330,18 +340,108 @@ func Serve(ctx context.Context, config *configuration.ArmadaConfig, healthChecks
...
+		schedulerIngester := &scheduler.Ingester{
+			PulsarClient: pulsarClient,
+			ConsumerOptions: pulsar.ConsumerOptions{
+				Topic:            config.Pulsar.JobsetEventsTopic,
+				SubscriptionName: "pulsar-scheduler-ingester",
+				Type:             pulsar.KeyShared,
+			},
+			MaxWriteInterval: 10 * time.Second,
+			MaxWriteRecords:  10000,
+			Db:               pool,
```
d80tb7, line 354: It looks like pool can be nil (there's a check in the deduplication code above for this) but Pulsar can be enabled.  In that case we'll pass a nil pool down to the scheduler.  How about we remove the feature flag for pulsar and assert that you have to provide postgres details.

internal/armada/server.go
```diff
@@ -330,18 +340,108 @@ func Serve(ctx context.Context, config *configuration.ArmadaConfig, healthChecks
...
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		pulsarExecutorApiServer = &scheduler.ExecutorApi{
+			Producer:       apiProducer,
+			Db:             pool,
+			MaxJobsPerCall: 100,
+		}
+
+		// // This producer does not need a name; having multiple producers is safe.
```
d80tb7, line 391: what's all this commented out code here?

internal/armada/server/submit_to_log.go
```diff
@@ -173,17 +173,27 @@ func (srv *PulsarSubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmi
...
 
 	// Check if the job can be scheduled on any executor,
-	// to avoid having users wait for a job that may never be scheduled
-	allClusterSchedulingInfo, err := srv.SubmitServer.schedulingInfoRepository.GetClusterSchedulingInfo()
-	if err != nil {
-		err = errors.WithMessage(err, "error getting scheduling info")
-		return nil, err
+	// to avoid having users wait for a job that may never be scheduled.
+	//
+	// We only perform this check for jobs submitted to the legacy scheduler.
```
d80tb7, line 178: I'd remove the blank line (and the full stop) here!

internal/armada/server/submit_to_log.go
```diff
@@ -173,17 +173,27 @@ func (srv *PulsarSubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmi
...
+	legacySchedulerJobs := make([]*api.Job, 0, len(apiJobs))
+	for _, apiJob := range apiJobs {
```
d80tb7, line 180: we could extract this out into a helper function as we have basically the same logic in `submit_from_log`

internal/common/eventutil/eventid/eventid.go
```diff
@@ -0,0 +1,89 @@
...
+		id.pulsarMessageId.PartitionIdx(),
+		id.pulsarMessageId.BatchIdx(),
+		id.index,
+	)
+}
+
+// Greater returns true if id occurred after other, or an error if the ids are not comparable
+// (i.e., if they are from different partitions).
+func (id *EventId) Greater(other *EventId) (bool, error) {
+	isGreater, err := id.pulsarMessageId.Greater(other.pulsarMessageId)
```
d80tb7, line 41: this will fail if either `id` or `other` is `nil`. This is probably what you want but I'd maybe throw a clear error if either is nil.

internal/pulsarutils/pulsar_to_channel.go
```diff
@@ -0,0 +1,35 @@
+package pulsarutils
+
+import (
+	"context"
+
+	"github.com/severinson/pulsar-client-go/pulsar"
+)
+
+// PulsarToChannel is a service for receiving messages from Pulsar and forwarding those on C.
```
d80tb7, line 9: It's not entirely clear why we need this- the client already lets you supply a channel that messages can be published to?

internal/scheduler/compaction.go
```diff
@@ -0,0 +1,126 @@
+package scheduler
```
d80tb7, line 1: Is this file in a usable state? The commented out code means that the structs don't comform to the interface and the todos mean that the implementations are not thread safe.

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+			// which are passed on to sub-functions.
+			messageCtx, ok := requestid.AddToIncomingContext(ctx, requestId)
+			if !ok {
+				messageCtx = ctx
+			}
+			messageLogger := log.WithFields(logrus.Fields{"messageId": msg.ID(), requestid.MetadataKey: requestId})
+			ctxWithLogger := ctxlogrus.ToContext(messageCtx, messageLogger)
+
+			// Any consumer can seek on any topic partition.
+			// An unexpected seek may cause messages to be delivered out of order,
```
d80tb7, line 240: What do you mean here by "An unexpected seek?"

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+			}
+			srv.lastMessageIdByPartition[partitionIdx] = msg.ID()
+
+			var err error
+			batches, err = srv.ProcessMessage(ctxWithLogger, msg, batches)
+			if err != nil {
+				logging.WithStacktrace(messageLogger, err).Warnf("processing message failed; ignoring")
+			}
+
+			// If we have more than 1 batch, apply all but the last immedtately.
```
d80tb7, line 274: minor (sp)

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+	// Ticker to trigger periodically writing to postgres.
+	ticker := time.NewTicker(5 * time.Second)
+	defer ticker.Stop()
+
+	// Run until ctx is canceled.
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case msg := <-messageChannel:
```
d80tb7, line 224: If I'm honest I'm not a massive fan of this because it mixes in the batching logic with other service-specific logic.    Now that we have generics we should be able to write a generic batching function and then write service-specific functions to operate on either a batch or single element as required.

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+			}
+			if lastMessageId, ok := srv.lastMessageIdByPartition[partitionIdx]; ok {
+				msgIsOufOfOrder, err := pulsarutils.FromMessageId(lastMessageId).GreaterEqual(msg.ID())
+				if err != nil {
+					return err
+				}
+				if msgIsOufOfOrder {
+					// pulsarutils.PulsarMessageId prints nicely, so we convert to those.
+					return fmt.Errorf(
+						"unexpected seek detected: received messages out of order for topic %s: id of received message is %s, but the previous message has id %s",
```
d80tb7, line 261: what happens to acks here?

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+						msg.Topic(), pulsarutils.FromMessageId(msg.ID()), pulsarutils.FromMessageId(lastMessageId),
+					)
+				}
+			}
+			srv.lastMessageIdByPartition[partitionIdx] = msg.ID()
+
+			var err error
+			batches, err = srv.ProcessMessage(ctxWithLogger, msg, batches)
+			if err != nil {
+				logging.WithStacktrace(messageLogger, err).Warnf("processing message failed; ignoring")
```
d80tb7, line 271: Do you want a `continue` here? I guess it'll ptrobably work because `batches` will probably be nil if there' s an errr here, but I think it'd be safer not to use a result from a function that's returned an error.

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+	// Store the message id in the batch.
+	// So we can ack messages once they've been written to postgres.
+	//
+	// It's fine to write to postgres and then not ack, since writes are idempotent.
+	// But it's not fine to ack messages not written to postgres,
+	// since skipping messages isn't detected.
+	//
+	// We defer adding the message id to ensure it's added to the last batch created
+	// from this sequence.
+	defer func() {
```
d80tb7, line 334: If I'm honest I'm not a fan of using a `defer` to modify the return value of the function after the return call.  I guess it's the principle of least surprise in that people don't expect their return object to be modified after the event.

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+			if e.SubmitJob.Scheduler != "pulsar" {
+				continue
+			}
+
+			// If there are job set operations to be applied,
+			// we can't add new jobs for any affected job sets to this batch.
+			// Since jobs submitted after those operations should not be affected.
+			// To that end, we create new batches as necessary here.
+			if _, ok := batch.JobSetsCancelled[sequence.GetJobSetName()]; ok {
+				batches = append(batches, NewBatch())
```
d80tb7, line 363: this is a minor thing but isn't it clearer to do:
```
batch = NewBatch()
batches = append(batches, batch)
```
This has the advantage of a) fewer operations and b) each line does only one thing.

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+				batch = batches[len(batches)-1]
+			}
+			for _, reprioritisation := range batch.Reprioritisations {
```
d80tb7, line 366: we could add:
```
func(*Batch) hasReprioritisation(jobSet string)
```
which could then turn this code into:
```
if batch.hasReproritisation(sequence.GetjobSetName){
  batch = NewBatch()
  batches = append(batches, batch)
}
```
The reason I say this is that this function is quite long and complex so anything we can do to shorten it is probably a good thing.

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+				if _, ok := reprioritisation.PrioritiesByJobSet[sequence.GetJobSetName()]; ok {
+					batches = append(batches, NewBatch())
+					batch = batches[len(batches)-1]
+					break
+				}
+			}
+
+			// Store the job submit message so that it can be sent to an executor.
+			submitJobBytes, err := proto.Marshal(e.SubmitJob)
```
d80tb7, line 375: How do we see this working long term.  I *think* what happens at present is that everything that's been processed so far in the batch is ignored but everything that comes after in the batch is thrown away.  This seems a bit weird.

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+		JobRunsRunning:    make(map[uuid.UUID]bool),
+		JobRunsSucceeded:  make(map[uuid.UUID]bool),
+	}
+}
+
+// Batch of reprioritisations.
+// PrioritiesByJobSet should always be applied before PrioritiesByJob.
+// ReprioritisationBatch
+// To ensure changes are applied in the correct order.
+type ReprioritisationBatch struct {
```
d80tb7, line 131: I'd be tempted to rename this `Reprioritisations` or something.  From the names I initially assumed this would be a `Batch` implementation

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+		case *armadaevents.EventSequence_Event_JobSucceeded:
+			jobId := armadaevents.UuidFromProtoUuid(e.JobSucceeded.GetJobId())
+			batch.JobsSucceeded[jobId] = true
+		case *armadaevents.EventSequence_Event_JobErrors:
+			eventId := eventid.New(msg.ID(), i).String()
+			for j, jobError := range e.JobErrors.GetErrors() {
+				bytes, err := proto.Marshal(jobError)
+				if err != nil {
+					err = errors.WithStack(err)
+					logging.WithStacktrace(log, err).Error("failed to marshal JobError")
```
d80tb7, line 449: I'd be tempted to set set `bytes` to nil here on the grounds that if a function returns an error I wouldn't use the value.

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+					Error:    bytes,
+					Terminal: jobError.GetTerminal(),
+				})
+			}
+		case *armadaevents.EventSequence_Event_JobRunAssigned:
+			runId := armadaevents.UuidFromProtoUuid(e.JobRunAssigned.GetRunId())
+			bytes, err := proto.Marshal(e.JobRunAssigned)
+			if err != nil {
+				err = errors.WithStack(err)
+				logging.WithStacktrace(log, err).Error("failed to marshal JobRunAssigned")
```
d80tb7, line 465: as above- I'd set `bytes` to nil

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+		case *armadaevents.EventSequence_Event_JobRunSucceeded:
+			jobId := armadaevents.UuidFromProtoUuid(e.JobRunSucceeded.GetJobId())
+			batch.JobRunsSucceeded[jobId] = true
+		case *armadaevents.EventSequence_Event_JobRunErrors:
+			eventId := eventid.New(msg.ID(), i).String()
+			for j, jobRunError := range e.JobRunErrors.GetErrors() {
+				bytes, err := proto.Marshal(jobRunError)
+				if err != nil {
+					err = errors.WithStack(err)
+					logging.WithStacktrace(log, err).Error("failed to marshal JobRunError")
```
d80tb7, line 483: again- let's set `bytes` to nil

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+			})
+			if err != nil {
+				return errors.WithStack(err)
+			}
+		}
+
+		// TODO: This will be slow if there's a large number of ids.
+		// Could be addressed by using a separate table for priority + upsert.
+		for jobId, priority := range reprioritisation.PrioritiesByJob {
+			err := queries.UpdateJobPriorityById(ctx, UpdateJobPriorityByIdParams{
```
d80tb7, line 566: I think a simple thing here is just to update all priorities in a single statement rather than doing them one by one by one  I'd think we'd need to fix this pretty soon because this will be extremely slow and current workflows do cause large number of reprios to get generated

internal/scheduler/ingester.go
```diff
@@ -0,0 +1,777 @@
...
+		records[i] = jobError
+		if jobError.Terminal {
+			failedJobIds = append(failedJobIds, jobError.JobID)
+		}
+	}
+	err = Upsert(ctx, srv.Db, "job_errors", JobErrorsSchema(), records)
+	if err != nil {
+		return err // TODO: Keep retrying on transient failures.
+	}
+	log.Infof("wrote %d errors into postgres", len(records))
```
d80tb7, line 624: why have we go log lines for this but not other operations?

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,302 @@
...
+	executors := srv.getActiveExecutors()
+	if len(executors) == 0 {
+		if len(srv.Executors) == 0 {
+			return errors.New("no executors available")
+		} else {
+			return errors.Errorf("all executors inactive: %+v", srv.Executors)
+		}
+	}
+
+	// Assign jobs in the queue to executors in a round-robin fashion.
```
d80tb7, line 254: I presume this is a trivial scheudling algorithm for now (not taking into account priority, capacity etc)?  If so we should put a comment in

internal/scheduler/unused/scheduler.go
```diff
@@ -0,0 +1,415 @@
...
+	item.priority = priority
+	heap.Fix(pq, item.index)
+}
+
+// scheduleForever continually schedules jobs.
+func (s *Scheduler) scheduleForever() {
+	for {
+		err := s.scheduleOne()
+		if err != nil {
+			fmt.Printf("error: %s\n", err)
```
d80tb7, line 249: shouldn't this be logging?

internal/scheduler/unused/scheduler.go
```diff
@@ -0,0 +1,415 @@
...
+	return pq[i].priority < pq[j].priority
+}
+
+func (pq JobPQ) Swap(i, j int) {
+	pq[i], pq[j] = pq[j], pq[i]
+	pq[i].index = i
+	pq[j].index = j
+}
+
+func (pq *JobPQ) Push(x interface{}) {
```
d80tb7, line 221: can this be simplified with `container/heap`?

internal/scheduler/unused/scheduler.go
```diff
@@ -0,0 +1,415 @@
...
+	if err != nil {
+		return err
+	}
+	fmt.Printf("suggested lease: %s\n", lease)
+	return nil
+}
+
+// selectQueue returns the highest priority queue from the scheduler priority queue.
+// Does not remove the queue from the priority queue.
+func (s *Scheduler) selectQueue() (*Queue, error) {
```
d80tb7, line 277: I'd be tempted to rename this `PeekQueue` and return nil rather than an error if nothing available. The caller can then check the result.

internal/scheduler/unused/scheduler.go
```diff
@@ -0,0 +1,415 @@
...
+	return q.Pq[0], nil
+}
+
+// suggestLease returns a suggested scheduling decision (i.e., a lease), or an error.
+// The caller is responsible for acting on that decision (or not) and updating the scheduler data structures accordingly.
+func (s *Scheduler) suggestLease(queue *Queue, job *Job) (*Lease, error) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	if len(s.clusters) == 0 {
+		err := errors.Errorf("no clusters available")
```
d80tb7, line 301: It feels a bit wrong to use errors to handle expected conditions (although maybe this is a go thing).

internal/scheduler/upsert.go
```diff
@@ -0,0 +1,323 @@
...
+	return db.BeginTxFunc(ctx, pgx.TxOptions{
+		IsoLevel:       pgx.ReadCommitted,
+		AccessMode:     pgx.ReadWrite,
+		DeferrableMode: pgx.Deferrable,
+	}, func(tx pgx.Tx) error {
+		return CopyProtocolUpsert(ctx, tx, tableName, schema, records)
+	})
+}
+
+func IdempotencyCheck(ctx context.Context, tx DBTX, topicName string, writeMessageIds map[int32]pulsar.MessageID) error {
```
d80tb7, line 124: What happens if this check fails?

d80tb7: More specifically what happens if this check fails- but some of the messageIds are greater than the last storedid and thus need to be applied.

internal/scheduler/upsert.go
```diff
@@ -0,0 +1,323 @@
...
+			Topic:       topicName,
+			StaleWrites: staleWrites,
+		}
+	}
+
+	// Otherwise, we have more recent data than what is already stored in the db that we should write.
+	// Update the message id in postgres to reflect the data to be written.
+	//
+	// TODO: Add a call to insert these in a single call.
+	for partitionIdx, writeMessageId := range writeMessageIds {
```
d80tb7, line 166: Are we sure this will work?  Having read the pulsar docs a bit more, I'm not convinced that that each topic is served by exactly one consumer for a keyShared subscription.  

e.g. from [here](https://streamnative.io/en/blog/engineering/2021-08-25-scalable-stream-processing-with-pulsars-key-shared-subscription/)
```
Pulsar achieves these guarantees by hashing the incoming key values and distributing the hashes evenly across all of the consumers on the subscription. Thus, we know that messages with the same key will generate the same hash value and consequently get sent to the same consumer as the previous messages with that key.

By ensuring that all messages with the same key are sent to the same consumer, that consumer is guaranteed to receive all of the messages for a given key and in the order they were received, which matches the streaming consumption semantics. Let’s explore a real-world use case where Pulsar’s Key_Shared subscription could be used effectively
```
 and
 ```
 Pulsar’s Key_Shared subscription type allows you to have both high throughput and stateful processing on a single topic.
 ```
  This implies to me that sharing of keys between subscribers isn't limited to the simple solution of having each subscriber listen to a subset of the partitions.

internal/scheduler/upsert.go
```diff
@@ -0,0 +1,323 @@
...
+		pgx.CopyFromSlice(len(records), func(i int) ([]interface{}, error) {
+			// TODO: Are we guaranteed that values always come in the order listed in the record? Otherwise we need to control the order.
+			_, values := NamesValuesFromRecord(records[i])
+			return values, nil
+		}),
+	)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	if n != int64(len(records)) {
```
d80tb7, line 214: Is this worth checking?  It's not clear to me how this could ever occur given the guarantees given by the db?

internal/scheduler/upsert.go
```diff
@@ -0,0 +1,323 @@
...
+		return errors.Errorf("only %d out of %d rows were inserted", n, len(records))
+	}
+
+	// Move those rows into the main table, using ON CONFLICT rules to over-write existing rows.
+	var b strings.Builder
+	fmt.Fprintf(&b, "INSERT INTO %s SELECT * from %s ", tableName, tempTableName)
+	fmt.Fprintf(&b, "ON CONFLICT (%s) DO UPDATE SET ", names[0])
+	for i, name := range names {
```
d80tb7, line 222: you might need to escape things here e.g. if `name` is some sql/postgres keyword

internal/scheduler/upsert.go
severinson: I think we may as well keep it. It doesn't cost us anything.

internal/scheduler/upsert.go
severinson: Because we get names from the sql schema, it should be fine here. But if we have escaping functions, we may as well use them.

internal/pulsarutils/pulsar_to_channel.go
d80tb7: I'm not sure- I guess the question is whether it logs errors and reconnects as expected.   If it does, what else would we do with the error? Maybe prometheus metrics?

internal/scheduler/query.sql.go
```diff
@@ -0,0 +1,868 @@
...
+	"context"
+
+	"github.com/google/uuid"
+)
+
+const getTopicMessageIds = `-- name: GetTopicMessageIds :many
+SELECT topic, ledger_id, entry_id, batch_idx, partition_idx FROM pulsar WHERE topic = $1
+`
+
+func (q *Queries) GetTopicMessageIds(ctx context.Context, topic string) ([]Pulsar, error) {
```
d80tb7, line 18: let's delete this (and other pulsar-db stuff)

severinson: Agreed. Deleted.

d80tb7 (approved):

PR #1439, opened 2022-08-30
Tooling for easily running full Armada stack locally

richscott:
This PR is a set of changes to enable newcomers (or even experienced Armada devs) quickly get the full suite of Armada services running locally, so they can do development on components like Lookout, etc.

The changes are mostly in 3 places:
- the makefile now has a number of `build-dev-xxxx` targets, to build Docker images for each of the constituent services with config files baked in that have settings appropriate for running locally, using docker-compose. There is a single overall `armada-dev` target that can be used to build all the images in one invocation. (It takes about 3 minutes to run on a c3.xlarge EC2 instance.)
- `docs/dev/armada-compose.yaml` has been introduced - this is sort of a superset of the existing `docs/dev/docker-compose.yaml`, but now also has the Armada services as well as the underlying infra services. It specifies config settings/files appropriate for running/testing the stack locally. The services all have host values specified, so networking between them is easy.  Several of the services used `depends_on` in order to ensure pre-requisite services are running first.
- `docs/dev/setup.sh` has been overhauled - it now checks for the existence of a Kind cluster controller, and starts it if necessary, then creates a `.env` file appropriate for use by the Job Service dev container (needed for UID/GID setting, so the container can write to a sqlite file in the host directory), and shows how to invoke docker-compose to start and run the services.

I expect there will be considerably more refinement of these scripts/configs, for easy use in CI/CD situations, automated testing, etc. - this is mostly intended for use by interns/newcomers in getting the Armada stack running locally.

To try this out:
- create a `.env` file in the root of the Armada repo, containing two lines: `APP_UID=nnnn` and `APP_GID=nnnn` (values taken from running `id -u` and `id -g`).
- run `make armada-dev` to build the dev containers
- run `bash docs/dev/setup.sh` and follow any instructions it gives.


docs/dev/config/armada/base.yaml
```diff
@@ -17,3 +17,14 @@ auth:
     execute_jobs: ["everyone"]
 pulsar:
   enabled: false
+  URL: "pulsar://pulsar:6650"
```
kannon92, line 20: If pulsar is disabled do we need this?

docs/dev/config/lookout/config.yaml
```diff
@@ -0,0 +1,44 @@
...
+  keepaliveParams:
+    maxConnectionIdle: 5m
+    time: 2h
+    timeout: 20s
+  keepaliveEnforcementPolicy:
+    minTime: 5m
+    permitWithoutStream: false
+
+uiConfig:
+  armadaApiBaseUrl: "http://localhost:8080"
```
kannon92, line 15: I don't think this is correct. Shouldn't we use container name for armada server or lookout?

docs/dev/config/lookout/config.yaml
```diff
@@ -0,0 +1,44 @@
...
+  userAnnotationPrefix: "armadaproject.io/"
+  binocularsEnabled: true
+  binocularsBaseUrlPattern: "http://localhost:8082" # {CLUSTER_ID} gets replaced by appropriate cluster id
```
kannon92, line 18: Same here. Should use binoculars

kannon92 (commented):
Small nits.  Will test tomorrow!

docs/dev/setup.sh
```diff
@@ -1,34 +1,60 @@
...
-if echo $DCLIENT_VERSION | grep -q '^2' ; then
+docker compose > /dev/null 2>&1
+if [ $? -eq 0 ] ; then
   COMPOSE_CMD='docker compose'
 fi
 
-kind create cluster --name demo-a --config ./docs/dev/kind.yaml
-
-OSTYPE=$(uname -s)
 if [ $OSTYPE == "Linux" ]; then
```
kannon92, line 13: You comment out pulsar in the compose file so I think the profile flag does nothing.

docs/dev/config/lookout/config.yaml
kannon92: I think this works somehow!

docs/dev/config/lookout/config.yaml
kannon92: Binoculars doesn't seem to work but unclear if that is due to Fake executor.  I think we can tackle this as a followup since binoculars isn't that important to us at this moment.

docs/dev/setup.sh
```diff
@@ -1,34 +1,60 @@
...
-echo "Using $stream_backend"
-
+OSTYPE=$(uname -s)
 COMPOSE_CMD='docker-compose'
+CLUSTER_NAME="demo-a"
+COMPOSE_FILE="./docs/dev/armada-compose.yaml"
 
-DCLIENT_VERSION=$(docker version -f '{{.Client.Version}}')
-if echo $DCLIENT_VERSION | grep -q '^2' ; then
+docker compose > /dev/null 2>&1
```
kannon92, line 8: Why is this here?

kannon92 (approved):
I tested this locally and it works pretty well.  I have a few nits about the script and some issues I see.  

Its up to you if you want to address!

kannon92 (changes requested):
We should change teardown.sh to reflect the new docker-compose file and we should remove the old docker-compose file since it is no longer used.

docs/dev/setup.sh
richscott: Because checking for Docker Version 20 or higher is not a reliable heuristic for the existence of "Docker Compose 2" (in which `docker-compose` (written in Python) was replaced by Go code directly as a `docker` sub-command).  So the only way to determine if an installation actually has the preferred new `docker compose` command is to run it, and check the exit status.

docs/dev/setup.sh
richscott: I was undecided if Pulsar bits should be left in or not - I'll rip all the Pulsar stuff out. If someone really wants it, it can be added later.

richscott: Removed in commit d7573ebd.

docs/dev/config/armada/base.yaml
richscott: Good catch - removed in commit 423a4022.

richscott:
> We should change teardown.sh to reflect the new docker-compose file and we should remove the old docker-compose file since it is no longer used.

I'm not sure if the original docker-compose.yaml is completely unused by any GR folks- I got the impression some of them use it to quickly start all the infra services, then they run the Armada services natively via 'go run', etc.

I'll update `teardown.sh`.

kannon92 (approved):
Looks really good!

docs/dev/config/jobservice/config.yaml
```diff
@@ -0,0 +1,14 @@
+grpcPort: 60003
+httpPort: 8090
+subscribeJobSetTime: 10000
+databasePath: "/run/jobservice.db"
```
jayofdoom, line 4: nit: /run seems like a really weird place for this?

docs/dev/setup.sh
```diff
@@ -1,34 +1,56 @@
...
+if [ $? -eq 0 ] ; then
```
jayofdoom, line 9: nit: you can delete line 8 and instead: 

```suggestion
if docker compose >/dev/null 2>&1 ; then
```

but then you'd have to reverse the logic so... probably not worth it?

jayofdoom (approved):

docs/dev/config/jobservice/config.yaml
richscott: `/run` is a not-super-common volume mount point, but is definitely used in various distros and virtualized systems - see https://duckduckgo.com/?t=ffab&q=%22%2Frun%22+directory&ia=web . I considered /var/run and /tmp but the former seemed not much improved, and using /tmp, I was nervous that perhaps other things in the container might be doing stuff in there.

docs/dev/setup.sh
richscott: Hm, yeah - I think reversing the logic could potentially be a bit confusing for devs taking a quick first glance at it.

docs/dev/config/jobservice/config.yaml
jayofdoom: My objection was mainly because of this use not fitting in the FHS (https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard), but it's in a container and doesn't really matter, which is why I marked it as a nit :D

richscott: That standard shows the purpose/requirements of `/run` as
```
Run-time variable data: Information about the running system since last boot, e.g., currently logged-in users and running [daemons](https://en.wikipedia.org/wiki/Daemon_(computer_software)). Files under this directory must be either removed or truncated at the beginning of the boot process, but this is not necessary on systems that provide this directory as a [temporary filesystem](https://en.wikipedia.org/wiki/Temporary_filesystem) ([tmpfs](https://en.wikipedia.org/wiki/Tmpfs)).
```
Isn't this exactly the use-case for an ephemeral sqlite3 db file?

PR #1452, opened 2022-09-02
Unfilter and compress job ownership groups

JamesMurkin:
The existing code filters the groups added to a job based on the groups used to auth to Armada in the name of efficiency

However this causes issues when the namespace being submitted to doesn't have the same perms as the queue being submitted to. As the group required to auth to the namespace was filtered out

Now we no longer filter out the groups to avoid this issue

To reduce the overhead of passing lots of groups around, we now compress these groups
 - We decompress them on the executor side where they are actually used


internal/armada/processor/event_nats.go
```diff
@@ -89,6 +89,8 @@ func (p *RedisEventProcessor) handleBatch(batch []*eventstream.Message) error {
 		event := msg.EventMessage
 		// For submitted events we null out the podspec(s)
 		if event.GetSubmitted() != nil {
+			event.GetSubmitted().Job.QueueOwnershipUserGroups = nil
```
d80tb7, line 92: probably should update the comment above

internal/armada/server/lease.go
```diff
@@ -303,6 +334,39 @@ func (q *AggregatedQueueServer) StreamingLeaseJobs(stream api.AggregatedQueue_St
...
 
+func (q *AggregatedQueueServer) decompressJobOwnershipGroups(jobs []*api.Job) error {
+	for _, j := range jobs {
+		// No need to decompress, if compressed groups not set
+		if len(j.CompressedQueueOwnershipUserGroups) == 0 {
+			continue
+		}
+		groups, err := q.decompressOwnershipGroups(j.CompressedQueueOwnershipUserGroups)
+		if err != nil {
+			return fmt.Errorf("failed to decompress ownership groups for job %s", j.Id)
```
d80tb7, line 345: do you lose the original error here?

d80tb7 (approved):

PR #1453, opened 2022-09-02
Adding pod checks to detect stuck node

### LINKED ISSUE #1293: Pods can become stuck forever on bad nodes
**Problem**

 Pods can go to Pending and then get stuck forever - **seems to only happens if the node is bad**
 
**Cause**
 
We typically detect stuck pods and retry them  https://github.com/G-Research/armada/blob/master/config/executor/config.yaml#L43

However if the pod is:
 - Assigned to a node
 - Has no events
 - Has no container statuses

Then it'll never meet any condition for retry

As this only happens on bad nodes, I think the best course of action is to review the code and check what will happen in the above scenario and add a way to detect it, so it can be handled (retried)

kannon92:
Closes #1293 

This PR addresses the above issue.  If a pod has both no container statuses and events, we should Retry the job.  

We should not Wait.  

Open Question:

Is it possible to add changes to fake executor so we can test this?  It isn't obvious to me in the code so I'm unsure if it is worth the effort.


internal/executor/podchecks/pod_checks.go
```diff
@@ -36,6 +37,9 @@ func NewPodChecks(cfg config.Checks) (*PodChecks, error) {
 
 func (pc *PodChecks) GetAction(pod *v1.Pod, podEvents []*v1.Event, timeInState time.Duration) (Action, string) {
 	messages := []string{}
+	if pc.isNodeBad(pod, podEvents) {
+		return ActionRetry, "Pod Statues and Pod Events are both empty. Retrying"
```
suprjinx, line 41: "Statues" -> "Statuses"

suprjinx (approved):
Looks good to me

JamesMurkin:
This could be suboptimal - although I can't easily confirm without testing myself. I know nodes can be slow to assign events/statuses.

You probably want a check that is like one of the below:
 - pending for X duration
 - pending for X duration with no events/statuses
 
 Otherwise a slow node will lead to us retrying the pod immediately.

kannon92:
> This could be suboptimal - although I can't easily confirm without testing myself. I know nodes can be slow to assign events/statuses.
> 
> You probably want a check that is like one of the below:
> 
> * pending for X duration
> * pending for X duration with no events/statuses
> 
> Otherwise a slow node will lead to us retrying the pod immediately - however maybe I'm wrong on how this code works.
> 
> Ideally this time would be configurable, like the rest of the checks.

Would this be a configuration in the executor?

internal/executor/podchecks/pod_checks.go
```diff
@@ -51,3 +60,10 @@ func (pc *PodChecks) GetAction(pod *v1.Pod, podEvents []*v1.Event, timeInState t
 	log.Infof("Pod checks for pod %s returned %s %s\n", pod.Name, resultAction, resultMessage)
 	return resultAction, resultMessage
 }
+
+// If a node is bad, we can have no pod status and no pod events.
+// We should retry the pod rather than wait
+func (pc *PodChecks) isNodeBad(pod *v1.Pod, podEvents []*v1.Event) bool {
```
JamesMurkin, line 66: I'm not sure I love the name of this, I'm not sure it'll always be true that the node is bad when it happens.

However I think it does the right thing

internal/executor/podchecks/pod_checks.go
```diff
@@ -31,11 +33,18 @@ func NewPodChecks(cfg config.Checks) (*PodChecks, error) {
...
 	}
 
-	return &PodChecks{eventChecks: ec, containerStateChecks: csc}, nil
+	return &PodChecks{eventChecks: ec, containerStateChecks: csc, timeWithoutEventsOrStatus: cfg.TimeWithoutEventsOrStatus}, nil
 }
 
 func (pc *PodChecks) GetAction(pod *v1.Pod, podEvents []*v1.Event, timeInState time.Duration) (Action, string) {
 	messages := []string{}
+
+	isNodeBad := pc.isNodeBad(pod, podEvents)
```
JamesMurkin, line 42: I'm in two minds on if this should just append the messages to the array instead of returning but functionally looks good

JamesMurkin (approved):

internal/executor/podchecks/pod_checks.go
kannon92: If this is true than we really only have 1 entry in the array.  If we have no pod events or pod status, we shouldn't run the getAction of these other steps.  At least in my testing, we hit a panic.  

So I made sure that if the node is bad (and stuck) than we Retry and I set a message.

internal/executor/podchecks/pod_checks.go
suprjinx: we could focus on the events and status of the pod, not the node: `hasNoEventsOrStatus()` or something like that

internal/executor/configuration/podchecks/types.go
```diff
@@ -26,8 +26,9 @@ const (
 )
 
 type Checks struct {
-	Events            []EventCheck
-	ContainerStatuses []ContainerStatusCheck
+	Events                    []EventCheck
+	ContainerStatuses         []ContainerStatusCheck
+	TimeWithoutEventsOrStatus time.Duration
```
suprjinx, line 31: 'TimeWithoutEventsOrStatus' might be clearer as 'DeadlineForUpdates'

kannon92: Done.

internal/executor/podchecks/pod_checks.go
kannon92: Done.

internal/executor/configuration/podchecks/types.go
```diff
@@ -26,8 +26,9 @@ const (
 )
 
 type Checks struct {
-	Events            []EventCheck
-	ContainerStatuses []ContainerStatusCheck
+	Events             []EventCheck
+	ContainerStatuses  []ContainerStatusCheck
+	DeadlineForUpdates time.Duration
```
suprjinx, line 31: I think i have one more suggestion here (the fault is mine for making the original suggestion, my apologies).  "deadline" is usually a time.Time set at some point in the future, and "timeout" a duration. Is it worth making this a Time, just to match the usual convention?

suprjinx (approved):

PR #1454, opened 2022-09-02
add support in armada server for preemptive jobs

dejanzele:


internal/armada/configuration/types.go
```diff
@@ -100,6 +101,12 @@ type SchedulingConfig struct {
 	MinJobResources                           v1.ResourceList
 }
 
+type PreemptionConfig struct {
+	Enabled              bool
+	PriorityClasses      map[string]int32
```
kannon92, line 106: Should we just use the Kubernetes API for PriorityClasses?

dejanzele: This is a mapping so Armada Server knows which Priority Classes are supported, and also to know the priorities so it can correctly calculate how much resources it can preempt for a prioritized job

kannon92: Maybe I wasn't clear.  I am asking if we should use the Kubernetes definition for PriorityClasses.  

https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#example-priorityclass

internal/common/validation/job.go
```diff
@@ -3,9 +3,77 @@ package validation
...
 	"fmt"
 
+	"github.com/pkg/errors"
+
+	"github.com/G-Research/armada/internal/common/armadaerrors"
+
 	"github.com/G-Research/armada/pkg/api"
 )
 
+func ValidateJobRequestItemPodSpec(r *api.JobSubmitRequestItem) error {
```
kannon92, line 13: Is there a reason why we didn't add tests for these functions?

internal/armada/configuration/types.go
dejanzele: Still not clear why, we are only interested in the Priority Class name (string) and the Priority value (int32)

severinson: One argument in favor of Kevin's suggestion is that we may later want to support features added to Kubernetes priority classes. For example, we may want to mark a priority class as non-preemptive. However, I think for now we can leave it like this and then update the config later if necessary.

severinson:
I've added a few comments. Overall, nice work. I like that you're cleaning up the pre-existing code as you go (adding contexts where it makes sense, attaching stack traces to errors, etc.).

internal/armada/scheduling/node_matching.go
```diff
@@ -155,14 +171,43 @@ func matchAnyNodeTypePodAllocation(
 	return nil, false, result
 }
 
+func getPreemptibleResources(
+	supportedPriorityClasses map[string]int32,
+	targetPriorityClass string,
+	allocatedResources map[int32]common.ComputeResourcesFloat,
+	availableResources common.ComputeResourcesFloat,
+) (common.ComputeResourcesFloat, error) {
+	targetPriority, exists := supportedPriorityClasses[targetPriorityClass]
```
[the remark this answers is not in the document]
d80tb7: I guess this is pretty minor but personally I dislike using `ok` as a blanket value for the key not being in the map.  `ok = false` has the semantic meaning of something being "incorrect" which is what you mean in somee cases, but not all.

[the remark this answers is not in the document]
d80tb7: having said that, in this case `ok` is actually a good name!

internal/armada/scheduling/lease.go
```diff
@@ -276,22 +301,30 @@ func (c *leaseContext) leaseJobs(
...
+					c.schedulingConfig.Preemption.PriorityClasses,
+				)
 				if ok {
 					slice = remainder
 					candidates = append(candidates, job)
 					candidatesLimit.RemoveFromRemainingLimit(job)
 					candidateNodes[job] = newlyConsumed
 					consumedNodeResources.Add(newlyConsumed)
+				} else {
+					log.WithError(err).Error("failed to match node")
```
d80tb7, line 327: Am I right in thinking this would mean that if a job was submitted with an invalid priority class  (e.g. because we changed priority classes between the job being submitted and it being scheduled), it would stick around for ever?

I'm tempted to say that in this case we should just submit the job as the default priority class.

internal/armada/scheduling/util.go
```diff
@@ -0,0 +1,7 @@
+package scheduling
+
+import v1 "k8s.io/api/core/v1"
+
+func isPriorityJob(spec *v1.PodSpec) bool {
```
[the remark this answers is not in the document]
d80tb7: +1 to that

d80tb7 (approved):

internal/common/validation/job.go
dejanzele: I'll add tests

internal/armada/scheduling/util.go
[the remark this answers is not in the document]
dejanzele: initially it was `hasPriorityClass`, but I didn't know will we have more complex logic for checking besides does it have a specified `priorityClassName`, so I opted for a more generic name.
With the current state of things, probably this is a better suggestion to rename it to `hasPriorityClass`

internal/armada/scheduling/node_matching.go
```diff
@@ -155,14 +171,43 @@ func matchAnyNodeTypePodAllocation(
...
+	preemptibleResources := make(common.ComputeResourcesFloat)
+
+	for priority, resources := range allocatedResources {
+		if priority < targetPriority {
+			preemptibleResources.Add(resources)
+		}
+	}
+	// resources which can be allocated to the current job are calculated as
+	// Max(sum of allocated resources for lower priority jobs, available resources)
+	preemptibleResources.Max(availableResources)
```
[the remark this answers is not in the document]
dejanzele: correct

internal/armada/scheduling/lease.go
dejanzele: Hmmm it will go to k8s scheduler, and k8s would error saying that pod cannot be scheduled due to non-existing PC.
Will Armada catch that error and handle it? (remove it from queue)

internal/armada/scheduling/lease.go
```diff
@@ -325,6 +363,31 @@ func (c *leaseContext) leaseJobs(
...
+// if not, default to DefaultPriorityClass if it is specified
+// otherwise default to no Priority Class
+func validateOrDefaultPriorityClass(podSpec *v1.PodSpec, preemptionConfig configuration.PreemptionConfig) {
+	_, ok := preemptionConfig.PriorityClasses[podSpec.PriorityClassName]
+	if !ok {
+		if preemptionConfig.DefaultPriorityClass != "" {
+			podSpec.PriorityClassName = preemptionConfig.DefaultPriorityClass
+		} else {
+			podSpec.PriorityClassName = ""
+		}
```
severinson, line 387:
```suggestion
		podSpec.PriorityClassName = preemptionConfig.DefaultPriorityClass
```

severinson: Both branches of the if statement produce the same result :)

internal/armada/scheduling/lease.go
```diff
@@ -325,6 +363,31 @@ func (c *leaseContext) leaseJobs(
...
+	isPreemptiveJob := isJobLargeEnough && hasPriorityClass(job.PodSpec)
+
+	return isRegularlySchedulable || isPreemptiveJob
+}
+
+// validateOrDefaultPriorityClass checks is the pod spec's priority class configured as supported in Server config
+// if not, default to DefaultPriorityClass if it is specified
+// otherwise default to no Priority Class
+func validateOrDefaultPriorityClass(podSpec *v1.PodSpec, preemptionConfig configuration.PreemptionConfig) {
+	_, ok := preemptionConfig.PriorityClasses[podSpec.PriorityClassName]
```
severinson, line 381:
```suggestion
       if preemptionConfig.PriorityClasses == nil {
                   podSpec.PriorityClassName = preemptionConfig.DefaultPriorityClass
        }
	_, ok := preemptionConfig.PriorityClasses[podSpec.PriorityClassName]
```

severinson: If no priority classes are defined, I guess the map would be nil. If that's the case, this will protect against panics.

severinson: If you agree with the change, you may need to fix the spacing here. I'm in a browser...

internal/armada/scheduling/lease.go
```diff
@@ -276,22 +303,33 @@ func (c *leaseContext) leaseJobs(
...
+					c.schedulingConfig.Preemption.PriorityClasses,
+				)
 				if ok {
 					slice = remainder
 					candidates = append(candidates, job)
 					candidatesLimit.RemoveFromRemainingLimit(job)
 					candidateNodes[job] = newlyConsumed
 					consumedNodeResources.Add(newlyConsumed)
+				} else {
+					log.WithError(err).Error("failed to match node")
```
JamesMurkin, line 332: I don't think we want this log line, or at the very least not as an error

I recently removed it from master as it spams the logs with errors but no real details. However that "error" isn't an error and purely is saying that the job doesn't fit onto the existing nodes.
 - This happens a lot, especially on clusters with multiple node types.

internal/armada/scheduling/lease.go
dejanzele: Clever is the arch enemy of clear, I like it :D

internal/armada/scheduling/lease.go
dejanzele: this is fixed

severinson (approved):

PR #1455, opened 2022-09-04
Event API Read Side

d80tb7:
This PR implements the read side for the new events from Pulsar.  Implementation is such that the api will default to serving the events from the old redis repository but can be forced to serve via the new repository (and code path) using a flag in the incoming grpc request.


.circleci/config.yml
```diff
@@ -92,6 +92,8 @@ commands:
           helm package deployment/lookout-migration/ -d charts/armada/
           helm package deployment/binoculars/ -d charts/armada/
           helm package deployment/jobservice/ -d charts/armada/
+          helm package deployment/lookout-ingester/ -d charts/armada/
```
d80tb7, line 95: The lookout-ingester stuff was missing from this file so I've added it along with the event-ingester.

internal/armada/repository/event.go
```diff
@@ -2,29 +2,27 @@ package repository
 
 import (
```
d80tb7, line 3: The diffs here are quite confusing.  This is the new event repository that loads up the stored messages from pulsar.  The previous repository has moved to `legacy_event.go`

internal/armada/repository/event_test.go
```diff
@@ -4,115 +4,222 @@ import (
 	"testing"
 	"time"
 
-	"github.com/G-Research/armada/internal/common/compress"
```
d80tb7, line 7: Ignore the diff here- this is a new file.  The previous file has been moved to `legacy_event_test.go`

internal/armada/repository/legacy_event.go
```diff
@@ -0,0 +1,300 @@
+package repository
```
d80tb7, line 1: This file has been renamed from `event.go`

internal/armada/repository/legacy_event_test.go
```diff
@@ -0,0 +1,118 @@
+package repository
```
d80tb7, line 1: this file has been renamed from `event_test.go`

pkg/api/event.proto
```diff
@@ -278,14 +278,16 @@ message JobSetRequest {
     string from_message_id = 3;
     string queue = 4;
     bool errorIfMissing = 5;
-    bool force_redis = 6;  // This field is for test purposes only
+    bool force_legacy = 6;  // This field is for test purposes only
```
d80tb7, line 281: we're safe to reuse field 6 because this field was never used in anger

internal/armada/repository/sequence/sequence.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	return seq.RedisString()
+}
+
+func (e *ExternalSeqNo) RedisString() string {
+	return fmt.Sprintf("%d-%d", e.Time, e.Seq)
+}
+
+// IsAfter returns true if this ExternalSeqNo is after the other.
+// ordering is by time then sequence, then sequence, then subsequence.  isLast is not considered.
```
JamesMurkin, line 111:
```suggestion
// ordering is by time then sequence,  then subsequence.  isLast is not considered.
```

internal/armada/repository/sequence/sequence.go
```diff
@@ -0,0 +1,135 @@
...
+	_, err := Parse(str)
+	return err == nil
+}
+
+func (e *ExternalSeqNo) String() string {
+	return fmt.Sprintf("%d:%d:%d:%d", e.Time, e.Seq, e.SubSeq, boolToInt(e.Last))
+}
+
+// PrevRedisId returns the redis id that we would have to query *from* in order to guarantee that we would
+// receive subsequent messages.  Not that this is somewhat complex as if this message is not the last message
```
JamesMurkin, line 86:
```suggestion
// receive subsequent messages.  Note that this is somewhat complex as if this message is not the last message
```

JamesMurkin (dismissed):

JamesMurkin (approved):

PR #1462, opened 2022-09-06
Fix 99keyrings/keyring warning on MacOS

carlocamurri:
Upgrading the library seems to fix the warnings
```
cgo-gcc-prolog:81:11: warning: 'SecKeychainCreate' is deprecated: first deprecated in macOS 12.0 - Custom keychain management is no longer supported [-Wdeprecated-declarations]
```
which show up on startup, as explained in the issue below

https://github.com/99designs/keyring/issues/107


richscott (approved):
I've seen this warning many times on my Mac - I look forward to this fix!

PR #1464, opened 2022-09-06
Add retry to SubscribeToJobSetId() until no-error exit

### LINKED ISSUE #1281: Add retry connection ability to the jobservice
The jobservice will connect to Armada via the EventClient.  There will be connection issues so we should handle reconnection in this service.

suprjinx:
Closes #1281


internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -38,20 +39,28 @@ func NewEventsToJobService(
...
-	}
-	defer conn.Close()
+	var conn *grpc.ClientConn
+	defer func() {
+		if conn != nil {
+			conn.Close()
+		}
+	}()
 	eventToJobService.jobServiceRepository.SubscribeJobSet(eventToJobService.queue, eventToJobService.jobSetId)
+	ctx, cancel := context.WithCancel(ctx)
```
kannon92, line 63: Should we be mixing cancelling context and error groups here?

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -60,42 +69,51 @@ func (eventToJobService *EventsToJobService) streamCommon(clientConnect *client.
...
-			return err
-		}
+		// this loop will run until the context is canceled or an error is encountered
 		for {
-
-			msg, err := stream.Recv()
-			if err != nil {
-				log.Error(err)
-				return err
+			conn, connErr := client.CreateApiConnection(clientConnect)
```
kannon92, line 81: I don't think we want this.  A big point of this was to only have a single connection for a given job-set.

We also want to infinitely listen on the stream of the GetJobSetEvents rather than keep calling it.

internal/jobservice/eventstojobs/eventstojobs.go
suprjinx: I thought about that but error group doesn't give you access to the cancel function which we kind of need in order to cause a no-error termination

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -4,101 +4,111 @@ import (
...
+	eventClient          events.JobEventReader
+	jobServiceRepository repository.JobTableUpdater
 }
 
 func NewEventsToJobService(
 	queue string,
 	jobSetId string,
 	jobId string,
+	eventClient events.JobEventReader,
 	jobServiceConfig *configuration.JobServiceConfiguration,
```
suprjinx, line 31: can remove this attribute, no longer used here

internal/jobservice/eventstojobs/eventstojobs.go
suprjinx: What I'm reading -- the parent context CancelFunc will also cancel the children/derived contexts. The inverse is not true. In this case, I think it's okay -- we use the errgroup to capture the goroutine return error but we're not relying on it's cancellation.

internal/jobservice/events/client.go
```diff
@@ -0,0 +1,60 @@
+//go:generate moq -out client_moq.go . JobEventReader
```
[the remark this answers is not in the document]
suprjinx: let me figure out the easiest way -- it's not needed unless the interface changes and we need to regen.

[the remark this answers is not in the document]
suprjinx: afaict it must be installed as standalone tool. Added comment to docs/developer.md about this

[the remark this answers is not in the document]
kannon92: Thanks for adding that!

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -4,101 +4,122 @@ import (
...
+		eventClient:          eventClient,
 		jobServiceRepository: jobServiceRepository,
 	}
 }
 
-// Subscribes to a JobSet from jobsetid
-func (eventToJobService *EventsToJobService) SubscribeToJobSetId(context context.Context) error {
-	return eventToJobService.streamCommon(&eventToJobService.jobServiceConfig.ApiConnection, context)
+// Subscribes to a JobSet from jobsetid. Will retry until there is a successful exit, up to the TTL
+func (eventToJobService *EventsToJobService) SubscribeToJobSetId(context context.Context, ttlSecs int64) error {
```
kannon92, line 42: I don't know if we want the same config as subscribe job set as the time to determine if our connection is bad.  

Should we just create a new config for TTL?

suprjinx: That sounds reasonable, but as a matter of course maybe it doesn't matter? If the subscribe job set time ttl expired, this will quit retrying. If this were larger than the subscribe time, it would continue trying but with certain failure.

kannon92: Yea I see your point.  But the SubscribeJobSetTime can be as long as how long it takes for a job-set to run.  So we were thinking that it should be close to the lifetime of a running job.  On the order of hours/days maybe.  

So if we keep the same config, we will be retrying connections for that long of time before we decide it is a failure.  

Right now, our SubscribeJobSetTime is 10000 seconds.

suprjinx: I probably don't understand it well enough, but if the job is still active we should still attempt to subscribe, right?

kannon92: Ideally but I don't think we have too.  

We unsubscribe if a job state doesn't get updated after x amount of time.  If a job is running for 2-3 weeks, we can unsubscribe and resubscribe without any issue.  All unsubscribed means is that we no longer have a connection to the job set.  It doesn't stop the job or anything like that.

suprjinx: this retry loop will exit when IsJobSetSubscribed() returns false.

suprjinx: So in this PR, we will attempt to retry for the duration of the SubscribeJobSetTime if we're getting errors from one of the go routines. This is the same amount of time that it would be subscribed. Once the subscribed jobs cleanup occurs, and the job is unsubscribed, this loop exits.

internal/jobservice/events/client.go
```diff
@@ -0,0 +1,61 @@
...
+	if ec.conn != nil {
+		ec.Close()
+	}
+}
+
+func (ec *EventClient) ensureApiConnection() error {
+	if ec.conn != nil {
+		return nil
+	}
+	conn, connErr := client.CreateApiConnection(ec.config)
```
d80tb7, line 54: so this won't be threadsafe.  I know there's a mutex in `JobSetSubscriptions` but it's not obvious if it'll be covering all use cases here.  Personally I'd make this class threadsafe.

internal/jobservice/events/client.go
```diff
@@ -0,0 +1,61 @@
...
+	if connErr != nil {
+		log.Warnf("Connection Issues with EventClient %v", connErr)
```
d80tb7, line 56: Asa a guide- I wouldn't log here.  Let whatever handles the error log it, else you end up with the same error logged at each level of the stack.

As an aside `log.withError(errr).Warn(....)` is quite nice becuase it makes sure you log the stack.

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -4,101 +4,122 @@ import (
...
+		for {
+			select {
+			case <-ctx.Done():
+				return nil
+			default:
+				msg, err := eventToJobService.eventClient.GetJobEventMessage(ctx, &api.JobSetRequest{
+					Id:             eventToJobService.jobSetId,
+					Queue:          eventToJobService.queue,
+					Watch:          true,
+					FromMessageId:  fromMessageId,
```
d80tb7, line 107: (nit) I think you can move `FromMessageId` into this loop to help with readability.

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -4,101 +4,122 @@ import (
...
+				})
+				if err != nil {
+					log.Error(err)
+					return err
+				}
+				fromMessageId = msg.GetId()
+				currentJobId := api.JobIdFromApiEvent(msg.Message)
+				jobStatus := EventsToJobResponse(*msg.Message)
+				if jobStatus != nil {
+					jobTable := repository.NewJobTable(eventToJobService.queue, eventToJobService.jobSetId, currentJobId, *jobStatus)
```
d80tb7, line 118: It seems a bit strange that you create a new Job "Table" on every message.  Should this be renamed `NewJobRow` or something?

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -4,101 +4,122 @@ import (
...
+				return nil
+			default:
+				msg, err := eventToJobService.eventClient.GetJobEventMessage(ctx, &api.JobSetRequest{
+					Id:             eventToJobService.jobSetId,
+					Queue:          eventToJobService.queue,
+					Watch:          true,
+					FromMessageId:  fromMessageId,
+					ErrorIfMissing: false,
+				})
+				if err != nil {
```
d80tb7, line 110: this is the error I think you should handle- simply log the error here, sleep for 5 seconds and then go round the loop again.

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -4,101 +4,122 @@ import (
...
+					log.Error(err)
+					return err
+				}
+				fromMessageId = msg.GetId()
```
d80tb7, line 114: If you go with my suggestion of handelling the error above, then move the update of `fromMessageId` to after the database insert.  That when when that function returns an error (right now it panics, but it shouldn't) we can only advance our position in the stream after we know the db insert was good.

internal/jobservice/events/client.go
d80tb7, line 54: won't this create a new connection every time?

suprjinx: thanks, good idea

suprjinx: only the first time around, or if ec.conn cleared due to an error.

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -4,101 +4,122 @@ import (
...
 	}
 }
 
-// Subscribes to a JobSet from jobsetid
-func (eventToJobService *EventsToJobService) SubscribeToJobSetId(context context.Context) error {
-	return eventToJobService.streamCommon(&eventToJobService.jobServiceConfig.ApiConnection, context)
+// Subscribes to a JobSet from jobsetid. Will retry until there is a successful exit, up to the TTL
+func (eventToJobService *EventsToJobService) SubscribeToJobSetId(context context.Context, ttlSecs int64) error {
+	expiresAt := time.Now().Add(time.Duration(ttlSecs) * time.Second)
+	for {
```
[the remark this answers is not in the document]
suprjinx: done

internal/jobservice/eventstojobs/eventstojobs.go
suprjinx: this is where the last-processed message id is passed into the EventClient so I'm not sure how to move it.

internal/jobservice/events/client.go
suprjinx: removed

internal/jobservice/events/client.go
suprjinx: added mutex

internal/jobservice/eventstojobs/eventstojobs.go
suprjinx: Yes can refactor, was just using the existing interface there

internal/jobservice/events/client.go
```diff
@@ -0,0 +1,85 @@
...
+
+// GetJobEventMessage performs all the steps for obtaining an event message
+func (ec *EventClient) GetJobEventMessage(ctx context.Context, jobReq *api.JobSetRequest) (*api.EventStreamMessage, error) {
+	err := ec.ensureApiConnection()
+	if err != nil {
+		return nil, err
+	}
+	eventClient := api.NewEventClient(ec.conn)
+	log.Debug("looking for job set events")
+	ctx, _ = context.WithTimeout(ctx, time.Duration(5)*time.Second)
```
kannon92, line 46: So does this timeout end the stream also?

internal/jobservice/eventstojobs/eventstojobs.go
suprjinx: done

internal/jobservice/eventstojobs/eventstojobs.go
suprjinx: done

internal/jobservice/events/client.go
suprjinx: changed to make this more clear with helper method

internal/jobservice/eventstojobs/eventstojobs.go
suprjinx: changed row represenation to JobStatus

internal/jobservice/events/client.go
suprjinx: yeah i bet you're right. I'll take another look.

suprjinx: removed this timeout for now, we need a gRPC healthcheck endpoint

kannon92 (approved):
@d80tb7 would like his approval before merging.  But it LGTM.

d80tb7 (approved):

PR #1465, opened 2022-09-07
Lookout fix race condition in Job table

carlocamurri:
There is a race condition in the Jobs table where if you make two (or more) queries in rapid succession (for example when typing quickly into a column filter), if the first query takes more time to complete than the second, stale data will be returned.

The fix is to use `AbortController` to cancel outbound request when the filters change, analogous to `context`s in go.
The current generated API client did not allow for `AbortController` signals to be passed in the functions, so I had to update the version of `openapi-generator-cli` to v5.4.0, which I found was the latest version which did not cause compilation errors.


internal/lookout/ui/src/services/MockJobService.ts
```diff
@@ -0,0 +1,122 @@
+import { v4 as uuidv4 } from "uuid"
```
kannon92, line 1: Oh man!  This is great!  Thanks for adding this.  Its been on my list of things to add so we can troubleshoot other API errors.

kannon92: So is this mock service used anywhere else?  I'd expect to see a unit test or something using it?

internal/lookout/ui/src/containers/JobsContainer.tsx
```diff
@@ -34,6 +34,7 @@ export type JobsContainerState = {
   jobDialogIsOpen: boolean
   clickedJob?: Job
   getJobsRequestStatus: RequestStatus
+  abortController: AbortController
```
kannon92, line 37: Why is there no import of this type/class anywhere in the code?  

I'd expect to see an import somewhere of AbortController?

kannon92: Its apparently part of react?

PR #1472, opened 2022-09-08
remove code-checks since golangci is checking this already

kannon92:
Both goimports and ineffassign are in golangci.  I think we can just remove the make target since these are duplicated.


richscott (dismissed):

jayofdoom (commented):
We probably shouldn't merge this until we fully understand why Geoff's change failed on make code-checks but not on golangci-lint step...

richscott (approved):

kannon92:
> We probably shouldn't merge this until we fully understand why Geoff's change failed on make code-checks but not on golangci-lint step...

Thinking about this, I may need to add a config for goimports that says to tread

Something like

goimports:
    local-prefixes: github.com/org/repo

To our linter and that should replicate code checks.

jayofdoom (approved):

suprjinx (approved):

PR #1473, opened 2022-09-08
Upload built docker image artifacts to GHA

jayofdoom:

- go-integration-tests builds and saves docker images to an artifact (#1473)
- go-integration-tests uses workflow_call to call upload-docker-images@master (change TODO, will be written once this and 1551 merge)
- upload-docker-images downloads those artifacts, and pushes them to docker #1551 
  - This is a separate workflow, so secrets can be isolated to an environment (`armada-dockerhub`), and then we can configure that environment to only work when called on a protected branch.
- (on release), armada-release re-tags these docker images (along with doing other work) and pushes them to dockerhub (#1552)


.github/workflows/armada-release.yml
```diff
@@ -0,0 +1,88 @@
...
+      - name: Upload artifact to github actions
+        uses: actions/upload-artifact@v3
+        with:
+          name: armadactl-release
+          path: dist/*
+
+  release-docker-images:
+    runs-on: buildjet-4vcpu-ubuntu-2204
+    strategy:
+      matrix:
```
jayofdoom, line 34: pull this, we don't run any go setup in this job

.github/workflows/armada-release.yml
```diff
@@ -0,0 +1,88 @@
...
+        # We obviously will not matrix releases; but
+        # the go-setup workflow expects this to be set
+        # via a matrix
+        go: [ '1.18' ]
+    steps:
+      - name: Pull and re-tag docker images
+        run: |
+          # TODO(JayF): Need to ensure we have a job that uploads docker images on push to master
```
jayofdoom, line 42: just making sure I call my own attention to this

.github/workflows/armada-release.yml
```diff
@@ -0,0 +1,88 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/workflows/go-setup
+      - name: Create version tag for dotnet
+        id: version
+        run: |
+          DOTNET_TAG=$(echo ${{ github.ref }} | sed 's/v//g')
+          echo "::set-output name=DOTNET_TAG::$DOTNET_TAG"
+      - run: make push-nuget
+        env:
```
jayofdoom, line 82: this needs ENV NUGET_API_KEY as well, unsure where that comes from in circle

.github/workflows/armada-release.yml
```diff
@@ -0,0 +1,88 @@
...
+    strategy:
+      matrix:
+        # We obviously will not matrix releases; but
+        # the go-setup workflow expects this to be set
+        # via a matrix
+        go: [ '1.18' ]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/workflows/go-setup
+      - run: make build-armadactl-release
```
jayofdoom, line 19: This needs env: 
- RELEASE_VERSION

.github/workflows/go-integration.yml
```diff
@@ -29,4 +29,42 @@ jobs:
         uses: actions/upload-artifact@v2.2.4
         with:
           name: junit.xml
-          path: test_reports/junit.xml
\ No newline at end of file
+          path: test_reports/junit.xml
+
+      - name: Assemble container tag
```
jayofdoom, line 34: I should probably just do this in bash to avoid the barely-useful `use`

.github/workflows/armada-release.yml
```diff
@@ -0,0 +1,88 @@
...
+      matrix:
+        # We obviously will not matrix releases; but
+        # the go-setup workflow expects this to be set
+        # via a matrix
+        go: [ '1.18' ]
+    steps:
+      - name: Pull and re-tag docker images
+        run: |
+          # TODO(JayF): Need to ensure we have a job that uploads docker images on push to master
+          # TODO(JayF): how to get these secrets in here safely?
```
kannon92, line 43: Can't you just add a secret to GitHub?  That seems to be the most secure option?

.github/workflows/go-integration.yml
```diff
@@ -29,4 +29,42 @@ jobs:
...
+        run: |
+          # Note(JayF): All the offical docker actions for uploading these
+          # containers also assume you need to rebuild them. That makes them
+          # ill-suited for our workflow.
+          TAG=$(echo -n "${{ steps.container_tag.outputs.long }}" | sed 's|/|-|g')
+
+          # TODO(JayF): how to get these secrets in here safely?
+          echo ${DOCKERHUB_PASS} | docker login -u ${DOCKERHUB_USER} --password-stdin
+
+          IMAGES_TO_UPLOAD=(
```
kannon92, line 51: Anyway to DRY this? Variable in GitHub actions?

.github/workflows/armada-release.yml
```diff
@@ -0,0 +1,88 @@
+name: Release Armada
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  release-armadactl:
+    runs-on: buildjet-4vcpu-ubuntu-2204
```
kannon92, line 9: Do we really need to pay money to use these runners for a release process?  What is Perf without these runners?

.github/workflows/go-integration.yml
jayofdoom: This isn't the exact set we're using in another place. It's subtly different. And frankly, I'm happy in this case to keep it all in bash.

jayofdoom:
The nature of this PR has to change, due to how the behavior has been restructured. Now, the implementation will be across multiple PRs:
- This one, which now adds code to go-integration-tests to export docker images, then upload a tarball of those exported images into github artifacts.
- One that permits uploading of docker images from previously-built artifacts from the ^^^ above change, using dockerhub secrets (in an environment limited to run on protected branches only). 
- Once that change merges, we'll have a small additional change to the go-integration-tests workflow: to call that docker image uploading workflow from master branch (must be run from protected branch) to upload those changes. This change will also need to remove the upload done by circleci it (or add some kind of prefix/suffix to the tag) to prevent them stepping on each other.
- Finally, add an additional workflow, that runs on: release, which re-tags the docker images already uploaded (by the above) to the proper released ones, along with other release items. 

I'll upload draft versions of these where possible -- mainly from where I've already written some of this code, but now has to move it around.

.github/workflows/go-integration.yml
```diff
@@ -26,7 +30,30 @@ jobs:
...
+          docker save armada-event-ingester | gzip > docker-images/armada-event-ingester.tar.gz
+          docker save armada-binoculars | gzip > docker-images/armada-binoculars.tar.gz
+          docker save armada-jobservice | gzip > docker-images/armada-jobservice.tar.gz
+          
+          tar -czf docker-images.tar.gz docker-images/*
+
+      - name: Upload docker image tarball to artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: docker-images
```
jayofdoom, line 57: this may need ${github.sha} or something else unique appended

jayofdoom: I don't think it needs this, artifacts are namespaced to their build.

PR #1490, opened 2022-09-09
Allow developers to run armada as separate processes outside of a container

### LINKED ISSUE #1489: We should add back a way to launch individual processes and community via localhost
For integrations, we may just want to simplify the network and communicate via localhost.

This issue is about adding back the way before we combined everything into docker-compose.

kannon92:
Fixes #1489

### DIFF
--- a/config/jobservice/config.yaml
+++ b/config/jobservice/config.yaml
@@ -10,5 +10,7 @@ grpc:
   keepaliveEnforcementPolicy:
     minTime: 5m
     permitWithoutStream: false
+apiConnection:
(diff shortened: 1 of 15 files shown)


richscott (approved):
I think the `docs/dev/local_config` naming might be confusing, since there is already a `docs/dev/config`, but I understand the motivation for making it. I'm going to think about maybe renaming the latter to something more specific/descriptive. And in general, I think it might be worth considering moving `docs/dev` into a top-level directory (with a new name), emphasizing it's really about tools for developing Armada and its ecosystem.

This is a good, very useful addition.

PR #1504, opened 2022-09-12
emit Failed event when pod gets preempted

dejanzele:


internal/executor/job/job_context.go
```diff
@@ -287,3 +295,22 @@ func (c *ClusterJobContext) handleDeletedPod(pod *v1.Pod) {
 		}
 	}
 }
+
+func (c *ClusterJobContext) checkForPreemption(pod *v1.Pod, jobId string) bool {
+	events := c.clusterContext.GetEvents()
+	for _, e := range events {
+		isCurrentDeletedPod := e.InvolvedObject.Name == pod.Name
+		if util.IsPreemptedEvent(e) && isCurrentDeletedPod {
+			c.registerIssue(jobId, &PodIssue{
```
severinson, line 304: I'd move this to the caller. So we don't have two different functions that may register an issue.

severinson: So the caller would be something like 
```
if preemptionEvent := c.checkForPreemption(pod, jobId); preemptionEvent {
// register preemption issue
} else {
// register ExternallyDeleted issue
}
```

internal/executor/job/job_context.go
```diff
@@ -276,6 +277,21 @@ func (c *ClusterJobContext) handleDeletedPod(pod *v1.Pod) {
 	if jobId != "" {
 		isUnexpectedDeletion := !util.IsMarkedForDeletion(pod) && !util.IsPodFinishedAndReported(pod)
 		if isUnexpectedDeletion {
+			// Preempted cluster event occurs after the pod gets killed (preempted)
+			// Wait for 3 seconds so the Preempted event gets recorded
+			time.Sleep(3 * time.Second)
```
d80tb7, line 282: What effect does sleeping for 3 seconds here have?  Does it block anything else?  

I guess I'm thinking about the case where a whole bunch of pods are pre-empted.  Doe this lead to the executor pausing processing events for any length of time?

internal/executor/job/job_context.go
```diff
@@ -287,3 +303,14 @@ func (c *ClusterJobContext) handleDeletedPod(pod *v1.Pod) {
 		}
 	}
 }
+
+func (c *ClusterJobContext) checkForPreemptionEvent(pod *v1.Pod) *v1.Event {
+	events := c.clusterContext.GetEvents()
```
d80tb7, line 308: Are we able to use an indexer here as in https://github.com/G-Research/armada/pull/1504/files#diff-67e3b00cf2b0b449d3dc3bee9dbf9b7ec3011c458b9bf79527dabdbecafa96bfR208

dejanzele: Yes, good catch

internal/executor/job/job_context.go
dejanzele: `	// AddEventHandler adds an event handler to the shared informer using the shared informer's resync
	// period.  Events to a single handler are delivered sequentially, but there is no coordination
	// between different handlers.
	AddEventHandler(handler ResourceEventHandler)`

I guess I should wrap this into a goroutine

internal/executor/job/job_context.go
```diff
@@ -269,21 +273,65 @@ func createStuckPodMessage(retryable bool, originalMessage string) string {
...
 	}
 }
+
+func (c *ClusterJobContext) checkForPreemptionEvent(pod *v1.Pod) (*v1.Event, error) {
+	events, err := c.clusterContext.GetPodEvents(pod)
+	if err != nil {
+		return nil, err
+	}
+	for _, e := range events {
+		isCurrentDeletedPod := e.InvolvedObject.Name == pod.Name
```
kannon92, line 331: Do you think we need a name place plus name here?

PR #1509, opened 2022-09-12
Update action.yml

shafink11:
Creating change to report error and line number in the CI.

### DIFF
--- a/.github/workflows/go-setup/action.yml
+++ b/.github/workflows/go-setup/action.yml
@@ -8,10 +8,6 @@ runs:
       with:
         go-version: ${{ matrix.go }}
         cache: true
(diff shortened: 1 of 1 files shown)


jayofdoom (changes requested):
So, the better place for this is probably `.github/workflows/go.yml` in the Go unit tests where the other golangci-lint action runs -- we have lots of jobs that run that go-setup that aren't strictly for go code.

shafink11:
Okay yeah I will submit a new PR that puts the code in the right section

kannon92:
> Okay yeah I will submit a new PR that puts the code in the right section
> 
> 
> 
> 

You should always commit your changes to the existing PR.

PR #1510, opened 2022-09-13
Update go.yml

shafink11:
Adding the code that states the line number and error type in the CI in the right place rather than the action.yaml file


.github/workflows/go.yml
```diff
@@ -62,3 +62,7 @@ jobs:
         with:
           name: junit.xml
           path: test_reports/junit.xml
+      - name: Display lint results
+        run: golangci-lint run
```
kannon92, line 66: Your CI job is failing because you don't have this tool installed before running this command.

PR #1512, opened 2022-09-13
add validation when submiting jobs via Pulsar route

dejanzele:


internal/armada/server/submit_to_log.go
```diff
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"time"
 
+	commonvalidation "github.com/G-Research/armada/internal/common/validation"
```
severinson, line 9: Armada imports should come last.

internal/common/validation/job.go
```diff
@@ -1,79 +1,63 @@
 package validation
 
 import (
-	"fmt"
-
+	"github.com/G-Research/armada/internal/armada/configuration"
```
severinson, line 4: Import ordering

severinson (approved):

PR #1521, opened 2022-09-15
update proto

d80tb7:
Add `KubernetesResourceInfo` to `JobRunSucceeded` and `JobSuceeded` messages.  The reason here is twofold:

- Because we're not reliably generating `JobRunRunning` messages you get into the position where there isn't a message that actually tells you what the machine the job ran on.
- The old (existing) public event stream has a `JobSucceeded` message that contains the run details (node, namespace etc).   At present I have to null these out which is a breaking api change.


internal/armada/repository/apimessages/conversions.go
```diff
@@ -239,15 +239,28 @@ func FromInternalJobSucceeded(queueName string, jobSetName string, time time.Tim
 	if err != nil {
 		return nil, err
 	}
+
+	apiEvent := &api.JobSucceededEvent{
```
severinson, line 243: Wouldn't it be better to have a function in eventutil? Like this but for log sequence to api sequence?
https://github.com/G-Research/armada/blob/master/internal/common/eventutil/eventutil.go#L519

internal/common/eventutil/eventutil.go
```diff
@@ -864,6 +864,24 @@ func EventSequenceFromApiEvent(msg *api.EventMessage) (sequence *armadaevents.Ev
 				JobRunSucceeded: &armadaevents.JobRunSucceeded{
 					RunId: runId,
 					JobId: jobId,
+					ResourceInfos: []*armadaevents.KubernetesResourceInfo{
+						{
+							ObjectMeta: &armadaevents.ObjectMeta{
+								Namespace:    m.Succeeded.NodeName,
```
severinson, line 870: NameSpace != NodeName

d80tb7: ah good spot.  I actually copied this from the equivalent `JobRunRunning` function and sure enough we have the same issue there- I've fixed both

internal/armada/repository/apimessages/conversions.go
d80tb7: I think it's a historical accident that the conversions from logEvent -> ApiEvent are here, while the reverse are in `EventUtil`.  That said I'm not convinced it's worth moving everything.  I hope the `EventUtil` ones will go away soon, while these ones are probably in the right place because the old `api.Event` should purely be confined to the serving of events to users.

severinson (approved):

PR #1536, opened 2022-09-20
Feature/executor preemption event separate

dejanzele:


internal/armada/server/event.go
```diff
@@ -49,13 +53,52 @@ func (s *EventServer) Report(ctx context.Context, message *api.EventMessage) (*t
 	if err := checkPermission(s.permissions, ctx, permissions.ExecuteJobs); err != nil {
 		return nil, status.Errorf(codes.PermissionDenied, "[Report] error: %s", err)
 	}
+	if event, ok := message.Events.(*api.EventMessage_Preempted); ok {
```
d80tb7, line 56: Are we able to do this earlier in the process (i.e. as soon as the event comes in from the executor)? I ask because the new event ingester doesn't have access to the jobs db so would have difficult filling in here.

internal/armada/server/event.go
```diff
@@ -49,13 +53,52 @@ func (s *EventServer) Report(ctx context.Context, message *api.EventMessage) (*t
...
+		if err := s.preemptedEventHandler(event); err != nil {
+			return &types.Empty{}, err
+		}
+	}
 	return &types.Empty{}, s.eventStore.ReportEvents([]*api.EventMessage{message})
 }
 
+func (s *EventServer) preemptedEventHandler(event *api.EventMessage_Preempted) error {
+	if event.Preempted.JobId != "" {
+		result, err := s.jobRepository.GetJobsByIds([]string{event.Preempted.JobId})
```
d80tb7, line 66: This might be slow because we'll end up doing a (potentially) large number of single redis lookups.  It's probably not worth optimizing right now, but worth bearing in mind.

kannon92 (commented):
Would preemption be considered a terminal event for other schedulers?  Thinking if I should add this state into job service/airflow?

dejanzele:
> Would preemption be considered a terminal event for other schedulers? Thinking if I should add this state into job service/airflow?

I don't think so, with the current implementation.
You will receive a terminal JobFailed event with a message `A pod of an active job has been killed`, and a separate JobPreempted event which serves as a notification what happened.

@severinson correct me if I am wrong

pkg/api/event.proto
```diff
@@ -132,6 +132,23 @@ message JobFailedEvent {
     Cause cause = 12;
 }
 
+message JobPreemptedEvent {
```
kannon92, line 135: You are having some python test failures because we generate some types from these event files.  You probably just need to add this event to the python tests.  

https://github.com/G-Research/armada/blob/master/client/python/tests/unit/test_gen.py

@Sharpz7 could help with this.

d80tb7:
> Would preemption be considered a terminal event for other schedulers? Thinking if I should add this state into job service/airflow?

Ideally it would be a terminal event, but due to the limitations of the current system it's not.  I'd ignore it for now, but as and when you let the airflow operator support preemption, you'll probably want make the operator resubmit the job.

internal/armada/server/event.go
```diff
@@ -49,16 +53,86 @@ func (s *EventServer) Report(ctx context.Context, message *api.EventMessage) (*t
...
+	for _, event := range preemptedEvents {
+		if err := s.enrichPreemptedEvent(event, results); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (s *EventServer) enrichPreemptedEvent(event *api.EventMessage_Preempted, results []*repository.JobResult) error {
```
d80tb7, line 106: I'd suggest giving the second argument more descriptive name than `results`.  Maybe `JobDetails`.  Also as you need to do lookoups on this, would it better to make it a map[string]JobResult?

internal/common/eventutil/eventutil.go
```diff
@@ -1043,6 +1043,45 @@ func EventSequenceFromApiEvent(msg *api.EventMessage) (sequence *armadaevents.Ev
 		// Do nothing; there's no corresponding Pulsar message.
 	case *api.EventMessage_Updated:
 		// Do nothing; we're not allowing arbitrary job updates.
+	// TODO: add case for Preempted event
```
d80tb7, line 1046: can get rid of this TODO!

internal/common/eventutil/eventutil.go
```diff
@@ -1043,6 +1043,45 @@ func EventSequenceFromApiEvent(msg *api.EventMessage) (sequence *armadaevents.Ev
...
+		sequence.JobSetName = m.Preempted.JobSetId
+
+		preemptedJobId, err := armadaevents.ProtoUuidFromUlidString(m.Preempted.JobId)
+		if err != nil {
+			return nil, err
+		}
+
+		event := &armadaevents.EventSequence_Event_JobPreempted{
+			JobPreempted: &armadaevents.JobPreempted{
+				PreemptedJobId:         preemptedJobId,
```
d80tb7, line 1058: Do we need `PreemptedJobSetId` and `PreemptedJobQueue` as these are presumably already available from the eventSequence

internal/executor/reporter/event.go
```diff
@@ -137,8 +141,87 @@ func CreateJobIngressInfoEvent(pod *v1.Pod, clusterId string, associatedServices
...
+
+func enrichPreemptedEventFromPreemptionMessage(event *api.JobPreemptedEvent, msg string) error {
+	info, err := util.ParsePreemptionMessage(msg)
+	if err != nil {
+		return err
+	}
+
+	if event.PreemptiveJobId == "" && util.IsArmadaJobPod(info.Name) {
+		jobId, err := util.ExtractJobIdFromName(info.Name)
+		if err != nil {
```
d80tb7, line 205: should this be a warning? I'm not sure what action ends up getting taken here but it's probably better to generate an event with the `PreemptiveJobId ` missing than it is to lose the event completely.

internal/executor/reporter/event.go
```diff
@@ -137,8 +141,87 @@ func CreateJobIngressInfoEvent(pod *v1.Pod, clusterId string, associatedServices
...
+			return errors.Errorf("error extracting preemptive job id from pod name: %v", err)
+		}
+		event.PreemptiveJobId = jobId
+	}
+	if event.PreemptivePodNamespace == "" {
```
d80tb7, line 210: why do we do the empty check here? It would be simpler if this function always set these fields?

pkg/armadaevents/events.proto
```diff
@@ -476,4 +474,19 @@ message SeqUpdates{
 
 message DatabaseSequence{
     EventSequence eventSequence = 1;
-}
\ No newline at end of file
+}
+
+message JobPreempted {
```
severinson, line 479: This should be `JobRunPreempted`, not `JobPreempted`. Since it's a specific run of the job that was preempted. Further, instead of adding all these fields at the top level of the message, we should have a oneof field. This oneof field would have as an option `PodPreempted`, which would contain the fields currently in the `JobPreempted` message. We do something similar for job submitted messages. This is to support preempting jobs that don't consist only of a single field in the future.

internal/common/eventutil/eventutil.go
dejanzele: well no, I will remove them

internal/executor/reporter/event.go
dejanzele: Sometimes the Cluster Event has the `Related` object which has info on the preemptive pod, sometimes no, and in that case I parse the preemption message and fill in the missing information

pkg/armadaevents/events.proto
severinson: I thought about this. Do we really need any more than this?
For now, we'd need to use the job id to populate the run id field.
```
message JobRunPreempted {
    // Id of the run that was preempted.
    Uuid preempted_run_id = 1;
    // Id of the job run that caused the preemption.
    Uuid preemptive_run_id = 2;
}
```

pkg/armadaevents/events.proto
```diff
@@ -476,4 +474,23 @@ message SeqUpdates{
...
+message PodPreempted {
+    // Uuid of the job that was preempted
+    Uuid preempted_job_id = 1;
+    // Uuid of the job run that was preempted.
+    Uuid preempted_run_id = 2;
+    // Uuid of the job that caused the preemption.
+    Uuid preemptive_job_id = 3;
+    // Uuid of the job run that caused the preemption.
+    Uuid preemptive_run_id = 4;
+}
```
severinson, line 496:
```suggestion

// Message to indicate that a JobRun has been preempted
message JobRunPreempted{
    // Uuid of the job that was preempted
    Uuid preempted_job_id = 1;
    // Uuid of the job run that was preempted.
    Uuid preempted_run_id = 2;
    // Uuid of the job that caused the preemption.
    Uuid preemptive_job_id = 3;
    // Uuid of the job run that caused the preemption.
    Uuid preemptive_run_id = 4;
}
```

d80tb7 (dismissed):

severinson (approved):

PR #1543, opened 2022-09-21
fix bug with wrong preemption event timestamp; parse preemption message

dejanzele:


internal/executor/reporter/event.go
```diff
@@ -187,6 +188,28 @@ func enrichPreemptedEventFromRelatedObject(event *api.JobPreemptedEvent, related
...
+	if err != nil {
+		return nil
+	}
+
+	if !util.IsArmadaJobPod(info.Name) {
+		return nil
+	}
+
+	if event.PreemptiveJobId == "" {
+		preemptiveJobId, err := util.ExtractJobIdFromName(info.Name)
```
severinson, line 202: This looks a bit aggressive. Isn't it better to just leave it empty if the preemptive job id is missing, instead of not generating a preempted event at all? Let me know if I've misunderstood.

dejanzele: My initial thoughts were that the event would have the `Related` field, which should be the reference to the preemptive pod.
After testing, the `Related` field in the Cluster Event is always `nil`.
My logic was, get info from the `Related` field, if it was `nil`, parse info from Cluster Event message (`Preempted by <namespace>/<name> on node <node>`.
Turned out `Related` is always nil, so this is the way to get at least the job id of the preemptive pod.

d80tb7 (approved):

internal/executor/reporter/event.go
severinson: Would related always be nil in all environment tho? Maybe for our clusters it's set.

PR #1545, opened 2022-09-22
Give Executor Patch perms on Events

d80tb7:
We need this because the executor now adds annotations to events to facilitate preemption.


theAntiYeti (approved):
LGTM

PR #1551, opened 2022-09-25
WIP: Upload docker images from artifacts

jayofdoom:

- go-integration-tests builds and saves docker images to an artifact (#1473)
- go-integration-tests uses workflow_call to call upload-docker-images@master (change TODO, will be written once this and 1473 merge)
- upload-docker-images downloads those artifacts, and pushes them to docker
  - This is a separate workflow, so secrets can be isolated to an environment (`armada-dockerhub`), and then we can configure that environment to only work when called on a protected branch.
- (on release), armada-release re-tags these docker images (along with doing other work) and pushes them to dockerhub (#1552)

### DIFF
--- a/.github/workflows/upload-docker-images.yml
+++ b/.github/workflows/upload-docker-images.yml
@@ -1,0 +1,56 @@
+name: Release Armada
+
+on:
+  workflow_call: {}
+
+jobs:
+  upload-docker-images:
+    runs-on: ubuntu-22.04
+    environment: armada-dockerhub
+    steps:
+      # trigger docker image upload workflow; downloading tarballs from artifacts and uploading them to docker
+      - name: Assemble container tag
+        id: container_tag
+        uses: pr-mpt/actions-commit-hash@v1
+        with:
+          commit: "${{ github.sha }}"
+          prefix: "${{ github.ref_name }}"
+      - name: Download saved docker-images artifact
+        uses: actions/download-artifact@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/upload-docker-images.yml
```diff
@@ -0,0 +1,56 @@
...
+          name: docker-images
+          path: docker-images.zip
+      - name: Unzip docker-images artifacts
+        run: |
+          unzip docker-images.zip
+          tar xf docker-images.tar.gz
+          rm docker-images.tar.gz
+      - name: Upload images to docker
+        env:
+            DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
```
jayofdoom, line 30:
```suggestion
            DOCKERHUB_PASS: ${{ secrets.DOCKERHUB_PASS }}
            DOCKERHUB_USER: ${{ secrets.DOCKERHUB_USER }}
```

.github/workflows/upload-docker-images.yml
```diff
@@ -0,0 +1,56 @@
...
+            DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
```
jayofdoom, line 31:
```suggestion
            DOCKERHUB_PASS: ${{ secrets.DOCKERHUB_PASS }}
            DOCKERHUB_USER: ${{ secrets.DOCKERHUB_USER }}
```

PR #1552, opened 2022-09-25
Add release action for armada

jayofdoom:

* go-integration-tests builds and saves docker images to an artifact (#1473)
* go-integration-tests uses workflow_call to call upload-docker-imagesmaster (change TODO, will be written once this and 1473 merge)
* upload-docker-images downloads those artifacts, and pushes them to docker
  * This is a separate workflow, so secrets can be isolated to an environment (`armada-dockerhub`), and then we can configure that environment to only work when called on a protected branch.
* (on release), armada-release re-tags these docker images (along with doing other work) and pushes them to dockerhub (#1552)

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-71) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/armada-release.yml
+++ b/.github/workflows/armada-release.yml
@@ -1,0 +1,83 @@
+name: Release Armada
+
+on:
+  release:
+    types: [published]
+
+# put secrets in an armada-release environment
+# ensure we use environment here
+jobs:
+  release-armadactl:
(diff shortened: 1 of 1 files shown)


kannon92:
hey @jayofdoom we ended up implementing this a different way once we had a better idea of the requirements.  Going to close.

PR #1553, opened 2022-09-26
WIP Enable submitting several Kubernetes objects

severinson:


pkg/api/submit.proto
```diff
@@ -26,6 +26,47 @@ message JobSubmitRequestItem {
...
+    // Main object that determines when an application has finished.
+    KubernetesMainObject mainObject = 12;
+    // Set of additional Kubernetes objects to create as part of the job.
+    repeated KubernetesObject objects = 13;
+}
+
+// Kubernetes objects that can serve as main objects for an Armada job.
+message KubernetesMainObject {
+    ObjectMeta objectMeta = 1;
+    oneof object {
```
kannon92, line 38: Can we just remove the oneof here since its only a single field?

severinson: No, it's necessary to support future extensions.

PR #1556, opened 2022-09-26
Adding Resources, Command and Arguments to the Job Detail View

### LINKED ISSUE #1476: Display resource information in job details view
User would like to be able to see request resource details (e.g. cpu, memory, gpu, disk) in the job details view without having to scan the job request yaml.

Related to https://github.com/G-Research/armada/issues/910.

### LINKED ISSUE #1477: Display command information in job details view
Users would like to see the executed command and command line parameters in job details view.

kannon92:
Closes #1477 
Closes #1476 

In this PR, we convert the JobYaml to a yaml definition and parse it.  We use this to get command, arguments, and resource requests.  The PodSpec can take a list of containers so we should probably render this as a list also.


internal/lookout/ui/src/services/ComputeResourcesService.ts
```diff
@@ -0,0 +1,89 @@
...
+  args: string[]
+  resources: {
+    limits: Limits
+  }
+}
+
+interface Limits {
+  cpu?: string
+  "nvidia.com/gpu"?: string
+  storage?: string
```
carlocamurri, line 20: This should be `ephemeral-storage` (or `ephemeralStorage`, worth double checking what the API returns)

internal/lookout/ui/src/components/job-dialog/JobDetails.tsx
```diff
@@ -59,6 +74,38 @@ export default function JobDetails(props: DetailsProps) {
             <DetailRow name="Job state" value={props.job.jobState} />
             <DetailRow name="Priority" value={props.job.priority.toString()} />
             <DetailRow name="Submitted" value={props.job.submissionTime} />
+            {command &&
+              command.map(
```
carlocamurri, line 78: Rather than displaying everything in the same table, I would have a separate expandable list of containers, with conatiner name, command, args, and requests for each container.

In this way you could have a function to get a list of Containers, instead of having many functions to get list of cpu requests, list of memory requests, etc... individually

internal/lookout/ui/src/services/ComputeResourcesService.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (val.args) {
+      stringArray.push(val.args.join(" "))
+    } else {
+      stringArray.push("")
+    }
+  })
+  return stringArray
+}
+
+export function getCpuFromJobYaml(yamlifed: JobYaml): string[] | undefined {
```
carlocamurri, line 57:
```suggestion
export function getCpuFromJobYaml(yamlified: JobYaml): string[] | undefined {
```
Likewise for the others

internal/lookout/ui/src/services/ComputeResourcesService.ts
```diff
@@ -0,0 +1,89 @@
...
+  if (checkContainerFieldExists(yamlifed)) return
+  const stringArray: string[] = []
+  yamlifed.podSpec.containers.map((val) => {
```
carlocamurri, line 60: Shouldn't you be able to do this with
```
return yamlified.podSpec.containers.map((val) => val.resources?.limits?.cpu ?? ""
```

internal/lookout/ui/src/services/ComputeResourcesService.ts
kannon92: Good call. Changed to ephemeral-storage.

internal/lookout/ui/src/components/job-dialog/JobDetails.tsx
kannon92: Good suggestion.  

I am not sure what you mean by "expandable list of containers".  Do you mean that this should be a react component of a List?  

I tried to combine all those separate functions into a single function so the usability of getting the container info is easy for the UI.  

Let me know if this is what you want.

internal/lookout/ui/src/services/ComputeResourcesService.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+}
+
+export interface ContainerInfo {
+  command?: string[]
+  arguments?: string[]
+  cpu?: string[]
+  memory?: string[]
+  gpu?: string[]
+  "ephermal-storage"?: string[]
```
carlocamurri, line 16:
```suggestion
  "ephemeral-storage"?: string[]
```
Please test this to make sure it works

internal/lookout/ui/src/services/ComputeResourcesService.ts
```diff
@@ -0,0 +1,111 @@
...
+export interface JobYaml {
+  podSpec: {
+    containers: Container[]
+  }
+}
+
+export interface ContainerInfo {
+  command?: string[]
+  arguments?: string[]
+  cpu?: string[]
```
carlocamurri, line 13: Why does a single `ContainerInfo` have a list of values for cpu, memory, gpu? Each container info should only have a single value. Don't think this model makes sense here, I'd probably use the `Container` interface below, and have a list of them.

internal/lookout/ui/src/services/ComputeResourcesService.ts
```diff
@@ -0,0 +1,111 @@
...
+  cpu?: string
+  "nvidia.com/gpu"?: string
+  "ephemeral-storage"?: string
+  memory?: string
+}
+
+export function getContainerInfoFromYaml(jobYaml: string): ContainerInfo {
+  const yaml = load(jobYaml) as JobYaml
+  return {
+    command: getCommandFromJobYaml(yaml),
```
carlocamurri, line 37: This logic should be changed based on previous comment, but in general you can implement this with a single map over the containers rather than having to get each list of fields individually.

internal/lookout/ui/src/services/ComputeResourcesService.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+function checkContainerFieldExists(yamlified: JobYaml): boolean {
+  return !yamlified.podSpec.containers
+}
+
+function getCommandFromJobYaml(yamlified: JobYaml): string[] | undefined {
+  if (checkContainerFieldExists(yamlified)) return
+  const stringArray: string[] = []
+  yamlified.podSpec.containers.map((val) => {
```
carlocamurri, line 53: Mentioned this already, but you shouldn't append values to an array in the map callback, rather return the values from the map itself. See the previous resolved comment for reference.

kannon92: Ah I see.

internal/lookout/ui/src/services/ComputeResourcesService.ts
kannon92: Done.

kannon92:
Talking with @dave-gantenbein, we think that the feature to expand based on container name is separate from this issue.

I added an accordion component that will optional expand the container items but it won't do smarter expansion based on name at the moment.

suprjinx (approved):
LGTM

PR #1557, opened 2022-09-26
Ensure tests run on consistent platform

jayofdoom:
For tests we actually run (e.g. not noop or codeql jobs), ensure they run on ubuntu-22.04 for consistency, instead of ubuntu-latest.


.github/workflows/airflow-operator-release-to-pypi.yml
```diff
@@ -5,7 +5,7 @@ on:
 
 jobs:
   run-python-tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
kannon92, line 8: The main drawback to this is what forces us to update to newer version of Ubuntu?  

When Ubuntu 22.04 is no longer in support, we should update this?

jayofdoom: Pretty much, yes. And if we don't do it proactively, things will kaboom.

On the flip side; if we use `ubuntu-latest` and it changes it can also kaboom things without warning.

There are tradeoffs but no universally good choice -- except making sure they're all consistent :D

kannon92 (approved):

jayofdoom:
Flakey test is failing circleci; but this change doesn't modify circleci. It should be a safe merge.

richscott (approved):

PR #1561, opened 2022-09-26
Expand locking to encompass entire SQLJobService.UpdateJobServiceDb func

ClifHouck:
This solves the flaky unit test TestConcurrentJobStatusUpdating which should, in turn, indicate this is a more robust way for UpdateJobServiceDb to operate. However, I'm not very happy with this because I don't understand *why* its better. I'm guessing it has something to do with including the `PrepareDb` call to sqlite, but I'm not certain.

### DIFF
--- a/internal/jobservice/repository/sql_job_service.go
+++ b/internal/jobservice/repository/sql_job_service.go
@@ -125,17 +125,19 @@ func jobStateStrToJSRState(jobState string) (js.JobServiceResponse_State, error)
 
 // Update database with JobTable.
 func (s *SQLJobService) UpdateJobServiceDb(jobTable *JobStatus) {
+	// SQLite only allows one write at a time. Therefore we must serialize
+	// writes in order to avoid SQL_BUSY errors.
+	s.writeLock.Lock()
+	defer s.writeLock.Unlock()
+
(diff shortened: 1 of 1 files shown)


internal/jobservice/repository/sql_job_service.go
```diff
@@ -125,17 +125,19 @@ func jobStateStrToJSRState(jobState string) (js.JobServiceResponse_State, error)
...
+	// SQLite only allows one write at a time. Therefore we must serialize
+	// writes in order to avoid SQL_BUSY errors.
+	s.writeLock.Lock()
+	defer s.writeLock.Unlock()
+
 	stmt, err := s.db.Prepare("INSERT OR REPLACE INTO jobservice VALUES (?, ?, ?, ?, ?, ?)")
 	if err != nil {
 		panic(err)
 	}
+	defer stmt.Close()
```
ClifHouck, line 137: I also added this since, according to go docs, you should close prepared statements once you're done with them. We may want to look into preparing these statements once during DB creation and then re-using them.

ClifHouck:
Apparently, this did not solve the issue. Need to look into a more robust solution.

ClifHouck:
Closing this in favor of #1568

PR #1567, opened 2022-09-27
Py gen test fix

### LINKED ISSUE #1560: Python: Latest Release of GRPC breaks unit tests
Our auto generation of types breaks if you use grpcio==1.49.1.  

tests/unit/test_gen.py::test_union_var FAILED                                                                     [ 93%]
tests/unit/test_gen.py::test_job_states PASSED                                                                    [ 95%]
tests/unit/test_gen.py::test_file_gen FAILED                                                                      [ 96%]

To resolve this, we pinned GRPC to 1.46 (as this was what we had before) but we should not have to pin to older version.

Sharpz7:
Closes #1560 

Basically, since grpc version 1.48, there has been a major change in how the stub files are generated, relying more heavily on the `DESCRIPTOR` tag for all the serialized infomation.

We now use this variable for all our generation, hopefully avoiding this issue going forward.

### DIFF
--- a/client/python/armada_client/gen/event_typings.py
+++ b/client/python/armada_client/gen/event_typings.py
@@ -10,12 +10,16 @@ def get_event_states():
 
 
 def get_all_job_event_classes():
+    job_events = []
+
     for possible_event in event_pb2.__dict__:
         try:
             if "job_id" in getattr(event_pb2, possible_event).DESCRIPTOR.fields_by_name:
-                yield possible_event
+                job_events.append(possible_event)
         except AttributeError:
             continue
+
+    return job_events
 
 
 def get_job_states():
--- a/client/python/armada_client/gen/event_typings.py
+++ b/client/python/armada_client/gen/event_typings.py
@@ -10,7 +10,12 @@ def get_event_states():
 
 
 def get_all_job_event_classes():
-    return [x for x in event_pb2.__dict__ if hasattr(getattr(event_pb2, x), "job_id")]
+    for possible_event in event_pb2.__dict__:
+        try:
+            if "job_id" in getattr(event_pb2, possible_event).DESCRIPTOR.fields_by_name:
+                yield possible_event
+        except AttributeError:
+            continue
 
 
 def get_job_states():
@@ -66,7 +71,7 @@ def main():
     states = get_event_states()
     print("Done creating EventStates")
 
-    classes = get_all_job_event_classes()
+    classes = list(get_all_job_event_classes())
     print("Done creating JobEvent classes")
 
     jobstates = get_job_states()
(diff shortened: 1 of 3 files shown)


client/python/armada_client/gen/event_typings.py
```diff
@@ -10,7 +10,16 @@ def get_event_states():
 
 
 def get_all_job_event_classes():
```
kannon92, line 12: Is this backwards compatible?

And can you update airflow operator grpc dependency?

client/python/pyproject.toml
```diff
@@ -4,7 +4,7 @@ version = "0.2.1"
 description = "Armada gRPC API python client"
 readme = "README.md"
 requires-python = ">=3.7,<3.9"
-dependencies = ["grpcio~=1.46.3", "grpcio-tools~=1.46.3", "mypy-protobuf>=3.2.0"]
+dependencies = ["grpcio~=1.49", "grpcio-tools~=1.49", "mypy-protobuf>=3.2.0"]
```
kannon92, line 7: @jayofdoom has suggested that we maybe not so strictly pinn our dependencies.  

```suggestion
dependencies = ["grpcio>=1.46.3", "grpcio-tools>=1.46.3", "mypy-protobuf>=3.2.0"]
```

If this is backwards compatible.  If not, then we can bump up the version and have it be greater than equal too.

client/python/armada_client/gen/event_typings.py
Sharpz7: Define backwards compatible? Will this fail if we lower the version down? I am not sure, I'll test it soon

Sharpz7:
Checked for backwards compatibility, works fine!

kannon92 (commented):
The logic looks good to me.  

I ask that you also apply the same dependency to the airflow operator and you push a release for both python client and airflow operator once this gets merged.

You should be able to do push releases by bumping the version (patch) and then running the action that releases the client and the airflow operator.

client/python/pyproject.toml
```diff
@@ -4,7 +4,7 @@ version = "0.2.1"
 description = "Armada gRPC API python client"
 readme = "README.md"
 requires-python = ">=3.7,<3.9"
-dependencies = ["grpcio~=1.46.3", "grpcio-tools~=1.46.3", "mypy-protobuf>=3.2.0"]
```
kannon92, line 7: On second thought, I think you should revert this change.  

I wasn't really saying that we have to update GRPC but that we were locked into the old version because we couldn't update.  

Talking with others, I think we should just keep the version set to the lowest we test with (1.46.3) so >=1.46.3.

jayofdoom: Yeah, I agree with this. If we are too strict with versions (e.g. ~= or == instead of <= or >=), it'll be difficult to co-install this.

Sharpz7:
@jayofdoom should I bump the versions up in this commit or make another PR?

PR #1568, opened 2022-09-27
Improve sqlite handling in Jobservice

### LINKED ISSUE #1563: Jobservice can still run into SQL_BUSY errors when trying to write to the job service table
Related to https://github.com/G-Research/armada/issues/1544

Locking around executing write queries seems to greatly reduce the chance of running into SQL_BUSY errors in the jobservice, but doesn't eliminate the chance. Need to figure out a more robust solution. 

Some possible avenues:

- Use [Write Ahead Logging](https://www.sqlite.org/wal.html)
- Add some retry logic to statements that error out due to SQL_BUSY.
- Use a different DB?

ClifHouck:
Does a few specific things:

- Enables [WAL](https://www.sqlite.org/wal.html) journaling for SQLite DB used by jobservice.
- Expands locking to encompass entire `SQLJobService.UpdateJobServiceDb` method.
- Returns an error from `SQLJobService.UpdateJobServiceDb` instead of `panic`ing on `stmt.Exec` error. Callers of `UpdateJobServiceDb` can handle the error (or not) as they see fit.

The move to WAL virtually eliminates the incidence of `SQLITE_BUSY` errors when used in conjunction with using a `sync.Mutex` around operations that would write to the DB. Additional load testing seems to indicate we're more likely to run into semaphore acquisition panics under extreme load under this regime. 

Closes #1563

### DIFF
--- a/internal/jobservice/eventstojobs/eventstojobs.go
+++ b/internal/jobservice/eventstojobs/eventstojobs.go
@@ -105,7 +105,12 @@ func (eventToJobService *EventsToJobService) streamCommon(ctx context.Context, t
 				jobStatus := EventsToJobResponse(*msg.Message)
 				if jobStatus != nil {
 					jobStatus := repository.NewJobStatus(eventToJobService.queue, eventToJobService.jobSetId, currentJobId, *jobStatus)
-					eventToJobService.jobServiceRepository.UpdateJobServiceDb(jobStatus)
+					err := eventToJobService.jobServiceRepository.UpdateJobServiceDb(jobStatus)
+					if err != nil {
+						log.WithError(err).Error("could not update job status, retrying")
+						time.Sleep(5 * time.Second)
+						continue
+					}
 				}
 				// advance the message id for next loop
 				fromMessageId = msg.GetId()
--- a/internal/jobservice/repository/sql_job_service.go
+++ b/internal/jobservice/repository/sql_job_service.go
@@ -45,6 +45,22 @@ func NewSQLJobService(jobSetSubscribe *JobSetSubscriptions, config *configuratio
 	return &SQLJobService{jobSetSubscribe: jobSetSubscribe, jobServiceConfig: config, db: db}
 }
 
+// Call on a newly created SQLJobService object to setup the DB for use.
+func (s *SQLJobService) Setup() {
+	s.useWAL()
+	s.CreateTable()
+}
+
+func (s *SQLJobService) useWAL() {
+	s.writeLock.Lock()
+	defer s.writeLock.Unlock()
+
+	_, err := s.db.Exec("PRAGMA journal_mode=WAL")
+	if err != nil {
+		panic(err)
+	}
+}
+
 type SubscribedTuple struct {
 	Queue  string
 	JobSet string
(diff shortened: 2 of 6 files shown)


internal/jobservice/repository/sql_job_service.go
```diff
@@ -124,23 +143,22 @@ func jobStateStrToJSRState(jobState string) (js.JobServiceResponse_State, error)
...
-func (s *SQLJobService) UpdateJobServiceDb(jobTable *JobStatus) {
+func (s *SQLJobService) UpdateJobServiceDb(jobTable *JobStatus) error {
+	// SQLite only allows one write at a time. Therefore we must serialize
+	// writes in order to avoid SQL_BUSY errors.
+	s.writeLock.Lock()
+	defer s.writeLock.Unlock()
+
 	stmt, err := s.db.Prepare("INSERT OR REPLACE INTO jobservice VALUES (?, ?, ?, ?, ?, ?)")
 	if err != nil {
 		panic(err)
```
kannon92, line 154: Let's remove this panic also.  Maybe just return this err?

internal/jobservice/repository/sql_job_service_test.go
```diff
@@ -282,8 +301,6 @@ func TestHealthCheck(t *testing.T) {
 // This test will fail if sqlite writes are not serialised somehow due to
```
kannon92, line 301: Is comment still correct?  I guess now we serialized and enable WAL?

internal/jobservice/repository/sql_job_service.go
ClifHouck: SGTM. Addressed in 262bad8.

internal/jobservice/repository/sql_job_service_test.go
ClifHouck: The locking in conjunction with WAL seems to eliminate them, at least in my testing.

kannon92 (approved):

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -105,7 +105,12 @@ func (eventToJobService *EventsToJobService) streamCommon(ctx context.Context, t
 				jobStatus := EventsToJobResponse(*msg.Message)
 				if jobStatus != nil {
 					jobStatus := repository.NewJobStatus(eventToJobService.queue, eventToJobService.jobSetId, currentJobId, *jobStatus)
-					eventToJobService.jobServiceRepository.UpdateJobServiceDb(jobStatus)
+					err := eventToJobService.jobServiceRepository.UpdateJobServiceDb(jobStatus)
+					if err != nil {
+						log.WithError(err).Error("could not update job status, retrying")
```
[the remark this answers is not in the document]
ClifHouck: Yes I think it could conceivably cause an infinite loop. I modeled this on the error handling just above at https://github.com/G-Research/armada/pull/1568/files/262bad849730c45e0575070af442f6bf1a84d1f1#diff-512415eae5bb0e754a65e8a3cdd55a69b04cd1a13a6e62b806b5fa8a33dd5d61L99

richscott (approved):

internal/jobservice/eventstojobs/eventstojobs.go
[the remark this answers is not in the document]
ClifHouck: I don't mind updating these to retry a few times and then either panic or just continue on.

PR #1573, opened 2022-09-28
Add terminationGracePeriod config, validation, and default application

suprjinx:


internal/armada/server/submit.go
```diff
@@ -897,8 +897,23 @@ func (server *SubmitServer) applyDefaultsToPodSpec(spec *v1.PodSpec) {
...
+	// add missing TerminationGracePeriod if needed
+	server.applyTerminationGracePeriodDefault(spec)
 }
 
+// applyTerminationGracePeriodDefault will give the podspec a default if needed
+func (server *SubmitServer) applyTerminationGracePeriodDefault(spec *v1.PodSpec) {
+	terminationGracePeriod := int64(server.schedulingConfig.TerminationGracePeriod.Default.Seconds())
+	preememptionEnabled := server.schedulingConfig.Preemption.Enabled
+	if preememptionEnabled &&
+		spec.PriorityClassName != "" &&
```
dejanzele, line 910: there is a util function called `hasPriorityClass`

internal/armada/server/submit.go
```diff
@@ -897,8 +897,23 @@ func (server *SubmitServer) applyDefaultsToPodSpec(spec *v1.PodSpec) {
...
+		spec.TerminationGracePeriodSeconds == nil &&
```
dejanzele, line 911: I'd extract the conditions into named variables so it is clearer, eg. this could be `terminationGracePeriodSecondsNotDefined`

internal/armada/server/submit.go
suprjinx: ah thanks will use

internal/armada/server/submit.go
```diff
@@ -897,8 +897,23 @@ func (server *SubmitServer) applyDefaultsToPodSpec(spec *v1.PodSpec) {
...
+		terminationGracePeriod > 0 {
```
ClifHouck, line 912: Should we check if `terminationGracePeriod` is bounded by the max and min? I guess this turns into the question if we should validate that the default is in between the max and min?

ClifHouck: I guess this would be caught anyway by `validateTerminationGracePeriod` but it might still be clearer to validate the default here so the user realizes the problem lies with the server configuration and not their own pod-spec.

ClifHouck: I don't think it's an unreasonable thing to do. Do we validate the server config elsewhere? I'm still new to most of this codebase.

suprjinx: I think we have to trust the configurer to put sane values in -- the whole system is easily broken by incorrect config and this is the least of those problems :-)

suprjinx: I went ahead and added a separate error message to the validation method when the default is outside the boundaries

internal/armada/configuration/types.go
```diff
@@ -97,13 +97,21 @@ type SchedulingConfig struct {
...
 	DefaultJobTolerations                     []v1.Toleration
+	TerminationGracePeriod                    TerminationGracePeriodConfig
 	MaxRetries                                uint // Maximum number of retries before a Job is failed
 	ResourceScarcity                          map[string]float64
 	PoolResourceScarcity                      map[string]map[string]float64
 	MaxPodSpecSizeBytes                       uint
 	MinJobResources                           v1.ResourceList
 }
 
+// TerminationGracePeriodConfig applied to preemptible jobs
```
d80tb7, line 108: So the comment says that this is for preempible jobs only, but the configuration is at the Scheduling level (i.e. `SchedulingConfig.TerminationGracePeriod  ` rather than at the Preemption level (i.e. `Scheduling.Preemption .TerminationGracePeriod ` )

I think we should either move this to be under `Premption` or have `TerminationGracePeriodConfig` applied to all jobs regardless.

d80tb7: FWIW My personal preference would be to just change the PremptionConfig to be:
```
type PreemptionConfig struct {
	Enabled bool
	PriorityClasses map[string]int32
	DefaultPriorityClass string
	// New Fields added for Grace period
        MinTerminationGracePeriod time.Duration
	MaxTerminationGracePeriod time.Duration
	DefaulTerminationGracePeriod time.Duration
}
```

internal/common/validation/podspec.go
```diff
@@ -54,6 +59,35 @@ func ValidatePodSpec(spec *v1.PodSpec, schedulingSpec *configuration.SchedulingC
...
+		exceedsBounds = (terminationGracePeriodSeconds < int64(config.TerminationGracePeriod.Minimum.Seconds()) ||
+			terminationGracePeriodSeconds > int64(config.TerminationGracePeriod.Maximum.Seconds()))
+		isDefault = terminationGracePeriodSeconds == int64(config.TerminationGracePeriod.Default)
+	}
+
+	if config.Preemption.Enabled &&
+		hasTerminationGracePeriod &&
+		exceedsBounds {
+
+		if isDefault {
```
d80tb7, line 77: I know you got asked to do this- but imo this is nuts!  We shouldn't validate this here but rather just validate the config on startup and bail if something crazy has been set.  I'm pretty sure we don't validate any config on startup right now so we probably need a separate issue to get this done.

internal/common/validation/podspec.go
```diff
@@ -54,6 +59,35 @@ func ValidatePodSpec(spec *v1.PodSpec, schedulingSpec *configuration.SchedulingC
...
 
+func validateTerminationGracePeriod(spec *v1.PodSpec, config *configuration.SchedulingConfig) error {
+	hasTerminationGracePeriod := spec.TerminationGracePeriodSeconds != nil
+	var terminationGracePeriodSeconds int64
+	var exceedsBounds, isDefault bool
+	if hasTerminationGracePeriod {
+		terminationGracePeriodSeconds = *spec.TerminationGracePeriodSeconds
+		exceedsBounds = (terminationGracePeriodSeconds < int64(config.TerminationGracePeriod.Minimum.Seconds()) ||
+			terminationGracePeriodSeconds > int64(config.TerminationGracePeriod.Maximum.Seconds()))
+		isDefault = terminationGracePeriodSeconds == int64(config.TerminationGracePeriod.Default)
```
d80tb7, line 70: So this isn't necessarily true?  Just becuase the `terminationGracePeriodSeconds ` is the same as our default, it doesn't mean we defaulted it.  In the case that the user happened to explicitly pick a value that matches our defualt, we'll get a misleading error.

Regardless, I've made a comment below, saying that we shouldn't be sanity checking the default value in this function.

internal/armada/configuration/types.go
suprjinx: Makes sense, can move it there

internal/common/validation/podspec.go
suprjinx: removing

suprjinx: removing here -- but in general the "applyDefaults" logic precedes validation, so it's not just this case where explicit vs default cannot be distinguished

internal/armada/configuration/types.go
```diff
@@ -123,6 +123,11 @@ type PreemptionConfig struct {
 	// Priority class assigned to pods that do not specify one.
 	// Must be an entry in PriorityClasses above.
 	DefaultPriorityClass string
+
+	// TerminationGracePeriod settings for preemptible pods
```
severinson, line 127: I disagree with having termination period be related to preemption. I saw Chris suggested moving it here, but I think these settings should be in ArmadaConfig and that they should be applied to all jobs, regardless of if they have a priority class etc.

severinson:
I believe you also need to update this function to leave GracePeriodSeconds as nil.
Otherwise the graceful termination period isn't respected.
https://github.com/G-Research/armada/blob/master/internal/executor/context/cluster_context.go#L432

Once you've done this, please test submitting pods with non-zero graceful termination period to ensure it works.

config/armada/config.yaml
```diff
@@ -49,6 +49,11 @@ scheduling:
   maxPodSpecSizeBytes: 65535
   minJobResources:
     memory: 1Mi
+  preemption:
+    enabled: false
+  minTerminationGracePeriod: 5s
+  maxTerminationGracePeriod: 300s
+  defaultTerminationGracePeriod: 60s
```
severinson, line 56: Let's make the default equal to the minimum, i.e., 5s, for now. Since the current behaviour is 0s, I'd prefer to use a low default initially.

internal/armada/server/submit.go
```diff
@@ -897,6 +897,22 @@ func (server *SubmitServer) applyDefaultsToPodSpec(spec *v1.PodSpec) {
...
+
+	// add missing TerminationGracePeriod if needed
+	server.applyTerminationGracePeriodDefault(spec)
+}
+
+// applyTerminationGracePeriodDefault will give the podspec a default if needed
+func (server *SubmitServer) applyTerminationGracePeriodDefault(spec *v1.PodSpec) {
+	specNeedsTerminationGracePeriod := spec.TerminationGracePeriodSeconds == nil
+	defaultTerminationGracePeriod := int64(server.schedulingConfig.DefaultTerminationGracePeriod.Seconds())
+	defaultTerminationGracePeriodAvailable := defaultTerminationGracePeriod > 0
```
severinson, line 909: I don't think this is right. We should allow setting the default to be 0s, which this check prevents.  I'd make this function do the following and nothing else:
- Check if `spec.TerminationGracePeriodSeconds` is `nil`.
- If it is, set it to `server.schedulingConfig.DefaultTerminationGracePeriod.Seconds()`.

internal/common/validation/podspec.go
```diff
@@ -54,6 +59,26 @@ func ValidatePodSpec(spec *v1.PodSpec, schedulingSpec *configuration.SchedulingC
 	return validatePorts(spec)
 }
 
+func validateTerminationGracePeriod(spec *v1.PodSpec, config *configuration.SchedulingConfig) error {
+	specHasTerminationGracePeriod := spec.TerminationGracePeriodSeconds != nil
+	configHasBounds := int64(config.MaxTerminationGracePeriod.Seconds()) > 0
```
severinson, line 64: Similar to the previous comment, we should allow `MaxTerminationGracePeriod`, which this check prevents. If you want to make a setting optional, you need to either
- Make the config value a pointer so you can compare against nil (k8s does this), or
- Add a separate boolean flag that controls whether a setting is active (we do this, e.g., for Pulsar).

Because we bundle default config when we build Armada, I don't think it's a problem to expect the graceful termination settings to be set. So I don't think we need to make this setting optional.

severinson:
Did you try this end-to-end? I.e., setup a local environment with Kind, submit a pod, and check that the graceful termination period is set correctly on the pod created in kind.

You can use `make tests-e2e-setup` and `make tests-e2e-teardown` to setup and tear down a local environment, respectively.

internal/executor/context/cluster_context.go
```diff
@@ -353,7 +354,10 @@ func (c *KubernetesClusterContext) ProcessPodsToDelete() {
 		}
 
 		if err == nil {
-			err = c.kubernetesClient.CoreV1().Pods(podToDelete.Namespace).Delete(context.Background(), podToDelete.Name, deleteOptions)
+			if podToDelete.DeletionGracePeriodSeconds != nil {
```
severinson, line 357: I don't think this is right.
As I suggested yesterday, I think we should update this function:
https://github.com/G-Research/armada/blob/master/internal/executor/context/cluster_context.go#L432
To look like this:
```
func createDeleteOptions() metav1.DeleteOptions {
	deleteOptions := metav1.DeleteOptions{
                // If left as nil, the default grace period set by the pod is used.
		GracePeriodSeconds: nil,
	}
	return deleteOptions
}
```
Is there a reason you didn't go with this option?

suprjinx: this function is used by ingress and service deletion as well so I didn't want to affect those.

internal/common/validation/podspec.go
suprjinx: Okay that makes sense, I'll just leave them as values and assume they are sane to avoid the pointer-checking.

config/armada/config.yaml
suprjinx: will do

internal/executor/context/cluster_context.go
severinson: Good catch. Sounds good.

config/armada/config.yaml
```diff
@@ -49,6 +49,11 @@ scheduling:
   maxPodSpecSizeBytes: 65535
   minJobResources:
     memory: 1Mi
+  preemption:
+    enabled: false
+  minTerminationGracePeriod: 5s
```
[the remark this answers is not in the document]
suprjinx: kk will do

PR #1575, opened 2022-09-29
Improve logging around auth attempts for grpc calls

### LINKED ISSUE #1370: Improve no perms error message
Currently, the error returned to users looks something like:
`{"error":"Request in not authenticated with any of the supported schemes.","code":16,"message":"Request in not authenticated with any of the supported schemes."}`
We should improve this to give more info. For example
- Info on which schemes were tried (type, principal name, etc.) and, for each scheme, if it resulted in "permission denied" or an error.
- What action the user was trying to make.

The goal is to help users understand if they're using the wrong creds, forgot to supply creds, if they're trying to perform some action other than what they intended, or if they just don't have perms. Currently, these events would all result in the same error message.

This is the part of the code that should be updated:
https://github.com/G-Research/armada/blob/master/internal/common/auth/authorization/common.go#L139

ClifHouck:
Closes #1370 

Improve auth middleware logging by recording all auth services attempted along with principal names.

This does not currently address logging the action taking place. That seems like it should already be covered by other logging middleware.

### DIFF
--- a/internal/common/auth/authorization/anonymous.go
+++ b/internal/common/auth/authorization/anonymous.go
@@ -4,6 +4,10 @@ import "context"
 
 type AnonymousAuthService struct{}
 
+func (authService *AnonymousAuthService) Name() string {
+	return "Anonymous"
+}
+
 func (AnonymousAuthService) Authenticate(ctx context.Context) (Principal, error) {
 	return anonymousPrincipal, nil
 }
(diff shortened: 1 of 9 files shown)


severinson:
I'd prefer we define specific error types in the `armadaerrors` package. 
https://github.com/G-Research/armada/blob/master/internal/common/armadaerrors/errors.go

If you do, instead of using `==` to check for known error types, use `errors.Is` and `errors.As`. See this link for an intro:
https://go.dev/blog/go1.13-errors

ClifHouck:
Closed in favor of https://github.com/G-Research/armada/pull/1592

PR #1577, opened 2022-09-30
New scheduler

severinson:
This PR introduces a new scheduler with the following benefits:

- It ensures there's a feasible mapping from pods to nodes before leasing jobs to an executor.
- It's better tested.
- It's fairer since it schedules one job per queue at a time.
- It understands priority classes and which jobs can preempt which.
- It tracks all node labels and taints, not only those given in the scheduler config.
- It records its scheduling decisions (successful or not) in reports that can be queried by users via a gRPC endpoint.

Like the existing scheduler, it's backed by Redis. However, the long-term goal is to have it store its state in-process.


internal/armada/server/lease.go
```diff
@@ -38,6 +45,13 @@ type AggregatedQueueServer struct {
 	eventStore               repository.EventStore
 	schedulingInfoRepository repository.SchedulingInfoRepository
 	decompressorPool         *pool.ObjectPool
+	// Map from "executorId-queue" to the most recent cluster usage report for that cluster.
+	// TODO: These should be stored in a database instead of in-memory.
+	// usageByExecutorId map[string]*schedulerobjects.QueueClusterResourceUsage
+	usageByQueueAndExecutor map[string]map[string]*schedulerobjects.QueueClusterResourceUsage
```
d80tb7, line 51: We'll need to do this before we merge to master won't we- I presume we'll just use redis initially?

internal/common/resource.go
```diff
@@ -166,6 +166,7 @@ func (a ComputeResources) AsFloat() ComputeResourcesFloat {
 }
 
 func QuantityAsFloat64(q resource.Quantity) float64 {
+	q.AsApproximateFloat64()
```
d80tb7, line 169: what's this line?

internal/executor/utilisation/cluster_utilisation.go
```diff
@@ -389,3 +389,37 @@ func GetAllocationByQueue(pods []*v1.Pod) map[string]common.ComputeResources {
...
+
+func GetAllocationByQueueAndPriority(pods []*v1.Pod) map[string]map[int32]common.ComputeResources {
+	rv := make(map[string]map[int32]common.ComputeResources)
+	for _, pod := range pods {
+
+		// Get the name of the queue this pod originated from,
+		// ignoring pods for which we can't find the queue.
+		queue, present := pod.Labels[domain.Queue]
+		if !present {
+			log.Errorf("Pod %s found not belonging to a queue, not reporting its allocation", pod.Name)
```
d80tb7, line 401: this should probably be a warn

internal/scheduler/legacyscheduler.go
```diff
@@ -0,0 +1,399 @@
...
+	"github.com/G-Research/armada/internal/armada/configuration"
+	"github.com/G-Research/armada/internal/armada/scheduling"
+	"github.com/G-Research/armada/internal/common"
+	"github.com/G-Research/armada/internal/common/logging"
+	"github.com/G-Research/armada/internal/scheduler/schedulerobjects"
+	"github.com/G-Research/armada/pkg/api"
+	"github.com/G-Research/armada/pkg/armadaevents"
+)
+
+type LegacyScheduler struct {
```
d80tb7, line 26: I think you're going to need some mutexes here (or make sure this code is marked as not threadsafe and get callers to use a mutex)

internal/scheduler/legacyscheduler.go
```diff
@@ -0,0 +1,399 @@
...
+	// Maps queue name to a bool, which is true if there are no more
+	// jobs in this queue, beyond those already downloaded.
+	gotAllQueuedJobsByQueue := make(map[string]bool)
+
+	// Random seed used when selecting queues.
+	// Provided as a paremeter to ensure the scheduler is deterministic when testing.
+	seed := c.InitialSeed
+
+	// Schedule jobs one at a time.
+	for (c.SchedulingConfig.MaximumJobsToSchedule == 0 || numJobsToLease < c.SchedulingConfig.MaximumJobsToSchedule) && len(totalResourcesByQueue) > 0 && consecutiveIterationsWithNoJobsLeased < len(totalResourcesByQueue) {
```
d80tb7, line 93: If `MaximumJobsToSchedule` is 0  then we schedule jobs forever?!

internal/scheduler/legacyscheduler.go
```diff
@@ -0,0 +1,399 @@
...
+		}
+
+		fmt.Println()
+		fmt.Println("=======")
+
+		fmt.Println("iteration ", seed)
+
+		// Select a queue to schedule job from.
+		// Queues with fewer resources allocated to them are selected with higher propability.
+		seed += 1
```
d80tb7, line 109: I'm not sure this is going to give as much randomness as we want.  I think it would be much better if we just made a threadsafe `rand.Rand()` and injected that.

internal/scheduler/legacyscheduler.go
```diff
@@ -0,0 +1,399 @@
...
+	for _, queue := range queues {
+		share := shares[queue]
+		current += share
+		if current >= pick {
+			return queue, share / sum
+		}
+
+	}
+	log.Error("Could not randomly pick a queue, this should not happen!")
+	queue := queues[len(queues)-1]
```
d80tb7, line 385: why do we pick a queue here when we know something has gone wrong? Why not return an error?

internal/scheduler/legacyscheduler.go
```diff
@@ -0,0 +1,399 @@
...
+	if len(shares) == 0 {
+		return "", 0
+	}
+
+	// Generate a random number between 0 and sum.
+	sum := 0.0
+	for _, share := range shares {
+		sum += share
+	}
+	pick := sum * rand.New(rand.NewSource(seed)).Float64()
```
d80tb7, line 368: we should probably assert  sum isn't  zero here else we will panic later on.  I think the current caller of this function  won't ever pass a zero, but better to be safe than sorry!

internal/scheduler/nodedb.go
```diff
@@ -0,0 +1,299 @@
...
+
+func (nodeDb *NodeDb) dominantResource(req *schedulerobjects.PodRequirements) string {
+	dominantResourceType := ""
+	dominantResourceFraction := 0.0
+	for t, q := range req.ResourceRequirements.Requests {
+		available, ok := nodeDb.totalResources[string(t)]
+		if !ok {
+			return string(t)
+		}
+		f := q.AsApproximateFloat64() / available.AsApproximateFloat64()
```
d80tb7, line 178: can this panic is available is zero?

internal/scheduler/schedulerobjects/nodetype.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+}
+
+// nodeTypeIdFromTaintsAndLabels generates an id that is unique for each combination
+// of taints, labels, and unset labels. We insert "$" characters to achieve this,
+// since taints and labels are not allowed to contain this character.
+// See:
+// https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
+// https://man.archlinux.org/man/community/kubectl/kubectl-taint.1.en
+func nodeTypeIdFromTaintsAndLabels(taints []v1.Taint, labels, unsetWellKnownLabels map[string]string) string {
```
d80tb7, line 48: ho bad does this end up being when you have say 10k with a thousand labels each?  I'm guessing  it's fine, but just thinking of the current case where we will be doing  this on every call?

internal/armada/server/lease.go
```diff
@@ -176,76 +187,37 @@ func (q *AggregatedQueueServer) StreamingLeaseJobs(stream api.AggregatedQueue_St
...
-		MinimumJobSize:      req.MinimumJobSize,
-		Nodes:               req.Nodes,
-	}
-	nodeResources := scheduling.AggregateNodeTypeAllocations(req.Nodes)
-	clusterSchedulingInfo := scheduling.CreateClusterSchedulingInfoReport(leaseRequest, nodeResources)
-	err = q.schedulingInfoRepository.UpdateClusterSchedulingInfo(clusterSchedulingInfo)
-	if err != nil {
-		return err
+	// Return no jobs if we don't have enough work.
+	var res common.ComputeResources = req.Resources
```
d80tb7, line 195: If we hit this then we're not going update usage?

internal/armada/server/lease.go
```diff
@@ -176,76 +187,37 @@ func (q *AggregatedQueueServer) StreamingLeaseJobs(stream api.AggregatedQueue_St
...
-	activePoolClusterReports := scheduling.FilterPoolClusters(req.Pool, activeClusterReports)
-	activePoolCLusterIds := scheduling.GetClusterReportIds(activePoolClusterReports)
-	clusterPriorities, err := q.usageRepository.GetClusterPriorities(activePoolCLusterIds)
-	if err != nil {
-		return err
-	}
+	// Get jobs to be leased.
+	// TODO: The lease call resource reporting logic should be moved here.
+	var jobs []*api.Job
+	if rand.Float64() < q.schedulingConfig.ProbabilityOfUsingNewScheduler {
```
d80tb7, line 203: we should log clearly what scheduler we've chosen

internal/armada/server/lease.go
```diff
@@ -176,76 +187,37 @@ func (q *AggregatedQueueServer) StreamingLeaseJobs(stream api.AggregatedQueue_St
...
+		// New scheduler.
+		// But we need to call the old scheduler with leaseJobs = false because other component rely on its reporting.
+		_, err = q.legacyGetJobs(stream.Context(), req, false)
```
d80tb7, line 206: what exactly does this do? specifically I'm worried that we already drive redis quite hard, and it looks like we might be doubling the load here?

d80tb7: I  checked and  it doesn't seem *too* bad

internal/armada/server/lease.go
d80tb7: Also if `q.schedulingConfig.ProbabilityOfUsingNewScheduler` is very low then we won't be saving usage reports often enough to stop them expiring.  I think the logic should be:

* Update usage reports (both new and old schedulers)
* Check if there's enough free resource to even bother scheduling
* If enough resource, choose a scheduler and scheudle

That way we get accurate usage reports regardless of what's going on

internal/armada/server/lease.go
```diff
@@ -334,6 +306,236 @@ func (q *AggregatedQueueServer) StreamingLeaseJobs(stream api.AggregatedQueue_St
...
+		return nil, err
+	}
+	sched.SchedulingReportsRepository = q.SchedulingReportsRepository
+	sched.MinimumJobSize = req.MinimumJobSize
+
+	// Give Schedule() a 3 second shorter deadline than ctx,
+	// to give it a chance to finish up before ctx is cancelled.
+	if deadline, ok := ctx.Deadline(); ok {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithDeadline(ctx, deadline.Add(-3*time.Second))
```
d80tb7, line 473: do we need to do some sanity checking here to see what ctx.Deadline is?  If it's less than 3 seconds then bad things  are going to happen.

internal/armada/server/lease.go
```diff
@@ -334,6 +306,236 @@ func (q *AggregatedQueueServer) StreamingLeaseJobs(stream api.AggregatedQueue_St
...
+		if err != nil {
+			logging.WithStacktrace(log, err).Error("failed to update cluster usage")
+		}
+	}
+
+	// If we successfully leased jobs, those have to be sent to the executor,
+	// since the leases will have been written into Redis.
+	if len(jobs) > 0 {
+		if err != nil {
+			logging.WithStacktrace(log, err).Error("failed to schedule jobs")
```
d80tb7, line 496: I think we should have a better error message here.  I *think* this means that we could schedule *some* jobs, but we encountered an error before we scheduled as many jobs as we wanted.

internal/armada/server/lease.go
```diff
@@ -334,6 +306,236 @@ func (q *AggregatedQueueServer) StreamingLeaseJobs(stream api.AggregatedQueue_St
...
+		return jobs, nil
+	} else if err != nil {
+		return nil, err
+	}
+
+	return jobs, nil
+}
+
+// aggregateUsage Creates a map of usage first by cluster and then by queue
+// Note that the desired cluster is excluded as this will be filled in later as are clusters that are not in the
```
d80tb7, line 507: this comment is out of date (my fault) we no longer filter the cluster out.

internal/armadactl/scheduling.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (a *App) GetQueueSchedulingReport(queue string) error {
+	return client.WithSchedulerReportingClient(a.Params.ApiConnectionDetails, func(c schedulerobjects.SchedulerReportingClient) error {
+		ctx, cancel := common.ContextWithDefaultTimeout()
+		defer cancel()
+		report, err := c.GetQueueReport(ctx, &schedulerobjects.Queue{Name: queue})
+		if err != nil {
+			return err
+		}
+		fmt.Fprint(a.Out, report.Report)
```
d80tb7, line 19: remove?  Change to log?

internal/common/eventutil/sequence_from_message_test.go
```diff
@@ -0,0 +1,73 @@
+package eventutil
+
+// import (
```
d80tb7, line 3: remove commented out code?

internal/executor/utilisation/cluster_utilisation_test.go
```diff
@@ -229,11 +331,17 @@ func TestGetAllocatedResourceByNodeName(t *testing.T) {
...
 }
 
-func hasKey(value map[string]common.ComputeResources, key string) bool {
-	_, ok := value[key]
+func hasKey[K comparable, V any](m map[K]V, key K) bool {
+	_, ok := m[key]
 	return ok
 }
 
+// TODO: Remove
```
d80tb7, line 339: remove?

internal/scheduler/api.go
```diff
@@ -28,7 +27,9 @@ type ExecutorApi struct {
 	api.UnimplementedAggregatedQueueServer
 	// Embed the Redis-backed event server.
 	// Provides methods for dual-publishing events etc.
-	*server.EventServer
+	//
+	// TODO: Can't import due to import cycle.
+	// *server.EventServer
```
d80tb7, line 32: remove?

internal/scheduler/legacyscheduler.go
```diff
@@ -0,0 +1,619 @@
...
+}
+
+func (it *QueuedJobsIterator) Next() (*api.Job, error) {
+	// Once this function has returned error,
+	// it will return this error on every invocation.
+	if it.err != nil {
+		return nil, it.err
+	}
+
+	// Get one job that was loaded asynchrounsly.
```
d80tb7, line 76: (minor) typo

internal/scheduler/legacyscheduler.go
```diff
@@ -0,0 +1,619 @@
...
+	// If a NodeDb is provided, try to find a node on which this job can be scheduled.
+	if it.NodeDb != nil {
+		podReport, err := it.selectNodeForPod(ctx, jobId, job)
+		if err != nil {
+			jobSchedulingReport.UnschedulableReason = err.Error()
+			return nil, err
+		}
+		jobSchedulingReport.PodSchedulingReports = append(jobSchedulingReport.PodSchedulingReports, podReport)
+		if podReport.Node == nil {
+			jobSchedulingReport.UnschedulableReason = "pod does not fit on any node"
```
d80tb7, line 293: will it be obvious why it doesn't fit?  If not we don't need to fix this now, just an idea for the future.

internal/scheduler/legacyscheduler.go
```diff
@@ -0,0 +1,619 @@
...
+	}
+	return report, nil
+}
+
+type LegacyScheduler struct {
+	SchedulingConfig configuration.SchedulingConfig
+	// Executor for which we're currently scheduling jobs.
+	ExecutorId string
+	// Used for matching pods with nodes.
+	NodeDb *NodeDb
```
d80tb7, line 363: do we need to export all this outside the module?

internal/scheduler/legacyscheduler.go
```diff
@@ -0,0 +1,619 @@
...
+	// Try to create leases.
+	jobs := make([]*api.Job, 0, numJobsToLease)
+	for queue, jobsToLease := range jobsToLeaseByQueue {
+
+		// TryLeaseJobs returns a list of jobs that were successfully leased.
+		// For example, jobs concurrently leased to another executor are skipped.
+		//
+		// TODO: Reports generated above will be incorrect if creating the lease fails.
+		successfullyLeasedJobs, err := c.JobRepository.TryLeaseJobs(c.ExecutorId, queue, jobsToLease)
+		if err != nil {
```
d80tb7, line 545: I think that you can fail to lease jobs if another instance beat you to it.  In this case logging errors with stacktraces is a bit disingenuous.  I guess the old implementation did something similar as the  problem  is with the JobRepo interface, but we should probably fix in a subsequent pr.

internal/scheduler/legacyscheduler.go
```diff
@@ -0,0 +1,619 @@
...
+			return queue, share / sum
+		}
+
+	}
+	log.Error("Could not randomly pick a queue, this should not happen!")
+	queue := queues[len(queues)-1]
+	return queue, shares[queue] / sum
+}
+
+func podSpecFromJob(job *api.Job) *v1.PodSpec {
```
d80tb7, line 609: don't we already have a function for this in `util` somewhere?

internal/scheduler/nodedb_test.go
```diff
@@ -0,0 +1,739 @@
...
+			Priority:             0,
+			ResourceRequirements: v1.ResourceRequirements{Requests: r},
+		})
+		assert.NoError(t, err)
+		assert.Nil(t, report.Node)
+	}
+}
+
+// Test that some resource we don't know about causes an error:
+// TODO:  Is returning an error here correct?
```
d80tb7, line 74: we get an  error in this test- is that correct?

internal/scheduler/nodedb_test.go
```diff
@@ -0,0 +1,739 @@
...
+		Tolerations: []v1.Toleration{
+			{
+				Key:   "gpu",
+				Value: "true",
+			},
+		},
+	}
+}
+
+// func testA100Job(priority int32) *schedulerobjects.PodRequirements {
```
d80tb7, line 709: remove?

internal/scheduler/legacyscheduler_test.go
```diff
@@ -0,0 +1,1017 @@
...
+				"A": append(testNSmallCpuJob(0, 1), testNLargeCpuJob(0, 1)...),
+			},
+			PriorityFactorsByQueue: map[string]float64{
+				"A": 1,
+			},
+			ExpectedIndicesByQueue: map[string][]int{
+				"A": {1},
+			},
+		},
+		"minimum job size": {
```
d80tb7, line 765: do we need tests for per-user limits and per-scheduling-rounnd limits?

internal/scheduler/nodedb.go
```diff
@@ -0,0 +1,332 @@
...
+	// Allowed pod priorities in sorted order.
+	// Because the number of database indices scales linearly with the number of distinct priorities,
+	// the efficiency of the NodeDb relies on the number of distinct priorities being small.
+	priorities       []int32
+	indexedResources map[string]interface{}
+	// Total amount of resources, e.g., "cpu", "memory", "gpu", managed by the scheduler.
+	// Computed approximately by periodically scanning all nodes in the db.
+	totalResources map[string]*resource.Quantity
+	// Set of node types for which there exists at least 1 node in the db.
+	NodeTypes map[string]*schedulerobjects.NodeType
```
d80tb7, line 33: do we need to export all this outside the module?

internal/scheduler/nodedb.go
```diff
@@ -0,0 +1,332 @@
...
+		return nil, err
+	}
+
+	for obj := it.Next(); obj != nil; obj = it.Next() {
+		node := obj.(*schedulerobjects.Node)
+		if node == nil {
+			break
+		}
+		// TODO: Use the score when selecting a node.
+		nodeDb.mu.Lock()
```
d80tb7, line 110: what's the mutex protecting here? Let's add a comment.

internal/scheduler/nodedb.go
```diff
@@ -0,0 +1,332 @@
...
+		NodesByJob:       make(map[uuid.UUID]map[string]interface{}),
+		JobsByNode:       make(map[string]map[uuid.UUID]interface{}),
+		AssignedByNode:   make(map[string]schedulerobjects.AssignedByPriorityAndResourceType),
+	}, nil
+}
+
+// Upsert will update the node db with the given nodes.
+func (nodeDb *NodeDb) Upsert(nodes []*schedulerobjects.Node) error {
+	txn := nodeDb.Db.Txn(true)
+	defer txn.Abort()
```
d80tb7, line 263: This is a bit weird- is it idiomatic to call commit() then abort() on success?

internal/scheduler/nodeiteration.go
```diff
@@ -0,0 +1,207 @@
...
+		})
+	}
+	return &NodeTypesResourceIterator{
+		priority: priority,
+		resource: resource,
+		pq:       pq,
+	}, nil
+}
+
+func (it *NodeTypesResourceIterator) WatchCh() <-chan struct{} {
```
d80tb7, line 47: Is this needed to conform to some interface?  If so le's doc what the interface is and why it's ok not to implement it.

internal/scheduler/nodeiteration_test.go
```diff
@@ -0,0 +1,536 @@
...
+
+func testNTaintedtGpuNode(n int, priorities []int32) []*schedulerobjects.Node {
+	rv := make([]*schedulerobjects.Node, n)
+	for i := 0; i < n; i++ {
+		rv[i] = testTaintedGpuNode(priorities)
+	}
+	return rv
+}
+
+// func testNA100Node(n int, priorities []int32) []*schedulerobjects.Node {
```
d80tb7, line 399: remove?

internal/scheduler/reports.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+func (repo *SchedulingReportsRepository) GetJobReport(ctx context.Context, jobId *schedulerobjects.JobId) (*schedulerobjects.JobReport, error) {
+	jobUuid, err := uuidFromUlidString(jobId.Id)
+	if err != nil {
+		return nil, err
+	}
+	report, ok := repo.GetJobSchedulingReport(jobUuid)
+	if !ok {
+		return nil, &armadaerrors.ErrNotFound{
```
d80tb7, line 66: should this (and the one above) be errors?

internal/scheduler/unused/scheduler.go
```diff
@@ -0,0 +1,278 @@
+package scheduler
+
+// Scheduler contains the current state of the Armada scheduler.
```
d80tb7, line 3: do we want to commit  this?

severinson: Let's not. I've deleted it.

internal/scheduler/nodeiteration_test.go
severinson: I want to add tests using these. So I'd prefer to keep them for now.

internal/scheduler/reports.go
severinson: Yes. So that gRPC error codes are set correctly.

internal/scheduler/nodeiteration.go
```diff
@@ -0,0 +1,207 @@
...
+	panic("not implemented")
```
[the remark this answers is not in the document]
severinson: I think this is fine. It should never be called and we want to ensure it crashes if it is. We just need it to satisfy the interface.

internal/scheduler/nodeiteration.go
severinson: Agreed. Let me to that.

internal/scheduler/nodedb.go
severinson: It's the way they recommend in their docs. `Abort` is a no-op if read-only or already committed.

internal/scheduler/nodedb.go
severinson: Agreed

internal/scheduler/nodedb.go
severinson: Not really. We can change it.

internal/scheduler/nodedb_test.go
severinson: Should be fixed now

internal/scheduler/legacyscheduler.go
severinson: Yes, I agree

internal/scheduler/legacyscheduler.go
severinson: Let's review it

internal/scheduler/legacyscheduler.go
severinson: Yes, a more detailed reason is stored in the `PodSchedulingReport` embedded in the `JobSchedulingReport`.

internal/armada/server/lease.go
severinson: The call will just time out I think? There should be no harm.

internal/armada/server/lease.go
severinson: Addressed

d80tb7 (approved):

PR #1580, opened 2022-09-30
localdev with pulsar

suprjinx:


kannon92:
I created this folder because I was blocked by testing armada with other external services.  I had to add back the local_config folder because I need Armada to be on my host network so airflow and other projects can talk to it.  

If we require any dev integration to be added via docker compose that could become unwiedely.  I support this change but there are different use-cases to keep in mind before we add everything to docker compose.

suprjinx:
I thought /docs/dev/local_config was only really referenced by the aramada-compose.yaml. I moved that compose file and the configs here (/localdev) to just kind of clean-slate the compose effort and make it simpler to "turn key" with minimal command line stuff (just `localhdev/run.sh`).

> I created this folder because I was blocked by testing armada with other external services. I had to add back the local_config folder because I need Armada to be on my host network so airflow and other projects can talk to it.
> 
> If we require any dev integration to be added via docker compose that could become unwiedely. I support this change but there are different use-cases to keep in mind before we add everything to docker compose.

kannon92:
> I thought /docs/dev/local_config was only really referenced by the aramada-compose.yaml. I moved that compose file and the configs here (/localdev) to just kind of clean-slate the compose effort and make it simpler to "turn key" with minimal command line stuff (just `localhdev/run.sh`).
> 
> > I created this folder because I was blocked by testing armada with other external services. I had to add back the local_config folder because I need Armada to be on my host network so airflow and other projects can talk to it.
> > If we require any dev integration to be added via docker compose that could become unwiedely. I support this change but there are different use-cases to keep in mind before we add everything to docker compose.

local_config is not referenced by the armada-compose file.

I believe that docs/dev/config is what is actually used in the armada-compose file.  The local_config is all about trying to run outside of a container, hence why I use localhost and not reference any containers by name.

suprjinx:
> 
> I believe that docs/dev/config is what is actually used in the armada-compose file. The local_config is all about trying to run outside of a container, hence why I use localhost and not reference any containers by name.

okay i'll restore that folder in this branch

localdev/scripts/pulsar.sh
```diff
@@ -0,0 +1,18 @@
...
+/pulsar/bin/pulsar-admin namespaces create armada/armada
+/pulsar/bin/pulsar-admin topics delete-partitioned-topic persistent://armada/armada/events -f || true
+/pulsar/bin/pulsar-admin topics create-partitioned-topic persistent://armada/armada/events -p 2
+
+# Disable topic auto-creation to ensure an error is thrown on using the wrong topic
+# (Pulsar automatically created the public tenant and default namespace).
+/pulsar/bin/pulsar-admin namespaces set-auto-topic-creation public/default --disable
+/pulsar/bin/pulsar-admin namespaces set-auto-topic-creation armada/armada --disable
+
+sleep infinity
```
kannon92, line 18: If this is needed, I think a comment may be needed.  An infinite sleep scares me.

.gitignore
```diff
@@ -68,3 +68,4 @@ typings.py
 third_party/airflow/docs/build/
 third_party/airflow/build/
 build/ssl/certs/ca-certificates.crt
+localdev/volumes/go/pkg/
```
kannon92, line 71: Maybe you should add volumes/go also

localdev/scripts/pulsar.sh
suprjinx: ok will add comment

.gitignore
suprjinx: good idea

suprjinx: well i wanted to make sure the volumes/go folder is created when you checkout -- there's a .keep file in there

kannon92: Ah. I thought the .keep was something that shouldn't be committed.  No worries.

makefile
```diff
@@ -78,7 +78,7 @@ DOCKER_GOPATH_DIR = $(word 1,$(DOCKER_GOPATH_TOKS))
 # This is used to generate published artifacts; to raise the golang version we publish and run
 # tests-in-docker against, you have to update the tag of the image used here.
 GO_CMD = docker run --rm $(DOCKER_RUN_AS_USER) -v ${PWD}:/go/src/armada -w /go/src/armada $(DOCKER_NET) \
-	-e GOPROXY -e GOPRIVATE -e GOCACHE=/go/cache -e INTEGRATION_ENABLED=true -e CGO_ENABLED=0 -e GOOS=linux -e GARCH=amd64 \
+	-e GOPROXY -e GOPRIVATE -e GOCACHE=/go/cache -e INTEGRATION_ENABLED=true -e CGO_ENABLED=0 -e GOOS=darwin -e GARCH=arm64 \
```
kannon92, line 81: This is probably causing your integration test failures.  We probably should not change this since our CI is running on linux with amd.

makefile
```diff
@@ -620,3 +620,8 @@ build-dev-jobservice:
 	docker build --build-arg APP_UID=$(shell id -u) --build-arg APP_GID=$(shell id -g) \
 		$(dockerFlags) -t armada-jobservice -f ./docs/dev/build/jobservice/Dockerfile ./.build/jobservice
 
+tests-dev:
```
kannon92, line 623: I think you should open this as separate PR.  This looks useful but unrelated to this PR?

suprjinx: thanks will do

localdev/config/armada/base.yaml
```diff
@@ -0,0 +1,29 @@
+corsAllowedOrigins: ["http://localhost:3000", "http://localhost:8089"]
```
Sharpz7, line 1: Is this going to work as expected being localhost?

localdev/config/binoculars/base.yaml
```diff
@@ -0,0 +1 @@
+corsAllowedOrigins: ["http://localhost:3000", "http://localhost:8089"]
```
Sharpz7, line 1: Is this going to work as expected being localhost?

localdev/run.sh
```diff
@@ -0,0 +1,30 @@
+#! /bin/bash
+INFRA_SVCS="redis postgres pulsar stan"
```
Sharpz7, line 2: Do we want this controlled from ENV vars or something? I feel like we want users to not have to edit scripts in the repo to stop stuff being accidentally commiteed

localdev/scripts/binoculars-start.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# this will update kind kube config copied from host to work inside docker
+if [ -f /.dockerenv ]; then
+    DOCKER_HOST=$(ping -c 1 host.docker.internal | grep -m 1 -o -E '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}')
```
Sharpz7, line 5: host.docker.internal has native support to only windows and mac by default, and for linux the Docker version installed needs to be relatively new. We probably want to document all this? https://medium.com/@TimvanBaarsen/how-to-connect-to-the-docker-host-from-inside-a-docker-container-112b4c71bc66

Sharpz7: Also, I do not 100% understand why this is needed. Is this to make sure kubectl points to the correct place?

suprjinx: yes, this is so armada services that talk to kubernetes control plane can do so using .kube/config of the host computer -- and they need to use a host name that matches the tls cert. Will add to comment (and extract for reuse)

localdev/run.sh
suprjinx: not sure what you mean?

localdev/config/binoculars/base.yaml
suprjinx: maybe? this will be the address used in your browser, likely localhost

localdev/config/armada/base.yaml
suprjinx: yes it should since those will be host names used in browser

localdev/config/binoculars/base.yaml
Sharpz7: We maybe want to support configuring this with ENV vars or something similar?

localdev/run.sh
Sharpz7: Just to check, these variables determine what components are run with docker-compose?

If a user decides they want to change it, and then writes a commit, those changes to this script will be sent in that PR. Ideally, we don't want these scripts to need any user-edits, and instead control them from enviromental variables / a config file that is local to the user. Does this make sense?

localdev/scripts/armada-start.sh
```diff
@@ -0,0 +1 @@
+go run ./cmd/armada/main.go --config ./localdev/config/armada/base.yaml --config ./localdev/config/armada/pulsar.yaml
```
Sharpz7, line 1: Again, do we want to use env vars or similar to support custom config? For example, my armada_jupyter stuff needs an option to be changed in the executor

makefile
suprjinx: removed -- not necessary to specify arch now

localdev/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
...
+  postgres:
+    container_name: postgres
+    image: postgres
+    environment:
+      - "POSTGRES_PASSWORD=psw"
+    ports:
+      - "5432:5432"
+
+  pulsar:
+    image: kezhenxu94/pulsar
```
richscott, line 23: What do you think about going ahead and using profiles for each of these - have a `profiles: ["macos"]` and `profiles: ["linux"]` block? People will have to get used to specifying, e.g. `--profile macos` when invoking `docker compose`, but shell command history will make that pretty easy after a while.

richscott (commented):
I had just a question about going ahead and using profiles for the pulsar image selection. Also, I don't understand the `scripts/pulsar.sh` empty file.

localdev/config/armada/base.yaml
```diff
@@ -0,0 +1,29 @@
...
+    create_queue: ["everyone"]
+    delete_queue: ["everyone"]
+    cancel_jobs: ["everyone"]
+    cancel_any_jobs: ["everyone"]
+    reprioritize_jobs: ["everyone"]
+    reprioritize_any_jobs: ["everyone"]
+    watch_events: ["everyone"]
+    watch_all_events: ["everyone"]
+    execute_jobs: ["everyone"]
+pulsar:
```
kannon92, line 18: I'm unsure what this option does since you also add pulsar.  I think we should remove this as its unclear to me if we actually enabled pulsar.

localdev/scripts/armada-start.sh
kannon92: I agree.  I think its fine to have dev input their own custom configs if they need to make changes or testing.

localdev/config/armada/base.yaml
suprjinx: okay will remvoe

localdev/docker-compose.yaml
suprjinx: ok I'll try it -- the run.sh script can default the right one for you, too

suprjinx: well profiles get a little messy -- we'd need "downstream" profile config for things that depend_on pulsar -- so alternative armada-servers, etc. Trying an env-var that defaults to linux but gets the arm64 image name in run.sh if needed.

kannon92 (approved):

richscott (approved):
Excellent work!

PR #1587, opened 2022-10-03
docs: document preemption in Armada

dejanzele:


severinson:
I'd prefer moving the docs for preemption config into the config itself (that includes everything except the example job). I.e., put it in the `PreemptionConfig` definition in the code. That way we ensure the code and docs are in synch. Creating standalone docs runs the risk of the docs becoming outdated without us noticing. For a mature project, that's less of an issue. Since there are fewer changes. But for Armada there are lots of changes.

Re the example job. I think it's better we but it into the user guide.

suprjinx:
I think it looks good but I can get behind Albin's suggestion. Example yaml might be better as standalone file which we can just directly submit to armada if wanted.

kannon92:
I think the config comments and examples look good.  I was hoping to see a design of how we implemented Preemption in Armada.  

I'd like to also see what the problems were with the current premption API so we can potentially bring this to the Kubernetes project as a potential feature request or fix.

dejanzele:
@kannon92 most probably the backoffice side of Preemption will get completely redesigned soon, so current diagrams won't be relevant, but the client-facing interface will remain the same.

I plan on updating the docs as soon as preemption is redesigned

docs/quickstart/job-queue-a-preemptive.yaml
```diff
@@ -0,0 +1,23 @@
+queue: queue-a
+jobSetId: job-set-1
+jobs:
+  - priority: 1
+    podSpec:
+      priorityClassName: armada-preemptive-low-priority
```
kannon92, line 6: The config above doesn't define this priorityClassName.  Is this a default one defined somewhere else?

dejanzele: @kannon92 
Does this make more sense - https://github.com/G-Research/armada/pull/1587/files#diff-566e8c3f80ad8aeb86911519a22c2e5d62e7cd341716e98bd21fd92b7cd2361cR1-R11 ?

kannon92: I think we should probably write something in our user guide about this. 

So I'm thinking we add a section to our user guide about preemption.  

1) Use that example for PriorityClass.
2) Reference the example job.

dejanzele: I agree , but we don't have a section for features, like User Guide or Manual, that would be a good addon for docs.

ArgoCD has a good example: https://argo-cd.readthedocs.io/en/stable/

kannon92 (approved):
The user guide looks great!

kannon92 (approved):

PR #1597, opened 2022-10-07
feat: add support for node reserved resources

dejanzele:


samclark (commented):
I spoke to @dejanzele on slack but wanted to commenting here to keep track of our thoughts. After reviewing this PR I'm fairly confident it would work fine but we have agreed that it could be improved by moving the logic / config from the server to the executor(s). The reserved amount can then simply be removed from the allocatable amounts reported by the executors (which we think stays true to the meaning on that property. i.e. the total capacity minus all the resources reserved by the system ... or in other words the resource available to be used). This also allows us to define the reserved amount on a per cluster basis which might be useful.

samclark (commented):
I moving this to the executor is the way to go but have some small requests.

- move the `NodeReservedResources` into the `KubernetesConfiguration` struct.
- leave the `UtilisationService` interface as it is and instead pass `NodeReservedResources` into the constructor of the concrete `ClusterUtilisationService` type. Store this value in that struct and use it in the existing (now unchanged signature) `GetAvailableClusterCapacity` method.

samclark (approved):

suprjinx (approved):
LGTM

kannon92 (approved):

PR #1610, opened 2022-10-11
Don't requeue job if we've already processed it

### LINKED ISSUE #1600: Don't Run A Job Again If Message Is Replayed
Currently when we receive a `SubmitJob` message in `SubmitFromLog` we do the following:

1. Put the Message onto the redis queue so that the scheduler can schedule it
2. Ack the message

What can happen here is that the message can be put onto redis, but the message is never acked (e.g. if the process is halted). In this case, the message will be redelivered, which can cause to to queue (and thus run) the same job twice.

We should ensure that we don't ever run a job twice, which most likely means ignoring the message, the second time we receive it.  One mechanism of doing this (not sure it would be the best) would be:
* at the same time we add the message to the redis queue, we also add a key to say that we queued the message
* Before queueing any messages, we check to see if this key exists.
For this to work reliably we'd need to ensure that the above two options were transactional (not sure if redis supports this)

d80tb7:
This fixes #1600 

I've modified the addJobs script so that:
- We create marker key (expiration 1 week) that shows that we've added the job to redis
- We check the presence of the marker key before adding the job to the queue.  If the key exists then we don't re-add.

Note that I use a marker key rather than the actual job key, because the job key is several KB in size and so we want to expire that much quicker than the marker key.

### DIFF
--- a/internal/armada/repository/job.go
+++ b/internal/armada/repository/job.go
@@ -27,6 +27,7 @@ const (
 	jobClusterMapKey   = "Job:ClusterId" //                    - map jobId -> cluster
 	jobRetriesPrefix   = "Job:Retries:"  // {jobId}            - number of retry attempts
 	jobClientIdPrefix  = "job:ClientId:" // {queue}:{clientId} - corresponding jobId
+	jobExistsPrefix    = "Job:added"     // {jobId}            - flag to say we've added the job
 	keySeparator       = ":"
 )
 
@@ -95,6 +96,7 @@ func NewRedisJobRepository(
 type SubmitJobResult struct {
 	JobId             string
 	SubmittedJob      *api.Job
+	AlreadyProcessed  bool
 	DuplicateDetected bool
 	Error             error
 }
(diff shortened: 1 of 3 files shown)


internal/armada/repository/job.go
```diff
@@ -1062,12 +1068,19 @@ local jobKey = KEYS[2]
...
 local jobClientIdKey = KEYS[5]
+local jobExistsKey = KEYS[6]
 
 local jobId = ARGV[1]
 local jobPriority = ARGV[2]
 local jobData = ARGV[3]
 local clientId = ARGV[4]
 
+
+local jobExists = redis.call('EXISTS', jobExistsKey, 'EX', 604800)
```
severinson, line 1079:
```suggestion
local jobExists = redis.call('EXISTS', jobExistsKey)
```

internal/armada/repository/job.go
```diff
@@ -1076,6 +1089,7 @@ if clientId ~= '' then
 	redis.call('SET', jobClientIdKey, jobId, 'EX', 14400)
 end
 
+redis.call('SET', jobExistsKey, '1')
```
severinson, line 1092:
```suggestion
redis.call('SET', jobExistsKey, '1', 'EX', 604800)
```

severinson (approved):

PR #1611, opened 2022-10-11
Add PGX Connection errors to list of retryable errors

d80tb7:
PGX  returns an unexported `connectionError` under various connection scenarios (e.g. validateConect fails).  We want to retry these errors so I've added a string match on the error string for now.  I'll raise an issue with PGX to get this type exported so we can make this more robust

### DIFF
--- a/internal/common/armadaerrors/errors.go
+++ b/internal/common/armadaerrors/errors.go
@@ -509,6 +509,12 @@ func IsRetryablePostgresError(err error) bool {
 		_, ok := retryablePostgresErrors[err.Code]
 		return ok
 	}
+
+	// This is quite nasty: the connectError reported by pgx isn't exported so instead we use a string match
+	if strings.Contains(err.Error(), "failed to connect") {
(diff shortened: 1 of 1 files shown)


kannon92:
LGTM: 

> I'll raise an issue with PGX to get this type exported so we can make this more robust

This is something that our organization (GR-OSS) could help address if its a big deal.

Nice job on being proactive on identifying issues in OSS projects.

internal/common/armadaerrors/errors.go
```diff
@@ -509,6 +509,12 @@ func IsRetryablePostgresError(err error) bool {
 		_, ok := retryablePostgresErrors[err.Code]
 		return ok
 	}
+
+	// This is quite nasty: the connectError reported by pgx isn't exported so instead we use a string match
```
kannon92, line 513: Is there a way to add some test coverage around this?

severinson (approved):

PR #1612, opened 2022-10-11
Unit tests for kerberos auth service

ClifHouck:
As requested by @dave-gantenbein. Came across the fact that the kerberos auth service had no unit tests part of my effort to improve auth error messages.


internal/common/auth/authorization/kerberos_test.go
```diff
@@ -0,0 +1,287 @@
...
+
+func getContextWithEncodedKerberosToken(rawToken []byte) context.Context {
+	nMD := metautils.NiceMD{}
+	nMD.Set(
+		"authorization",
+		spnego.HTTPHeaderAuthResponseValueKey+" "+base64.StdEncoding.EncodeToString(rawToken))
+	return nMD.ToIncoming(context.Background())
+}
+
+func TestKerberosAuthenticate(t *testing.T) {
```
kannon92, line 119: Nit: Usually I see individual test cases should follow SRP.  It seems like this one function tests a lot of cases.  It makes it hard to deduce the failure in the logs with all these different test cases.  

Would you be willing to separate these into different test cases?

ClifHouck: Yep, I'll split them up.

kannon92:
1) What does your tooling say about coverage for kerberos?  I'm not familiar with this code but it'd be nice to see if we are missing some tests.

2) Can we find a subject matter expert on Kerberos to aide in the review?

I think the code LGTM but I don't really know what Kerberos is so I feel nervious approving this PR.

dave-gantenbein:
> 1. What does your tooling say about coverage for kerberos?  I'm not familiar with this code but it'd be nice to see if we are missing some tests.
> 2. Can we find a subject matter expert on Kerberos to aide in the review?
> 
> I think the code LGTM but I don't really know what Kerberos is so I feel nervious approving this PR.

Hey Kevin - I asked Clif to get the tests up from zero to at least cover the work he did. We can continue to clean it up incrementally.

kannon92 (approved):

PR #1620, opened 2022-10-12
health check

### LINKED ISSUE #1508: Expose health check as gRPC endpoint in services
`pkg/client/connection.go` includes an ArmadaHealthCheck() function -- but this appears to invoke the Armada /health endpoint via REST.

For this ticket, make health an endpoint we can invoke with the gRPC client, for both Submit and Event services.

kannon92:
Closes #1508 

1) Add Health endpoint to Submit, Event and JobService.  
2) Add some logging to jobservice to deduce what is going wrong with auth.


internal/jobservice/server/server.go
```diff
@@ -43,3 +44,15 @@ func (s *JobServiceServer) GetJobStatus(ctx context.Context, opts *js.JobService
 
 	return response, err
 }
+
+func (s *JobServiceServer) Health(ctx context.Context, _ *types.Empty) (*js.HealthCheckResponse, error) {
+	log.Info("JobService Health Check Starting")
```
suprjinx, line 49: I think this might create a lot of log noise if there are frequent health checks -- maybe shouldn't log at all

suprjinx (changes requested):
Looks good to me except perhaps the logging will be excessive

kannon92:
> Looks good to me except perhaps the logging will be excessive

@iJanki-gr asked that we add more logging to jobservice as it currently logs very little.  I was added these so we can deduce failures with health check to narrow down issues.  I'm unclear if this health check will be called a lot or just when something is setup.

internal/jobservice/server/server.go
richscott: This might get noisy, but if it gets on people's nerves, we can quickly dial it down to a `log.Debug()`.

richscott (approved):

internal/jobservice/server/server.go
kannon92: Removed.

richscott (approved):

PR #1621, opened 2022-10-13
[WIP] Store Cluster Usage in Redis

d80tb7:


internal/armada/server/lease.go
```diff
@@ -676,3 +678,48 @@ func (q *AggregatedQueueServer) getJobById(jobId string) (*api.Job, error) {
 	}
 	return jobs[0], err
 }
+
+// createUsageByQueueAndExecutor Creates a map of usage first by queue and then by cluster
+// This exists because the ClusterResourceUsageReports we store in redis have to be keyed by cluster first
+// and here we  need to index by queue first.
+// Note that the desired cluster is excluded as this will be filled in later as are clusters that are not in the
+// same pool as the desired cluster
+func (q *AggregatedQueueServer) createUsageByQueueAndExecutor(reportsByCluster map[string]*schedulerobjects.ClusterResourceUsageReport, desiredCluster string, pool string) map[string]map[string]*schedulerobjects.QueueClusterResourceUsage {
```
severinson, line 687: I guess we can remove this now?

internal/armada/server/lease.go
```diff
@@ -676,3 +678,48 @@ func (q *AggregatedQueueServer) getJobById(jobId string) (*api.Job, error) {
...
+				}
+			}
+		}
+	}
+	return usageByQueueAndExecutor
+}
+
+// createClusterUsageReport creates a schedulerobjects.ClusterResourceUsageReport suitable for storing in redis
+// from the map of QueueClusterResourceUsage which is indexed by queue and cluster
+func createClusterUsageReport(reportsByQueue map[string]map[string]*schedulerobjects.QueueClusterResourceUsage, desiredCluster string, pool string) *schedulerobjects.ClusterResourceUsageReport {
```
severinson, line 712: We can remove this now?

internal/armada/repository/usage.go
```diff
@@ -161,6 +167,41 @@ func (r *RedisUsageRepository) UpdateClusterLeased(report *api.ClusterLeasedRepo
 	return err
 }
 
+func (r *RedisUsageRepository) GetClusterQueueResourceUsage() (map[string]*schedulerobjects.ClusterResourceUsageReport, error) {
+	result, err := r.db.HGetAll(clusterQueueResourceUsageKey).Result()
+	if err != nil {
+		return nil, fmt.Errorf("[RedisUsageRepository.GetClusterQueueResourceUsage] error reading from database: %s", err)
+	}
+	reports := make(map[string]*schedulerobjects.ClusterResourceUsageReport)
```
severinson, line 175: Is this a map from queue name to report? May be good to comment.

internal/armada/repository/usage.go
```diff
@@ -161,6 +167,41 @@ func (r *RedisUsageRepository) UpdateClusterLeased(report *api.ClusterLeasedRepo
...
+		reports[k] = report
+	}
+	return reports, nil
+}
+
+func (r *RedisUsageRepository) UpdateClusterQueueResourceUsage(cluster string, resourceUsage *schedulerobjects.ClusterResourceUsageReport) error {
+
+	data, err := proto.Marshal(resourceUsage)
+	if err != nil {
+		return fmt.Errorf("[RedisUsageRepository.UpdateClusterQueueResourceUsage] error marshalling report: %s", err)
```
severinson, line 192: We should use `errors.Errorf`. Then we don't need the `[]` part either.

internal/armada/repository/usage.go
```diff
@@ -161,6 +167,41 @@ func (r *RedisUsageRepository) UpdateClusterLeased(report *api.ClusterLeasedRepo
...
+	}
+
+	pipe := r.db.TxPipeline()
+	pipe.HSet(clusterQueueResourceUsageKey, cluster, data)
```
severinson, line 196: Can we expire these automatically? In case a cluster goes away we'd like the reports to be expired.

severinson: Also, doesn't this need to somehow be keyed by both executor and queue? Since there are many per-queue reports per executor.

internal/armada/server/lease.go
severinson: Or is it just that the docstring hasn't been updated?

internal/scheduler/schedulerobjects/schedulerobjects.proto
```diff
@@ -60,6 +60,13 @@ message QueueClusterResourceUsage {
     map<int32, ResourceList> resourcesByPriority = 4 [(gogoproto.nullable) = false];
 }
 
+// A collection of QueueClusterResourceUsage
+// This is only needed to brige the gap between the redis based scheduler and the new scheduler
+message ClusterResourceUsageReport {
```
severinson, line 65: Instead of defining a new object, I'd prefer to just add the pool field to `QueueClusterResourceUsage`.

internal/armada/repository/usage.go
d80tb7: Ideally yes but the rest of the  file is using this format and I didn't want to make this change as part of this pr
I think it would be better to have a separate pr where all that's done is to change to the new way of doing errors.

internal/armada/repository/usage.go
d80tb7: `clusterQueueResourceUsageKey` is a map like object where `cluster` is the key and `data` the value.  In this case the `data` is itlsef a map  of queue -> cluster so I think we're good.

I don't think we want this key to expire, rather `Data` contains a timestamp saying when it was last updated.  If the time is too far in the past we don't consider the usage valid and therefore exclude it from scheduling (this is consistent with how we do things in the legacy usage)

internal/armada/server/lease.go
d80tb7: We still need it (but it is much  simpler).  Leave it in for now and when e2e passes I'll integrate the new ones

internal/scheduler/schedulerobjects/schedulerobjects.proto
d80tb7: I think we need a new object here- in this case because we need to store a collection (in my latest code this is a map) of `QueueClusterResourceUsage` not just a single one. I also want to add a timestamp here so that we can check if the data is out of date.  I agree that having to have a new object defined here isn't ideal but we should be able to retire it once we go to postgres.

PR #1623, opened 2022-10-13
Double defaultPulsarTimeout

ClifHouck:
As a minimum-effort way to avoid flaky pulsar test failures.


ClifHouck:
Not sure how to prove the effectiveness of this change other than to merge and see if CI runs become more stable.

ClifHouck:
I'm also happy to pursue more robust ways of improving e2e test reliability, but hopefully this stops the flaky pulsar e2e TestService test from failing sporadically.

richscott (approved):
I agree that we should work on making the integration tests much more robust in general (and investigate some sort of polling logic for verifying Pulsar's aliveness), but this is a good pragmatic fix for right now.

PR #1627, opened 2022-10-14
add support for metrics in lookout ingester

dejanzele:


internal/common/startup.go
```diff
@@ -125,7 +125,7 @@ func ServeMetricsFor(port uint16, gatherer prometheus.Gatherer) (shutdown func()
 	return ServeHttp(port, mux)
 }
 
-// ServeHttp starts a HTTP server listening on the given port.
+// ServeHttp starts an HTTP server listening on the given port.
```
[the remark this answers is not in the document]
dejanzele: Actually @kannon92, this seems to be pretty [opinionated](https://www.pristineword.com/grammar-an-http/#:~:text=It's%20an%20HTTP.,takes%20the%20article%20%22an%22.) :D

kannon92 (approved):
LGTM

kannon92 (dismissed):

internal/eventingester/ingester.go
```diff
@@ -7,6 +7,8 @@ import (
 	"regexp"
 	"sync"
 
+	"github.com/G-Research/armada/internal/lookoutingester/metrics"
```
suprjinx, line 10: maybe we should move shared metrics to pkg/ instead  of having cross-service dependency here?

internal/lookoutingester/instructions/instructions.go
```diff
@@ -55,7 +65,12 @@ func Convert(
 // resulting InstructionSet will contain all events that could be parsed, along with the mesageId of the original message.
 // In the case that no events can be parsed (e.g. the message is not valid protobuf), an empty InstructionSet containing
 // only the messageId will be returned.
-func ConvertMsg(ctx context.Context, msg *pulsarutils.ConsumerMessage, userAnnotationPrefix string, compressor compress.Compressor) *model.InstructionSet {
+func (i *Service) ConvertMsg(
```
suprjinx, line 68: "i" is a bit cryptic as a name for service reference

internal/eventingester/ingester.go
dejanzele: actually I removed it, as we have better metrics in pulsar itself

kannon92 (approved):

PR #1636, opened 2022-10-17
Yarn integration

headphonejames:
Switching to using yarn to downland and assemble dependencies. Addresses https://github.com/G-Research/armada/issues/1486


docs/dev/setup.sh
```diff
@@ -3,7 +3,7 @@
 OSTYPE=$(uname -s)
 COMPOSE_CMD='docker-compose'
 CLUSTER_NAME="demo-a"
-COMPOSE_FILE="./docs/dev/armada-compose.yaml"
```
kannon92, line 6: The previous version is correct.  @richscott changed this file to use his docker-file.  And I reference this file in a different part of the build.  

We could remove this and only use @suprjinx localdev code but let's leave this alone for this PR.

kannon92: Didn't realize that we removed this in a previous PR.  This is actually correct!

kannon92 (approved):

kannon92:
Switching from npm to yarn looks to save about 40-50s.  

4:49 to 4:01
5:07 to 4:21s.

kannon92:
LGTM @carlocamurri is going to follow up with team to see if we can merge this.  It will require internal changes to build system so let's hold on merge until we get a Good to go.

kannon92 (approved):

PR #1639, opened 2022-10-17
Fixes to Event Ingester

d80tb7:
* Default `eventRetention` in EventIngester to 336 hours (matches the current default in the server)
* EventIngester no longer has a hard coded list of retryable redis errors. Instead by default, it considers all errors as retryable, and only gives up if the error matches one of a (config-defined) number of regexes.

### DIFF
--- a/internal/eventingester/store/insertion.go
+++ b/internal/eventingester/store/insertion.go
@@ -3,7 +3,7 @@ package store
 import (
 	"context"
 	"fmt"
-	"strings"
+	"regexp"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


internal/eventingester/store/insertion.go
```diff
@@ -18,20 +18,20 @@ import (
 // InsertEvents takes a channel of armada events and inserts them into the event db
 // the events are republished to an output channel for further processing (e.g. Ackking)
 func InsertEvents(ctx context.Context, db EventStore, msgs chan *model.BatchUpdate, bufferSize int,
```
d80tb7, line 20: This piece of code should probably refactored to have a struct storing some common stuff (e.g. list of fatal errors).  I didn't do it as part of this change, just because I wanted to get this in relatively quickly.

JamesMurkin (approved):

Issue #908, opened 2022-04-01
Allow exact match on Jobs tab in Lookout


robertdavidsmith:
Problem

If my queue name is a prefix of someone else's

 - Such as batch-andrews which is a prefix of batch-andrewsi

Then when I search job by my queue name, I get this other persons jobs.

 

Proposal

Could we implement some way to alleviate this, possibly allowing for exact match (with via UI option or quoting the string).

 - I.e "batch-andrews" would only give matches for batch-andrews queue because it is quoted so we use an exact match

@JamesMurkin has full context, suggest getting in touch with him before starting work

dave-gantenbein:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Armada-62c744dc3b819100132ca032/issues/G-Research/armada/908?planning-poker) with ZenHub @dejanzele @jayofdoom @kannon92 @richscott

ClifHouck:
Closing this as #1665 addresses this issue. Opened #1679 to track drop-down work if want to go that route.

PR #1652, opened 2022-10-17
Allow the use of quotes to specify job queue name exactly

ClifHouck:
Lookout's UI allows a user to filter jobs by job queue name via the table displayed on the jobs tab. This commit enables a quoted string within `lookout.GetJobsRequest.Queue` to enable an exact match search and not just a prefix search on job queue name.

Partial #908

### DIFF
--- a/internal/lookout/repository/jobs.go
+++ b/internal/lookout/repository/jobs.go
@@ -103,7 +103,11 @@ func (r *SQLJobRepository) createWhereFilters(opts *lookout.GetJobsRequest) []go
 	var filters []goqu.Expression
 
 	if opts.Queue != "" {
-		filters = append(filters, StartsWith(job_queue, opts.Queue))
+		if quotedString, ok := getQuotedString(opts.Queue); ok {
+			filters = append(filters, Exactly(job_queue, quotedString))
+		} else {
+			filters = append(filters, StartsWith(job_queue, opts.Queue))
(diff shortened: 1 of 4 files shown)


ClifHouck:
A couple of thoughts/questions after thinking about this a little more:
1. Do we want all searchable fields to have this same behavior? I'd learn toward yes. Consistency is good.
2. Does the react table have this type of search/filtering capability built-in? I'm still woefully ignorant of front-end land.
3. It appears that each time the search field changes it sends a query to lookout's backend, possibly after some delay. Is this desired behavior?

kannon92:
> A couple of thoughts/questions after thinking about this a little more:
> 
> 1. Do we want all searchable fields to have this same behavior? I'd learn toward yes. Consistency is good.

Yea I think that is a good call.  If it is too much work, we can make it as a followup.  But we reuse a lot of this code so it may be handled for free.  
> 2. Does the react table have this type of search/filtering capability built-in? I'm still woefully ignorant of front-end land.

We manually add filtering compability in the UI.  Basically it creates an event whenever a user types and it gets saved as a filter.  
> 3. It appears that each time the search field changes it sends a query to lookout's backend, possibly after some delay. Is this desired behavior?

Yea I believe so.  If you look at the URL, you can see it generates a rest-api to the lookout server using query params.

PR #1653, opened 2022-10-17
Closes #1482

### LINKED ISSUE #1482: Lookout: time in state column
Users would like a new job view column which displays how long the job had been in its current state (queued, running, succeeded, failed, etc...).

suprjinx:


richscott (approved):

internal/lookout/ui/src/containers/JobsContainer.tsx
```diff
@@ -109,6 +109,16 @@ class JobsContainer extends React.Component<JobsContainerProps, JobsContainerSta
...
         },
+        {
+          id: "jobStateDuration",
+          name: "Time in State",
+          accessor: "jobStateDuration",
+          urlParamKey: "job_state_duration",
+          isDisabled: false,
+          filter: [],
+          defaultFilter: [],
+          width: 0.7,
```
kannon92, line 120: @carlocamurri told me that all these width values need to add up to 1.

internal/lookout/repository/jobs.go
```diff
@@ -248,6 +252,31 @@ func sortJobsByJobId(jobInfos []*lookout.JobInfo, descending bool) {
...
+	switch {
+	case row.Finished.Valid:
+		timeStamp = row.Finished.Time
+	case row.Started.Valid:
+		timeStamp = row.Started.Time
+	case row.Created.Valid:
+		timeStamp = row.Created.Time
+	case row.Cancelled.Valid:
+		timeStamp = row.Cancelled.Time
+	case row.Submitted.Valid:
```
kannon92, line 270: what about other states like PENDING or Queued

internal/lookout/ui/src/containers/JobsContainer.tsx
suprjinx: I'll adjust

internal/lookout/repository/jobs.go
suprjinx: I believe the other states are captured here -- if nothing but "submitted" has a value, it is queued, etc

kannon92: Don’t we want time in queued or pending?

This looks to map state based on Time.
https://github.com/G-Research/armada/blob/master/internal/lookout/repository/jobs.go

kannon92: This could be right but I find it confusing.  Started and Created don't match the states in the lookout code.  I think Started corresponds to Running and Created would correspond to Queued.  But could you use the JobState as the key for the switch instead to make this a bit more clear

internal/lookout/repository/jobs_test.go
```diff
@@ -8,12 +8,63 @@ import (
...
+		jobInfo := jobInfos[0]
+
+		timeInState := time.Now().Sub(pendingTime)
+	        expectedDuration := duration.ShortHumanDuration(timeInState)
+		assert.Equal(t, expectedDuration, jobInfo.JobStateDuration)
+	})
+}
+
+
+func TestGetJobs_GetRunning(t *testing.T) {
```
kannon92, line 43: Can you add tests for the missing States? Queued, Cancelled

suprjinx: I added tests for each timestamp we are considering but will double check.

internal/lookout/repository/jobs.go
```diff
@@ -248,6 +252,29 @@ func sortJobsByJobId(jobInfos []*lookout.JobInfo, descending bool) {
...
 }
 
+func determineJobStateDuration(state JobState, row *JobRow) string {
+	if row == nil {
+		return ""
+	}
+	var timeStamp time.Time
+
+	switch state {
+	case JobSucceeded, JobFailed:
```
kannon92, line 262: Thank you!  This looks a lot better to me!

internal/lookout/ui/src/containers/JobsContainer.tsx
kannon92: I was wrong.  JobSet colums add to 1.  Looks like Jobs don't.

internal/lookout/repository/jobs.go
kannon92: Looks good!

kannon92 (dismissed):
Nice work! Once you get the linter fixed, I think this is good to go!

kannon92 (approved):

PR #1665, opened 2022-10-18
Default to exact matches when specifying job queue name

ClifHouck:
But also allow the use of `*` to denote wildcard/like searching.

Partial #908.


ClifHouck:
Probable alternative to #1652

internal/lookout/repository/util.go
```diff
@@ -50,6 +50,16 @@ func StartsWith(field exp.IdentifierExpression, pattern string) goqu.Expression
 	return field.Like(pattern + "%")
 }
 
+// Do a glob (sql like) search if the pattern contains any * characters,
+// otherwise do an exact match.
+func GlobSearchOrExact(field exp.IdentifierExpression, pattern string) goqu.Expression {
```
kannon92, line 55: Is this a special thing to do because wildcards are special in SQL?

ClifHouck: Part of the idea is speed. If the user hasn't indicated they want a wider match, we default to an exact match, which should be a faster query. 

The use of `*` is just inertia, I doubt most users want to use `%` to denote a wildcard.

kannon92: Ah TIL.  Should we support % as a wildcard also then?

ClifHouck: I don't think so. I don't think I've ever used it as such outside of SQL. But I'm willing to be persuaded. @JamesMurkin any thoughts?

JamesMurkin: I'm not sure users would expect % to be a wildcard and I'm not sure we gain much out of allowing multiple - although not massively harmful

I personally think we should stick to just using *, but it wouldn't be terrible if we wanted to support both

JamesMurkin: I'm not sure users would expect % to be a wildcard and I'm not sure we gain much out of allowing multiple - although not massively harmful

I personally think we should stick to just using *, but it wouldn't be terrible if we wanted to support both

kannon92: Sounds good to me.

JamesMurkin (approved):

kannon92 (approved):

PR #1678, opened 2022-10-20
Fix cluster event handlers on fake executor

JamesMurkin:
Currently cluster event handlers are being mixed up with pod event handlers when they are registered

This results in them receiving pod events that they can't handle
 - Resulting in a lot of errors in logs, making it very hard to read the log for anything useful

Splitting out the event handlers, so that each handler is only can be passed the right type of events

### DIFF
--- a/internal/executor/fake/context/context.go
+++ b/internal/executor/fake/context/context.go
@@ -50,7 +50,8 @@ var DefaultNodeSpec = []*NodeSpec{
 type FakeClusterContext struct {
 	clusterId             string
 	pool                  string
-	handlers              []*cache.ResourceEventHandlerFuncs
+	podEventHandlers      []*cache.ResourceEventHandlerFuncs
+	clusterEventHandlers  []*cache.ResourceEventHandlerFuncs
 	rwLock                sync.RWMutex
 	pods                  map[string]*v1.Pod
 	events                map[string]*v1.Event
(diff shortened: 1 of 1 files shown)


internal/executor/fake/context/context.go
```diff
@@ -50,7 +50,8 @@ var DefaultNodeSpec = []*NodeSpec{
 type FakeClusterContext struct {
 	clusterId             string
 	pool                  string
-	handlers              []*cache.ResourceEventHandlerFuncs
+	podEventHandlers      []*cache.ResourceEventHandlerFuncs
+	clusterEventHandlers  []*cache.ResourceEventHandlerFuncs
```
ClifHouck, line 54: Maybe this doesn't matter since this is a fake executor, but `clusterEventHandlers` doesn't seem to be used anywhere, other than assignment. Is that expected?

JamesMurkin: That is expected, I just added it so if anyone wanted to use it in future they don't make the same mistake

JamesMurkin: Alternatively we could just do nothing on AddClusterEventEventHandler and remove that field

d80tb7: I think it's fine to keep it.  Maybe just add a comment to say it's not used right now?

d80tb7 (approved):

internal/executor/fake/context/context.go
ClifHouck: It's fine, I just lack context around how the fake executor works. A comment would be nice but not necessary.

PR #1683, opened 2022-10-20
1494 lookout use orig job spec

### LINKED ISSUE #1494: [Lookout] Use orig_job_spec rather than job column for displaying the job yaml
Lookout uses the `job.job` database column to display the job yaml.  This is a binary json column, which is unfortunately slow to ingest.

To solve this, lookout ingester has started populating the `job.orig_job_spec` column which is the same data contains gzip compressed proto data stored in a postres bytea column,

We are now populating both columns, and have been for some time, so we can move over to the new column and delete the old one.  To do this we need to:

a) change lookout to take the job yaml from the `orig_job_spec` column.
b) change lookout ingester to no longer populate the old `job` column
c) change the lookout schema to drop the old `job` column.

suprjinx:
Closes #1494


internal/lookout/repository/store.go
```diff
@@ -38,7 +38,7 @@ func NewSQLJobStore(db *goqu.Database, annotationPrefix string) *SQLJobStore {
 }
 
 func (r *SQLJobStore) RecordJob(job *api.Job, timestamp time.Time) error {
-	jobJson, err := json.Marshal(job)
+	jobJson, err := job.Marshal()
```
suprjinx, line 41: "jobJson" isn't correct here, it is protobuf marshalling -- i'll rename

suprjinx: ultimately, this was just for the out-of-date tests -- we are storing job in lookout-ingester now

internal/lookoutingester/lookoutdb/insertion.go
```diff
@@ -229,12 +227,12 @@ func (l *DB) CreateJobsBatch(ctx context.Context, instructions []*model.CreateJo
...
 func (l *DB) CreateJobsScalar(ctx context.Context, instructions []*model.CreateJobInstruction) {
-	sqlStatement := `INSERT INTO job (job_id, queue, owner, jobset, priority, submitted, job, orig_job_spec, state, job_updated)
-		 VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10)
+	sqlStatement := `INSERT INTO job (job_id, queue, owner, jobset, priority, submitted, orig_job_spec, state, job_updated)
+		 VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)
          ON CONFLICT DO NOTHING`
 	for _, i := range instructions {
 		err := withDatabaseRetryInsert(func() error {
-			_, err := l.db.Exec(ctx, sqlStatement, i.JobId, i.Queue, i.Owner, i.JobSet, i.Priority, i.Submitted, i.JobJson, i.JobProto, i.State, i.Updated)
+			_, err := db.Exec(ctx, sqlStatement, i.JobId, i.Queue, i.Owner, i.JobSet, i.Priority, i.Submitted, i.JobProto, i.State, i.Updated)
```
suprjinx, line 235: that's probably supposed to be l.db .. will check

suprjinx: fixed

internal/lookout/repository/jobs.go
```diff
@@ -280,11 +286,9 @@ func makeJobFromRow(row *JobRow) (*api.Job, error) {
 		return nil, nil
 	}
 
-	var jobFromJson api.Job
-	jobJson := ParseNullString(row.JobJson)
-	err := json.Unmarshal([]byte(jobJson), &jobFromJson)
+	jobSpec, err := makeApiJob([]byte(row.OrigJobSpec.String))
 	if err != nil {
-		log.Errorf("error while parsing job %s json: %v", ParseNullString(row.JobId), err)
+		log.Warnf("unable to load orig job spec. %+v", err)
```
kannon92, line 291: Is this temporary?

internal/lookout/repository/store_test.go
```diff
@@ -979,7 +980,7 @@ func getPriority(t *testing.T, db *goqu.Database, jobId string) float64 {
 
 func getJob(t *testing.T, db *goqu.Database, jobId string) *api.Job {
 	var job api.Job
-	jobJson := ParseNullString(selectNullString(t, db, fmt.Sprintf("SELECT job FROM job WHERE job_id = '%s'", jobId)))
+	jobJson := ParseNullString(selectNullString(t, db, fmt.Sprintf("SELECT orig_job_spec FROM job WHERE job_id = '%s'", jobId)))
```
suprjinx, line 983: this should probably switch to proto marshall/unmarshall instead of jamming json in there

internal/lookout/repository/jobs.go
kannon92: Error -> Warn I mean?

suprjinx: not sure why i changed that, will switch back

kannon92 (approved):

PR #1685, opened 2022-10-21
Clean up termination config/defaults

severinson:
For pods that set no grace period or explicitly set it to zero, automatically change it to the minimum allowed. Remove the default grace period setting and instead use the minimum as the default.

### DIFF
--- a/internal/armada/server/submit.go
+++ b/internal/armada/server/submit.go
@@ -903,17 +903,26 @@ func (server *SubmitServer) applyDefaultsToPodSpec(spec *v1.PodSpec) {
 		spec.PriorityClassName = defaultPriorityClass
 	}
 
-	// add missing TerminationGracePeriod if needed
 	server.applyTerminationGracePeriodDefault(spec)
 }
 
-// applyTerminationGracePeriodDefault will give the podspec a default if needed
+// applyTerminationGracePeriodDefault sets the termination grace period
+// of the pod equal to the minimum if
+// - the pod does not explicitly set a termination period, or
+// - the pod explicitly sets a termination period of 0.
 func (server *SubmitServer) applyTerminationGracePeriodDefault(spec *v1.PodSpec) {
-	specNeedsTerminationGracePeriod := spec.TerminationGracePeriodSeconds == nil
-	defaultTerminationGracePeriod := int64(server.schedulingConfig.DefaultTerminationGracePeriod.Seconds())
-
-	if specNeedsTerminationGracePeriod {
-		spec.TerminationGracePeriodSeconds = &defaultTerminationGracePeriod
+	var podTerminationGracePeriodSeconds int64
+	if spec.TerminationGracePeriodSeconds != nil {
+		podTerminationGracePeriodSeconds = *spec.TerminationGracePeriodSeconds
+	}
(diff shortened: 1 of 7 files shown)


internal/armada/server/submit.go
```diff
@@ -903,17 +903,26 @@ func (server *SubmitServer) applyDefaultsToPodSpec(spec *v1.PodSpec) {
...
+	var podTerminationGracePeriodSeconds int64
+	if spec.TerminationGracePeriodSeconds != nil {
+		podTerminationGracePeriodSeconds = *spec.TerminationGracePeriodSeconds
+	}
+	if podTerminationGracePeriodSeconds == 0 {
+		defaultTerminationGracePeriodSeconds := int64(
+			server.schedulingConfig.MinTerminationGracePeriod.Seconds(),
+		)
+		if defaultTerminationGracePeriodSeconds < 0 {
+			panic("MinTerminationGracePeriod is negative")
```
d80tb7, line 923: Let's not panic here- if we want to validate config let's do it at startup

internal/armada/server/submit.go
```diff
@@ -903,17 +903,26 @@ func (server *SubmitServer) applyDefaultsToPodSpec(spec *v1.PodSpec) {
...
+// - the pod explicitly sets a termination period of 0.
 func (server *SubmitServer) applyTerminationGracePeriodDefault(spec *v1.PodSpec) {
-	specNeedsTerminationGracePeriod := spec.TerminationGracePeriodSeconds == nil
-	defaultTerminationGracePeriod := int64(server.schedulingConfig.DefaultTerminationGracePeriod.Seconds())
-
-	if specNeedsTerminationGracePeriod {
-		spec.TerminationGracePeriodSeconds = &defaultTerminationGracePeriod
+	var podTerminationGracePeriodSeconds int64
+	if spec.TerminationGracePeriodSeconds != nil {
+		podTerminationGracePeriodSeconds = *spec.TerminationGracePeriodSeconds
```
d80tb7, line 916: we don't verify that  *spec.TerminationGracePeriodSeconds is > 0

internal/armada/server/submit.go
```diff
@@ -903,17 +903,26 @@ func (server *SubmitServer) applyDefaultsToPodSpec(spec *v1.PodSpec) {
...
+	}
+	if podTerminationGracePeriodSeconds == 0 {
+		defaultTerminationGracePeriodSeconds := int64(
```
d80tb7, line 919: why don't we do the conversion once at startup rather than every time?  Or even make the config value be an int representing seconds rather than a duration.

d80tb7 (dismissed):

internal/armada/server/submit.go
severinson: We have another check that rejects pods with a value less than min.

internal/armada/server/submit.go
severinson: I think this is fine. I like having a duration. Since then we can put things like "5m".

internal/armada/server/submit.go
severinson: Agreed. Let's remove the panic and later add proper config validation at startup.

d80tb7 (approved):

PR #1686, opened 2022-10-21
Add per-priority class default tolerations

severinson:


internal/armada/server/submit.go
```diff
@@ -882,25 +882,17 @@ func (server *SubmitServer) applyDefaultsToPodSpec(spec *v1.PodSpec) {
...
 
-	// Each pod must have some default tolerations
-	// Here, we add any that are missing
-	podTolerations := make(map[string]v1.Toleration)
-	for _, podToleration := range spec.Tolerations {
-		podTolerations[podToleration.Key] = podToleration
-	}
-	for _, defaultToleration := range server.schedulingConfig.DefaultJobTolerations {
-		podToleration, ok := podTolerations[defaultToleration.Key]
-		if !ok || !defaultToleration.MatchToleration(&podToleration) {
```
severinson, line 893: This is an optimisation to avoid adding redundant tolerations. I don't think it's necessary and have removed it. This optimisation would not even remove all duplicate taints. Since there may be several tolerations with the same key.

d80tb7 (approved):

internal/armada/server/submit.go
severinson: Duplicate tolerations should we safe.

PR #1694, opened 2022-10-24
Update Job Deletion Logic To Support Force Kill

d80tb7:
This changes the executor's job deletion logic so that:

- The first attempt to delete a job attempts to delete the job while respecting the graceful termination period set on the pod
- Any subsequent delete attempts that occur within the graceful termination period + killTimeout (killTimeout defaults to 5 minutes but is configurable) are no-ops.
- Any subsequent delete attempts that occur after graceful termination period + killTimeout attempts to delete the job with a graceful termination of 0 (i.e. force kills the job)

### DIFF
--- a/internal/executor/context/cluster_context.go
+++ b/internal/executor/context/cluster_context.go
@@ -343,9 +343,6 @@ func (c *KubernetesClusterContext) DeleteIngress(ingress *networking.Ingress) er
 
 func (c *KubernetesClusterContext) ProcessPodsToDelete() {
 	pods := c.podsToDelete.GetAll()
-
-	deletionGracePeriod := 5 * time.Minute
-
 	util.ProcessPodsWithThreadPool(pods, c.deleteThreadCount, func(podToDelete *v1.Pod) {
 		if podToDelete == nil {
 			return
@@ -354,9 +351,11 @@ func (c *KubernetesClusterContext) ProcessPodsToDelete() {
 			// We've never tried to delete this pod before.  Delete using the grace period
 			c.doDelete(podToDelete, false)
 		} else {
+			// we've tried to delete this pod before. If we're after the kill period then force delete
+			// else it's a no-op
 			killTime := podToDelete.DeletionTimestamp.
 				Add(util.GetDeletionGracePeriodOrDefault(podToDelete)).
-				Add(deletionGracePeriod)
+				Add(c.killTimeout)
 			if c.clock.Now().After(killTime) {
 				log.Warnf("Pod %s/%s was requested deleted at %s, but is still present.  Force killing.", podToDelete.Namespace, podToDelete.Name, podToDelete.DeletionTimestamp)
 				c.doDelete(podToDelete, true)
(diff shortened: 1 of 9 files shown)


internal/executor/context/cluster_context_test.go
```diff
@@ -308,6 +313,33 @@ func TestKubernetesClusterContext_ProcessPodsToDelete_CallPatchOnClient_WhenPods
...
+	pod = createSubmittedBatchPod(t, clusterContext)
+	pod.DeletionGracePeriodSeconds = pointer.Int64(10)
+	pod.DeletionTimestamp = &metav1.Time{Time: testClock.Now().Add(-310 * time.Second)}
+	client.Fake.ClearActions()
+
+	// submitPodsWithWait(t, clusterContext, pod)
+	client.Fake.ClearActions()
+	clusterContext.DeletePods([]*v1.Pod{pod})
+	clusterContext.ProcessPodsToDelete()
+	assert.Equal(t, len(client.Fake.Actions()), 2)
```
d80tb7, line 340: Here I want to assert that we using a timout of 0 to kill the pod, but it doesn't look like the fake client tells me this.

internal/executor/context/cluster_context.go
```diff
@@ -334,36 +343,55 @@ func (c *KubernetesClusterContext) DeleteIngress(ingress *networking.Ingress) er
...
+			// We've never tried to delete this pod before.  Delete using the grace period
+			c.doDelete(podToDelete, false)
+		} else {
+			// we've tried to delete this pod before. If we're after the kill period then force delete
+			// else it's a no-op
+			killTime := podToDelete.DeletionTimestamp.
+				Add(util.GetDeletionGracePeriodOrDefault(podToDelete)).
+				Add(c.killTimeout)
+			if c.clock.Now().After(killTime) {
+				log.Warnf("Pod %s/%s was requested deleted at %s, but is still present.  Force killing.", podToDelete.Namespace, podToDelete.Name, podToDelete.DeletionTimestamp)
```
samclark, line 360: shall we make this INFO level?

internal/executor/context/cluster_context.go
```diff
@@ -82,6 +86,8 @@ type KubernetesClusterContext struct {
 	eventInformer            informer.EventInformer
 	// If provided, stops object creation while EtcdMaxFractionOfStorageInUse or more of etcd storage is full.
 	etcdHealthMonitor healthmonitor.EtcdLimitHealthMonitor
+	killTimeout       time.Duration
```
samclark, line 89: s/killTimeout/podKillTimeout

samclark (dismissed):

internal/executor/context/cluster_context.go
d80tb7: done

internal/executor/context/cluster_context.go
d80tb7: done

samclark (approved):

PR #1697, opened 2022-10-25
Queue Limits By Priority

d80tb7:
Implements queue limits by priority.  This enables lower (preemptible) pirotiy classes to be given higher queue limits in order to incentivize preemption.

Preemption configuration is now of the form:
```
PreemptionConfig:
PriorityClasses:
  armada-default:
    priority: 30000
    maximalResourceFractionPerQueue:
      cpu: 20
      memory: 20  
  armada-preemptible:
    priority: 20000
    maximalResourceFractionPerQueue:
      cpu: 90
      memory: 90  
```

As part of this PR I removed the preemption logic from the old scheduler.


internal/armada/configuration/types.go
```diff
@@ -248,7 +252,7 @@ type JetstreamConfig struct {
 
 type QueueManagementConfig struct {
 	AutoCreateQueues       bool
-	DefaultPriorityFactor  queue.PriorityFactor
```
d80tb7, line 251: I changed this because it was causing a circular reference.  Essentially other armada module imported by the `configuration` package is problematic because configuration is imported everywhere.  In this case all we were using this for was a type alias so it shouldn't really matter.

internal/armada/scheduling/lease.go
```diff
@@ -6,8 +6,6 @@ import (
 	"math/rand"
 	"time"
 
-	v1 "k8s.io/api/core/v1"
```
d80tb7, line 9: this is the old scheduler where I have removed the preemption logic

internal/armada/scheduling/node_matching.go
```diff
@@ -108,13 +108,12 @@ func matchAnyNodeTypeAllocation(
 	job *api.Job,
 	nodeAllocations []*nodeTypeAllocation,
 	alreadyConsumed nodeTypeUsedResources,
-	supportedPriorityClasses map[string]int32,
 ) (nodeTypeUsedResources, bool, error) {
 	newlyConsumed := nodeTypeUsedResources{}
 
 	for _, podSpec := range job.GetAllPodSpecs() {
```
d80tb7, line 114: This is the old scheduler where I have removed preemption;

internal/armada/server.go
```diff
@@ -476,10 +483,73 @@ func createRedisClient(config *redis.UniversalOptions) redis.UniversalClient {
 	return redis.NewUniversalClient(config)
 }
 
-// TODO Is this all validation that needs to be done?
-func validateArmadaConfig(config *configuration.ArmadaConfig) error {
+// TODO: Is this all validation that needs to be done?
+func validateCancelJobsBatchSizeConfig(config *configuration.ArmadaConfig) error {
```
d80tb7, line 487: I've renamed this function because it claimed to validate all the armada config but all  it actually did was check that the `cancelJobsBatchSize` was greater than 0!  Clealry we need to have a better strategy for validating our configuration but that's a job for another day.

internal/scheduler/legacyscheduler.go
```diff
@@ -385,6 +426,9 @@ type LegacyScheduler struct {
 	JobRepository SchedulerJobRepository
 	// Minimum quantity allowed for jobs leased to this cluster.
 	MinimumJobSize map[string]resource.Quantity
+	// Limit (as a fraction of the cluster) of resource types at each priority
+	// This is cumulative i.e. faction at level x includes all previous levels
```
JamesMurkin, line 430:
```suggestion
	// This is cumulative i.e. fraction at level x includes all previous levels
```

JamesMurkin (approved):

PR #1699, opened 2022-10-25
Airflow annotation for task ID

### LINKED ISSUE #1608: ArmadaOperator: Add airflow task id to Armada annotations
When a user submits an Airflow task, we should add the airflow task id to the Armada annotations.

An example can be below: 
```
jobSetId: test
priority: 100
namespace: batch-test
annotations:
  armadaproject.io/taskRunId: 0a016adf-8325-1d6a-8183-268de7a625ea 
  ````

### PR DESCRIPTION
**Pull request by suprjinx**
Closes #1608

### CODE REVIEW

File: third_party/airflow/armada/operators/armada.py [hunk 1]
```diff
@@ -102,3 +104,23 @@ def execute(self, context) -> None:
...
+    :param job_request_items: The job request items we plan to send to armada
+
+    :return: annotated job request items for armada
+    """
+    task_instance = context["ti"]
+    task_id = task_instance.task_id
+
+    for item_list in job_request_items:
+        for item in item_list:
+            item.annotations["armadaproject.io/taskId"] = task_id
```
ClifHouck, line 124: Should annotations be stored as constants? Is there a spec somewhere for them?

third_party/airflow/tests/unit/test_airflow_operator_mock.py
```diff
@@ -69,65 +71,19 @@ def sleep_job():
         ],
     )
     return [submit_pb2.JobSubmitRequestItem(priority=0, pod_spec=pod)]
-
-
-def test_job_service_health():
```
ClifHouck, line 74: Why'd a bunch of tests disappear? They seem unrelated to your change.

suprjinx: hmm that's not right -- fixing

third_party/airflow/armada/operators/armada.py
suprjinx: re: spec, there is only the ticket (#1608) AFAIK. 

re: const -- i usually think that a const is unecessary for singular usage, but I moved this to a const var.

third_party/airflow/tests/unit/test_airflow_operator_mock.py
suprjinx: errant deletion, fixed

ClifHouck (dismissed):
LGTM

third_party/airflow/armada/operators/armada.py
```diff
@@ -29,6 +29,8 @@
 
 armada_logger = logging.getLogger("airflow.task")
 
+ANNOTATION_TASK_ID = "armadaproject.io/taskId"
```
kannon92, line 32: This is actually configurable from lookout so we should look more into this. Env variable or read from lookout confit.

third_party/airflow/armada/operators/armada.py
```diff
@@ -102,3 +106,22 @@ def execute(self, context) -> None:
             "Armada Job finished with %s and message: %s", job_state, job_message
         )
         airflow_error(job_state, self.name, job_id)
+
+
+def annotate_job_request_items(context, job_request_items):
```
kannon92, line 111: Can you add return types

third_party/airflow/armada/operators/armada.py
kannon92: No, I mean that lookout will decide what annotations to show up if that match the part before '/'.  

https://github.com/G-Research/armada/blob/master/config/lookout/config.yaml#L16

So annotations that are displayed to user in lookout have to have a form that is present in the config.  Otherwise, we ignore them.  

The pod object has to add the annotation that matches how we deploy lookout to display annotations.  We should probably see if this is really just a constant value or changes depending on production environment in GR.

suprjinx: it doesn't seem like airflow module should pull in lookout config

third_party/airflow/armada/operators/armada.py
suprjinx: added

localdev/docker-compose.yaml
```diff
@@ -4,6 +4,9 @@ networks:
   kind:
     external: true
```
kannon92, line 6: Can we remove these changes from this PR?

suprjinx: oops, yes will remove -- should be in master now

third_party/airflow/armada/operators/armada.py
```diff
@@ -120,3 +127,43 @@ def _get_lookout_url(self, job_id: str) -> str:
...
+    :param job_request_items: The job request items to be sent to armada
+
+    :return: annotated job request items for armada
+    """
+    task_instance = context["ti"]
+    task_id = task_instance.task_id
+    run_id = context["run_id"]
+    dag_id = context["dag"].dag_id
+
+    for item in job_request_items:
```
kannon92, line 149: My python is weak.  Does this work if annotations are not defined?  I think they are optional fields?

suprjinx: i think it's an empty dictionary when empty and this should work -- would you mind running in your airflow setup to confirm? I have not set that up yet

kannon92: Sounds like a perfect time to do that.. ;)

suprjinx: the sleep_job() test function provides the default annotations value in the unit test, and this works

kannon92 (approved):

PR #1700, opened 2022-10-25
Make job set search/filter use GlobSearchOrExact

### LINKED ISSUE #1698: Make search behavior of "job set" search field consistent with "queue" and "owner" search fields in the Jobs lookout tab.
Make search behavior of "job set" search field consistent with "queue" and "owner" search fields in the Jobs lookout tab.

Related to 
- https://github.com/G-Research/armada/issues/1679
- https://github.com/G-Research/armada/issues/908
- https://github.com/G-Research/armada/pull/1665

ClifHouck:
Closes #1698


severinson:
How does this affect performance? Starts with, unlike glob, doesn't necessarily require a full scan of the database.

ClifHouck:
> How does this affect performance? Starts with, unlike glob, doesn't necessarily require a full scan of the database.

Certain queries will be slower, for sure. Any search missing the wildcard operator `*` will be an exact search, which should be very fast. Searches such as `test*` will be equivalent in performance to `StartsWith`. Other searches such as `*test` will likely be less performant. I believe in the most likely search cases, performance will be better or equivalent.

severinson (approved):

severinson:
I've approved.

Issue #1506, opened 2022-09-12
Write Pulsar message-receiving client for testing


richscott:
Using the data defintions and logic in `internal/pulsarutils/pulsarclient.go` and `messageid.go`, write a basic Pulsar client program that (optionally, authenticates first) to a Pulsar cluster, and continuously polls (select's) from a channel as supplied by the `Receive()` func in internal/pulsar/async.go.  Connection configuration should be specifiable via command-line flags.

Interrupting the program (ctrl+c / SIGINT) should close the context and terminate the program.

This issue is part of the suite of tools specified in #1460.

dave-gantenbein:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Armada-62c744dc3b819100132ca032/issues/G-Research/armada/1506?planning-poker) with Zenhub @dejanzele @jayofdoom @kannon92 @richscott @suprjinx

PR #1702, opened 2022-10-25
Add 'watch' subcommand to pulsartest, to receive Pulsar events (#1506)

richscott:
Add new `watch` sub-command to the `pulsartest` utility, to observe Pulsar events. For now, it does not consume any events, nor does it continually poll for events (however, that may be added).


cmd/pulsartest/cmd/watch.go
```diff
@@ -0,0 +1,59 @@
...
+func watchCmd() *cobra.Command {
+	a := &pulsartest.App{}
+
+	cmd := &cobra.Command{
+		Use:   "watch",
+		Short: "Watch for Pulsar events",
+		Long:  "Watch for Pulsar events",
+		Args:  cobra.ExactArgs(0),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			url, err := cmd.Flags().GetString("url")
```
dave-gantenbein, line 19: I know all the err'ing is a golang-ism, but could this be hidden in a private getArgs(...) method to make the rest of this func read cleaner?

internal/pulsartest/app.go
```diff
@@ -36,19 +37,39 @@ func New(params Params) (*App, error) {
...
-		CompressionType:  compressionType,
-		CompressionLevel: compressionLevel,
-		Topic:            params.Pulsar.JobsetEventsTopic,
-	})
-	if err != nil {
-		return nil, errors.Wrapf(err, "error creating pulsar producer %s", producerName)
+	var producer pulsar.Producer
+	var reader pulsar.Reader
+
+	if cmdType == "submit" {
```
dave-gantenbein, line 43: Testme?

cmd/pulsartest/cmd/watch.go
richscott: I've applied a desiccant in commit 6396b0ce - the new `processCmdFlags()` func in `cmd/pulsartest/cmd/util.go` handles this processing for both `watch.go` and `submit.go`, and total net line-count should be down a bit.

internal/pulsartest/app.go
richscott: Tests to exercise the parameters-checking logic in `New(...)` have been added in `e2e/pulsartest_client/app_test.go` at commit 27472bc4 - the test is there because the Pulsar client library needs a running Pulsar server when instantiating a `Producer` object (it checks if the topic is valid, etc). This test is run as part of the `tests-e2e` makefile target.

cmd/pulsartest/cmd/root.go
```diff
@@ -21,6 +21,7 @@ func RootCmd() *cobra.Command {
 
 	cmd.AddCommand(
 		submitCmd(),
+		watchCmd(),
```
suprjinx, line 24: nice

suprjinx (approved):

PR #1705, opened 2022-10-26
Lookout fix container details if job yaml is undefined, actually return job yaml from API

carlocamurri:


suprjinx (approved):
LGTM

internal/lookout/ui/src/components/job-dialog/JobDetails.tsx
```diff
@@ -82,19 +82,20 @@ export default function JobDetails(props: DetailsProps) {
...
-          <Accordion>
-            <AccordionSummary expandIcon={<ExpandMore />}>
-              <h3>Job YAML</h3>
-            </AccordionSummary>
-            <AccordionDetails>
-              <p className="details-yaml">{props.job.jobYaml}</p>
-            </AccordionDetails>
-          </Accordion>
-        </div>
+        <>
```
ClifHouck, line 86: What does adding an empty, unnamed tag here accomplish?

carlocamurri: https://reactjs.org/docs/fragments.html#short-syntax
An expression must have a single React element. In this case we need to return two, so am wrapping them in a Fragment

ClifHouck (approved):
LGTM

severinson (approved):

PR #1711, opened 2022-10-28
Lookout API make querying with empty fields sane

carlocamurri:


internal/common/util/list.go
```diff
@@ -35,3 +35,13 @@ func DeepCopyListUint32(list []uint32) []uint32 {
 	}
 	return result
 }
+
+func Filter[T any](list []T, predicate func(val T) bool) []T {
```
kannon92, line 39: Can you add some unit tests that live outside of job_test?  I’d like to see some tests around list directly?

JamesMurkin (approved):

JamesMurkin (approved):

PR #1714, opened 2022-10-31
feat: fix issues with armada-bundle helm chart; add pulsar init

dejanzele:


deployment/armada-bundle/templates/pulsar-init-job.yaml
```diff
@@ -0,0 +1,43 @@
...
+            - |
+              echo "Waiting for Pulsar... ({{ .Values.pulsar.init.brokerHost }}:{{ .Values.pulsar.init.port }})"
+
+              while ! nc -z {{ .Values.pulsar.init.brokerHost }} {{ .Values.pulsar.init.port }}; do
+                sleep 1
+              done
+
+              echo "Pulsar started!"
+      containers:
+        - name: init
```
kannon92, line 24: This name is kinda confusing.  You have an init container and then you name a normal container init?

deployment/armada-bundle/templates/pulsar-init-job.yaml
```diff
@@ -0,0 +1,43 @@
...
+          image: apachepulsar/pulsar:2.10.2
```
kannon92, line 25: Could we make this a template since mac has a different image?

deployment/armada-bundle/values.yaml
```diff
@@ -143,8 +150,76 @@ armada-lookout:
...
+        dbname: postgres
+        sslmode: disable
+    pulsar:
+      enabled: true
+      URL: "pulsar://pulsar-broker.armada.svc.cluster.local:6650"
+      jobsetEventsTopic: "persistent://armada/armada/events"
 
 pulsar:
+  init:
+    enabled: false
```
severinson, line 213: Is it right that this is disabled?

deployment/armada-bundle/values.yaml
```diff
@@ -25,40 +25,10 @@ postgresql:
...
-        dbname: postgres
-    uiConfig:
-      armadaApiBaseUrl: "http://armada.default.svc.cluster.local:8080"
-    prunerConfig:
-      daysToKeep: 42
-      batchSize: 1000
-
 stan:
-  nameOverride: "nats"
+  stan:
```
severinson, line 29: We should be able to drop stan now. Since we're populating lookout from Pulsar I don't think we have anything left using stan/nats. Also, have a look at the most recent version of the quickstart to see what they're doing there.

kannon92 (commented):
Small items to fix but LGTM.  

Can we look into ways to incorporate this chart in our CI?  Maybe we should look into using this for e2e tests or release testing?  @jimbobby5 did say that we should consider having GR use this helm chart as their deployments.  Maybe we should start this conversation so we can at least have single way of doing deployments.

deployment/armada-bundle/templates/pulsar-init-job.yaml
dejanzele: Hmm the main container initalizes Pulsar, but I don't have a better name currently :/ I am open to suggestions tho

deployment/armada-bundle/values.yaml
dejanzele: Hmmm, so you can configure this chart to use external components for Pulsar, NATS/STAN, Redis, Postgres... or use install them from this umbrella chart.
Should we in all cases initialize Pulsar? Maybe we can actually, I think the Puslar init script is idempotent

deployment/armada-bundle/values.yaml
dejanzele: On Friday we had tons of issues due to STAN and cluster id. Can Armada Server and Lookout work with Pulsar only now (without STAN as streaming backend)?

deployment/armada-bundle/templates/pulsar-init-job.yaml
dejanzele: makes sense

deployment/armada-bundle/values.yaml
severinson: Yes, it can. Set `disableEventProcessing` (I think that's the right name) to `true` in lookout and make sure to start the `lookout-ingester`. I think the quickstart does this now.

deployment/armada-bundle/values.yaml
severinson: Ah. In that case I guess we leave it disabled? Better to only initialise once.

kannon92 (dismissed):

PR #1715, opened 2022-10-31
Lookout Ingestor v2

carlocamurri:
* Define database schema for new Lookout
* Add ingestion logic (largely based on previous one)
* Add benchmarks for ingestion logic performance


cmd/lookoutingesterv2/main.go
```diff
@@ -0,0 +1,33 @@
...
+func init() {
+	pflag.StringSlice(
+		CustomConfigLocation,
+		[]string{},
+		"Fully qualified path to application configuration file (for multiple config files repeat this arg or separate paths with commas)",
+	)
+	pflag.Parse()
+}
+
+func main() {
```
d80tb7, line 23: I'd be tempted to have the main ingester just take an extra argument to run in benchark mode rather than adding a whole new executable.

config/lookoutingesterv2/config.yaml
```diff
@@ -0,0 +1,28 @@
+postgres:
+  maxOpenConns: 100
```
d80tb7, line 2: we can probably reduce this- the new iingester  will only have a single thread ingesting into the database at any one time so should (in theory) only need a single connection.  We can have a pool to give us flexibility later if we need it, but I think we can reduce it to something like a fixed pool of 5 connections.

(I realise you probably just copied this from the original ingester config!)

config/lookoutingesterv2/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+metrics:
+  port: 9003
+
+pulsar:
+  enabled: true
+  URL: "pulsar://localhost:6650"
+  jobsetEventsTopic: "persistent://armada/armada/events"
+
+paralellism: 1
```
d80tb7, line 21: we don't need this- I had it in the old ingester, but the new framework only supports a single level of parallelism fornow.

config/lookoutingesterv2/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+subscriptionName: "lookout-ingester"
+batchSize: 10000
+batchDuration: 500ms
+pulsarReceiveTimeout: 5s
```
d80tb7, line 25: I think this (and the below) needs to move into the pulsar config block.

config/lookoutingesterv2/config.yaml
```diff
@@ -0,0 +1,28 @@
...
+  port: 9003
+
+pulsar:
+  enabled: true
+  URL: "pulsar://localhost:6650"
+  jobsetEventsTopic: "persistent://armada/armada/events"
+
+paralellism: 1
+subscriptionName: "lookout-ingester"
+batchSize: 10000
```
d80tb7, line 23: `batchSize` and `BatchDuration` should move into the pulsar block.

internal/common/database/db_testutil.go
```diff
@@ -8,48 +8,65 @@ import (
...
 	"github.com/G-Research/armada/internal/common/util"
 )
 
-func WithTestDb(migrations []migration, action func(db *pgxpool.Pool) error) error {
+// WithTestDb spins up a Postgres database for testing
+// migrations: perform the list of migrations before entering the action callback
+// configOverride: optional PostgresConfig to specify which instance to connect to. Defaults to localhost
+// note - if an override is specified, the database will not be cleaned up after the test
+// action: callback for client code
+func WithTestDb(migrations []Migration, configOverride *configuration.PostgresConfig, action func(db *pgxpool.Pool) error) error {
```
d80tb7, line 20: I'd probably have two functions, one that takes an override, one that doesn't (on the grounds that most unit tests won't want an override and passing a random nil everywhere is a bit ugly)

internal/common/database/jobstates.go
```diff
@@ -0,0 +1,43 @@
+package database
+
+type JobState string
+
+type JobRunState string
+
+const (
```
d80tb7, line 7: I'm not sure these can go in the `common/database` package because they are lookout specific.  I'm not entirely sure what the best thing to do is here though because they *are* common between ingester and lookout?  Maybe  put them in  `common/database/lookout`?

internal/common/util/job.go
```diff
@@ -0,0 +1,19 @@
+package util
+
+import (
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/G-Research/armada/pkg/api"
+)
+
+func PodSpecFromJob(job *api.Job) *v1.PodSpec {
```
d80tb7, line 9: I'm sure I've seen this exact function somewhere else in our codebase!

internal/lookoutingesterv2/ingester.go
```diff
@@ -0,0 +1,50 @@
...
+// Run will create a pipeline that will take Armada event messages from Pulsar and update the
+// Lookout database accordingly.  This pipeline will run until a SIGTERM is received
+func Run(config *configuration.LookoutIngesterV2Configuration) {
+	log.Infof("Opening connection pool to postgres")
+	m := metrics.Get()
+	db, err := database.OpenPgxPool(config.Postgres)
+	if err != nil {
+		panic(errors.WithMessage(err, "Error opening connection to postgres"))
+	}
+	lookoutDb := lookoutdb.NewLookoutDb(db, m, 10, 60)
```
d80tb7, line 27: let's comment what the magic numbers are here!

internal/lookoutingesterv2/instructions/instructions.go
```diff
@@ -0,0 +1,553 @@
...
+	event *armadaevents.SubmitJob,
+	update *model.InstructionSet,
+) error {
+	jobId, err := armadaevents.UlidStringFromProtoUuid(event.GetJobId())
+	if err != nil {
+		c.metrics.RecordPulsarMessageError(metrics.PulsarMessageErrorProcessing)
+		return err
+	}
+
+	// Try and marshall the job Json. This shouldn't go wrong but if it does, it'c not a fatal error
```
d80tb7, line 128:
```suggestion
	// Try and marshall the job Json. This shouldn't go wrong but if it does, it's not a fatal error
```

internal/lookoutingesterv2/instructions/instructions.go
```diff
@@ -0,0 +1,553 @@
...
+	// Rather it means that the json won't be available in the ui
+	var jobProto []byte
+	apiJob, err := eventutil.ApiJobFromLogSubmitJob(owner, []string{}, queue, jobSet, ts, event)
+	if err == nil {
+
+		jobProtoUncompressed, err := proto.Marshal(apiJob)
+		if err != nil {
+			log.Warnf("Couldn't marshall job %s in jobset %s as json.  %+v", jobId, jobSet, err)
```
d80tb7, line 136:
```suggestion
			log.Warnf("Couldn't marshall job %s in jobset %s as proto.  %+v", jobId, jobSet, err)
```

internal/lookoutingesterv2/instructions/instructions.go
```diff
@@ -0,0 +1,553 @@
...
+		}
+
+		jobProto, err = c.compressor.Compress(jobProtoUncompressed)
+		if err != nil {
+			log.Warnf("Couldn't compress proto for job %s in jobset %s as json.  %+v", jobId, jobSet, err)
```
d80tb7, line 141:
```suggestion
			log.Warnf("Couldn't compress proto for job %s in jobset %s.  %+v", jobId, jobSet, err)
```

internal/lookoutingesterv2/instructions/instructions.go
d80tb7, line 136: albin converted me to using `log.WithError(err).Warn(...)` rather than `log.Warnf("%+v", err)`

internal/lookoutingesterv2/lookoutdb/insertion.go
```diff
@@ -0,0 +1,884 @@
...
+	}
+}
+
+func (l *LookoutDb) CreateJobsBatch(ctx context.Context, instructions []*model.CreateJobInstruction) error {
+	return l.withDatabaseRetryInsert(func() error {
+		tmpTable := uniqueTableName("job")
+
+		createTmp := func(tx pgx.Tx) error {
+			_, err := tx.Exec(ctx, fmt.Sprintf(`
+				CREATE TEMPORARY TABLE %s
```
d80tb7, line 134: somethinn weird is going on with the formatting here!

d80tb7: and with all the other sql queries!

internal/lookoutingesterv2/instructions/instructions.go
```diff
@@ -0,0 +1,553 @@
...
+		Memory:                    resources.Memory,
+		EphemeralStorage:          resources.EphemeralStorage,
+		Gpu:                       resources.Gpu,
+		Priority:                  int64(event.Priority),
+		Submitted:                 ts,
+		LastTransitionTime:        ts,
+		LastTransitionTimeSeconds: ts.Unix(),
+		State:                     database.JobQueuedOrdinal,
+		JobProto:                  jobProto,
+		PriorityClass:             priorityClass,
```
d80tb7, line 166: do we need to trim any of these string fields to ensure they fit in the varchar cols?

internal/lookoutingesterv2/instructions/instructions.go
```diff
@@ -0,0 +1,553 @@
...
+			}
+			update.JobRunsToUpdate = append(update.JobRunsToUpdate, jobRunUpdate)
+			break
+		}
+	}
+	return nil
+}
+
+func tryCompressError(jobId string, errorString string, compressor compress.Compressor) []byte {
+	compressedError, err := compressor.Compress([]byte(util.RemoveNullsFromString(errorString)))
```
d80tb7, line 471: out of interest, why do we need to remove the nulls here?

internal/lookoutingesterv2/instructions/instructions.go
d80tb7: this goes for all the string values in this file eg. the jobset in user annotation lookup. Apologies if you are trimming them- I haven't got this in an ide so it's not the easiest thing to check.

internal/scheduler/legacyscheduler.go
```diff
@@ -785,15 +785,3 @@ func pickQueueRandomly(weights map[string]float64, random *rand.Rand) (string, f
 	queue := queues[len(queues)-1]
 	return queue, weights[queue] / sum
 }
-
-func podSpecFromJob(job *api.Job) *v1.PodSpec {
```
d80tb7, line 789: ah this might be where I was thinking of it!

config/lookoutingesterv2/config.yaml
carlocamurri: Is this correct? Should the `PulsarConfig` be changed then?

internal/common/util/job.go
carlocamurri: Indeed, I took it from here: https://github.com/G-Research/armada/blob/aa7cad0c3ec6acb485161277ff464f43a42732e0/internal/scheduler/legacyscheduler.go#L789

Just moved it to util package, as it could be useful elsewhere

internal/lookoutingesterv2/lookoutdb/insertion.go
carlocamurri: Very weird - looks correct in GoLand

internal/lookoutingesterv2/instructions/instructions.go
carlocamurri: In my understanding, for the same reason we should remove it from the Job Proto (double checking now, ti seems that isn't happening despite the test being there). This caused problems at insertion time in Postgres when trying to store null chars, but maybe it is not relevant anymore for byte array fields. Will test to confirm

internal/lookoutingesterv2/ingester.go
carlocamurri: Extracted into config and commented there, think it's sensible

internal/lookoutingesterv2/instructions/instructions.go
carlocamurri: Not needed, just works for byte arrays. Have removed

internal/lookoutingesterv2/instructions/instructions.go
carlocamurri: Truncating all fields (except for job id/run id, as those are passed in as UUIDs)

d80tb7 (approved):

PR #1719, opened 2022-11-01
WIP Add endpoint for submitting standard K8s objects

severinson:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-76) by [Unito](https://www.unito.io)


pkg/api/apply.proto
```diff
@@ -0,0 +1,54 @@
...
+    repeated KubernetesObject objects = 1 [(gogoproto.nullable) = false];
+}
+
+message ApplyResponse {
+    string jobId = 1;
+}
+
+service Apply {
+    rpc Apply (ApplyRequest) returns (ApplyResponse) {
+        option (google.api.http) = {
```
kannon92, line 49: Since we are trying to not expose REST API endpoints, should we just make this a GRPC call?  No reason to add more rest api endpoints if we don't need them?

PR #1723, opened 2022-11-01
Add lookout_url_template to ArmadaOperator

### LINKED ISSUE #1603: ArmadaOperator: Lookout url feature request
Hello,

I have a feature request for armada operator in Airflow.
It would be desirable if in the task log a link to the Armada Lookout UI job view is printed.

ArmadaOperator could accept an addition parameter and/or read it from an ENV variable that would look like:

lookout_url_template: "https://lookout.armada.xxx/jobs?job_id=<JOB_ID>&newest_first=true"

That would allow the user to quickly check job status in armada and also to see the logs from the Lookout interface.

ClifHouck:
Setting lookout_url_template properly causes ArmadaOperator to log the lookout URL of the job ID created upon execution.

Closes #1603

### DIFF
--- a/third_party/airflow/tests/unit/test_armada_operator.py
+++ b/third_party/airflow/tests/unit/test_armada_operator.py
@@ -30,4 +30,4 @@ def test_get_lookout_url(lookout_url_template, job_id, expected_url):
         lookout_url_template=lookout_url_template,
     )
 
-    assert operator.get_lookout_url(job_id) == expected_url
+    assert operator._get_lookout_url(job_id) == expected_url
--- a/third_party/airflow/tests/unit/test_armada_operator.py
+++ b/third_party/airflow/tests/unit/test_armada_operator.py
@@ -1,0 +1,21 @@
+from armada.operators.armada import ArmadaOperator
+import pytest
+
+get_lookout_url_test_cases = [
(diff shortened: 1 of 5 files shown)


third_party/airflow/tests/unit/test_armada_operator.py
```diff
@@ -0,0 +1,33 @@
...
+    ),
+    ("", "test_id", ""),
+]
+
+
+@pytest.mark.parametrize(
+    "lookout_url_template, job_id, expected_url", get_lookout_url_test_cases
+)
+def test_get_lookout_url(lookout_url_template, job_id, expected_url):
+    operator = ArmadaOperator(
```
kannon92, line 23: Hmm.  I think this should error out if it is running.  

We assume that someone is able to submit a job using the ArmadaClient and then use the JobService for polling in the Operator.  Looking at the code, I think there is a bug if you are able to pass tests with a None ArmadaClient and JobService.

I created test_airflow_operator_mock so I can test with mock grpc objects.  I would also suggest adding tests their with the lookout url.

kannon92 (changes requested):
Nice Work!  I think there are some more tests you can write and I think you exposed a bug if you pass a None object for the ArmadaClient.  I don't think this code should work with that.

kannon92 (approved):

kannon92:
You should bump the version in the Tom file and publish a new version. It’s an action you can run after this is merged.

PR #1730, opened 2022-11-02
[WIP] Common Framework For Ingesters

d80tb7:
We have a couple of ingesters and are likely to have a couple more.  Now that go has generics I've written a common ingestion pipeline and converted both event ingester and lookout ingester to use it.


internal/common/ingest/batch.go
```diff
@@ -1,27 +1,22 @@
...
-// Batch batches up events from a channel.  Batches are created whenever maxItems InstructionSets have been
+// Batch batches up events from a channel.  Batches are created whenever maxItems have been
 // received or maxTimeout has elapsed since the last batch was created (whichever occurs first).
-// This function has a lot in common with lookoutingester.batch.  Hopefully when generics become available we can
-// factor out most of the common code
-func Batch(values <-chan *pulsarutils.ConsumerMessage, maxItems int,
-	maxTimeout time.Duration, bufferSize int, clock clock.Clock,
-) chan []*pulsarutils.ConsumerMessage {
-	out := make(chan []*pulsarutils.ConsumerMessage, bufferSize)
+func Batch[T any](values <-chan T, maxItems int,
```
severinson, line 11: Did you consider basing the batcher on a struct with a type parameter? Similar to
https://github.com/ninedraft/batch

The above batcher uses callbacks, which makes it a bit more generic too.

Why do we need to pass in a clock instead of using time functions?

severinson: In this function, I think we can avoid the goto. Also, by passing in a context and not spawning a new goroutine inside the function, we can enable running the batcher in an errgroup. The caller would be responsible for spawning the goroutine (or not). That would also give the caller the choice of running the batcher async or sync.

internal/common/ingest/ingestion_pipeline.go
```diff
@@ -0,0 +1,211 @@
...
+// EventSequencesWithIds consists of a batch of Event Sequences along with the corresponding Pulsar Message Ids
+type EventSequencesWithIds struct {
+	EventSequences []*armadaevents.EventSequence
+	MessageIds     []pulsar.MessageID
+}
+
+// IngestionPipeline is an application tha reads message from pulsar and inserts them into a sink. The pipeline will
+// handle the following automatically:
+//   - Receiving messages from pulsar
+//   - Splitting messages into batches for efficient processing
```
severinson, line 51:
```suggestion
//   - Combining messages into batches for efficient processing
```

internal/common/ingest/ingestion_pipeline.go
```diff
@@ -0,0 +1,211 @@
...
+//   - Unmarshalling into event sequences
+//   - Acking processed messages
+//   - Gracefully terminating on a SIGTERM
+//
+// Callers must supply two structs, an InstructionConverter for converting event sequences into something that can be
+// exhausted and a Sink capable of exhausting these objects
+type IngestionPipeline[T HasPulsarMessageIds] struct {
```
severinson, line 58: I like this.

internal/common/ingest/ingestion_pipeline.go
```diff
@@ -0,0 +1,211 @@
...
+		pulsarSubscriptionName: pulsarSubscriptionName,
+		pulsarbatchSize:        pulsarbatchSize,
+		pulsarbatchDuration:    pulsarbatchDuration,
+		converter:              converter,
+		sink:                   sink,
+	}
+}
+
+// Run will run the ingestion pipeline until a SIGTERM is received
+func (ingester *IngestionPipeline[T]) Run() {
```
severinson, line 88: It's better to pass in a context. That way the caller is responsible for closing this function. The contexts used throughout the app make up a tree. The root context is created in the main function and is already set up to cancel on sigterm. When the root context is cancelled, the cancellation propagates to all child contexts.

internal/common/ingest/ingestion_pipeline.go
```diff
@@ -0,0 +1,211 @@
...
+	defer shutdownMetricServer()
+
+	// Waitgroup that wil fire when the pipeline has been torn down
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+
+	// Subscribe to Pulsar and receive messages
+	pulsarClient, err := pulsarutils.NewPulsarClient(&ingester.pulsarConfig)
+	if err != nil {
+		panic(errors.WithMessage(err, "Error creating pulsar client"))
```
severinson, line 102: I prefer this function to return an error.

internal/common/ingest/ingestion_pipeline.go
```diff
@@ -0,0 +1,211 @@
...
+// Run will run the ingestion pipeline until a SIGTERM is received
+func (ingester *IngestionPipeline[T]) Run() {
+	// create a context that will end on a sigterm
+	ctx := createContextWithShutdown()
+
+	shutdownMetricServer := common.ServeMetrics(ingester.metricsConfig.Port)
+	defer shutdownMetricServer()
+
+	// Waitgroup that wil fire when the pipeline has been torn down
+	wg := &sync.WaitGroup{}
```
severinson, line 96: I recommend using an errgroup. Run all child services in that errgroup.

internal/common/ingest/ingestion_pipeline.go
```diff
@@ -0,0 +1,211 @@
...
+	})
+	defer consumer.Close()
+
+	if err != nil {
+		panic(errors.WithMessage(err, "Error creating pulsar consumer"))
+	}
+	pulsarMsgs := pulsarutils.Receive(ctx, consumer, ingester.pulsarConfig.ReceiveTimeout, ingester.pulsarConfig.BackoffTime)
+
+	// Batch up messages
+	batchedMsgs := Batch(pulsarMsgs, ingester.pulsarbatchSize, ingester.pulsarbatchDuration, clock.RealClock{})
```
severinson, line 120: Here, I think batch should run as a child service in an errgroup. That way cancellation and error propagation just works.

internal/common/ingest/ingestion_pipeline.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// Convert to event sequences
+	eventSequences := make(chan *EventSequencesWithIds)
+	go func() {
```
severinson, line 124: Run in errgroup.

internal/common/ingest/ingestion_pipeline.go
```diff
@@ -0,0 +1,211 @@
...
+	}()
+
+	log.Info("Ingestion pipeline set up. Running until shutdown event received")
+	// wait for a shutdown event
+	wg.Wait()
+	log.Info("Shutdown event received - closing")
+}
+
+// createContextWithShutdown returns a context that will report done when a SIGTERM is received
+func createContextWithShutdown() context.Context {
```
severinson, line 163: Sigterm should only be caught in the program main function.

internal/common/ingest/retry.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/G-Research/armada/internal/common/util"
+)
+
+// WithRetry executes the supplied action until it either completes successfully or it returns false, indicating that
+// the error is fatal
+func WithRetry(action func() (error, bool), maxBackoff int) error {
+	backOff := 1
+	for {
+		err, retry := action()
```
severinson, line 16: It's more idiomatic to have err as the second argument.

internal/common/ingest/ingestion_pipeline.go
d80tb7: done

internal/lookoutingester/instructions/instructions.go
```diff
@@ -1,185 +1,134 @@
...
 	GetNodeName() string
 }
 
-type Service struct {
-	m *commonmetrics.Metrics
-}
-
-func New(m *commonmetrics.Metrics) *Service {
-	return &Service{m: m}
+type InstructionCenverter struct {
```
carlocamurri, line 30:
```suggestion
type InstructionConverter struct {
```
Should be changed everywhere it's used as well

internal/common/ingest/ingestion_pipeline.go
d80tb7: I haven't used an errgroup because I don't think we need it (yet) but I don't think it will be hard to add one if we need one

carlocamurri (approved):

PR #1733, opened 2022-11-02
Fix the column widths in the jobs table. Enable horizontal scrolling.

headphonejames:
Move from having fractional column width sizes in the jobs table columns to having a fixed width, populated with sensible default values. Add horizontal scrolling for table overflow in the browser when the table exceeds the browser size.

Fix step towards addressing: https://github.com/G-Research/armada/issues/1480


internal/lookout/ui/src/components/jobs/JobTableColumns.tsx
```diff
@@ -90,27 +82,14 @@ export default function createJobTableColumns(props: JobTableColumnsProps) {
           />
         )
       }
```
kannon92, line 84: I don't think we should remove this.

kannon92 (changes requested):
The UI logic looks good.  Going to test more today.  But I think you deleted some code that you shouldn't have so lets add back the JobDuration.

dave-gantenbein:
Will resolve under #1750

PR #1741, opened 2022-11-04
Localdev tweaks for lookout, armada-server to work; add priority classes

suprjinx:


localdev/config/armada/config.yaml
```diff
@@ -25,8 +38,9 @@ eventapi:
   postgres:
     connection:
       host: "postgres"
+defaultToLegacyEvents: true
 pulsar:
-  enabled: true
+  enabled: false
```
Sharpz7, line 43: Should pulsar be getting disabled?

localdev/config/executor/config.yaml
```diff
@@ -36,7 +36,7 @@ kubernetes:
   stuckTerminatingPodExpiry: 1m
   podDefaults:
     ingress:
-      hostnameSuffix: "svc"
+      hostnameSuffix: "localhost"
```
Sharpz7, line 39: Can I ask what the behavior of this is?

Like for example, say you have a pod on {pod-id}.{etc etc}.localhost, will that resolve?

localdev/config/lookout/config.yaml
```diff
@@ -13,10 +13,10 @@ grpc:
     permitWithoutStream: false
 
 uiConfig:
-  armadaApiBaseUrl: "http://armada-server:8080"
+  armadaApiBaseUrl: "http://localhost:8080"
```
Sharpz7, line 16: Why is this being done? Shouldn't we default to the docker network namespace?

localdev/config/lookout/config.yaml
```diff
@@ -13,10 +13,10 @@ grpc:
...
   userAnnotationPrefix: "armadaproject.io/"
   binocularsEnabled: true
-  binocularsBaseUrlPattern: "http://binoculars:8082" # {CLUSTER_ID} gets replaced by appropriate cluster id
+  binocularsBaseUrlPattern: "http://localhost:8082" # {CLUSTER_ID} gets replaced by appropriate cluster id
```
Sharpz7, line 19: Same as above

localdev/config/armada/config.yaml
suprjinx: armadactl watch does not seem to work with this set true. now matches the config in e2e

localdev/config/executor/config.yaml
suprjinx: it may or may not, but was just changing to match e2e config

localdev/config/lookout/config.yaml
suprjinx: this url is used in web ui (browser) so the internal/docker name doesn't work

localdev/config/lookout/config.yaml
suprjinx: same answer

localdev/config/armada/config.yaml
Sharpz7: Interesting. I'll see if this changes anything for me.

localdev/config/executor/config.yaml
Sharpz7: I doubt it will resolve. But if it does, @kannon92 this might let you get the armada-jupyter stuff working?

localdev/config/lookout/config.yaml
Sharpz7: Ah yeah good point, because those names won't resolve on a MacOS Host.

Sharpz7 (approved):
LGTM, definitely need to add comments / documentaition on these features soon.

PR #1750, opened 2022-11-07
Column fixed with horizontal scrolling

carlocamurri:


kannon92 (approved):
Did a test with annotations and all that.  LGTM.

PR #1761, opened 2022-11-08
Fix Lookout bug time in state using the first Job Run rather than latest one

carlocamurri:


internal/lookout/repository/jobs.go
```diff
@@ -262,27 +262,55 @@ func sortJobsByJobId(jobInfos []*lookout.JobInfo, descending bool) {
...
-		return ""
-	}
+func (r *SQLJobRepository) updateTimeInState(jobInfo *lookout.JobInfo) {
 	var timeStamp time.Time
 
+	state := JobState(jobInfo.JobState)
 	switch state {
 	case JobSucceeded, JobFailed:
-		timeStamp = row.Finished.Time
+		possibleTime := findLatest(jobInfo.Runs, func(run *lookout.RunInfo) *time.Time { return run.Finished })
```
kannon92, line 271: Could you combine the if else loop into a common function?

```
		if possibleTime == nil {
			log.Warnf("No finished timestamp found for job with id %s", jobInfo.Job.Id)
		} else {
			timeStamp = *possibleTime
		}
```
The main difference between all these blocks of code is the error message which I think you can easily parametrize.

PR #1763, opened 2022-11-08
Make template_lookout_url an optional argument to ArmadaOperator.__init__

ClifHouck:


third_party/airflow/armada/operators/armada.py
```diff
@@ -58,7 +58,7 @@ def __init__(
         job_service_client: JobServiceClient,
         armada_queue: str,
         job_request_items,
-        lookout_url_template: str,
+        lookout_url_template: str = "",
```
kannon92, line 61:
```suggestion
        lookout_url_template: Optional[str] = None,
```
This is typically the way of declaring types optional

third_party/airflow/armada/operators/armada.py
```diff
@@ -58,7 +59,7 @@ def __init__(
         job_service_client: JobServiceClient,
         armada_queue: str,
         job_request_items,
-        lookout_url_template: str,
+        lookout_url_template: Optional[str] = "",
```
kannon92, line 62: I think I'd set this to None.  Otherwise you will never hit the code in your check ever.

kannon92 (approved):

PR #1766, opened 2022-11-09
Use a "docker native" instead of host-mounted volume for go mods etc

suprjinx:


localdev/docker-compose.yaml
```diff
@@ -4,6 +4,9 @@ networks:
   kind:
     external: true
 
+volumes:
```
kannon92, line 7: Could you explain this as a comment?  

I don't really understand how this fixes the issues you have.  I'll admit that I am not familar with "docker native".

suprjinx: well the volume was previously mounted from the host computer file system, macos in my case, which has performance issues on macos

i'll add a comment about what it does here

kannon92 (approved):

Sharpz7 (approved):

PR #1778, opened 2022-11-10
Move the scheduler ingester to the new common framework.

d80tb7:
This PR moves the scheduler ingester out into its own component and moves it to the common ingestion framework. Other things done as part of this PR are:
* Moved Upsert to `common` as it is generally useful and not ingester specific
* Added a set of tests for conversion between event sequences and db ops

This PR makes the ingester a lot more prod ready but we're not there quite yet.  As a result I haven't yet created a helm chart etc.


internal/scheduleringester/instructions_test.go
```diff
@@ -0,0 +1,227 @@
...
+				assert.Equal(t, expectedSubmit, actualSubmit)
+			}
+		default:
+			assert.Equal(t, expectedOp, actualOp)
+		}
+	}
+}
+
+func assertSchedulingInfoEqual(t *testing.T, expectedBytes []byte, actualBytes []byte) {
+	actualSchedInfo, err := unmarshalSchedulingInfo(actualBytes)
```
carlocamurri, line 194: Nice to add `t.Helper()` here

internal/scheduler/api.go
```diff
@@ -52,10 +55,10 @@ func (srv *ExecutorApi) StreamingLeaseJobs(stream api.AggregatedQueue_StreamingL
 	}
 
 	// Get leases assigned to this executor.
-	queries := New(srv.Db)
+	queries := sqlc.New(srv.Db)
```
d80tb7, line 58: the `sqlc` directory here is probably temporary.  I've moved the database stuff here partly to avoid cyclic references and partly because calling `scheduler.New()` didn't create a new scheduler, but rather just the database and I found that confusing.

One of the next taks is to start using database migrations for this database so I expect this will change again at that point.

internal/scheduler/leaderelection.go
```diff
@@ -5,6 +5,9 @@ import (
 	"fmt"
```
d80tb7, line 5: I'm planning to remove this file in a subsequent PR as we plan to do leader election via K8s rather than by database

internal/scheduler/leaderelection_test.go
```diff
@@ -6,6 +6,9 @@ import (
 	"testing"
 	"time"
```
d80tb7, line 7: When we do leader election via K8s, this file will be removed.

internal/scheduleringester/instructions_test.go
d80tb7: done

internal/scheduler/sqlc/query.sql.go
```diff
@@ -1,13 +1,12 @@
-// Code generated by sqlc. DO NOT EDIT.
```
severinson, line 1: Did you manually move and rename package files instead of re-generating using sqlc into a new directory?

severinson (approved):

internal/scheduler/sqlc/query.sql.go
severinson: Btw. You can install `sqlc` and other tools using this make target.
https://github.com/G-Research/armada/blob/master/makefile#L590

internal/scheduler/api.go
severinson: Using sqlc as a package name is a bit confusing I think. Since we import another package named sqlc.

internal/scheduler/sqlc/query.sql.go
d80tb7: yeah I did- I'm going to change this again as the next bit of work to move the database to migration scripts so that time I'll do the migration properly.

internal/scheduler/api.go
d80tb7: Yeah I'll fix this in the next pr

PR #1779, opened 2022-11-10
WIP: Test Lookout Ingester as part of E2E tests

### LINKED ISSUE #1495: Make Lookout Part of E2E Tests
Currently the E2E tests don't test any of lookout.  Testing the UI might be hard, but we should at least be able to test the state of the database fairly easily.  I think to do this we need to:

1. Make lookout-ingester be instantiated as part of e2e tests
2. Write a test that asserts the state of the database (job in right state, correct number of job runs, job yaml as expected etc)

ClifHouck:
Closes #1495


ClifHouck:
Probably need to add some more tests, but this is functional at least.

makefile
```diff
@@ -439,6 +440,12 @@ tests-e2e-setup: setup-cluster
 		-e ARMADA_APICONNECTION_ARMADAURL="server:50051" \
 		-e ARMADA_APICONNECTION_FORCENOTLS=true \
 		armada-executor --config /e2e/setup/insecure-executor-config.yaml
+	docker run -d --name lookout-ingester-migrate  --network=kind -v ${PWD}/e2e:/e2e \
+		armada-lookout-ingester --config /e2e/setup/lookout-ingester-config.yaml --migrateDatabase
+	sleep 3
```
kannon92, line 445: Is there anything we can check to see if this job worked without sleeping?

makefile
```diff
@@ -439,6 +440,12 @@ tests-e2e-setup: setup-cluster
...
+	docker logs lookout-ingester-migrate
```
kannon92, line 446: Why do we need to get logs here?

e2e/lookout_ingester_test/lookout_ingester_test.go
```diff
@@ -0,0 +1,200 @@
...
+	})
+	assert.NoError(t, err)
+}
+
+type PartialJob struct {
+	JobId string
+	State int
+}
+
+func awaitJobEntry(db *sql.DB, jobId string, pollWait time.Duration, expiresAfter time.Duration, expectedState int) (*PartialJob, error) {
```
d80tb7, line 110: It's worth noting that in general this can only be guaranteed to work for terminal states.  For non-terminal states there are two cases where it can fail:
- The state is updated in the database twice (or more) between polls
- The state is conflated out by the lookout ingester and thus is never written to the db at all.
I think so far you've only used it for terminal states so all is good, but probably warrants a comment (and maybe an error if the user passes a non terminal state).

The same also goes for `awaitJobRunEntry`

e2e/lookout_ingester_test/lookout_ingester_test.go
```diff
@@ -0,0 +1,200 @@
...
+		assert.NoError(t, err)
+		assert.NotNil(t, pJob)
+		assert.Equal(t, pJob.JobId, submitResponse.JobResponseItems[0].JobId)
+
+		return nil
+	})
+	assert.NoError(t, err)
+}
+
+func TestLookoutIngesterUpdatesPostgresWithJobInfoFailedJob(t *testing.T) {
```
d80tb7, line 68: I like this but note there's a couple of different ways that a job can fail and each way determines what is written into lookout.  Specifically if a job actually starts up (which it won't in this case) we have to write the exit code(s) into the db.

makefile
d80tb7: assuming this runs synchronously I'm not sure why we need to sleep here? As soon as the process exists we know that the db is in a good state. Is it to make sure the logs are available.

makefile
d80tb7: Are we able to check the exit code of the migration?  If  it's non-zero then we should bail here and print the logs.  If not, we can just assume the migration code worked and not print anything.

d80tb7:
I like the way this is going.

ClifHouck:
Maybe still want some more tests, but this feels like its in a much better spot now.

kannon92 (commented):
Could you refactor the tests to work like a table driven test?

PR #1794, opened 2022-11-14
[Ingesters] Bug In Batcher When No Message Received In Timeout

d80tb7:
The batcher was only resetting when it flushed its buffer. In cases where the buffer was empty the batcher was not being reset which meant it only released its messages when the buffer became full and not when the time condition fired.

This PR addresses the above issue and adds a test confirming the fix

As a bonus fix- I've also corrected the default config for `receiveTimeout` and `backoffTime` for the lookout and event ingesters.  Previously we moved this config into `PulsarConfig` but the config files were not updated.

### DIFF
--- a/internal/common/ingest/batch.go
+++ b/internal/common/ingest/batch.go
@@ -58,8 +58,8 @@ func (b *Batcher[T]) Run(ctx context.Context) {
 				b.mutex.Lock()
 				if len(b.buffer) > 0 {
 					b.callback(b.buffer)
-					appendToBatch = false
 				}
+				appendToBatch = false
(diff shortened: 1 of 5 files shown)


internal/common/ingest/batch.go
```diff
@@ -58,8 +58,8 @@ func (b *Batcher[T]) Run(ctx context.Context) {
 				b.mutex.Lock()
 				if len(b.buffer) > 0 {
 					b.callback(b.buffer)
-					appendToBatch = false
```
d80tb7, line 61: this was the bug

PR #1798, opened 2022-11-16
Localdev airflow

### LINKED ISSUE #1774: Localdev: Add airflow support
There are two possible methods for this:

- Install airflow as a container using docker-compose
- Install airflow through kind using https://airflow.apache.org/docs/helm-chart/1.0.0/quick-start.html

suprjinx:
closes #1774


richscott (approved):

third_party/airflow/README.md
```diff
@@ -36,6 +36,16 @@ python3.8 -m pip install armada-airflow
 
 From the top level of the repo, you should run `make airflow-operator`.  This will generate proto/grpc files in the jobservice folder.
 
+Airflow with the Armada operator can be run alongside the other Armada services via the localdev docker-compose
```
Sharpz7, line 39: When you refer to an "armada operator" here, what do you mean?

localdev/scripts/airflow-start.sh
```diff
@@ -0,0 +1,20 @@
+# download the needful
```
Sharpz7, line 1: Has there been testing on whether this works with the armada python code?

third_party/airflow/README.md
suprjinx: well Armada operator is the airflow plugin that we provide. So this runs an airflow instance (other software) with our operator installed.

localdev/scripts/airflow-start.sh
suprjinx: yes i have tested

richscott (approved):

third_party/airflow/README.md
Sharpz7: Do we maybe want to be more specific to what we mean hear since we are building an armada operator seperately to handle installs?

At the very least create an issue to track that this change will be needed to our docs in future.

kannon92 (approved):

PR #1801, opened 2022-11-18
Gang scheduling

severinson:


config/armada/config.yaml
```diff
@@ -54,6 +54,8 @@ scheduling:
   maximumJobsToSchedule: 5000
   maxQueueReportsToStore: 1000
   MaxJobReportsToStore: 100000
+  resourceScarcity:
```
d80tb7, line 57: is adding this going to change the behaviour in prod clusters- e.g. for a gpu pool which only had scarcity set for gpus, it'll now start considering cpus?

internal/common/validation/job.go
```diff
@@ -5,15 +5,95 @@ import (
...
-	return ValidatePodSpecPriorityClass(j.PodSpec, config.Enabled, config.PriorityClasses)
+	for _, job := range jobs {
+		if err := ValidateApiJob(job, config); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func validateGangs(jobs []*api.Job, gangIdAnnotation, gangCardinalityAnnotation string) error {
```
d80tb7, line 27: we probably also have to validate that tolerations, node affinities and node selectors are equal for everything in the gang.  Otherwise you could end up with a gang that could never be scheduled (specifically the case where all members of the gang can be scheduled,  but only only disjoint clusters.)

internal/scheduler/legacyscheduler.go
```diff
@@ -200,110 +226,166 @@ func (it *QueueCandidateJobsIterator) schedulingReportFromJob(ctx context.Contex
...
+		it.MaximalResourceFractionPerQueue,
 	); exceeded {
 		jobSchedulingReport.UnschedulableReason = reason + " (total limit for this queue)"
 		return jobSchedulingReport, nil
 	}
 
-	// Check per-round resource limits for this queue.
-	if exceeded, reason := it.exceedsResourceLimits(
+	// MaximalCumulativeResourceFractionPerQueueAndPriority check.
+	if exceeded, reason := exceedsPerPriorityResourceLimits(
```
d80tb7, line 284: as discussed- this needs to be checked for the whole gang.

internal/scheduler/legacyscheduler.go
```diff
@@ -318,151 +400,180 @@ func uuidFromUlidString(ulid string) (uuid.UUID, error) {
...
+type CandidateGangIterator struct {
+	ctx context.Context
+	CandidateJobsIterator
+	// Jobs are grouped into gangs by this annotation.
+	gangIdAnnotation string
+	// Jobs in a gang specify the number of jobs in the gang via this annotation.
+	gangCardinalityAnnotation string
+	// Map from queue to the gangs for that queue.
+	// Each gang is made up of a slice of jobs with equal value for gangIdAnnotation.
+	gangsByQueue map[string]map[string][]*JobSchedulingReport
```
d80tb7, line 473: It looks like the `jobSpechedulingReport` seems to be moving from being a "Report" (i.e. a record of what happened), to a "context" (i.e. something that's storing the state of the entire scheduling round).  Is it worth renaming (or even refactoring) this.

internal/scheduler/legacyscheduler.go
```diff
@@ -318,151 +400,180 @@ func uuidFromUlidString(ulid string) (uuid.UUID, error) {
...
+		}
+		if report == nil {
+			return nil, nil
+		}
+
+		gangId, gangCardinality, isGangJob, err := GangIdAndCardinalityFromJob(report.Job, it.gangIdAnnotation, it.gangCardinalityAnnotation)
+		if err != nil {
+			log := ctxlogrus.Extract(it.ctx)
+			logging.WithStacktrace(log, err).Errorf("failed to get gang cardinality for job %s", report.Job.Id)
+			gangCardinality = 1 // Schedule jobs with invalid gang cardinality one by one.
```
d80tb7, line 509: I don't like this- i.e. catching arbitrary errors by defaulting a value, unless there is good reason to think it safe in all cases.  What sort of errors are we expecting here?

internal/scheduler/legacyscheduler.go
```diff
@@ -318,151 +400,180 @@ func uuidFromUlidString(ulid string) (uuid.UUID, error) {
...
+		}
+		if isGangJob {
```
d80tb7, line 511: do we write in the report anywhere that the gang couldn't be scheduled because we don't have all the members?

internal/scheduler/legacyscheduler.go
```diff
@@ -318,151 +400,180 @@ func uuidFromUlidString(ulid string) (uuid.UUID, error) {
...
+			jobsByGangId, ok := it.gangsByQueue[report.Job.Queue]
```
d80tb7, line 512: I wonder if it's possible to write a default multimap now we have generics, in which case all this would reduce to a single line.  Isuspectwe probably can't due to maps (and slices) being special cased int  he compiler.

internal/scheduler/legacyscheduler.go
```diff
@@ -318,151 +400,180 @@ func uuidFromUlidString(ulid string) (uuid.UUID, error) {
...
+func (sched *LegacyScheduler) podRequirementsFromJobs(jobs []*api.Job) []*schedulerobjects.PodRequirements {
+	rv := make([]*schedulerobjects.PodRequirements, 0, len(jobs))
+	for _, job := range jobs {
+		rv = append(rv, sched.podRequirementsFromJob(job)...)
+	}
+	return rv
+}
+
+func (sched *LegacyScheduler) podRequirementsFromJob(job *api.Job) []*schedulerobjects.PodRequirements {
+	rv := make([]*schedulerobjects.PodRequirements, 0, 1+len(job.PodSpecs))
```
d80tb7, line 539: this is a weird function:
- It allows multiple pod specs (which we don't elsewhere)
- It allows both podspec and podspecs (which we don't elsewhere)

internal/scheduler/legacyscheduler.go
```diff
@@ -474,205 +585,127 @@ func (sched *LegacyScheduler) String() string {
...
 	jobRepository SchedulerJobRepository,
 	priorityFactorByQueue map[string]float64,
+	initialResourcesByQueueAndPriority map[string]schedulerobjects.QuantityByPriorityAndResourceType,
 ) (*LegacyScheduler, error) {
-	priorities := make([]int32, 0)
-	cumulativePriorityLimits := make(map[int32]map[string]float64, 0)
-	for _, priority := range schedulingConfig.Preemption.PriorityClasses {
-		priorities = append(priorities, priority.Priority)
-		cumulativePriorityLimits[priority.Priority] = priority.MaximalResourceFractionPerQueue
+	if ResourceListAsWeightedApproximateFloat64(constraints.ResourceScarcity, constraints.TotalResources) == 0 {
```
d80tb7, line 596: I'm not in an ide here so it's a bit difficuly to check, but comparing the rsults of a function  called `ResourceListAsWeightedApproximateFloat64` with an integer (in this case 0) seems wrong.

internal/scheduler/legacyscheduler.go
```diff
@@ -691,6 +724,32 @@ func (sched *LegacyScheduler) Schedule(
...
+	}
+	gangCardinalityString, ok := annotations[gangCardinalityAnnotation]
+	if !ok {
+		return "", 0, false, errors.Errorf("missing annotation %s", gangCardinalityAnnotation)
+	}
+	gangCardinality, err := strconv.Atoi(gangCardinalityString)
+	if err != nil {
+		return "", 0, false, errors.WithStack(err)
+	}
+	if gangCardinality == 0 {
```
d80tb7, line 747: should this be `if gangCardinality <= 0`

internal/scheduler/legacyscheduler.go
```diff
@@ -765,7 +824,12 @@ func pickQueueRandomly(weights map[string]float64, random *rand.Rand) (string, f
 	for _, share := range weights {
 		sum += share
 	}
-	pick := sum * random.Float64()
+	var pick float64
+	if random != nil {
```
d80tb7, line 828: why do we support passing a nil random into this function?

internal/scheduler/nodeiteration_test.go
```diff
@@ -392,13 +392,13 @@ func testNGpuNode(n int, priorities []int32) []*schedulerobjects.Node {
...
 }
 
-func testNTaintedtGpuNode(n int, priorities []int32) []*schedulerobjects.Node {
-	rv := make([]*schedulerobjects.Node, n)
-	for i := 0; i < n; i++ {
-		rv[i] = testTaintedGpuNode(priorities)
-	}
-	return rv
-}
+// func testNTaintedtGpuNode(n int, priorities []int32) []*schedulerobjects.Node {
```
d80tb7, line 395: why si this commented out?

internal/scheduler/nodeiteration_test.go
```diff
@@ -483,38 +483,38 @@ func testGpuNode(priorities []int32) *schedulerobjects.Node {
...
-			priorities,
-			map[string]resource.Quantity{
-				"cpu":    resource.MustParse("64"),
-				"memory": resource.MustParse("1024Gi"),
-				"gpu":    resource.MustParse("8"),
-			},
-		),
-	}
-}
+// func testTaintedGpuNode(priorities []int32) *schedulerobjects.Node {
```
d80tb7, line 486: and this!

internal/scheduler/reports.go
```diff
@@ -89,102 +89,107 @@ func (report *SchedulingRoundReport) String() string {
...
-	} else {
-		if m, ok := schedulingRoundReport.UnsuccessfulJobSchedulingReportsByQueue[queue]; ok {
-			m[jobSchedulingReport.JobId] = jobSchedulingReport
-		} else {
-			schedulingRoundReport.UnsuccessfulJobSchedulingReportsByQueue[queue] = map[uuid.UUID]*JobSchedulingReport{
-				jobSchedulingReport.JobId: jobSchedulingReport,
-			}
-		}
+func (report *SchedulingRoundReport) AddJobSchedulingReport(r *JobSchedulingReport) {
+	if report == nil {
```
d80tb7, line 104: Why is it ok for this function to be null safe but others not?  In general I'm against making public functions null safe other than  in some special case (appending to slic es being a good example), because it risks an unintended null going unrecognised and then  surfacing as a problem in  some other random part of the program.

internal/scheduler/schedulerobjects/resourcelist.go
```diff
@@ -33,13 +33,17 @@ func (a QuantityByPriorityAndResourceType) String() string {
...
-		rsa := a[p]
-		rsa.Add(rsb)
-		a[p] = rsa
+	for p, rlb := range b {
+		a.AddResouceList(p, rlb)
 	}
 }
 
+func (a QuantityByPriorityAndResourceType) AddResouceList(priority int32, rlb ResourceList) {
+	rla := a[priority]
```
d80tb7, line 42: what happens if a[priority]  doesn't exist?

config/armada/config.yaml
severinson: Per-pool constraints trump the global setting.

internal/scheduler/schedulerobjects/resourcelist.go
severinson: It's initialised to the zero value, which is safe.

internal/scheduler/legacyscheduler.go
severinson: Done

internal/scheduler/legacyscheduler.go
severinson: I agree. It may be best to do in a subsequent pr.

internal/scheduler/legacyscheduler.go
severinson: This should never happen. I put it in to ensure jobs don't get stuck if we make a mistake in the future. I.e., if a job with an invalid gang cardinality annotation gets into the scheduler somehow. But maybe it's better to just not schedule it...

internal/scheduler/legacyscheduler.go
severinson: For this function, I think it's reasonable to fall back to the default if a rand isn't provided.

internal/scheduler/reports.go
severinson: Agreed. I added it while refactoring and forgot to remove it.

internal/scheduler/legacyscheduler.go
severinson: Yes :)

internal/scheduler/legacyscheduler.go
severinson: The compiler should understand that the `0` is of type `float64`. I don't think there are any floating point comparison issues re comparing with 0.

internal/scheduler/legacyscheduler.go
severinson: I agree that it's a bit weird. We may want to leave it as-is for now since it'd be good to be moving towards a scheduler that supports multi-pod jobs.

internal/scheduler/legacyscheduler.go
severinson: Doing so isn't necessary anymore, since we collect gangs on a per-queue basis now.

internal/common/validation/job.go
severinson: My preference is to leave it as-is for now and to add a proper check that runs the scheduler against the job in a subsequent PR.

severinson: We need a way to run the `NodeDb` matching logic in a special mode that doesn't consider resources consumed by running jobs. If we then also cache the most recent `NodeDb` for each cluster, we can run submitted jobs against these cached `NodeDb` instances and check if any could schedule the job.

We'd need to do this anyway if we are to keep a submission-time check when we remove core Redis and the old scheduler.

internal/scheduler/legacyscheduler.go
severinson: We'd need a bit of extra code. But it's doable.

internal/scheduler/legacyscheduler.go
```diff
@@ -114,196 +177,300 @@ func queuedJobsIteratorLoader(ctx context.Context, jobIds []string, ch chan *api
...
+		if err != nil {
+			return nil, err
+		}
+		if job == nil {
+			return nil, nil
+		}
+
+		gangId, gangCardinality, isGangJob, err := GangIdAndCardinalityFromJob(job, it.gangIdAnnotation, it.gangCardinalityAnnotation)
+		if err != nil {
+			log := ctxlogrus.Extract(it.ctx)
```
d80tb7, line 224: as discussed we should just fail the job here, but if we can't do this quickly, let's return `[]*api.Job{job}, nil` as otherwise we end up making decisions based on values returned from a function that has errored.

d80tb7: this can be done as a subsequent pr

d80tb7 (approved):

PR #1809, opened 2022-11-21
Provide Breakdown of Key Prometheus Metrics By PriorityClass

d80tb7:


internal/armada/cache/queue.go
```diff
@@ -28,11 +28,9 @@ type QueueCache struct {
 	jobRepository            repository.JobRepository
 	schedulingInfoRepository repository.SchedulingInfoRepository
 
-	refreshMutex     sync.Mutex
-	queuedDurations  map[string]map[string]*metrics.FloatMetrics
```
d80tb7, line 32: This change is why the code changes are a bit more extensive than one might expect.  Previously we stored 4 separate values of `map[string]map[string]*metrics` where the first key is the queue and the second key the pool.  I really didn't fancy adding priority class by making this a `map[string]map[string]map[string]*metrics` where the final map is keyed by priority class so I've changed it to be ` map[string][]*metrics` where the map is keyed by queue and the metrics object contains the values for pool and priorityClass.  This is fine because:

- We update the map atomically by queue
- The queue cache itself ensures that each slice doesn't contain duplicates for a given priority class, pool combination.
-  The consumers of the cache want to consume this data as a slice (i.e thy want to iterate over it) rather than a map (i.e. making arbitrary lookoups by key).

JamesMurkin (approved):

PR #1828, opened 2022-11-26
[CleanUp] Remove Deprecated Code

d80tb7:
* Old Non-Streaming Lease requests
* Jetstream
* Old Scheduler (only half of this could be removed, because many of the functions are used in metrics calculation)


internal/lookout/application.go
```diff
@@ -55,34 +54,20 @@ func StartUp(config configuration.LookoutConfiguration, healthChecks *health.Mul
...
-				panic(err)
-			}
-			eventStream = eventstream.NewStanEventStream(
-				config.Nats.Subject,
-				stanClient,
-				stan.SetManualAckMode(),
-				stan.StartWithLastReceived())
-
-			healthChecks.Add(stanClient)
+		stanClient, err := eventstream.NewStanClientConnection(
```
suprjinx, line 57: can we remove this nats ingestion from lookout, since lookoutingester should handle via pulsar?

suprjinx (approved):

internal/lookout/application.go
d80tb7: we can- but I plan to do this in a follow up where we rip out large amounts of nats code

JamesMurkin (approved):

PR #1842, opened 2022-11-28
add airflow integration test

### LINKED ISSUE #1773: Add E2E Tests for Airflow Operator [Without Airflow]
We have https://github.com/G-Research/armada/blob/master/third_party/airflow/tests/integration/test_airflow_operator_logic.py which functions as a local integration test.  

We should add this into our github actions as part of our integration tests.  

The work needed:

1) Setup the e2e environment using localdev
2) Install all the python dependencies (armada-client, airflow-operator, etc)
3) Write a github action that runs pytest on the test_airflow_operator.

Monitor test to make sure it is not fickle.

kannon92:
Fixes #1773


localdev/run.sh
```diff
@@ -7,26 +7,31 @@ ARMADA_SVCS="armada-server lookout lookout-ingester executor binoculars jobservi
...
+    echo "Detecting Linux so will use normal pulsar image."
 fi
 
 # see if pulsar is already up, in which case we don't need to sleep
-SLEEP_TIME=0
-docker-compose ps | grep -E "pulsar.+running" &> /dev/null
-if [ $? -ne 0 ];
+SLEEP_TIME=50
+compose_list="$(docker-compose ps)"
+if [[ $compose_list == *"pulsar.+running"* ]];
```
suprjinx, line 31: does this regex pattern work for the bash matching?

makefile
```diff
@@ -382,7 +382,7 @@ rebuild-executor: build-docker-executor
 
 .ONESHELL:
 tests-e2e-teardown:
-	docker rm -f nats redis pulsar server executor postgres || true
+	docker rm -f nats redis pulsar server executor postgres lookout-ingester-migrate lookout-ingester jobservice || true
```
suprjinx, line 385: this is tearing down services that are not started by the "-setup" target

suprjinx (approved):

makefile
kannon92: Good call.

localdev/run.sh
kannon92: I think that is the wildcard part

makefile
kannon92: This is actually a bug fix that I notice tests-e2e-setup creates but does not teardown.  

But for this work, I should actually call teardown-localdev.

localdev/run.sh
kannon92: I see your point.  Is it a bad assumpition to just assume that if pulsar is running in docker environment it is running?

suprjinx (approved):

PR #1847, opened 2022-11-30
[WIP] Lookout v2 API

carlocamurri:


pkg/api/api.swagger.go
```diff
@@ -5047,6 +5047,14 @@ func SwaggerJsonTemplate() string {
 		"    \"v1Volume\": {\n" +
 		"      \"type\": \"object\",\n" +
 		"      \"title\": \"Volume represents a named volume in a pod that may be accessed by any container in the pod.\",\n" +
+		"      \"required\": [\n" +
```
carlocamurri, line 5050: Assuming these changes got picked up, but were introduced by something else

internal/armada/server/submit.go
```diff
@@ -915,8 +915,8 @@ func (server *SubmitServer) applyTerminationGracePeriodDefault(spec *v1.PodSpec)
 // fillContainerRequestsAndLimits updates resource's requests/limits of container to match the value of
 // limits/requests if the resource doesn't have requests/limits setup. If a Container specifies its own
 // memory limit, but does not specify a memory request, assign a memory request that matches the limit.
-// Similarly, if a Container specifies its own CPU limit, but does not specify a CPU request, automatically
-// assigns a CPU request that matches the limit.
+// Similarly, if a Container specifies its own Cpu limit, but does not specify a Cpu request, automatically
```
kannon92, line 918: I think CPU is actually correct in this case.

internal/common/database/functions.go
```diff
@@ -27,3 +33,49 @@ func OpenPgxPool(config configuration.PostgresConfig) (*pgxpool.Pool, error) {
 	err = db.Ping(context.Background())
 	return db, err
 }
+
+func UniqueTableName(table string) string {
```
kannon92, line 37: Can you add tests for these functions?

internal/lookoutv2/generate/main.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/mount"
+	"github.com/docker/docker/client"
+	"github.com/docker/docker/pkg/stdcopy"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	armadaMountDir  = "/mnt/armada"
```
kannon92, line 20: I think this may be specific for wsl?  Not sure this would work on native linux or mac?

internal/common/util/list.go
```diff
@@ -45,3 +45,11 @@ func Filter[T any](list []T, predicate func(val T) bool) []T {
 	}
 	return out
 }
+
+func Map[T any, U any](list []T, fn func(val T) U) []U {
```
d80tb7, line 49: this is a nit but I think really this should be in a package called `slices` so its `slices.Map()` (as opposed to `arrays.Map()` or `maps.Map()`).

internal/common/util/map.go
```diff
@@ -58,3 +58,11 @@ func FilterKeys(a map[string]string, keys []string) map[string]string {
 
 	return result
 }
+
+func InverseMap[K comparable, V comparable](a map[K]V) map[V]K {
```
d80tb7, line 62: We probably should put some doc on this function to say what it does in the case of duplicate keys,

internal/lookoutv2/conversions/convert.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func FromSwaggerOrder(order *models.Order) *model.Order {
+	return &model.Order{
+		Direction: order.Direction,
+		Field:     order.Field,
+	}
+}
+
+func toSwaggerTimePtr(swaggerTs *time.Time) *strfmt.DateTime {
```
d80tb7, line 82: why is the input called `swaggerTs` when this is converting into a swagger type?

internal/lookoutv2/repository/common.go
```diff
@@ -0,0 +1,739 @@
...
+	queryOrd, err := qb.makeQueryOrder(order, queryTables)
+	if err != nil {
+		return nil, err
+	}
+	fromSql, err := qb.queryTablesToSql(queryTables, jobIdCol)
+	if err != nil {
+		return nil, err
+	}
+	valuesMap := make(map[string]interface{})
+	whereSql, err := qb.queryFiltersToSql(queryFilters, valuesMap)
```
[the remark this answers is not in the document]
d80tb7: More generally I think I'd document every function in this file as our unique requirements means that there's a lot going on here.

internal/lookoutv2/conversions/convert.go
carlocamurri: Should be the other way around (input should be `ts`, output should be `swaggerTs`)
`strfmt.DateTime` is defined in go-openapi/strfmt

internal/lookoutv2/repository/common.go
```diff
@@ -0,0 +1,739 @@
...
+	return &Query{
+		Sql:  sql,
+		Args: []interface{}{},
+	}, tempTable
+}
+
+func (qb *QueryBuilder) JobCount(filters []*model.Filter) (*Query, error) {
+	err := qb.validateFilters(filters)
+	if err != nil {
+		return &Query{}, errors.Wrap(err, "filters are invalid")
```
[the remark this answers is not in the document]
carlocamurri: My error, doing this a few times so fixed all instances of this

internal/lookoutv2/repository/common_test.go
```diff
@@ -0,0 +1,94 @@
+package repository
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/G-Research/armada/internal/common/util"
+)
+
```
d80tb7, line 10: There's fewer tests here than I would expect given the complexity of the code in `common.go`

internal/lookoutv2/repository/getjobs.go
```diff
@@ -0,0 +1,376 @@
...
+	tx, err := r.db.BeginTx(ctx, pgx.TxOptions{
+		IsoLevel:       pgx.RepeatableRead,
+		AccessMode:     pgx.ReadWrite,
+		DeferrableMode: pgx.Deferrable,
+	})
+	if err != nil {
+		log.WithError(err).Error("failed to start transaction")
+		return nil, err
+	}
+	defer func() {
```
d80tb7, line 88: `r.db.BeginTxFunc` might make this a bit easier as you won't have to sort out handling the transaction cleanup in a defer.

internal/lookoutv2/repository/getjobs.go
```diff
@@ -0,0 +1,376 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &GetJobsResult{
+		Jobs:  jobs,
+		Count: count,
+	}, nil
+}
+
+func rowsToJobs(jobRows []*jobRow, runRows []*runRow, annotationRows []*annotationRow, order *model.Order) ([]*model.Job, error) {
```
d80tb7, line 158: so this probably a dumb question, but why do we have both `model.Job` and the swagger job.  Could we convert to swagger job straight away or does that lose useful infomation due to the limitations of json?

internal/lookoutv2/repository/groupjobs.go
```diff
@@ -0,0 +1,148 @@
...
+	ctx context.Context,
+	filters []*model.Filter,
+	order *model.Order,
+	groupedField string,
+	aggregates []string,
+	skip int,
+	take int,
+) (*GroupByResult, error) {
+	var err error
+	tx, err := r.db.BeginTx(ctx, pgx.TxOptions{
```
d80tb7, line 56: As above `BeginTxFunc` might make this easier

makefile
```diff
@@ -503,7 +503,7 @@ junit-report:
 	rm -f test_reports/junit.xml
 	$(GO_TEST_CMD) bash -c "cat test_reports/*.txt | go-junit-report > test_reports/junit.xml"
 
-setup-proto: download
+setup-proto:
```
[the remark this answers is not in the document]
kannon92: May be breaking the python steps actually

d80tb7 (approved):

PR #1853, opened 2022-12-01
Fix EventIngester Logging BackOffTime Incorrectly

d80tb7:
EventIngester is logging the BackOffTime as seconds when in fact the value it's using is nanoseconds.  This leads to alarming log lines such as "Retryable error encountered, will wait 200000000 seconds before retrying".


samclark (approved):
Approved ... even better imho would be a more human readable unit

d80tb7:
Yeah you're right- let me change this

ClifHouck (approved):
LGTM

PR #1854, opened 2022-12-01
Add magefile.go for compiling proto

severinson:


magefile.go
```diff
@@ -0,0 +1,350 @@
...
+	case "darwin/amd64":
+		return "osx-x86_64", nil
+	case "darwin/arm64":
+		return "osx-aarch_64", nil
+	case "linux/386":
+		return "linux-x86_32", nil
+	case "linux/amd64":
+		return "linux-x86_64", nil
+	case "linux/arm64":
+		return "linux-aarch_64", nil
```
dejanzele, line 330: should it be `linux-aarch_64` (double a)

magefile.go
```diff
@@ -0,0 +1,350 @@
...
+	case "windows/386":
+		return "win32", nil
+	case "windows/amd64":
+		return "win64", nil
+	}
+	return "", fmt.Errorf("protoc not supported on %s/%s", runtime.GOOS, runtime.GOARCH)
```
dejanzele, line 336: I'd rather not use `fmt.Errorf` as it is deprecated, `github.com/pkg/errors` is better but not maintained.
I think https://github.com/go-faster/errors is the successor.

magefile.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	return sh.OutCmd("protoc"), nil
+}
+
+func exists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err == nil {
+		return true, nil
+	}
+	if os.IsNotExist(err) {
```
dejanzele, line 313: can you replace this block with just `return false, err`, caller anyways need to check is `err != nil` and you return false in both cases.

magefile.go
```diff
@@ -0,0 +1,350 @@
...
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.ContentLength < 1000 {
+		contents, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
+		return fmt.Errorf("failed to download protoc: unexpected response '%s'", string(contents))
```
dejanzele, line 230: same for `fmt.Errorf`

magefile.go
severinson: There's already a double a?

magefile.go
severinson: Let's use `github.com/pkg/errors` since we use that everywhere else.

magefile.go
severinson: Done

magefile.go
severinson: Done

magefile.go
severinson: Done

.github/dependabot.yml
```diff
@@ -1,49 +1,49 @@
-# Basic dependabot.yml file with
```
severinson, line 1: Fixed line endings

.github/workflows/codeql-analysis.yml
```diff
@@ -1,77 +1,77 @@
-# For most projects, this workflow file will not need changing; you simply need
```
severinson, line 1: Fixed line endings

.gitpod.Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM golang:1.18.5-buster
```
severinson, line 1: Fixed line endings

.gitpod.yml
```diff
@@ -1,16 +1,16 @@
-image:
```
severinson, line 1: Fixed line endings

client/DotNet/ArmadaProject.Io.Client/ArmadaProject.Io.Client.csproj
```diff
@@ -1,28 +1,28 @@
-<Project Sdk="Microsoft.NET.Sdk">
```
severinson, line 1: Fixed line endings

severinson:
E2e tests failed due to image pull limits. This shouldn't have any impact on e2e tests.

dejanzele (approved):

PR #1856, opened 2022-12-01
testsuite improvements

severinson:
Improving the testsuite. In particular, enabling it to run several tests in parallel.
Disables getting job logs and benchmark reports until these can be made to work with the parallelised approach.


cmd/testsuite/cmd/test.go
```diff
@@ -164,22 +159,22 @@ func writeJUnitReport(junitPath string, testSuites *junit.Testsuites) error {
...
-	}
-
-	_, err = benchmarkFile.Write(data)
-	if err != nil {
-		return errors.WithStack(err)
-	}
-
-	return nil
-}
+// func writeBenchmarkReport(benchmarkPath string, report *eventbenchmark.GlobalBenchmarkReport) error {
```
suprjinx, line 162: do we need to keep this in the source?

cmd/testsuite/cmd/test.go
```diff
@@ -87,63 +79,66 @@ func testCmdRunE(app *testsuite.App) func(cmd *cobra.Command, args []string) err
...
 				return errors.WithMessage(err, "error writing junit report")
 			}
 		}
 
-		if benchmarkPath != "" {
-			if err := writeBenchmarkReport(benchmarkPath, app.GetBenchmarkReport()); err != nil {
-				return errors.WithMessage(err, "error writing benchmark report")
-			}
-		}
+		// if benchmarkPath != "" {
```
suprjinx, line 129: do we need to keep this?

d80tb7 (approved):

cmd/testsuite/cmd/test.go
severinson: We don't. But I'm hoping to re-enable it shortly.

PR #1882, opened 2022-12-08
Testsuite in CI

severinson:
Using an old version of goreleaser for now to avoid needing to upgrade `k8s.io` packages.

Setup + all e2e tests take between 6 and 7 mins using this approach.


.github/workflows/go-integration.yml
```diff
@@ -20,18 +19,46 @@ jobs:
...
-      - run: make tests-e2e-no-setup
-        env:
-          INTEGRATION_ENABLED: true
-      - run: make junit-report
-      - name: Upload junit report
+      - uses: actions/checkout@v3
+      # Needed for docker/setup-buildx-action; see
+      # https://github.com/moby/buildkit#github-actions-cache-experimental
+      - name: Expose GitHub Runtime
+        uses: crazy-max/ghaction-github-runtime@v2
```
d80tb7, line 38: guess we've done our due diligence on this image!

.goreleaser.yml
```diff
@@ -122,6 +133,9 @@ dockers:
       - "{{ .Env.DOCKER_REPO }}armada:{{ .Major }}.{{ .Minor }}.{{ .Patch }}"
       - "{{ .Env.DOCKER_REPO }}armada:latest"
     build_flag_templates:
+      - --builder={{ .Env.DOCKER_BUILDX_BUILDER }}
```
d80tb7, line 136: does this mean we will need to set these if using goreleaser locally?

go.mod
```diff
@@ -100,131 +100,255 @@ require (
...
+	cloud.google.com/go/iam v0.3.0 // indirect
+	cloud.google.com/go/kms v1.4.0 // indirect
+	cloud.google.com/go/storage v1.22.1 // indirect
+	code.gitea.io/sdk/gitea v0.15.1 // indirect
 	github.com/99designs/go-keychain v0.0.0-20191008050251-8e49817e8af4 // indirect
 	github.com/99designs/keyring v1.2.1 // indirect
+	github.com/AlekSi/pointer v1.2.0 // indirect
 	github.com/AthenZ/athenz v1.10.39 // indirect
+	github.com/Azure/azure-pipeline-go v0.2.3 // indirect
+	github.com/Azure/azure-sdk-for-go v59.3.0+incompatible // indirect
```
d80tb7, line 114: not something to fix in this PR, but we really should work out if we really need to be bringing in stuff like this.

magefile.go
```diff
@@ -15,15 +15,147 @@ import (
...
+	}
+	mg.Deps(BootstrapTools)
+	mg.Deps(DockerBundle, Kind)
+	err := sh.Run("docker-compose", "up", "-d", "redis", "postgres", "pulsar", "stan")
+	if err != nil {
+		return err
+	}
+
+	// TODO: Necessary to avoid connection error on Armada server startup.
+	time.Sleep(10 * time.Second)
```
d80tb7, line 46: are we able to check some condition, to make this less likely to fail in the case of slow startup?

.goreleaser.yml
severinson: No, using it locally will use the default runner.

magefile.go
severinson: I think we should fix Armada to retry for some time.

.github/workflows/go-integration.yml
severinson: Let me try removing this to see if it still works.

magefile.go
```diff
@@ -15,15 +15,147 @@ import (
...
+	bytes, err := yaml.Marshal(minimalConfig)
+	if err != nil {
+		return err
+	}
+	f, err := os.Create(".goreleaser-docker.yml")
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	if n, err := f.Write(bytes); err != nil {
```
d80tb7, line 150: I think you can save yourself a couple of lines here by importing `io/ioutil` and then ` ioutil.WriteFile`

arguably we should make a function in common that marshals a struct to yaml and writes to a file, but that's probbaly not worth doing right now.

d80tb7 (approved):

PR #1883, opened 2022-12-08
Add cancel by multiple job ids

carlocamurri:
Adds `jobIds` field to CancelJobs request payload, enabling cancellation by multiple job ids.

Creates an event sequence for each queue-job set pair, and submits them to pulsar.


internal/armada/server/submit_to_log.go
```diff
@@ -330,6 +337,101 @@ func (srv *PulsarSubmitServer) CancelJobs(ctx context.Context, req *api.JobCance
...
+			idQueueJobSetMap[q][jobSet] = append(idQueueJobSetMap[q][jobSet], job.Job.Id)
+		}
+	}
+
+	var cancelledIds []string
+	var eventSequences []*armadaevents.EventSequence
+	for q, jobSets := range idQueueJobSetMap {
+		userId, groups, err := srv.Authorize(ctx, q, permissions.CancelAnyJobs, queue.PermissionVerbCancel)
+		if err != nil {
+			log.WithError(err).Errorf("user cannot cancel jobs in queue %s", q)
```
JamesMurkin, line 370: Probably we should just return the error back to the user here and not do anything

PR #1886, opened 2022-12-08
Fixes mypy=0.991

Sharpz7:
Closes https://github.com/G-Research/armada/issues/1872

General fixes so that mypy works. This PR will not however put it back into the CI


client/python/armada_client/client.py
```diff
@@ -49,7 +49,7 @@ def get_job_events_stream(
         queue: str,
         job_set_id: str,
         from_message_id: Optional[str] = None,
```
kannon92, line 51: Could you just assign this to ""?  So you can remove the check below?

client/python/armada_client/client.py
```diff
@@ -188,6 +209,20 @@ def reprioritize_jobs(
...
+
+        if job_ids:
+            job_set_id = ""
+            queue = ""
+
+        elif job_set_id and queue:
+            job_ids = []
+
+        else:
+            raise ValueError("Either job_ids or job_set_id and queue must be provided.")
```
kannon92, line 224: It'd be nice to see a test for this.

client/python/armada_client/client.py
kannon92: You could just default initialize it to “” rather than optional?  Not sure if that’s normal in python

Sharpz7: Its because the function that we pass that to expects that type, so my hands are tied I think.

kannon92 (changes requested):
Wherever you added raises, I would like to see some kind of testing around this.

client/python/tests/unit/test_client.py
```diff
@@ -161,7 +161,15 @@ def test_get_queue_info():
...
     test_create_queue()
     test_submit_job()
-    tester.cancel_jobs(queue="test", job_set_id="job-set-1", job_id="job1")
+
+    # Test that the right combination of jobid or jobsetid and queue is used
+    # also check that the Value error is raised
+    with pytest.raises(ValueError):
+        tester.cancel_jobs(queue="test", job_id="job-1", job_set_id="job-set-1")
+
+    tester.cancel_jobs(job_id="job-1")
```
kannon92, line 170: Wouldn’t the first call cancel the job and the second wouldn’t be a no opp?

when writing tests, it’s usually good to follow S
RP. You don’t want side effects of one test influencing another.

kannon92 (commented):
Looks good. I’d like to see the tests separated and I’m not sure if you are checking the behavior of some of these tests. Cancelling should probably verify output to make sure it worked.


And same for repriortize

client/python/tests/unit/test_client.py
Sharpz7: Oh yeah. I didn't think about this. I will fix it.

Sharpz7: In hindsight, I do not think this applies for the unit tests - these are only testing that the gRPC server accepts the request as valid - there is no armada behavior checks and therefore no way to test if the cancelation is successful.

kannon92: You are calling a mock for these unit tests. Nothing is stopping you from implementing a test cancel function

kannon92: You also could check returns here.

Sharpz7: Yeah okay - I am just confused why we didn't do this earlier. I will redo the mock class to support full return statement checking across the board

kannon92 (approved):

Issue #1850, opened 2022-11-30
Add a flag to lookout ingester to disable conflation


ClifHouck:
**Is your feature request related to a problem? Please describe.**

Currently e2e tests for lookout ingester expect to see all applicable job state changes reflected in the db. This is not guaranteed when lookout ingester has conflation enabled. Tests are stable at the moment, but it would be nice to have the ability to disable conflation altogether for some tests.

**Describe the solution you'd like**

Add flag to lookout ingester to disable conflation. 

**Describe alternatives you've considered**

**Additional context**

PR #1889, opened 2022-12-08
Add debug configuration to lookout ingester

ClifHouck:
Adds two debug configuration flags:
1. `debug.disableConflateDBUpdates`
2. `debug.forceScalarDBUpdates`

These flags help ensure that e2e tests can "see" all changes happening to the database.

Also removes `LookoutIngesterConfiguration.PulsarReceiveTimeout` and `LookoutIngesterConfiguration.PulsarBackoffTime` since they were unused.

Fixed all lookout ingester configurations to use
`LookoutIngesterConfiguration.PulsarConfig.{ReceiveTimeout,BackoffTime}` since those values are actually used by the code in lookout ingester.

Fixes #


ClifHouck:
Closes #1850

internal/lookout/configuration/types.go
```diff
@@ -60,10 +73,6 @@ type LookoutIngesterConfiguration struct {
 	BatchSize int
 	// Maximum time since the last batch before a batch will be inserted into the database
 	BatchDuration time.Duration
-	// Time for which the pulsar consumer will wait for a new message before retrying
```
ClifHouck, line 63: These two config options were unused by lookoutingester code. The equivalent options in LookoutIngesterConfiguration.Pulsar were used instead.

e2e/setup/lookout-ingester-config.yaml
```diff
@@ -15,12 +15,16 @@ pulsar:
...
 subscriptionName: "lookout-ingester"
 batchSize: 10000
-batchDuration: 500ms
-pulsarReceiveTimeout: 5s
-pulsarBackoffTime: 1s
+batchDuration: 500ms  
 minJobSpecCompressionSize: 1024
 userAnnotationPrefix: "armadaproject.io/"
+
+debug:
```
ClifHouck, line 28: So these config options are strictly necessary to get stable e2e tests for lookoutingester. They *do*, however, offer a stronger guarantee that the tests will observe every update to the jobs table.

internal/lookout/configuration/types.go
```diff
@@ -45,13 +45,26 @@ type LookoutConfiguration struct {
 	DisableEventProcessing bool
 }
 
+type LookoutIngesterDebugConfig struct {
+	// Disables DB update conflation. If conflation is disabled then update
+	// instructions	will not conflatined/coalesed. Meaning update instructions
```
suprjinx, line 50: typos on conflation / coalesced

suprjinx (dismissed):

e2e/lookout_ingester_test/lookout_ingester_test.go
```diff
@@ -456,3 +472,27 @@ func generateNotificationConnectionEventHandler(channelName string) func(pq.List
 		}
 	}
 }
+
+var (
+	createShadowTableSql = "CREATE TABLE %s_shadow (LIKE %s INCLUDING ALL, update_num SERIAL);"
+	alterShadowTableSql  = "ALTER TABLE %s_shadow DROP CONSTRAINT %s_shadow_pkey, ADD PRIMARY KEY (job_id, update_num);"
+)
+
+func createShadowTables(db *sql.DB) {
```
d80tb7, line 481: you might be able to remove some of this boilerplate with https://www.postgresql.org/docs/14/sql-notify.html. I've actually never used this so I'm not 100% sure whether it will actually make your life easier though.

e2e/setup/lookout-ingester-config.yaml
d80tb7: Do you need `forceScalarDbUpdates` because if you turn conflation off then we produce incorrect output (I could see how this could happen if you had multiple updates for the same job)?

If this is the case then I'd just have `disableConflateDBUpdates` as an option and have that internally do the forcing to scalar. Thais is because just setting `disableConflateDBUpdates`  alone would be a bad config.

internal/lookout/configuration/types.go
d80tb7: good spot- yes I moved them when I went to the common framework code but forgot to remove the config options

e2e/lookout_ingester_test/lookout_ingester_test.go
ClifHouck: I think I prefer this way instead of using notify. This has the added benefit of creating a durable log of what changed during test execution. Notify is more ephemeral.

e2e/setup/lookout-ingester-config.yaml
ClifHouck: You hit the nail on the head. Yes, I can simplify this to get rid of `forceScalarDbUpdates`.

suprjinx (approved):

PR #1916, opened 2022-12-12
adding errcheck to linter

### LINKED ISSUE #1911: Consider integrating the Errcheck linter
A little while back I found the issue that was fixed by https://github.com/G-Research/armada/pull/1735, and just now (by dint of complaining about Golang error handling on Hacker News) I learned of the existence of https://github.com/kisielk/errcheck, a linter to point out when you've forgotten to handle an error. I just ran errcheck on the Armada repo as of the commit before #1735, and it pointed out precisely that bug (among many other places where errors are silently dropped).

(Specific instructions for me were extremely simple, because I use Nix, but I presume it's also easy using standard Go tooling. For me, `nix-shell -p errcheck go` and then `cd armada && git checkout --hard 6239d744d8676aa798ddf5784dd08ace802974bc^ && errcheck ./...` was enough.)

I have no particular skin in this game, but wanted to suggest incorporating errcheck into the pipelines.

(I was also pointed to https://github.com/golangci/golangci-lint as an aggregation of lots of important linters.)

kannon92:
Fixes #1911


pkg/client/auth/oidc/device.go
```diff
@@ -48,7 +48,10 @@ func AuthenticateDevice(config DeviceDetails) (*TokenCredentials, error) {
 		if err != nil {
 			return nil, err
 		}
-		defer cmd.Process.Kill()
+		defer func() {
+			err := cmd.Process.Kill()
+			fmt.Printf("Error killing your process: %s", err)
```
kannon92, line 53: Bug.

pkg/client/auth/oidc/pkce.go
```diff
@@ -89,10 +89,20 @@ func AuthenticatePkce(config PKCEDetails) (*TokenCredentials, error) {
...
+		err := server.Serve(listener)
+		if err != nil {
+			panic(err)
+		}
+	}()
 
 	cmd, err := openBrowser("http://" + localUrl)
-	defer cmd.Process.Kill()
+	defer func() {
+		err := cmd.Process.Kill()
```
kannon92, line 101: What should we do if we can't kill the browser that does oidc sign in?

internal/scheduleringester/schedulerdb.go
```diff
@@ -37,7 +37,10 @@ func (s *SchedulerDb) Store(ctx context.Context, instructions *DbOperationsWithM
 			shouldRetry := armadaerrors.IsNetworkError(err) || armadaerrors.IsRetryablePostgresError(err)
 			return shouldRetry, err
 		}, s.initialBackOff, s.maxBackOff)
-		multierror.Append(result, err)
+		multiAppendErr := multierror.Append(result, err)
+		if multiAppendErr != nil {
```
kannon92, line 41: Not sure what we should do here.  If we can't append a multierror, not sure if we should panic or just return that error.  Decided to return an error here.

internal/armada/server/submit_from_log.go
```diff
@@ -47,7 +47,10 @@ func (srv *SubmitFromLog) Run(ctx context.Context) error {
 		if err := recover(); err != nil {
 			log.WithField("error", err).Error("unexpected panic; restarting")
 			time.Sleep(time.Second)
-			go srv.Run(ctx)
+			go func() {
+				err := srv.Run(ctx)
+				log.WithField("error", err).Error("unexpected panic restarting again")
```
kannon92, line 52: Bug.  And this could become a recursive call if run keeps failing.

suprjinx: what's the bug? it doesn't look potentially recursive, since the go func is not calling panic()

suprjinx: srv.Run() returning an error is not a panic

cmd/scheduler/main.go
```diff
@@ -43,7 +43,8 @@ func main() {
 	if viper.GetBool(MigrateDatabase) {
 		migrateDatabase(&config)
 	} else {
-		scheduler.Run(&config)
+		err := scheduler.Run(&config)
+		panic(errors.WithMessage(err, "Failed to run scheduler"))
```
suprjinx, line 47: panic() will change behavior -- I think prefer "log it" approach used elsewhere for this lint fixup PR

pkg/client/auth/oidc/pkce.go
```diff
@@ -89,10 +89,20 @@ func AuthenticatePkce(config PKCEDetails) (*TokenCredentials, error) {
 
 	server := &http.Server{}
 
-	go server.Serve(listener)
+	go func() {
+		err := server.Serve(listener)
+		if err != nil {
+			panic(err)
```
suprjinx, line 95: i don't think we should introduce panic

pkg/client/auth/oidc/pkce.go
suprjinx: we didn't do anything before, so log and move on seems fine

internal/armada/server/submit_from_log.go
kannon92: The bug is on my end but its fixed.  I log an error no matter if err is nil or not.  I did that in a few places so bug is mostly a note to me to fix that.  

The recursive call is that this function is already srv.Run so we are recovering from a panic and then calling the same function again.  If we keep failing then it could be a problem.

pkg/client/auth/oidc/pkce.go
kannon92: I guess I thought that if we can't start the http server that would be bad but yea, I can add a log instead of a panic.  I think that makes sense

internal/armada/server/eventsprinter.go
```diff
@@ -43,7 +43,12 @@ func (srv *EventsPrinter) Run(ctx context.Context) error {
 		if err := recover(); err != nil {
 			log.WithField("error", err).Error("unexpected panic; restarting")
 			time.Sleep(time.Second)
-			go srv.Run(ctx)
+			go func() {
+				err := srv.Run(ctx)
+				if err != nil {
+					log.WithField("error", err).Error("unexpected panic; restarting")
```
kannon92, line 49: Fix the comment for this.  Not really restarting here but we get here if we already tried to restart.

internal/executor/service/job_lease.go
```diff
@@ -161,7 +161,10 @@ func (jobLeaseService *JobLeaseService) requestJobLeases(leaseRequest *api.Strea
 
 	// Get received jobs on the channel and send back acks.
 	g.Go(func() error {
-		defer stream.CloseSend()
+		defer func() {
+			err := stream.CloseSend()
```
kannon92, line 165: Bug.

e2e/lookout_ingester_test/lookout_ingester_test.go
```diff
@@ -442,7 +442,10 @@ func createTestTriggers(db *sql.DB) {
 func dropTestTriggers(db *sql.DB) {
 	for _, trigger := range testTriggers {
 		// We don't care about the result. Just drop them.
-		db.Exec(fmt.Sprintf("DROP TRIGGER %s on %s", trigger.Name(), trigger.Table))
+		_, err := db.Exec(fmt.Sprintf("DROP TRIGGER %s on %s", trigger.Name(), trigger.Table))
```
kannon92, line 445: @ClifHouck I tried to panic as you do in these tests but it failed the test.  Is there anything bad happening that this shows?  Should we be getting errors during dropping triggers in the e2e tests?

ClifHouck: I realized the 'right' way to deal with this is to add a `IF EXISTS` clause to these statements. I can fix this in another PR or push a commit to your branch (although I don't know if I have permission to do this).

internal/jobservice/application.go
```diff
@@ -73,7 +78,8 @@ func (a *App) StartUp(ctx context.Context, config *configuration.JobServiceConfi
 			for _, value := range sqlJobRepo.GetSubscribedJobSets() {
 				log.Infof("Subscribed job sets : %s", value)
 				if sqlJobRepo.CheckToUnSubscribe(value.Queue, value.JobSet, config.SubscribeJobSetTime) {
-					sqlJobRepo.CleanupJobSetAndJobs(value.Queue, value.JobSet)
```
kannon92, line 76: Bug

internal/armada/server/submit_from_log.go
suprjinx: maybe we shouldn't recover from panic with restart? or some sort backoff on restarts by counting (?)

pkg/client/auth/oidc/pkce.go
kannon92: Should these be log.Warn or log.Error rather than Printf?

pkg/client/auth/oidc/device.go
```diff
@@ -48,7 +48,12 @@ func AuthenticateDevice(config DeviceDetails) (*TokenCredentials, error) {
 		if err != nil {
 			return nil, err
 		}
-		defer cmd.Process.Kill()
+		defer func() {
+			err := cmd.Process.Kill()
+			if err != nil {
+				fmt.Printf("Error killing your process: %s", err)
```
kannon92, line 54: Same here.  Should probably be logging these as warnings and not Printf?

internal/testsuite/app.go
```diff
@@ -188,7 +188,10 @@ func (a *App) RunTests(ctx context.Context, testSpecs []*api.TestSpec) (*TestSui
 			eventLogger:          eventLogger,
 		}
 		go func() {
-			testRunner.Run(ctx)
+			err := testRunner.Run(ctx)
+			if err != nil {
+				fmt.Fprintf(a.Out, "Error with running %s", err)
```
kannon92, line 193: In this case, I think this would be correct.  If we get an error, we can output it to the test file.

internal/scheduler/scheduler.go
```diff
@@ -340,6 +340,9 @@ func Run(config *Configuration) error {
 	defer producer.Close()
 
 	scheduler := NewScheduler(producer, db)
-	scheduler.Run(app.CreateContextWithShutdown())
+	runErr := scheduler.Run(app.CreateContextWithShutdown())
```
kannon92, line 343: @severinson would this be sufficient for handling errors from scheduler?  Not sure what to do in the case that we can't run the scheduler.

e2e/lookout_ingester_test/lookout_ingester_test.go
ClifHouck: Addressed as part of https://github.com/G-Research/armada/pull/1889

internal/armada/server/submit_from_log.go
```diff
@@ -47,7 +47,12 @@ func (srv *SubmitFromLog) Run(ctx context.Context) error {
 		if err := recover(); err != nil {
 			log.WithField("error", err).Error("unexpected panic; restarting")
 			time.Sleep(time.Second)
-			go srv.Run(ctx)
+			go func() {
+				err := srv.Run(ctx)
+				if err != nil {
+					log.WithField("error", err).Error("unexpected panic after restarting")
```
severinson, line 53: Same as the above comment.

internal/common/ingest/batch_test.go
```diff
@@ -79,7 +79,8 @@ func TestBatch_Time(t *testing.T) {
 
 	inputChan <- 1
 	inputChan <- 2
-	waitForBufferLength(ctx, batcher, 2)
+	err := waitForBufferLength(ctx, batcher, 2)
+	assert.Nil(t, err)
```
severinson, line 83:
```suggestion
	err := waitForBufferLength(ctx, batcher, 2)
	if !assert.NoError(t, err) {
	    return
	}
```

severinson: Same for all other instances. Use this trick for non-recoverable errors.

internal/common/ingest/batch_test.go
```diff
@@ -103,12 +104,15 @@ func TestBatch_Time_WithIntialQuiet(t *testing.T) {
...
```
severinson, line 108: Use `assert.NoError` + `return`.

internal/common/ingest/batch_test.go
```diff
@@ -103,12 +104,15 @@ func TestBatch_Time_WithIntialQuiet(t *testing.T) {
...
 	testClock.Step(5 * time.Second)
 	waitForExpectedEvents(ctx, result, 1)
 	inputChan <- 3
 	inputChan <- 4
-	waitForBufferLength(ctx, batcher, 2)
+	err = waitForBufferLength(ctx, batcher, 2)
+	assert.Nil(t, err)
```
severinson, line 114: Use `assert.NoError` + `return`.

internal/common/logging/interceptors_test.go
```diff
@@ -37,7 +38,8 @@ func TestUnaryServerInterceptor(t *testing.T) {
 	}
 
 	f := UnaryServerInterceptor()
-	f(ctx, nil, nil, handler)
+	_, err := f(ctx, nil, nil, handler)
+	assert.Nil(t, err)
```
severinson, line 42: Use `assert.NoError` + `return`.

internal/common/logging/interceptors_test.go
```diff
@@ -66,5 +68,6 @@ func TestStreamServerInterceptor(t *testing.T) {
 	}
 
 	f := StreamServerInterceptor()
-	f(nil, stream, nil, handler)
+	err := f(nil, stream, nil, handler)
+	assert.Nil(t, err)
```
severinson, line 72: Use `assert.NoError` + `return`.

internal/common/pgkeyvalue/pgkeyvalue_test.go
```diff
@@ -217,7 +217,10 @@ func TestCleanup(t *testing.T) {
 		// Then try adding baz twice more to make sure it gets cleaned up both times.
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
-		go store.PeriodicCleanup(ctx, time.Microsecond, time.Microsecond)
+		go func() {
+			err := store.PeriodicCleanup(ctx, time.Microsecond, time.Microsecond)
+			assert.Nil(t, err)
```
severinson, line 222: Use `assert.NoError` + `return`.

internal/common/pgkeyvalue/pgkeyvalue.go
```diff
@@ -87,7 +90,10 @@ func (c *PGKeyValueStore) LoadOrStoreBatch(ctx context.Context, batch []*KeyValu
 	// If the table doesn't exist, create it and try again.
 	var pgErr *pgconn.PgError
 	if errors.As(err, &pgErr) && pgErr.Code == pgerrcode.UndefinedTable { // Relation doesn't exist; create it.
-		c.createTable(ctx)
+		err = c.createTable(ctx)
+		if err != nil {
```
severinson, line 94: I'd use the shorter form `if err := f(); err != nil { ...`.

internal/common/pgkeyvalue/pgkeyvalue_test.go
```diff
@@ -249,7 +252,8 @@ func BenchmarkStore(b *testing.B) {
 			b.FailNow()
 		}
 		for i := 0; i < b.N; i++ {
-			store.AddKey(context.Background(), "foo")
+			_, err := store.AddKey(context.Background(), "foo")
+			assert.Nil(b, err)
```
severinson, line 256: Use `assert.NoError` + `return`.

internal/common/requestid/interceptors_test.go
```diff
@@ -61,11 +62,13 @@ func TestUnaryServerInterceptor(t *testing.T) {
 
 	replace := false
 	f := UnaryServerInterceptor(replace)
-	f(ctx, nil, nil, handler)
+	_, err := f(ctx, nil, nil, handler)
+	assert.Nil(t, err)
```
severinson, line 66: Use `assert.NoError` + `return`.

internal/common/requestid/interceptors_test.go
```diff
@@ -61,11 +62,13 @@ func TestUnaryServerInterceptor(t *testing.T) {
...
 
 	replace = true
 	f = UnaryServerInterceptor(replace)
-	f(ctx, nil, nil, handler)
+	_, err = f(ctx, nil, nil, handler)
+	assert.Nil(t, err)
```
severinson, line 71: Use `assert.NoError` + `return`.

internal/common/requestid/interceptors_test.go
```diff
@@ -97,11 +100,13 @@ func TestUnaryServerInterceptorWithExisting(t *testing.T) {
 
 	replace = false
 	f := UnaryServerInterceptor(replace)
-	f(ctx, nil, nil, handler)
+	_, err := f(ctx, nil, nil, handler)
+	assert.Nil(t, err)
```
severinson, line 104: Use `assert.NoError` + `return`.

internal/common/requestid/interceptors_test.go
```diff
@@ -97,11 +100,13 @@ func TestUnaryServerInterceptorWithExisting(t *testing.T) {
...
 
 	replace = true
 	f = UnaryServerInterceptor(replace)
-	f(ctx, nil, nil, handler)
+	_, err = f(ctx, nil, nil, handler)
+	assert.Nil(t, err)
```
severinson, line 109: Use `assert.NoError` + `return`.

internal/common/requestid/interceptors_test.go
```diff
@@ -123,11 +128,13 @@ func TestStreamServerInterceptor(t *testing.T) {
 
 	replace := false
 	f := StreamServerInterceptor(replace)
-	f(nil, stream, nil, handler)
+	err := f(nil, stream, nil, handler)
+	assert.Nil(t, err)
```
severinson, line 132: Use `assert.NoError` + `return`.

internal/common/requestid/interceptors_test.go
```diff
@@ -123,11 +128,13 @@ func TestStreamServerInterceptor(t *testing.T) {
...
 
 	replace = true
 	f = StreamServerInterceptor(replace)
-	f(nil, stream, nil, handler)
+	err = f(nil, stream, nil, handler)
+	assert.Nil(t, err)
```
severinson, line 137: Use `assert.NoError` + `return`.

internal/executor/utilisation/job_utilisation_reporter_test.go
```diff
@@ -35,7 +35,8 @@ func TestUtilisationEventReporter_ReportUtilisationEvents(t *testing.T) {
 	fakeUtilisationService := &fakePodUtilisationService{data: &testPodResources}
 
 	reporter := NewUtilisationEventReporter(clusterContext, fakeUtilisationService, fakeEventReporter, reportingPeriod)
-	submitPod(clusterContext)
+	_, err := submitPod(clusterContext)
+	assert.Nil(t, err)
```
severinson, line 39: Use `assert.NoError` + `return`.

internal/jobservice/application.go
```diff
@@ -73,7 +78,10 @@ func (a *App) StartUp(ctx context.Context, config *configuration.JobServiceConfi
...
+					_, err := sqlJobRepo.CleanupJobSetAndJobs(value.Queue, value.JobSet)
+					if err != nil {
+						log.Warnf("Error detected from CleanupJobSetAndJobs: %v", err)
```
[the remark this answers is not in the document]
severinson: Attach a stack trace (see the `logging.WithStackTrace` method) instead of adding the method name to the log line.

internal/jobservice/repository/sql_job_service_test.go
```diff
@@ -186,7 +186,8 @@ func TestCheckToUnSubscribeWithoutSubscribing(t *testing.T) {
 func TestUpdateJobSetTime(t *testing.T) {
 	WithSqlServiceRepo(func(r *SQLJobService) {
 		r.SubscribeJobSet("test", "job-set-1")
-		r.UpdateJobSetTime("test", "job-set-1")
+		err := r.UpdateJobSetTime("test", "job-set-1")
+		assert.Nil(t, err)
```
severinson, line 190: Use `assert.NoError` + `return`.

internal/jobservice/repository/sql_job_service_test.go
```diff
@@ -263,12 +264,14 @@ func TestDeleteJobsBeforePersistingRaceError(t *testing.T) {
...
-
+		var actualNumberOfJobs int64 = 1
 		jobStatus1 := NewJobStatus("test-race", "job-set-race", "job-race", *responseSuccess)
 		err := r.UpdateJobServiceDb(jobStatus1)
 		assert.Nil(t, err)
 		r.SubscribeJobSet("test-race", "job-set-race")
-		r.CleanupJobSetAndJobs("test-race", "job-set-race")
+		numberOfJobs, deleteErr := r.CleanupJobSetAndJobs("test-race", "job-set-race")
+		assert.Equal(t, numberOfJobs, actualNumberOfJobs)
+		assert.Nil(t, deleteErr)
```
severinson, line 274: Use `assert.NoError` + `return`.

internal/jobservice/repository/sql_job_service_test.go
```diff
@@ -263,12 +264,14 @@ func TestDeleteJobsBeforePersistingRaceError(t *testing.T) {
...
 		noExist := &jobservice.JobServiceResponse{State: jobservice.JobServiceResponse_JOB_ID_NOT_FOUND}
-
+		var actualNumberOfJobs int64 = 1
 		jobStatus1 := NewJobStatus("test-race", "job-set-race", "job-race", *responseSuccess)
 		err := r.UpdateJobServiceDb(jobStatus1)
 		assert.Nil(t, err)
 		r.SubscribeJobSet("test-race", "job-set-race")
-		r.CleanupJobSetAndJobs("test-race", "job-set-race")
+		numberOfJobs, deleteErr := r.CleanupJobSetAndJobs("test-race", "job-set-race")
+		assert.Equal(t, numberOfJobs, actualNumberOfJobs)
```
severinson, line 273:
```suggestion
		assert.Equal(t, expectedNumberOfJobs, actualNumberOfJobs)
```

internal/lookoutingester/lookoutdb/insertion_test.go
```diff
@@ -591,8 +592,8 @@ func TestUpdate(t *testing.T) {
 	err := testutil.WithDatabasePgx(func(db *pgxpool.Pool) error {
 		ldb := &LookoutDb{db: db, metrics: metrics.Get()}
 		// Do the update
-		ldb.Store(context.Background(), defaultInstructionSet())
-
+		storeErr := ldb.Store(context.Background(), defaultInstructionSet())
+		assert.NoError(t, storeErr)
```
severinson, line 596: Also return.

internal/scheduler/scheduler.go
severinson: Just return the error here; don't wrap it. By capturing a stack trace at the point where the error occurs, it's clear what went wrong so there's no need to add additional messages.

Further, you should use `WithMessage` instead of `Wrap` to attach a message to an error generated from within the Armada code, since `Wrap` also attaches a stack trace. We should only attach a stack trace where the error is generated or, if originating in a library outside the Armada code base, at the point at which the error is first encountered within the Armada code.

It's the responsibility of the caller of this function to handle this error (probably by shutting down).

internal/scheduleringester/schedulerdb.go
```diff
@@ -37,7 +37,10 @@ func (s *SchedulerDb) Store(ctx context.Context, instructions *DbOperationsWithM
 			shouldRetry := armadaerrors.IsNetworkError(err) || armadaerrors.IsRetryablePostgresError(err)
 			return shouldRetry, err
 		}, s.initialBackOff, s.maxBackOff)
-		multierror.Append(result, err)
+		multiAppendErr := multierror.Append(result, err)
```
severinson, line 40: This should be `multierror.Append(result, err)`. Don't check if the error isn't `nil` here. By doing so the function would return prematurely at the first error. We want to collect all errors before returning.
`multierror.Append` works like the regular `append` for slices in that it adds information to an existing error. But we need to update the variable in-place (like with the regular `append`), which we weren't doing.

internal/testsuite/app.go
severinson, line 193:
```suggestion
				fmt.Fprintf(a.Out, "error in test case %s: %s", <name of test case>, err)
```

pkg/client/auth/oidc/device.go
severinson: Agreed. This should be a log.
And please start error messages with lower-case letters.
Since they get embedded in sequences of errors like `error 1: error 2: error 3`.

pkg/client/auth/oidc/pkce.go
severinson: `logging.WithStacktrace.Error`

internal/scheduler/scheduler.go
dave-gantenbein: @severinson - is this something that should be standard throughout the code base? If so, it would be good to document that somewhere publicly.

e2e/armadactl_test/armadactl_test.go
```diff
@@ -221,8 +221,14 @@ jobs:
 		t.Fatalf("error creating jobfile: %s", err)
 	}
 	jobPath := jobFile.Name()
-	jobFile.Write(jobData)
-	jobFile.Sync()
+	_, err = jobFile.Write(jobData)
+	if err != nil {
+		t.Fatalf("error writing jobFile: %s", err)
```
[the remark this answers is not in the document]
kannon92: So I followed this pattern because this entire file uses this for the test.  Should I change all instances of this in this file?  Or change just what I've changed?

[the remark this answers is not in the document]
kannon92: So it looks like we could use require:

```
require.NoError(t, err) which will force a failure if the test fails.

File: pkg/client/command_line.go [hunk 37]
```diff
@@ -110,7 +114,11 @@ func LoadCommandlineArgsFromConfigFile(cfgFile string) error {
 // e.g., by calling LoadCommandlineArgsFromConfigFile.
 func ExtractCommandlineArmadaApiConnectionDetails() *ApiConnectionDetails {
 	apiConnectionDetails := &ApiConnectionDetails{}
-	viper.Unmarshal(apiConnectionDetails)
+	err := viper.Unmarshal(apiConnectionDetails)
+	if err != nil {
+		log.Errorf("Could not unmarshal apiConnectionDetails: %s", err)
```
kannon92, line 119:
```suggestion
		log.Errorf("could not unmarshal apiConnectionDetails: %s", err)
```

cmd/armadactl/cmd/commands.go
```diff
@@ -21,7 +23,10 @@ func createCmd(a *armadactl.App) *cobra.Command {
 		},
 	}
 	cmd.Flags().StringP("file", "f", "", "specify file for resource creation.")
-	cmd.MarkFlagRequired("file")
+	if err := cmd.MarkFlagRequired("file"); err != nil {
+		log.Errorf("error detected with MarkFlagRequired for file %v", err)
+		return nil
```
suprjinx, line 28: should this just proceed as before instead of returning nil?

e2e/armadactl_test/armadactl_test.go
```diff
@@ -234,29 +213,21 @@ jobs:
 		UserOwners:     owners,
 		GroupOwners:    groups,
 	})
-	if err != nil {
-		t.Fatalf("failed to instantiate queue: %s", err)
-	}
+	require.NoError(t, err, "failed to instantiate queue: %s", err)
```
suprjinx, line 216: require makes this a lot cleaner -- nice

internal/common/auth/authorization/kerberos_test.go
```diff
@@ -57,12 +57,20 @@ func generateTempKeytab() (string, error) {
...
+		return "", err
+	}
 	// Write to temp file and close it.
-	kt.Write(f)
-	defer f.Close()
+	_, err = kt.Write(f)
+	if err != nil {
+		return "", err
+	}
+	err = f.Close()
```
suprjinx, line 70: should Close() move back to the defer function so it's always attempted?

internal/common/ingest/batch_test.go
suprjinx: or one-liner require?

internal/jobservice/application.go
```diff
@@ -73,23 +83,30 @@ func (a *App) StartUp(ctx context.Context, config *configuration.JobServiceConfi
...
 			}
 		}
 		return nil
 	})
 	g.Go(func() error {
-		defer log.Infof("Stopping server.")
+		defer log.Infof("stopping server.")
 
-		log.Info("JobService service listening on ", config.GrpcPort)
+		log.Info("jobService service listening on ", config.GrpcPort)
```
suprjinx, line 98: jobservice all lowercase?

internal/scheduler/scheduler.go
severinson: I think the general principle is to capture the current state of the program at the point at which an error occurs using a stack trace. Exception-based languages (e.g., Python and Java) do that for you, but in go it has to be done manually. I think the approach we take is fairly standard and that it's sufficient to leave it implicit from the code.

internal/testsuite/app.go
```diff
@@ -188,7 +188,10 @@ func (a *App) RunTests(ctx context.Context, testSpecs []*api.TestSpec) (*TestSui
 			eventLogger:          eventLogger,
 		}
 		go func() {
-			testRunner.Run(ctx)
+			err := testRunner.Run(ctx)
+			if err != nil {
+				fmt.Fprintf(a.Out, "error in test case %s: %s", rv.TestCaseReports[i].TestSpec.Name, err)
```
severinson, line 193: This will result in a nil reference panic.
You need to get the test case name from the `testSpecs` passed into the `RunTests` function.

internal/jobservice/application.go
kannon92: It would be two words so I think JobService or jobService would be correct?

cmd/armadactl/cmd/queue_test.go
```diff
@@ -233,7 +236,10 @@ func TestUpdate(t *testing.T) {
 
 			// Set CLI flags; falls back to default values if not set
 			for _, flag := range test.Flags {
-				cmd.Flags().Set(flag.name, flag.value)
+				err := cmd.Flags().Set(flag.name, flag.value)
```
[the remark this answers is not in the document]
kannon92: Yep, require.NoError.  Made modifcations in this test to use it.

e2e/lookout_ingester_test/lookout_ingester_test.go
```diff
@@ -442,7 +442,10 @@ func createTestTriggers(db *sql.DB) {
...
+		if err != nil {
+			fmt.Println("Error with dropping test triggers", err)
```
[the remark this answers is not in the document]
kannon92: I think @ClifHouck is working on this so I'll leave as is.  We were actually hitting these errors in the test.

internal/testsuite/app.go
kannon92: Good call.

internal/common/auth/authorization/kerberos_test.go
kannon92: Done

severinson (approved):

internal/jobservice/application.go
```diff
@@ -9,14 +9,17 @@ import (
 	"path/filepath"
 	"time"
 
-	log "github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 
+	log "github.com/sirupsen/logrus"
```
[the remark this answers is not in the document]
kannon92: Good catch

severinson (approved):

PR #1917, opened 2022-12-13
Add Lookout V2 reprioritisation dialog

isaac-jordan:
#### What type of PR is this?

This is a UI PR for Lookout V2.

- Adds reprioritisation dialog
  - This is similar to Cancel dialog, but different enough that I don't think they should share the same component
  - Reprioritisation API calls are batched into 5000 jobs per call, all made in parallel
- Extracts out some of Cancel dialog code to utils to be re-used
- Adds tests for Reprioritisation dialog
- Fixes a bug where you couldn't use the "select all" toggle in the table header 

This PR is part of https://github.com/G-Research/armada/issues/1869

**Loads up job information**
image

**User can start once they enter a priority**
image

**Shows a popover, request status, and retrieves updated job state on success**
image


internal/lookout/ui/src/components/lookoutV2/JobStatusTable.tsx
```diff
@@ -0,0 +1,55 @@
...
+  jobsToRender: Job[]
+  totalJobCount: number
+  additionalColumnsToDisplay: {
+    displayName: string
+    formatter: (job: Job) => string
+  }[]
+  showStatus: boolean
+  jobStatus: Record<string, string>
+}
+export const JobStatusTable = memo(
```
kannon92, line 16: Just curious, why no tests for this component?

isaac-jordan: @kannon92 It's simple enough that I didn't see there being particularly strong added value in testing it, and it's indirectly tested as part of the two dialog's tests. Would you disagree?

kannon92: Nope that sounds good! Just wanted your thoughts on why

carlocamurri (approved):

PR #1921, opened 2022-12-13
add matrix to airflow and client release

kannon92:
Fix client/airflow release action

### DIFF
--- a/.github/workflows/airflow-operator-release-to-pypi.yml
+++ b/.github/workflows/airflow-operator-release-to-pypi.yml
@@ -9,14 +9,22 @@ jobs:
     strategy:
       matrix:
         go: [ '1.18' ]
+        python: [ '3.8', '3.9', '3.10' ]
+        include:
+          - tox-env: 'py38'
+          - tox-env: 'py39'
+            python: '3.9'
+          - tox-env: 'py310'
+            python: '3.10'
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/airflow-operator-release-to-pypi.yml
```diff
@@ -9,14 +9,22 @@ jobs:
     strategy:
       matrix:
         go: [ '1.18' ]
+        python: [ '3.8', '3.9', '3.10' ]
+        include:
+          - tox-env: 'py38'
```
ClifHouck, line 14: What allows you to omit `python: '3.8'` here, while needing it for the other `tox-env` entries?

suprjinx (approved):

.github/workflows/airflow-operator-release-to-pypi.yml
kannon92: This came from another PR that was just merged.  I'm mostly replicating it in the release action as our unit test build now has this.  

The reason we have these two is that tox requires py38 and the setup-action for python requires a version like '3.8'..  So..

kannon92: Oh I see your question.  I think maybe we include tox py38 as the default and the others are extra?

PR #1965, opened 2022-12-21
Reconcile config

severinson:
We have lots on config files; let's reconcile.


config/armada/config.yaml
```diff
@@ -17,60 +17,99 @@ grpc:
     permitWithoutStream: false
 redis:
   addrs:
-    - "localhost:6379"
+    - "redis:6379"
```
d80tb7, line 20: I don't understand this- how does this work if you're running armada in an ide?  Or is this config no longer used for that?

config/armada/config.yaml
```diff
@@ -17,60 +17,99 @@ grpc:
...
-    # if enabled, Armada will accept submitted jobs which have a priorityClassName field
-    # otherwise it will refuse the jobs on submit
-    enabled: false
-    # priorityClasses field is a map of priority class names to priority values
-    # defined priority classes should exist with the same priority in all executor clusters
-    priorityClasses: {}
-    # if preemption is enabled, Armada server will set the priorityClassName in the job's podSpec
-    # with the value of defaultPriorityClass, if the podSpec does not already reference a priority class
+    enabled: true
+    priorityClasses:
```
d80tb7, line 40: assuming that this file remains the default config we'll need to ensure we create these in the helm chart

config/armada/config.yaml
```diff
@@ -17,60 +17,99 @@ grpc:
...
+      armada-default:
+        priority: 1000
+        maximalResourceFractionPerQueue:
+          memory: 0.99
+          cpu: 0.99
+      "": # Treat missing PC the same as armada-default.
```
d80tb7, line 46: I really don't like this!  Can we solve this more elegantly in the code?

config/armada/config.yaml
```diff
@@ -17,60 +17,99 @@ grpc:
...
+    # Default to empty PC to not require PCs be set up in worker clusters.
     defaultPriorityClass: ""
-  useProbabilisticSchedulingForAllResources: true
-  queueLeaseBatchSize: 200
-  maximumLeasePayloadSizeBytes: 7340032 # 1024 * 1024 * 7
+  queueLeaseBatchSize: 1000
   minimumResourceToSchedule:
-    memory: 100000000 # 100Mb
-    cpu: 0.25
+    memory: 1000000
```
d80tb7, line 60: let's keep the comment to say what this is as it's not obvious what the units are

config/armada/config.yaml
severinson: I doesn't, the default config is for running in containers.
To run in an IDE, this value needs to be overridden, e..g, as shown here
https://github.com/G-Research/armada/blob/master/docs/development_guide.md#running-the-armada-server-and-executor-in-visual-studio-code

Alternatively, we could move the overrides from env variables into a separate config file with only overrides.

config/armada/config.yaml
severinson: I think this was only necessary while having old jobs in the system and that it can now be removed.

config/armada/config.yaml
severinson: I added the PCs to the executor helm. Could you please verify that it's correct?

config/armada/config.yaml
severinson: I've removed it.

config/armada/config.yaml
severinson: Fixed

d80tb7 (approved):

PR #1971, opened 2022-12-30
Remove Nats

d80tb7:
Completely Removes the NATS dependency  from Armada:

* Removes all NATS processing from server
* Removes the old "Legacy Event" Api that was populated from NATS
* Removes NATS processing from Lookout.
* Removes instantiation of Nats in E2E tests and LocalDev.


internal/armada/processor/event_nats.go
```diff
@@ -2,24 +2,20 @@ package processor
...
 
 type StreamEventStore struct {
-	stream             eventstream.EventStream
 	PulsarSubmitServer *server.PulsarSubmitServer
 }
 
-func NewEventStore(stream eventstream.EventStream) *StreamEventStore {
-	return &StreamEventStore{stream: stream}
+func NewEventStore() *StreamEventStore {
+	return &StreamEventStore{}
```
JamesMurkin, line 18: This file should probably be renamed

Also in ReportEvents - we should assume pulsar is enabled

d80tb7: done

docker-compose.yaml
```diff
@@ -1,4 +1,4 @@
-version: "3.8"
+xversion: "3.8"
```
JamesMurkin, line 1: Is this an intentional change?

d80tb7: no- a typo

config/armada/config.yaml
```diff
@@ -20,19 +20,12 @@ redis:
...
-  password: ""
-  db: 0
-  poolSize: 1000
 eventsApiRedis:
   addrs:
     - redis:6379
   password: ""
   db: 1
   poolSize: 1000
-forceNewEvents: true
```
severinson, line 35: Do we no longer need this flag?

makefile
```diff
@@ -486,7 +485,7 @@ tests-e2e: build-armadactl build-docker-no-lookout tests-e2e-setup
 		docker logs executor
 		echo -e "\nserver logs:"
 		docker logs server
-		docker rm -f nats redis pulsar server executor postgres lookout-ingester-migrate lookout-ingester jobservice
+		docker rm -f redis pulsar server executor postgres lookout-ingester-migrate lookout-ingester jobservice
```
severinson, line 488: event-ingester missing from this.

severinson (approved):

config/armada/config.yaml
d80tb7: we do not- the old event store was populated from nats so will now be state- you *have* to use the new events

e2e/setup/event-ingester-config.yaml
```diff
@@ -0,0 +1,13 @@
+redis:
```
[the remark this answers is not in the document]
d80tb7: ok let me remove this as one of my next PRs

makefile
d80tb7: ok I'll fix this as part of a new PR

PR #1972, opened 2023-01-01
Created seperate coverage action.

### LINKED ISSUE #1956: Fix coverage in PR
A required step according to:

https://about.codecov.io/blog/python-code-coverage-using-github-actions-and-codecov/

>One great feature that Codecov offers is the ability to add coverage reports right in your github pull requests. All you need to do is add a codecov.yml file in the root of your project. The file can be empty, it just needs to exist. Now, next time you make a PR a coverage report will appear that looks something like this.

Sharpz7:
Closes #1956.

It does so by allowing the action to run across all PR's, and more importantly, on main so that the correct coverage percentages are shown.

It also:
- Only runs the python coverage upload once, instead of multiple times for each python version
- Seperates the python client coverage and armada-server coverage by a custom codecov "flag"

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,54 @@
+name: Go Coverage
+
+on:
+  push:
+  pull_request:
+
+go-unit-coverage:
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        go: [ '1.18' ]
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/workflows/go-setup
+
+      - name: make tests
+        run: make tests
+
+      - name: Go Test Coverage
+        uses: codecov/codecov-action@v3
+        with:
+          files: ./internal_coverage.xml,./cmd_coverage.xml # optional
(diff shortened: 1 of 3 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: make tests
+        run: make tests
+
+      - name: Go Test Coverage
+        uses: codecov/codecov-action@v3
+        with:
+          files: ./internal_coverage.xml,./cmd_coverage.xml # optional
+          flags: armada-server
+
+python-unit-coverage:
```
kannon92, line 27: Missing airflow operator?

Sharpz7: Yep, good spot. Will get on this soon.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,54 @@
+name: Code Coverage
+
+on:
+  push:
+  pull_request:
+
+go-unit-coverage:
```
kannon92, line 7: I think you are missing a top level job command btw.

kannon92 (approved):

PR #1974, opened 2023-01-04
[Pulsar Scheduler] Core Scheduler

d80tb7:
This is the first PR for the new Pulsar Scheduler.  It introduces the core scheduler which runs a periodic scheduling loop.  This scheduling loop handles the mechanics of updating internal job state, determining  which jobs should be scheduled, and publishing new event messages.  The core scheduler itself relies on several other components to make this happen (`SchedulingAlgo`, `LeaderController`, `Publisher` etc.) which are introduced only as interfaces for now.  Implementations of these interfaces will follow in subsequent PRs.

Some notes on the PR:

* The code is not yet prod ready.  This is largely because of the need to balance PRs containing representative code with the neeed to get constant feedback.  
* The  core scheduling loop is quite well tested, but I plan to add more tests particularly around duplicate messages
* The original scheduler prototype had all JobIds as UUIDs.  In this PR I have changed them to be ULIDs, which defacto means representing them as strings most of the time. the  reason for this is that there will be a period of time where we have to run both the (old) Redis scheduler and the (new) Pulsar scheduler.  If we make the change to UUIDs now then this means we will have to deal with JobIds being either a UUId or a ULID.  In  the intertests of simplification, I've decideed this is a problem we could do without and so it's simnpler if all JobIds remain ULIDs. I think we will change JobIds to be UUIDs a some point, but this change is best made when we've decommissioned the Redis scheduler.
* Because of the aforementioned change this PR contains a large number of files where the only changes are moving UUIDs -> strings.  Please don't spend to much time on these- they correspond to code that is not in production and will be be anyway modified in subsequent PRs.
* Make sure you review `scheduler.go`- it's a large file so github is hiding it, but it's the most important thing here!


internal/scheduler/api.go
```diff
@@ -3,6 +3,7 @@ package scheduler
 import (
 	"context"
```
d80tb7, line 4: Please ignore changes in this file- this is not yet used by the scheduler and I plan to update it in a subsequent PR
For now, I've done enough to make compilation work but no more.

internal/scheduleringester/dbops.go
```diff
@@ -82,17 +82,17 @@ func discardNilOps(ops []DbOperation) []DbOperation {
 	return rv
```
d80tb7, line 82: only changes here are uuid -> string

internal/scheduleringester/dbops_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
```
d80tb7, line 4: only changes here are uuid -> string

internal/scheduleringester/instructions.go
```diff
@@ -140,7 +140,10 @@ func (c *InstructionConverter) handleSubmitJob(job *armadaevents.SubmitJob, meta
 		return nil, err
```
d80tb7, line 140: only changes here are uuid -> string

internal/scheduleringester/instructions_test.go
```diff
@@ -31,8 +31,8 @@ func TestConvertSequence(t *testing.T) {
 	}{
```
d80tb7, line 31: only changes here are uuid -> string

internal/scheduleringester/schedulerdb_test.go
```diff
@@ -11,13 +11,14 @@ import (
 	"github.com/stretchr/testify/assert"
```
d80tb7, line 11: uuid -> string

internal/scheduler/scheduler.go
```diff
@@ -2,344 +2,657 @@ package scheduler
...
+			log.Infof("Received a leaderToken. Leader status is %t", leaderToken.leader)
+
+			// If we are becoming leader then we must ensure we have caught up to all Pulsar messages
+			if leaderToken.leader && leaderToken != prevLeaderToken {
+				log.Infof("Becoming leader")
+				fullUpdate = true
+				err := s.ensureDbUpToDate(ctx, 1*time.Second)
+				if err != nil {
+					log.WithError(err).Error("Could not become master")
+					leaderToken = InvalidLeaderToken()
```
severinson, line 121: How is the k8s leader election logic notified that we're no longer leader so it can trigger a failover?

internal/scheduler/jobdb.go
```diff
@@ -56,31 +199,91 @@ func NewJobDb() (*JobDb, error) {
...
+	result := iter.Next()
+	if result != nil {
+		job = result.(*SchedulerJob)
+	}
+	return job, err
+}
+
+// GetAll returns all jobs in the database.
+// The Jobs returned by this function *must not* be subsequently modified
+func (jobDb *JobDb) GetAll(txn *memdb.Txn) ([]*SchedulerJob, error) {
```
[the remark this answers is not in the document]
d80tb7: So for actual scheduler calls I want to get rid of it- there's a comment in `scheduler.go` to that effect.  For unit testing purposes it's quite useful.  I don't think the performance will be that bad. Remember it's a slice of pointers so the memory shouldn't be too bad.  The cpu time needed to construct it should be roughly equivalent to iterating over the entire db- which also shouldn't be that bad and something I'd assume you'd need to do anyway if you call this function.

I think keep it for now and once we update `scheduler.go`  to not need it, we can revisit.

internal/scheduler/scheduler.go
```diff
@@ -2,344 +2,657 @@ package scheduler
...
-func (srv *Scheduler) getActiveExecutors() []string {
-	activeExecutors := make([]string, 0)
-	for executorName, lastSeen := range srv.Executors {
-		if time.Since(lastSeen) < srv.ExecutorAliveDuration {
-			activeExecutors = append(activeExecutors, executorName)
+// removeTerminalJobs takes the supplied list of jobs and removes any that are in a terminal state from the Job Db
+func (s *Scheduler) removeTerminalJobs(txn *memdb.Txn, updatedJobs []*SchedulerJob) error {
+	idsToDelete := make([]string, 0)
+	for _, job := range updatedJobs {
+		if job.Succeeded || job.Cancelled || job.Failed {
```
JamesMurkin, line 354: Could you just use Terminal() here?

internal/scheduler/database/migrations/001_initialize_schema.up.sql
```diff
@@ -29,7 +33,7 @@ ALTER TABLE jobs ALTER COLUMN submit_message SET STORAGE EXTERNAL;
 
 CREATE TABLE runs (
     run_id UUID PRIMARY KEY,
-    job_id UUID NOT NULL,
+    job_id TEXT NOT NULL,
```
JamesMurkin, line 36: Consistent column type casing. We jobs uses lowercase, runs uses upper

internal/scheduler/scheduler.go
d80tb7: It's not.  As far as K8's is concerned you are still leader.  Net time round the scheduling cycle you'll be told you're leader again and you can have another go at syncing you db state.

We could try and notify  K8s that your not leader, but I haven't implemented that yet.  A poor man's version of that would be to just crash at this point.

internal/scheduler/scheduler.go
```diff
@@ -2,344 +2,657 @@ package scheduler
...
+			start := s.clock.Now()
+			leaderToken := s.leaderController.GetToken()
+			fullUpdate := false
+			log.Infof("Received a leaderToken. Leader status is %t", leaderToken.leader)
+
+			// If we are becoming leader then we must ensure we have caught up to all Pulsar messages
+			if leaderToken.leader && leaderToken != prevLeaderToken {
+				log.Infof("Becoming leader")
+				fullUpdate = true
+				err := s.ensureDbUpToDate(ctx, 1*time.Second)
```
JamesMurkin, line 118: Possibly worth having some more defined timeout here

Currently if any of the events never come through on pulsar - it'll block forever waiting - or until the overarching context expires (which I imagine only happens on shutdown)

internal/scheduler/jobdb.go
```diff
@@ -111,33 +316,37 @@ func (it *JobQueueIterator) NextJobItem() *SchedulerJob {
...
 	return it.NextJobItem()
 }
 
+// jobDbSchema() creates the database schema.
+// This is a simple schema consisting of a single "jobs" table with indexes for fast lookups
 func jobDbSchema() *memdb.DBSchema {
 	indexes := make(map[string]*memdb.IndexSchema)
-	indexes["id"] = &memdb.IndexSchema{
-		Name:    "id",
+	indexes[idIndex] = &memdb.IndexSchema{
```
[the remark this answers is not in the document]
d80tb7: so the name of the index is still `id` it's just been broken out into a const variable called `idIndex`.  We could rename the variable but I'm mildly opposed because although it's obvious from the context here, it's not obvious in other places where we need to use it.

internal/scheduler/scheduler.go
d80tb7: yes I meant to do this but forgot- doh!

internal/scheduler/database/migrations/001_initialize_schema.up.sql
d80tb7: done- I've made everything lower case for now

internal/scheduler/scheduler.go
d80tb7: yes I can!

internal/scheduler/database/executor_repository.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+
+	"github.com/G-Research/armada/internal/scheduler/schedulerobjects"
+)
+
+// Executor is a representation of an Armada Executor
+type Executor struct {
```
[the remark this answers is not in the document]
d80tb7: Yeah- I think that's a very long term goal though.  For now we definitely need to stick with executors!

internal/scheduler/scheduler.go
ClifHouck: Maybe we should blow up/return with an error here then?

internal/scheduler/scheduler.go
```diff
@@ -2,344 +2,660 @@ package scheduler
...
-	for _, job := range jobs {
-		job := job
 
-		// Keep the serial number updated.
-		if job.Serial > srv.JobsSerial {
-			srv.JobsSerial = job.Serial
-		}
+	// Start a transaction- we'll roll this back if we don't publish
+	txn := s.jobDb.WriteTxn()
+	defer txn.Abort()
```
ClifHouck, line 164: I'm assuming `Abort()` is a no-op if `Commit()` has already been called?

ClifHouck: Looks like it https://github.com/hashicorp/go-memdb/blob/main/txn.go#L137

internal/scheduler/scheduler.go
```diff
@@ -2,344 +2,660 @@ package scheduler
...
+
+// syncState updates the state of the jobs in jobDb to match the state in postgres
+// It returns all jobs that have been updated
+func (s *Scheduler) syncState(ctx context.Context) ([]*SchedulerJob, error) {
+	updatedJobs, updatedRuns, err := s.jobRepository.FetchJobUpdates(ctx, s.jobsSerial, s.runsSerial)
 	if err != nil {
-		return errors.WithStack(err)
+		return nil, err
 	}
+	log.Infof("Received %d updated jobs and %d updated job runs", len(updatedJobs), len(updatedRuns))
```
ClifHouck, line 221: Nit: breaking out updateJobs and updatedRuns into logrus fields would enable slightly better log structure/analysis.

internal/scheduler/scheduler.go
```diff
@@ -2,344 +2,660 @@ package scheduler
...
+		default:
+			numReceived, err := s.jobRepository.CountReceivedPartitions(ctx, groupId)
+			if err != nil {
+				log.WithError(err).Error("Error querying the database  or marker messages")
+			}
+			if numSent == numReceived {
+				log.Infof("Successfully ensured that database state is up to date")
+				return nil
+			}
+			log.Infof("Recevied %d partitions, still waiting on  %d", numSent, numSent-numReceived)
```
ClifHouck, line 607: Nit: recommend breaking out formatting arguments into logrus fields.

internal/scheduler/database/modelsext.go
```diff
@@ -0,0 +1,16 @@
+package database
+
+// Terminal returns true if Job is in a terminal state
+func (job Job) Terminal() bool {
```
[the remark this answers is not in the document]
d80tb7: lets park this one for now- I'm planning to do some perf testing once we've got all the bits in place so we can see if it's an issue then.

internal/scheduler/jobdb.go
```diff
@@ -23,27 +31,162 @@ type JobDb struct {
...
+	return job.GetAnnotations()
+}
+
+// GetId returns the id of the Job.
+func (job *SchedulerJob) GetId() string {
+	return job.JobId
+}
+
+// Terminal returns true if the job  is in a terminal state
+func (job *SchedulerJob) Terminal() bool {
```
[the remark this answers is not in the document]
d80tb7: yeah that's a better name- let me update to use that

internal/scheduler/scheduler.go
d80tb7: > Maybe we should blow up/return with an error here then?
Yeah we could- the disadvantage is that we'd put more load on the db, because K8s would just restart us and then we'd hit the db again to re-initialise from scratch.

I think the current logic is fine- but We can re-evaluate this once I've done some E2E testing with real k8s.

internal/scheduler/api.go
```diff
@@ -68,7 +68,7 @@ func (srv *ExecutorApi) StreamingLeaseJobs(stream api.AggregatedQueue_StreamingL
 
 	// Get data stored in sql for these jobs.
 	// In particular, the Pulsar submit job message for each job.
-	jobIds := make([]uuid.UUID, len(runs))
+	jobIds := make([]string, len(runs))
```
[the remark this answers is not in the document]
d80tb7: yes that's a good idea- I'll do that

internal/scheduler/jobdb.go
```diff
@@ -23,27 +31,162 @@ type JobDb struct {
...
+	return job.Succeeded || job.Cancelled || job.Failed
+}
+
+// NumReturned returns the number of times this job has been returned by executors
+// Note that this is O(N) on Runs, but this should be fine as the number of runs should be small
```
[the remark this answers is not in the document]
d80tb7: The common case for runs is that it's 1, with a max of 5 so I think this is fine for now.

internal/scheduler/jobdb.go
```diff
@@ -56,31 +199,91 @@ func NewJobDb() (*JobDb, error) {
...
 		if err != nil {
 			return errors.WithStack(err)
 		}
 	}
-	txn.Commit()
 	return nil
 }
 
+// GetById returns the job with the given Id or nil if no such job exists
+// The Job returned by this function *must not* be subsequently modified
```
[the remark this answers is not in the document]
d80tb7: > When you say "must not be subsequently modified" you mean the structs returned by this method must not be modified by callers of this method?

Yes this is correct. It's a requirement of HashiCorp's memDb lib which we are using under the covers.

I like the idea of making `SchedulerJob` immutable (I worked for a long time as a scala dev and I miss things being immyutable by default!) but I don't think it's trivial because `SchedulerJob` has is a relatively complex objec t so it'll be quite a bit of code.  Any  objections if we do this in a follow-up PR once we've got this working E2E?

internal/scheduler/scheduler.go
d80tb7: As above- although I'm less convinced this is a useful stat to expose.

internal/scheduler/api.go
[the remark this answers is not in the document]
d80tb7: Actually although it's a good idea  the implementation turns out to be horrendous because the rest of the system expects JobIds to be strings, which in turn means we have to either change a gazillion functions to take our new ulid type, or we end up having to cast everywhere in the scheduler.

I think it's best to  kepp it as is for now.  Hopefully once the new scheduler is ourt, we can change the JobIds to be a proper UUID.  That should clear up this mess as everything has UUID support.

ClifHouck:
Should this merge to master or a longer lived feature branch? It seems to function in a localdev environment but you've stated there's more to flesh out.

internal/scheduler/api.go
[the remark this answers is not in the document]
severinson: I agree. Leave it as strings and then move towards UUIDs.

severinson (approved):

PR #1981, opened 2023-01-05
Magefile refactor, use magefile for proto

severinson:
`mage proto` is now more consistent with `make proto` as it worked prior to this PR.

`FileDescriptorProto` generated from `mage proto` differs from that of `make proto`. This is because `make proto` changes the generated go code in-place, causing the `FileDescriptorProto` to be inconsistent with the generated code.

There are some inconsistents of this form that I'm not sure where they're coming from. The first one is the original checked into the repo and the second that generated by `mage proto`. `make proto` puts some extra stuff (`,1,opt,name=key,proto3`) in the json tag.
```
-       AllocatedResources map[int32]ComputeResource `protobuf:"bytes,7,rep,name=allocated_resources,json=allocatedResources,proto3" json:"allocatedResources,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
+       AllocatedResources map[int32]ComputeResource `protobuf:"bytes,7,rep,name=allocated_resources,json=allocatedResources,proto3" json:"allocatedResources" protobuf_key:"varint,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
```

There are also inconsistencies of this form. `mage proto` consistently adds `,omitempty` for nullable fields, which I think is correct.
```
-       CancelledIds []string `protobuf:"bytes,1,rep,name=cancelled_ids,json=cancelledIds,proto3" json:"cancelledIds"`
+       CancelledIds []string `protobuf:"bytes,1,rep,name=cancelled_ids,json=cancelledIds,proto3" json:"cancelledIds,omitempty"`
```


magefiles/protoc.go
```diff
@@ -52,3 +61,66 @@ func protocCheck() error {
 	}
 	return nil
 }
+
+func protocInstall() error {
+	if ok, err := exists("./protoc"); ok || err != nil {
```
samclark, line 66: replace `"./protoc`"with `protocBinary()`

magefiles/go.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func goModuleVersion(name string) (string, error) {
+	out, err := goOutput("list", "-m", name)
+	if err != nil {
+		return "", err
+	}
+	fields := strings.Fields(out)
+	if len(fields) != 2 {
+		return "", errors.Errorf("unexpected protoc version output: %v", fields)
```
samclark, line 67: replace `unexpected protoc version` with `unexpected go list output`

makefile
```diff
@@ -566,34 +566,8 @@ airflow-operator:
...
-
-	# embed swagger json into go binary
-	$(GO_TEST_CMD) templify -e -p=api -f=SwaggerJson  pkg/api/api.swagger.json
-	$(GO_TEST_CMD) templify -e -p=lookout -f=SwaggerJson  pkg/api/lookout/api.swagger.json
-	$(GO_TEST_CMD) templify -e -p=binoculars -f=SwaggerJson  pkg/api/binoculars/api.swagger.json
-
-	# fix all imports ordering
-	$(GO_TEST_CMD) goimports -w -local "github.com/G-Research/armada" ./pkg/api/
-	$(GO_TEST_CMD) goimports -w -local "github.com/G-Research/armada" ./pkg/armadaevents/
-	$(GO_TEST_CMD) goimports -w -local "github.com/G-Research/armada" ./internal/scheduler/schedulerobjects/
```
samclark, line 596: we should be able to remove the `./scripts/proto.sh` and ` ./build/proto/Dockerfile` files if we're using the mage code path exclusively

pkg/client/auth/kerberos/spnego_windows.go
```diff
@@ -3,7 +3,10 @@ package kerberos
 import (
```
samclark, line 3: these changes seem unrelated to the pr

samclark (commented):
Overall looks great ... have made a few comments

makefile
samclark: looks like `./scripts/proto.sh` has been removed in this pr

pkg/client/auth/kerberos/spnego_windows.go
severinson: The errcheck linter complained about these.
I see now that pkg is excluded from the errcheck running in CI. I can move these to a separate PR.

severinson:
I've left `merge-swagger.go` separate from the magefiles for now. We can in a separate PR do one of
- Merge into magefiles
- Move all generation to use `go:generate`
- Remove the swagger altogether

samclark (approved):

PR #1984, opened 2023-01-06
Fix errcheck in pkg

severinson:


pkg/client/auth/kerberos/spnego_windows.go
```diff
@@ -21,18 +25,29 @@ func NewSPNEGOCredentials(serverUrl string, config ClientConfig) (credentials.Pe
...
+	log := ctxlogrus.Extract(ctx)
 
 	cred, err := negotiate.AcquireCurrentUserCredentials()
 	if err != nil {
 		return nil, err
 	}
-	defer cred.Release()
+	defer func() {
+		if err := cred.Release(); err != nil {
+			err = errors.WithStack(err)
```
kannon92, line 36: Do you need the `errors.WithStack(err)` if you are doing `logging.WithStacktrace`?  Seems redundant?

kannon92 (approved):

pkg/client/auth/kerberos/spnego_windows.go
severinson: `logging.WithStacktrace` prints the stack trace captured via `errors.WithStack`.

PR #1997, opened 2023-01-10
Support Lower Python Client Versions

Sharpz7:
Closes https://github.com/G-Research/armada/issues/1996


client/python/pyproject.toml
```diff
@@ -4,7 +4,7 @@ version = "0.2.6"
 description = "Armada gRPC API python client"
 readme = "README.md"
 requires-python = ">=3.7"
-dependencies = ["grpcio>=1.46.3", "grpcio-tools>=1.46.3", "mypy-protobuf>=3.2.0"]
+dependencies = ["grpcio>=1.48", "grpcio-tools>=1.48", "mypy-protobuf>=3.2.0", "protobuf>=3.20.3"]
```
kannon92, line 7: Why the upgrade of grpcio and grpcio-tools?

Sharpz7: I don't think there's a negative, and as well as that it makes it easier to track the version that current breaks protobuf being below v4.

kannon92 (dismissed):

theAntiYeti (approved):

kannon92 (approved):

PR #2002, opened 2023-01-11
Reconcile config

severinson:
In preparation for using goreleaser in CI.
Images are included because I've renormalised line endings.


deployment/executor/Chart.yaml
```diff
@@ -3,11 +3,3 @@ description: A helm chart for armada-executor component
 name: armada-executor
 version: LATEST
 appVersion: LATEST
-
-priorityClasses:
```
severinson, line 7: This was moved into `values.yaml` in another PR.

internal/lookout/ui/src/components/lookoutV2/ColumnSelect.module.css
```diff
@@ -1,30 +1,30 @@
-.columnMenu {
```
severinson, line 1: Normalised line endings.

PR #2008, opened 2023-01-12
Lookout V2 use batch cancellation endpoint

carlocamurri:
* Uses batch cancellation endpoint to cancel jobs in batches
* Specifies queue and jobset with each request (as the backend will require this in the future)
* Changes node services port in kind from 3000 to 11000, as 3000 is reserved for React local development

Next step is to make sure jobs are loaded and cancelled incrementally, rather than load all -> cancel all. This requires larger refactorings, and will require this PR.


internal/lookout/ui/src/services/lookoutV2/UpdateJobsService.ts
```diff
@@ -14,73 +14,99 @@ export interface UpdateJobsResponse {
...
 
-  reprioritiseJobs = async (jobIds: JobId[], newPriority: number): Promise<UpdateJobsResponse> => {
+  reprioritiseJobs = async (jobs: Job[], newPriority: number): Promise<UpdateJobsResponse> => {
     const response: UpdateJobsResponse = { successfulJobIds: [], failedJobIds: [] }
 
-    const maxJobsPerRequest = 5000
-    const jobIdChunks = _.chunk(jobIds, maxJobsPerRequest)
+    const maxJobsPerRequest = 10000
+    const chunks = createJobBatches(jobs, maxJobsPerRequest)
+    console.log(chunks)
```
d80tb7, line 75: should this be here?

internal/lookout/ui/src/utils/jobsDialogUtils.ts
```diff
@@ -1,12 +1,12 @@
...
-import { JobFilter, Job } from "models/lookoutV2Models"
+import { Job, JobFilter } from "models/lookoutV2Models"
 import { IGetJobsService } from "services/lookoutV2/GetJobsService"
 
 export const getAllJobsMatchingFilters = async (
   filters: JobFilter[],
   getJobsService: IGetJobsService,
 ): Promise<Job[]> => {
-  const MAX_JOBS_PER_REQUEST = 1000
+  const MAX_JOBS_PER_REQUEST = 10000
```
d80tb7, line 9: it's ok to increase this by a factor of 10x?

carlocamurri: I believe so - API should be able to serve 10000 jobs at a time. It will result in fewer requests being made for the same requests.

Other option is having this be configurable

PR #2012, opened 2023-01-12
Run testsuite in circleci

severinson:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-102) by [Unito](https://www.unito.io)


.circleci/config.yml
```diff
@@ -1,4 +1,7 @@
 version: 2.1
+orbs:
```
kannon92, line 2: I thought the general consensus was the phase out CircleCI for GHA.

Kotwic4: yet armada was not moved away from CircleCI in last 4 months and this PR was fixing releasing c# client

kannon92: Yea we have been working on this over time.  Some areas we find difficult are how to address release artifacts without leaking secrets in github.  We also were given a series of tasks to research before we fully replace circle ci.  

One of the last steps is to remove the release stage from circle CI but we have a few items that are of higher priority.  

Thank you for looking into this!

kannon92:
We removed circle ci so this PR is no longer relevant.

PR #2016, opened 2023-01-13
First set of Scheduler Interface Implementations

d80tb7:
Adds Implementations for the `LeaderController` and `Publisher` interfaces


pkg/armadaevents/events.proto
```diff
@@ -463,19 +463,6 @@ message JobDuplicateDetected {
     Uuid old_job_id = 2;
 }
 
-message SeqUpdate{
```
d80tb7, line 466: this was part of the aborted postgres based events service

internal/scheduler/publisher.go
```diff
@@ -2,20 +2,193 @@ package scheduler
...
 	PublishMessages(ctx context.Context, events []*armadaevents.EventSequence, token LeaderToken) error
 
 	// PublishMarkers publishes a single marker message for each Pulsar partition.  Each marker
 	// massage contains the supplied group id, which allows nall marker messages for a given call
 	// to be identified.  The uint32 returned is the number of messages published
 	PublishMarkers(ctx context.Context, groupId uuid.UUID) (uint32, error)
 }
+
+// PulsarPublisher is the default implementation of Publisher
+type PulsarPublisher struct {
```
severinson, line 35: Is this the right abstraction? I'd prefer leader election and pulsar publishing be separate abstractions, i.e., 1) check if leader and 2) publish. Imo this abstraction is too specialised.

severinson (dismissed):

severinson (approved):

PR #2028, opened 2023-01-16
Update CircleCI config to use new GH org.

richscott:


.circleci/config.yml
```diff
@@ -48,8 +48,8 @@ commands:
 
   checkout-charts:
     description: |
-      Checkout G-Research charts repo so we can update them later
-      Checkout from: https://github.com/G-Research/charts
```
kannon92, line 52: I don't think this is correct.  I don't believe we migrated charts yet.

richscott: Thanks - fixed in commit 682f2c4d.

kannon92 (dismissed):

kannon92 (dismissed):

jayofdoom (approved):
I don't love that this mixes resource_class changes and working_directory changes.

But YOLO! If it works it works! Migration day :partying_face:

PR #2035, opened 2023-01-18
Improved issue handling

JamesMurkin:
Improved handling stuck pending pods

 - Improving how we handle non-retryable jobs. So we don't mark the jobs as done  and then fail  to send a JobFailedEvent causing jobs to end with no terminal event
 - Improve handling retryable pending jobs 
   -   Try very hard to not delete pods that have changed state 
   -  Try very hard to not ReturnLease of now Running/Succeeded/Failed jobs


internal/executor/context/cluster_context.go
```diff
@@ -317,6 +316,59 @@ func (c *KubernetesClusterContext) AddClusterEventAnnotation(event *v1.Event, an
...
+			return err
+		}
+		currentPod, err = c.waitForPodUpdateInLocalCache(pod, time.Second*5)
+		if err != nil {
+			return err
+		}
+	}
+
+	if !condition(currentPod) {
+		return fmt.Errorf("pod does not match provided condition")
```
d80tb7, line 337: I think `errors.Errorf` is marginally better as it captures a stacktrace.

d80tb7 (approved):

internal/executor/context/cluster_context.go
```diff
@@ -317,6 +316,59 @@ func (c *KubernetesClusterContext) AddClusterEventAnnotation(event *v1.Event, an
 	return nil
 }
 
+func (c *KubernetesClusterContext) DeletePodWithCondition(pod *v1.Pod, condition func(pod *v1.Pod) bool, pessimistic bool) error {
```
dave-gantenbein, line 319: It would be great to have tests for all this logic.

PR #2036, opened 2023-01-18
Add new table for locking migrations exclusively to one session

### LINKED ISSUE #2033: Migration DB locking
I think we can do locking on the database_version sequence so only one migration instance can access

suprjinx:
<!--  Thanks for sending a pull request!  Here are some tips for you:

#### What type of PR is this?

Code change 

Changes Lookout migrations code to prevent potential concurrency issues during migration

Fixes #2033 

Expect no change but concurrent runs will be prevented

### DIFF
--- a/internal/lookout/repository/schema/migrations.go
+++ b/internal/lookout/repository/schema/migrations.go
@@ -2,11 +2,13 @@ package schema
 
 import (
 	"bytes"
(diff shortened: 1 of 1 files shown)


suprjinx:
Not needed, using Batch Job for this

PR #2037, opened 2023-01-18
Simplify NewZlibDecompressor return

ClifHouck:
`NewZlibDecompressor` doesn't need to have an `error` in its return signature since it only returns `nil`. This corrects that.


d80tb7 (approved):

d80tb7:
lgtm!

d80tb7 (approved):

PR #2048, opened 2023-01-20
Make scheduler job order deterministic

severinson:
Necessary for future work on preemption to fair share.
Also, various simplifications.


internal/armada/repository/job.go
```diff
@@ -56,7 +56,7 @@ type JobResult struct {
 
 type JobRepository interface {
 	PeekQueue(queue string, limit int64) ([]*api.Job, error)
-	TryLeaseJobs(clusterId string, queue string, jobs []*api.Job) ([]*api.Job, error)
+	TryLeaseJobs(clusterId string, jobIdsByQueue map[string][]string) (map[string][]string, error)
```
d80tb7, line 59: can we document what's going on here?  Specifically what does the returned `map[string][]string` represent

internal/armada/repository/job.go
```diff
@@ -139,17 +139,28 @@ func (repo *RedisJobRepository) AddJobs(jobs []*api.Job) ([]*SubmitJobResult, er
...
 
 func (repo *RedisJobRepository) RenewLease(clusterId string, jobIds []string) (renewedJobIds []string, e error) {
+	// TODO: If we can pass in the queue, we don't need to load jobs from Redis.
 	jobs, err := repo.GetExistingJobsByIds(jobIds)
 	if err != nil {
 		return nil, err
 	}
-
-	leasedJobs, err := repo.leaseJobs(clusterId, jobs)
+	jobsById := make(map[string]*api.Job)
```
d80tb7, line 147: you can assign this map with the correct length

internal/armada/scheduling/pod_matching_context.go
```diff
@@ -2,29 +2,28 @@ package scheduling
 
 import (
```
d80tb7, line 3: do we still need this? I had a feeling this was used by the old scheduler but do w e use it in the new one?

internal/scheduler/legacyscheduler.go
```diff
@@ -317,99 +358,248 @@ func (it *QueueCandidateGangIterator[T]) schedulingReportsFromJobs(ctx context.C
...
+}
+
+func (pq *QueueCandidateGangIteratorPQ) Push(x any) {
+	n := len(*pq)
+	item := x.(*QueueCandidateGangIteratorItem)
+	item.index = n
+	*pq = append(*pq, item)
+}
+
+func (pq *QueueCandidateGangIteratorPQ) Pop() any {
```
d80tb7, line 417: will this panic if you try  and Pop from an empty queue?

d80tb7: Just to be clear, I'd expect that popping from an empty queue would be a valid operation and would return nil (assuming reference types only) or a (zeroValue, bool) (assuming we can have value types too)

internal/scheduler/legacyscheduler.go
```diff
@@ -533,15 +709,23 @@ func (sched *LegacyScheduler[T]) String() string {
...
 	constraints SchedulingConstraints,
 	config configuration.SchedulingConfig,
 	nodeDb *NodeDb,
-	jobRepository SchedulerJobRepository[T],
+	jobIteratorsByQueue map[string]JobIterator,
 	priorityFactorByQueue map[string]float64,
 	initialResourcesByQueueAndPriority map[string]schedulerobjects.QuantityByPriorityAndResourceType,
-) (*LegacyScheduler[T], error) {
+) (*LegacyScheduler, error) {
+	if len(jobIteratorsByQueue) != len(priorityFactorByQueue) {
```
d80tb7, line 721: I'd be tempted to combine `jobIteratorsByQueue` and `priorityFactorByQueue` to a `map[string][Queue]` where `Queue` is:
```
 Struct{
    jobIterator JobIterator
    priorityFactor float64    
 }
  ```
That way this function doesn't have to check that they are consistent.  Moreover I'm 99% sure we generate these at the same time (if we don't we have a problem!) so it'll be fewer objects to pass around.

File: internal/scheduler/legacyscheduler.go [hunk 6]
```diff
@@ -758,100 +942,45 @@ func queueSelectionWeights(priorityFactorByQueue map[string]float64, aggregateRe
 	return rv
 }
 
+func max(a, b float64) float64 {
```
d80tb7, line 945: think this exists in the standard library now

internal/scheduler/submitcheck.go
```diff
@@ -56,10 +56,11 @@ func (srv *SubmitChecker) CheckApiJobs(jobs []*api.Job) (bool, string) {
 		if gangId == "" {
 			continue
 		}
-		reqs, err := PodRequirementsFromJobs(srv.priorityClasses, jobs)
-		if err != nil {
-			return false, fmt.Sprintf("error encountered extracting scheduling requirements")
-		}
+		reqs := PodRequirementsFromLegacySchedulerJobs(jobs, srv.priorityClasses)
+		// TODO: Fix
```
d80tb7, line 60: does this need fixing/removing before merge?

internal/armada/repository/job.go
severinson: Done

internal/armada/repository/job.go
severinson: Fixed

internal/armada/scheduling/pod_matching_context.go
severinson: It's still used by the submit-time checks I believe.

internal/scheduler/legacyscheduler.go
severinson: Yes, it'll panic on calling `Pop()` on an empty queue. I think doing so is analogous to trying to access an out of bounds index on a slice and that it should cause a panic.

internal/scheduler/legacyscheduler.go
severinson: Fixed

internal/scheduler/submitcheck.go
severinson: This function doesn't even return an error. I've removed the todo.

internal/scheduler/legacyscheduler.go
severinson: I agree. Fixed.

internal/armada/scheduling/pod_matching_context.go
severinson: In any case, let's just fix compiler errors resulting from me moving things around now and remove the code subsequent PRs.

internal/scheduler/legacyscheduler.go
severinson: Additionally, this is the standard PQ implementation given in the docs. Not that we should necessarily use that implementation if we think it's wrong though.
https://pkg.go.dev/container/heap#example-package-PriorityQueue

d80tb7 (approved):

PR #2050, opened 2023-01-22
Initialise New Scheduler

d80tb7:
Code Necessary to Intialise new scheduler and start the scheduling cycle.  Main changes in the this PR:

* Scheduler now has subcommands (one for runnign the scheduler, another to migrate the db etc)
* Implementation of `QueueRepository` that is just backed by the existing Redis code.
* Scheduler has its config validated by https://github.com/go-playground/validator. If this works well we can do the same for other components
*  Full initialisation code for the scheduler (see `schedulerapp` for this)
* Added viper hooks for the pulsar config enums so they are parsed as part of viper rather than in app code


cmd/scheduler/cmd/migrate_database.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return err
+	}
+	start := time.Now()
+	log.Info("Beginning scheduler database migration")
+	db, err := database.OpenPgxConn(config.Postgres)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to connect to database")
+	}
+	err = schedulerdb.Migrate(context.Background(), db)
```
severinson, line 36: Should we have a configurable timeout?

cmd/scheduler/cmd/migrate_database.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "Failed to migrate scheduler database")
```
severinson, line 38:
```suggestion
		return errors.WithMessage(err, "Failed to migrate scheduler database")
```

severinson: `Wrap` attaches a stack trace, but we should only attach a stack trace at the error origin and only a message here.

cmd/scheduler/cmd/migrate_database.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	taken := time.Now().Sub(start)
```
severinson, line 40: Use `time.Since` instead.

config/scheduler/config.yaml
```diff
@@ -1,25 +1,94 @@
+cyclePeriod: 10s
+executorTimeout: 1h
```
severinson, line 2: 1h sounds like a lot.

internal/common/config/hooks.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/apache/pulsar-client-go/pulsar"
+	"github.com/mitchellh/mapstructure"
+	"github.com/pkg/errors"
+	"github.com/spf13/viper"
+	"k8s.io/apimachinery/pkg/api/resource"
+
+	"github.com/armadaproject/armada/internal/common/armadaerrors"
+)
+
+var CustomHooks = []viper.DecoderConfigOption{
```
severinson, line 17: Validation! So nice.

internal/scheduler/api.go
```diff
@@ -27,25 +27,24 @@ type ExecutorApi struct {
...
+	maxJobsPerCall uint,
 ) *ExecutorApi {
 	return &ExecutorApi{
 		producer:             producer,
 		jobRepository:        jobRepository,
 		executorRepository:   executorRepository,
 		allowedPriorities:    allowedPriorities,
 		maxJobsPerCall:       maxJobsPerCall,
-		maxPulsarMessageSize: maxPulsarMessageSize,
+		maxPulsarMessageSize: 1024 * 1024 * 2,
```
severinson, line 47: Should this be hardcoded?

internal/scheduler/schedulerapp.go
```diff
@@ -0,0 +1,146 @@
...
+	// Pulsar
+	//////////////////////////////////////////////////////////////////////////
+	log.Infof("Setting up Pulsar connectivity")
+	pulsarClient, err := pulsarutils.NewPulsarClient(&config.Pulsar)
+	defer pulsarClient.Close()
+	if err != nil {
+		return errors.WithMessage(err, "Error creating pulsar client")
+	}
+	pulsarPublisher, err := NewPulsarPublisher(pulsarClient, pulsar.ProducerOptions{
+		Name:             fmt.Sprintf("armada-scheduler-%s", uuid.New()),
```
severinson, line 61: `uuid.NewString()` ?

internal/scheduler/schedulerapp.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return errors.WithMessage(err, "Error creating pulsar publisher")
+	}
+
+	//////////////////////////////////////////////////////////////////////////
+	// Leader Election
+	//////////////////////////////////////////////////////////////////////////
+	leaderController, err := createLeaderController(config.Leader)
+	if err != nil {
+		return errors.WithMessage(err, "Error creating leader controller")
```
severinson, line 76:
```suggestion
		return errors.Wrap(err, "Error creating leader controller")
```

internal/scheduler/schedulerapp.go
```diff
@@ -0,0 +1,146 @@
...
+		CompressionLevel: config.Pulsar.CompressionLevel,
+		BatchingMaxSize:  config.Pulsar.MaxAllowedMessageSize,
+		Topic:            config.Pulsar.JobsetEventsTopic,
+	})
+	if err != nil {
+		return errors.Wrapf(err, "Error creating pulsar producer for executor api")
+	}
+	authServices := auth.ConfigureAuth(config.Auth)
+	grpcServer := grpcCommon.CreateGrpcServer(config.Grpc.KeepaliveParams, config.Grpc.KeepaliveEnforcementPolicy, authServices)
+	defer grpcServer.GracefulStop()
```
severinson, line 96: You may want to close forcefully on a a timeout (like we do in `server.go`), since I think `GracefulStop` may hang indefinitely while waiting for connections to close.

internal/scheduler/schedulerapp.go
```diff
@@ -0,0 +1,146 @@
...
+	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", config.Grpc.Port))
+	if err != nil {
+		return errors.WithMessage(err, "Error setting up grpc server")
```
severinson, line 99:
```suggestion
		return errors.WithMessage(err, "failed to setup grpc server")
```

severinson: Nit. But ut's recommended to start error messages with lower-case letters since thet're often strung together like "message 1: message 2` etc.

internal/scheduler/schedulerapp.go
```diff
@@ -0,0 +1,146 @@
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "Error creating kubernetes client")
+		}
+		clientSet, err := kubernetes.NewForConfig(clusterConfig)
+		if err != nil {
+			return nil, errors.Wrapf(err, "Error creating kubernetes client")
+		}
+		return NewKubernetesLeaderController(LeaderConfig{}, clientSet.CoordinationV1()), nil
+	default:
+		return nil, errors.New(fmt.Sprintf("%s is not a value leader mode", config.Mode))
```
severinson, line 144: use `errors.Errorf`.

internal/scheduler/schedulerapp.go
severinson: Same for other parts of this file.

internal/scheduler/schedulerapp.go
d80tb7: I think this should be fine- the armada server has to deal with unbounded streams and rpcs that last hours.  This should be dealing with rpcs that last seconds at most.

internal/scheduler/schedulerapp.go
d80tb7: done

severinson (approved):

PR #2053, opened 2023-01-23
Initial implementation of executor using executor API

JamesMurkin:
Implements the changes required to use the executor API 

Only enabled if `useExecutorApi` is set to true

The executor can only talk to either the existing armada server api OR the executor api - - it cannot do both at once

**Changes to existing executor functionality**

 - We no longer support multi pod Jobs on the executor. This has been the case server side for a while so should be a noop. Executor will now only submit the first pod in podSpecs
 - We now rely on the server side to produce most of the existing ingress/service objects for existing flows. I have removed where we populate them if they aren't given to us by the API

**ReportEvents**

For this call we simply take the events we get now and convert them to the new pulsar events
 - We also add the RunId

Longer term, we will move to natively creating the pulsar events and remove the conversion step


internal/executor/reporter/event_sender.go
```diff
@@ -0,0 +1,100 @@
...
+			return err
+		}
+		runId, err := armadaevents.ProtoUuidFromUuidString(e.JobRunId)
+		if err != nil {
+			return fmt.Errorf("Failed to convert uuid string %s to uuid because %s", e.JobRunId, err)
+		}
+		populateRunId(sequence, runId)
+		sequences = append(sequences, sequence)
+	}
+	sequences = eventutil.CompactEventSequences(sequences)
```
d80tb7, line 49: You might want to limit the size here else you could end up wit a mega event! I think there's something in eventutil to do this.

internal/executor/reporter/event_sender.go
```diff
@@ -0,0 +1,111 @@
...
+		case *armadaevents.EventSequence_Event_JobRunAssigned:
+			runEvent.JobRunAssigned.RunId = jobRunId
+		case *armadaevents.EventSequence_Event_JobRunRunning:
+			runEvent.JobRunRunning.RunId = jobRunId
+		case *armadaevents.EventSequence_Event_JobRunSucceeded:
+			runEvent.JobRunSucceeded.RunId = jobRunId
+		case *armadaevents.EventSequence_Event_JobRunErrors:
+			runEvent.JobRunErrors.RunId = jobRunId
+		case *armadaevents.EventSequence_Event_JobRunPreempted:
+			//TODO set preemptive run id correctly
```
JamesMurkin, line 77:
```suggestion
			// TODO set preemptive run id correctly
```

d80tb7 (approved):

PR #2054, opened 2023-01-23
Add JobsMultiIterator

severinson:
Will be used to first iterate over running jobs before queued jobs in the scheduler.


internal/scheduler/jobiteration.go
```diff
@@ -85,3 +95,33 @@ func queuedJobsIteratorLoader(ctx context.Context, jobIds []string, ch chan *api
 	}
 	return nil
 }
+
+// MultiJobsIterator chains several JobIterators together,
+// emptying them in the order provided.
+type MultiJobsIterator struct {
+	i   int
```
d80tb7, line 102: minor not but I'd probably use a more descriptive name than `i` here.

d80tb7 (approved):

PR #2056, opened 2023-01-23
Configurable utilization metrics scraping

robertdavidsmith:
Allow scraping of custom utilisation metrics for pods, directly from a configurable prometheus (In addition to the existing cpu and memory which are scraped from the kublet).

Immediate need is for gpu utilisation, which will be configured with the following

```
metric:
  exposeQueueUsageMetrics: true
  customUsageMetrics:
  - namespace: gpu-operator
    endpointSelectorLabelName: app
    endpointSelectorLabelValue: nvidia-dcgm-exporter
    metrics:
    - name: gpu
      prometheusMetricName: DCGM_FI_DEV_GPU_UTIL
      prometheusPodNameLabel: pod
      aggregateType: Sum
      multiplier: 0.01
```

Others will be added in future, definitely memory and maybe power usage.

The implementation involves
- Get list of endpoints from EndpointSlice informer
- Map these to urls
- Download from the urls, and reuse code under https://github.com/prometheus/common to parse the returned text
- Aggregate the samples into the required values.

These is also some refactoring to make this easier, e.g. break kubelet metrics fetching out of pod_utilisation.go.


internal/executor/utilisation/pod_utilisation_kubelet_metrics.go
```diff
@@ -0,0 +1,69 @@
+package utilisation
```
robertdavidsmith, line 1: Relocated from pod_utilisation.go

internal/executor/utilisation/pod_utilisation.go
```diff
@@ -24,26 +24,39 @@ type PodUtilisationService interface {
...
 	podUtilisationData map[string]*domain.UtilisationData
 	dataAccessMutex    sync.Mutex
 }
 
-func NewMetricsServerQueueUtilisationService(
+type podUtilisationFetcher interface {
+	fetch(nodes []*v1.Node, podNameToUtilisationData map[string]*domain.UtilisationData, clusterContext clusterContext.ClusterContext)
+}
+
+func NewPodUtilisationService(
```
robertdavidsmith, line 39: Consistent naming - rename NewMetricsServerQueueUtilisationService to NewPodUtilisationService as that's what it's called elsewhere

JamesMurkin (approved):

PR #2057, opened 2023-01-24
Add nodeIdLabel config and optimisations

severinson:
With this PR, we add a node label the value of which uniquely identifies each node. This is necessary to have a common notion of what identifies a node. We also add optimisations for node selectors that include this label. These optimisations will be important for future scheduler work that require efficiently targeting a specific node for scheduling a pod.


pkg/api/util.go
```diff
@@ -5,32 +5,48 @@ import (
...
 	"github.com/armadaproject/armada/internal/scheduler/schedulerobjects"
 )
 
-func NewNodeFromNodeInfo(nodeInfo *NodeInfo, executor string, allowedPriorities []int32, lastSeen time.Time) *schedulerobjects.Node {
+func NewNodeFromNodeInfo(nodeInfo *NodeInfo, nodeIdLabel string, allowedPriorities []int32, lastSeen time.Time) (*schedulerobjects.Node, error) {
 	allocatableByPriorityAndResource := schedulerobjects.NewAllocatableByPriorityAndResourceType(allowedPriorities, nodeInfo.TotalResources)
 	for p, rs := range nodeInfo.AllocatedResources {
 		allocatableByPriorityAndResource.MarkAllocated(p, schedulerobjects.ResourceList{Resources: rs.Resources})
 	}
+	nodeId, ok := nodeInfo.Labels[nodeIdLabel]
```
d80tb7, line 24: this now means that you *have* to have label that uniquely identifies a node.  Whap happens if the user doesn't set this?  Does Armada fail fast or do confusing things happen?

severinson: Armada will ignore and log a warning for any nodes it can't get the id of.

PR #2058, opened 2023-01-24
Scheduler Db Cleanup Job

d80tb7:
Cleanup Job for the scheduler Db. This removes information for all jobs that have been in a terminal state longer than some configurable tine period


internal/scheduler/database/db_pruner.go
```diff
@@ -0,0 +1,103 @@
...
+	"k8s.io/apimachinery/pkg/util/clock"
+)
+
+// PruneDb removes completed jobs (and related runs and errors) from the database if their `lastUpdateTime`
+// is more than `keepAfterCompletion` in the past.
+// Jobs are deleted in batches across transactions. This means that if this job fails midway through, it still
+// may have deleted som jobs.
+// The function will run until the supplied context is cancelled.
+func PruneDb(ctx ctx.Context, db *pgx.Conn, batchLimit int, keepAfterCompletion time.Duration, clock clock.Clock) error {
+	start := time.Now()
```
carlocamurri, line 19: Should this be `clock.Now`?

d80tb7: So I don't know if there's a clear answer to this.  It doesn't *have* to be `clock.Now` because this variable is only used to log out how long the migration took to run and we don't need to manipulate that in unit tests.  Basically the tradeoff is:
- If we use `clock.Now` the benefit  is that it means that all times in the function are consistent and reduces the risk that someone will  start using e.g. `start` for something else in future and then get weird things happening in tests.
- The disadvantage is that the log lines saying how long each function took to run will be nonsense when we're running tests.

internal/scheduler/database/db_pruner.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/jackc/pgx/v4"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/clock"
+)
+
+// PruneDb removes completed jobs (and related runs and errors) from the database if their `lastUpdateTime`
+// is more than `keepAfterCompletion` in the past.
+// Jobs are deleted in batches across transactions. This means that if this job fails midway through, it still
+// may have deleted som jobs.
```
JamesMurkin, line 16:
```suggestion
// may have deleted some jobs.
```

internal/scheduler/database/db_pruner.go
d80tb7: I spoke to @JamesMurkin and we came to an executive decision that leaving it as is was best- please shout if you disagree!

internal/scheduler/database/db_pruner.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+
+			if batchSize == 0 {
+				// nothing more to delete
+				keepGoing = false
+				return nil
+			}
+
+			// Delete everything that's present in the batch table
+			// Do this all in oe call so as to be more terse with the syntax
```
JamesMurkin, line 81:
```suggestion
			// Do this all in one call so as to be more terse with the syntax
```

JamesMurkin (approved):

PR #2059, opened 2023-01-25
Scheduler plumbing necessary for preemption to fair share

severinson:


internal/scheduler/schedulerobjects/schedulerobjects.proto
```diff
@@ -9,47 +9,49 @@ import "github.com/gogo/protobuf/gogoproto/gogo.proto";
...
     ResourceList TotalResources = 7 [(gogoproto.nullable) = false];
-	  // Resources available for jobs of a given priority.
-	  // E.g., AvailableResources[5]["cpu"] is the amount of CPU available to jobs with priority 5,
-	  // where available resources = unused resources + resources assigned to lower-priority jobs.
+    // Resources available for jobs of a given priority.
+    // E.g., AvailableResources[5]["cpu"] is the amount of CPU available to jobs with priority 5,
+    // where available resources = unused resources + resources assigned to lower-priority jobs.
     map<int32, ResourceList> AllocatableByPriorityAndResource = 8 [(gogoproto.nullable) = false];
-    // run ids of jobs running on this node
+    // Ids of the jobs runs currently assigned to this this.
```
d80tb7, line 50:
```suggestion
    // Ids of the jobs runs currently assigned to this node.
```

internal/scheduler/testfixtures_test.go
```diff
@@ -0,0 +1,427 @@
...
+	for _, req := range reqs {
+		maps.Copy(
+			req.ResourceRequirements.Requests,
+			schedulerobjects.V1ResourceListFromResourceList(rl),
+		)
+	}
+	return reqs
+}
+
+func testNSmallCpuJob(priority int32, n int) []*schedulerobjects.PodRequirements {
```
d80tb7, line 161: Can we rename this `testNOneCpuJob`

d80tb7: Same for the other methods here.  It also means that someone is less likely to change what "small" means without thinking about it.

internal/scheduler/testfixtures_test.go
```diff
@@ -0,0 +1,427 @@
...
+				Value: "true",
+			},
+		},
+		Annotations: map[string]string{
+			JobIdAnnotation: uuid.NewString(),
+		},
+	}
+}
+
+func testNodeItems1() []*schedulerobjects.Node {
```
d80tb7, line 243: testCpuCluster?

d80tb7 (approved):

internal/scheduler/testfixtures_test.go
severinson: That'd be good. I'll improve these fixtures as I write more tests.

internal/scheduler/testfixtures_test.go
severinson: Agreed

PR #2060, opened 2023-01-25
Bring Scheduler Ingester Up To Date

d80tb7:
In the course of the writing the scheduler, some features have been implemented which needed work to be done on the scheduler ingester.  This PR address that. Specific features are:

* Scheduler ingester now persists terminal job run errors.
* Scheduler ingester no longer now handles job cancel requested messages
* Scheduler ingester no longer cancels job runs when on receipt of a job cancel message
* When a job run error indicates that the lease was returned, the ingester marks it as such in the db.

I've also added some indexes to the scheduler database which we will need for efficient access.  The only one that worries me is the indexes on `serial` which we have added to the `jobs` and `runs` table.  That's because we update the `serial` on *every* row insert so it's likely that we'll be constantly rebuilding this index.  Whether or not this is is an issue in practise we'll have to test.


internal/common/compress/list.go
```diff
@@ -36,3 +44,11 @@ func DecompressStringArray(input []byte, decompressor Decompressor) ([]string, e
 	err = dec.Decode(&data)
 	return data, err
 }
+
+func MustDecompressStringArray(input []byte, compressor Decompressor) []string {
```
severinson, line 48: Is called string but takes as input a `[]byte`.

internal/common/compress/list.go
```diff
@@ -20,6 +20,14 @@ func CompressStringArray(input []string, compressor Compressor) ([]byte, error)
 	return compressor.Compress(bs)
 }
 
+func MustCompressStringArray(input []string, compressor Compressor) []byte {
```
kannon92, line 23: Possibly add some tests to these functions?

internal/common/compress/list.go
severinson: I'm assuming we're sure this'll never panic :)

internal/common/ingest/testfixtures/event.go
```diff
@@ -8,6 +8,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 
+	"github.com/armadaproject/armada/internal/armada/configuration"
```
severinson, line 11: We may want to add a `_test` suffix to the file name to avoid these being built into the final binary.

internal/common/proto/protoutil_test.go
```diff
@@ -0,0 +1,65 @@
...
+var (
+	msg              = &armadaevents.CancelJob{JobId: armadaevents.ProtoUuidFromUuid(uuid.New())}
+	compressor       = compress.NewThreadSafeZlibCompressor(1024)
+	decompressor     = compress.NewThreadSafeZlibDecompressor()
+	marshalledMsg, _ = proto.Marshal(msg)
+	compressedMsg, _ = compressor.Compress(marshalledMsg)
+	invalidMsg       = []byte{0x3}
+)
+
+func TestUnmarshall_Valid(t *testing.T) {
```
kannon92, line 24: One test pattern I've started following is to use table driven tests.  

https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

internal/scheduleringester/config.go
```diff
@@ -13,6 +13,8 @@ type Configuration struct {
 	Metrics configuration.MetricsConfig
 	// General Pulsar configuration
 	Pulsar configuration.PulsarConfig
+	// Map of allowed priority classes to priority
```
kannon92, line 16: Not sure I really understand what this comment is trying to say.

internal/common/proto/protoutil.go
```diff
@@ -0,0 +1,74 @@
...
+	b, err := proto.Marshal(msg)
+	if err != nil {
+		return nil, errors.Wrap(err, "Error marshalling object")
+	}
+	return compressor.Compress(b)
+}
+
+// MustMarshall marshalls a proto message and panics if the unmarshall fails.  The main use case here is for unit tests.
+// Think carefully if you intend to use this elsewhere
+func MustMarshall(msg proto.Message) []byte {
```
severinson, line 57: Is this safe? Seems somewhat risky to panic on a marshalling error.

severinson (approved):

kannon92 (approved):
LGTM

internal/common/compress/list.go
d80tb7: let me break this out into a separate test file so it can only be used by test code.

internal/scheduleringester/config.go
d80tb7: I'll change it to `Map of allowed priority classes name to priority value`

internal/common/proto/protoutil_test.go
d80tb7: I do that but I'm not sure it's useful here as we only have a couple of tests

internal/common/compress/list.go
d80tb7: actually no I won't because if I do that you can't import it from any other module!  I think it's ok as `MustX` is pretty idiomatic in go for something that will panic and should be used with great caution

internal/common/compress/list.go
d80tb7: done

internal/common/proto/protoutil_test.go
kannon92: Yea I see that.  No worries then.

PR #2065, opened 2023-01-26
Add Pod Issue handling for jobs created using Executor API

JamesMurkin:
As part of the migration to using the Executor API, we now have 2 flows through the executor

The new flow had no handling of stuck pods, which is what this PR introduces

It was simpler to split this off as the handling is slightly different to existing code
 - Existing code is tied in with lease handling
 - The existing code has to make additional calls (return lease, report done) - where the new executor api is purely event based


internal/executor/service/pod_issue_handler.go
```diff
@@ -0,0 +1,348 @@
...
+	"github.com/armadaproject/armada/internal/executor/reporter"
+	"github.com/armadaproject/armada/internal/executor/util"
+	"github.com/armadaproject/armada/pkg/api"
+)
+
+type IssueType int
+
+const (
+	UnableToSchedule  IssueType = iota
+	StuckStartingUp   IssueType = iota
```
d80tb7, line 24: you can get rid of all the `iota`s after the first if you want

internal/executor/service/pod_issue_handler.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+func createIssues(managedPods []*v1.Pod, podIssues map[string]*podIssue) []*issue {
+	podsByRunId := make(map[string]*v1.Pod, len(managedPods))
+
+	for _, pod := range managedPods {
+		runId := util.ExtractJobRunId(pod)
+		if runId != "" {
+			podsByRunId[runId] = pod
+		}
```
d80tb7, line 195: what does it mean if `runId` is not present here?  Should we log something?

internal/executor/service/pod_issue_handler.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+
+	result := make([]*issue, 0, len(podIssues))
+
+	for _, podIssue := range podIssues {
+		relatedPod := podsByRunId[podIssue.RunId]
```
d80tb7, line 201: what happens if `podsByRunId` doesn't have the key?

internal/executor/service/pod_issue_handler.go
```diff
@@ -0,0 +1,348 @@
...
+
+	if issue.CurrentPodState != nil {
+		p.clusterContext.DeletePods([]*v1.Pod{issue.CurrentPodState})
+	} else {
+		p.markIssuesResolved(issue.Issue)
+	}
+}
+
+// For retryable issues we must:
+//   - Report JobUnableToScheduleEvent
```
d80tb7, line 261: I'm not sure the scheduler cares about `unableToScheduleEvent`.  What's this used for?

internal/executor/util/process.go
```diff
@@ -0,0 +1,37 @@
...
+	"sync"
+
+	commonUtil "github.com/armadaproject/armada/internal/common/util"
+)
+
+func ProcessItemsWithThreadPool[K any](ctx context.Context, maxThreadCount int, itemsToProcess []K, processFunc func(K)) {
+	wg := &sync.WaitGroup{}
+	processChannel := make(chan K)
+
+	for i := 0; i < commonUtil.Min(len(itemsToProcess), maxThreadCount); i++ {
```
d80tb7, line 14: pretty sure there's a `min` function in the standard library now

d80tb7 (approved):

internal/executor/service/pod_issue_handler.go
JamesMurkin: That is fine, it'll just be nil - which is often expected - such as if the pod was deleted and we are reporting it getting deleted unexpectedly

internal/executor/service/pod_issue_handler.go
JamesMurkin: It is informational

Really it should just be "StartUpIssueEvent"

We could remove it - and we probably should if we had better state management. 

It is sometimes helpful when debugging to see that the issue was detected and attempted to be handled
 - Especially as part of handling the issue is deleting the pod - which could mean the executor loses state during this process. (i.e executor deletes it and then shuts down before it can return lease)

internal/executor/util/process.go
JamesMurkin: Are you sure? I can't seem to find it - would you mind linking to anywhere we use it?

New go does allow writing a generic min function - so you're probably correct

internal/executor/service/pod_issue_handler.go
JamesMurkin: Done

internal/executor/util/process.go
d80tb7: ah I'm wrong- it doesn't yet exist.  Hopefully it will soon

d80tb7 (approved):

PR #2066, opened 2023-01-26
Populate RunIds on NodeInfo during lease call

JamesMurkin:
RunIds need to be populated so that the scheduler knows what pod is running on which node

### DIFF
--- a/internal/executor/util/pod_util_test.go
+++ b/internal/executor/util/pod_util_test.go
@@ -308,6 +308,60 @@ func makePodsWithJobIds(jobIds []string) []*v1.Pod {
 	return pods
 }
 
+func TestExtractJobRunIds(t *testing.T) {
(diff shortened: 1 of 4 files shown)


internal/executor/util/pod_util_test.go
```diff
@@ -308,6 +308,60 @@ func makePodsWithJobIds(jobIds []string) []*v1.Pod {
 	return pods
 }
 
+func TestExtractJobRunIds(t *testing.T) {
```
kannon92, line 311: I think these individual test cases could be refactored to be a table driven test.  

I find them nice because you write the boilerplate code once and then give multiple cases.

kannon92 (approved):
LGTM. One comment about table driven tests but it’s mostly a nit

d80tb7 (approved):

PR #2068, opened 2023-01-29
Ingest Partition Markers

d80tb7:
When trying to get the next scheduler e2e tests working, I immediately realised that I hadn't persisted the partition markers to the DB (doh!).  This PR addresses that.  Summary is:

- Partition Markers are now `EventMessageEvents` so that they can be handled by the `SchedulerIngester`.  
- Code In Scheduler Ingester to insert these into the db.
- Code in other event handlers to ensure these messages are ignored without triggering warnings etc.


internal/scheduleringester/schedulerdb.go
```diff
@@ -158,10 +158,19 @@ func (s *SchedulerDb) WriteDbOp(ctx context.Context, op DbOperation) error {
...
+		return database.Upsert(ctx, s.db, "job_run_errors", records)
+	case InsertPartitionMarker:
+		for _, marker := range o.markers {
+			err := queries.InsertMarker(ctx, schedulerdb.InsertMarkerParams{
+				GroupID:     marker.GroupID,
+				PartitionID: marker.PartitionID,
+				Created:     marker.Created,
+			})
+			if err != nil {
+				return errors.Wrapf(err, "error inseerting partition merker")
```
JamesMurkin, line 170:
```suggestion
				return errors.Wrapf(err, "error inserting partition merker")
```

PR #2077, opened 2023-02-01
Remove option to cancel  multiple jobns without supplying a queue and jobset

d80tb7:
Before Christmas we added a bulk cancel API for use by Lookout.  This supported two code paths for bulk cancelling jobs:

* User provides just a list of job ids but not a queue and jobset
* User provides a list of job ids and queue/jobset

The first of these is not very performant as all jobs have to be fetched in order to validate the user has perms to cancel them.  Lookout has now been migrated to use only the second api so we can delete the non performant code paths.

Although this is a public api- it was only added recently and lookout is (was) the only user so this should be safe to remove.

### DIFF
--- a/internal/armada/server/submit_to_log.go
+++ b/internal/armada/server/submit_to_log.go
@@ -214,7 +214,6 @@ func selectJobsForLegacyScheduler(jobs []*armadaevents.SubmitJob) []*armadaevent
 }
 
 func (srv *PulsarSubmitServer) CancelJobs(ctx context.Context, req *api.JobCancelRequest) (*api.CancellationResult, error) {
-
 	if len(req.JobIds) > 0 {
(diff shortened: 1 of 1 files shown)


carlocamurri (approved):

kannon92:
The python CI :(

should we remove the api call from submit.proto?

d80tb7:
> The python CI :(
> 
> should we remove the api call from submit.proto?

We haven't actually changed the api call signature.  There was a single api call that allowed users to specify a list of job ids, a queue and a jobset.  The previous code allowed queue and jobset to be nil, but the new code returns an error to the user if this is the case.

kannon92:
The only concern I have is that we do have this api in the python client so are we sure no python or c# person isn’t using it?

PR #2081, opened 2023-02-02
Preemption to fair share

severinson:


internal/common/slices/slices.go
```diff
@@ -1,9 +1,69 @@
...
+// such that each such slice contains at most maxLen elements.
+func PartitionToMaxLen[S ~[]E, E any](s S, maxLen int) []S {
+	n := int(math.Ceil(float64(len(s)) / float64(maxLen)))
+	if n == 0 {
+		n = 1
+	}
+	return Partition(s, n)
+}
+
+// Partition partitions the elements of s into n non-overlapping slices,
```
d80tb7, line 28: Do we need the behaviour that some slices are len/n and some len/n+1?  The more common behaviour is to make a slices length 1/n and the last slice have between 1 and 1/n elements.  That has the benefit that the implementation is simpler.

internal/scheduler/legacyscheduler.go
```diff
@@ -791,6 +954,204 @@ func NewLegacyScheduler(
...
+	ctx context.Context,
+	jobRepo JobRepository,
+	constraints SchedulingConstraints,
+	config configuration.SchedulingConfig,
+	nodeDb *NodeDb,
+	priorityFactorByQueue map[string]float64,
+	initialResourcesByQueueAndPriority map[string]schedulerobjects.QuantityByPriorityAndResourceType,
+	nodeFairShareEvictionProbability float64,
+	nodeOversubscriptionEvictionProbability float64,
+) ([]LegacySchedulerJob, []LegacySchedulerJob, map[string]*schedulerobjects.Node, error) {
```
d80tb7, line 967: at the point we're returning a tuple of 4 objects I'd be tempted to make a wrapper struct.
I'd also put some doc on the fucntion to say what it does

internal/scheduler/nodedb.go
```diff
@@ -383,45 +533,72 @@ func (nodeDb *NodeDb) dominantResource(req *schedulerobjects.PodRequirements) st
...
-			return errors.WithStack(err)
-		}
+	// Add the node to the db.
+	if err := txn.Insert("nodes", node); err != nil {
+		return errors.WithStack(err)
 	}
+
+	// TODO: Doing this at every node insert seems wasteful.
+	nodeDb.mu.Lock()
+	nodeDb.timeOfMostRecentUpsert = time.Now()
```
d80tb7, line 599: why do we need this- it used to be used to check liveness for the submit check but I've removed the need for that

internal/scheduler/nodeiteration.go
```diff
@@ -172,44 +267,78 @@ func (it *NodeTypeResourceIterator) NextNodeItem() *schedulerobjects.Node {
...
+	node, ok := obj.(*schedulerobjects.Node)
 	if !ok {
 		panic(fmt.Sprintf("expected *NodeItem, but got %T", obj))
 	}
-	if nodeItem.NodeTypeId != it.nodeType.Id {
-		// The index is sorted by NodeType first.
-		// So we've seen all nodes of this NodeType if this comparison fails.
+	// If it.dominantQueue != *, the index is sorted by node.DominantQueue() first.
+	// Otherwise, the index is sorted by NodeTypeId first.
+	if it.dominantQueue != "*" {
```
d80tb7, line 276: I'd move * into some constant so that it's obvious what it is in all contexts

internal/scheduler/scheduling_algo.go
```diff
@@ -162,7 +162,7 @@ func (l *LegacySchedulingAlgo) scheduleOnExecutor(
 ) ([]*SchedulerJob, error) {
 	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
 	defer cancel()
-	nodeDb, err := l.constructNodeDb(executor.Nodes, l.priorityClassPriorities)
+	nodeDb, err := l.constructNodeDb(executor.Nodes, l.config.Preemption.PriorityClasses)
```
d80tb7, line 165: there are lots of changes to `lease.go` but hardly any changes here?  Is that because this doesn't need changes or is it because they still need to be made?

internal/common/slices/slices.go
severinson: When used for load-balancing this would be slightly better. But it may be an unnecessary optimisation. We don't need these functions at all anymore and could remove them entirely.

internal/scheduler/nodedb.go
severinson: In that case we don't need it anymore. Let me remove it.

internal/scheduler/nodeiteration.go
severinson: Fixed

internal/scheduler/scheduling_algo.go
severinson: I haven't made `scheduling_algo.go` work with preemption-to-fair-share yet. I think it may be best to do that in a follow-up PR. The existing scheduling (without preemption-to-fair-share) should "just work" though.

internal/scheduler/legacyscheduler.go
severinson: Agreed. Something for a later PR I think. I'd like to rework this whole file a bit.

d80tb7 (approved):

PR #2084, opened 2023-02-03
Lookout v2 DB Pruner

carlocamurri:
Pruning logic (CronJob), based on scheduler DB pruner, acting similar to Lookout v1 pruner


internal/common/util/list.go
```diff
@@ -45,3 +45,16 @@ func Filter[T any](list []T, predicate func(val T) bool) []T {
 	}
 	return out
 }
+
+func Concat[T any](slices [][]T) []T {
```
d80tb7, line 49: I think there's a case for this to be `Concat[T any](slices ...[]T) []T`  I think the current function sig is a bit more like a `flatmap` (albeit without the `map`)!

internal/lookoutv2/pruner/pruner_test.go
```diff
@@ -0,0 +1,168 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.testName, func(t *testing.T) {
+			err := lookout.WithLookoutDb(func(db *pgxpool.Pool) error {
+				converter := instructions.NewInstructionConverter(metrics.Get(), "armadaproject.io/", &compress.NoOpCompressor{})
+				store := lookoutdb.NewLookoutDb(db, metrics.Get(), 3, 10)
+
+				ctx := context.TODO()
```
d80tb7, line 114: I reckon put a timeout here- that way if  the unit test ends up triggering some crazy database loop won't hang.

d80tb7 (approved):

PR #2085, opened 2023-02-04
Api To Direct Jobs To Pulsar Scheduler

d80tb7:
Changes to allow Legacy (Redis) and New (Pulsar) schedulers to run in parallel.  Basic strategy is as follows:

* Pulsar messages are now tagged with an attribute, `schedulerName` that indicates whether the message is for the legacy scheduler, the new scheduler or both. Compoenents folter out messages they are not interested in (e.g. the scheduler ingester doesn't care about events for the legacy scheduler so can just ignore them).

* The Submit API (`submit_to_log.go`) has been updated so that on submission, a job is allocated to either the new scheduler or the old scheduler.  This is done by first checking if the job is schedulable on either scheduler.  If the job is schedulable only on one of the schedulers, it is assigned to that one, if it is schedulable on both then it is assigned to the puslar scheduler at a probability defined in config (default is 0%).

* For cancelling and reprioritising  jobs we have various code paths that require the queue and jobset to be retrieved from redis.  For Jobs assigned to the Legacy scheduler this is not an issue, because all jobs are stored in redis and therefore can be retived by the api when the cancel request is retrieved.  For jobs assigned to the Pulsar scheduler, this is an issue as the info was not previously in redis and We do not want the api hitting the scheduler (or it's db)  to retrieve this.  For now I've solved this by storing a minimal Job in Redis which the api can retrieve.

* Both the legacy scheduler and the pulsar scheduler store `schedulerobjects.Executor`s in redis when they get lease calls.  By  storing this in redis this means we can reconstruct the nodedb for both legacy and pulsar scheduler in the api and therefore validate that Jobs can fit on any node at submission time.  Note that this means I have *disabled the old "does the job fit"* code path and now we only use Albin's new code path that utilises JobDb.  The other benefit of storing this data in Redis is that it means that both the Pulsar scheduler and the Legacy Scheduler can see each other's utilisation and thus calculate fair share properly.  Note that I haven't actually added the code to do this yet though.

Note that the use of redis for storing the data above (minimal Pulsar jobs + executor info) is hopefully temporary as for as long as we do this we cannot remove the redis dependency from the api.  I think we can takle this after we have switrched schedulers, however.


cmd/eventsprinter/logic/logic.go
```diff
@@ -33,11 +33,6 @@ func PrintEvents(url, topic, subscription string, verbose bool) error {
 			}
 			consumer.Ack(msg)
 
-			// We're only interested in control messages.
```
d80tb7, line 36: I've removed the concept of "control messages" as it dates from a time when we were going to keep utilisation messages separate.  As we no longer do this, it's not being used for anything.

e2e/pulsar_test/pulsar_test.go
```diff
@@ -324,7 +324,7 @@ func TestSubmitCancelJobSet(t *testing.T) {
 		}
 
 		// Test that we get submit, cancel job set, and cancelled messages.
-		numEventsExpected := numJobs + 1 + numJobs
+		numEventsExpected := numJobs + numJobs + numJobs
```
d80tb7, line 327: So this is a change in the api.  Previously we had two ways of cancelling a jobset via the api:
- Call CancelJob with an empty jobId
- Call CancelJobset
The two went down different code paths with the first sending a CacnelJobset message and the second sending n cancel jobs messages (one for each job in the jobset).  I've unified the codepaths so now we just do the latter.

internal/armada/repository/job.go
```diff
@@ -80,6 +83,9 @@ type JobRepository interface {
 	GetQueueActiveJobSets(queue string) ([]*api.JobSetInfo, error)
 	AddRetryAttempt(jobId string) error
 	GetNumberOfRetryAttempts(jobId string) (int, error)
+	StorePulsarSchedulerJobDetails(jobDetails []*schedulerobjects.PulsarSchedulerJobDetails) error
```
d80tb7, line 86: These are the new functions for storing a minimal set of information in redis (just queue and jobset) for each job assigned to the pulsar scheduler

internal/armada/server/reporting.go
```diff
@@ -60,22 +59,6 @@ func reportDuplicateDetected(repository repository.EventStore, results []*reposi
 	return nil
 }
 
-func reportDuplicateFoundEvents(repository repository.EventStore, events []*api.JobDuplicateFoundEvent) error {
```
d80tb7, line 63: we don't need to do this any more because we just generate the duplicate message at submit time and put it into the event sequence

internal/scheduler/submitcheck.go
```diff
@@ -1,71 +1,122 @@
...
+	for i, job := range jobs {
+		reqs := PodRequirementsFromLegacySchedulerJob(job, srv.priorityClasses)
+		canSchedule, reason := srv.check([]*schedulerobjects.PodRequirements{reqs})
+		if !canSchedule {
+			return canSchedule, fmt.Sprintf("%d-th job unschedulable:\n%s", i, reason)
+		}
+	}
 	// Then, check if all gangs can be scheduled.
-	for gangId, jobs := range groupJobsByAnnotation(srv.gangIdAnnotation, jobs) {
+	for gangId, jobs := range GroupJobsByAnnotation(srv.gangIdAnnotation, jobs) {
```
severinson, line 106: You could use `GroupJobsByAnnotation` to group by gang id in `submit_to_log.go`. Avoid duplicating some code.

internal/scheduler/submitcheck.go
```diff
@@ -79,29 +130,30 @@ func groupJobsByAnnotation(annotation string, jobs []*api.Job) map[string][]*api
...
-	if len(nodeDbByExecutor) == 0 {
+	executorById = srv.filterStaleNodeDbs(executorById)
+	if len(executorById) == 0 {
 		return false, "no executor clusters available"
 	}
 
 	canSchedule := false
 	var sb strings.Builder
-	for executor, nodeDb := range nodeDbByExecutor {
+	for id, executor := range executorById {
```
severinson, line 150: We can return early if scheduling on some cluster succeeds.

internal/scheduler/submitcheck_test.go
```diff
@@ -0,0 +1,175 @@
...
+							"cpu": resource.MustParse("1"),
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func testNJobGang(n int) []*api.Job {
```
severinson, line 141: Move to `testfixtures_test.go` please.

severinson: Even better, write tests to use existing test fixtures instead of writing new ones.

internal/scheduleringester/ingester.go
```diff
@@ -3,6 +3,8 @@ package scheduleringester
 import (
 	"time"
 
+	"github.com/armadaproject/armada/internal/common/schedulers"
```
severinson, line 6: Import ordering.

severinson (approved):

internal/scheduler/submitcheck.go
d80tb7: this function is different- the function there groups all jobs without an annotation into a single "gang" with no id which subsequent code in that file knows to ignore.  The function here groups all jobs without an annotation into a gang of 1.

config/armada/config.yaml
```diff
@@ -6,6 +6,9 @@ corsAllowedOrigins:
   - http://localhost:8089
   - http://localhost:10000
 cancelJobsBatchSize: 1000
+pulsarSchedulerEnabled: false
+probabilityOfUsingPulsarScheduler: 0
+ignoreJobSubmitChecks: true
```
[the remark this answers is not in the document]
d80tb7: I originally called it `submitJobSubmitChecks` but changed it to `ignore` because that's closer to what this actually does- the check is still performed it just ignores the result.  I wouldn't worry about this too much- it's only staying in for one release, once we have migrated everything over to the new check mechanism (and thus redis is populated) we can remove this flag.

internal/scheduler/submitcheck_test.go
d80tb7: These do the use the test fixtures- `test1CoreCpuJob()`, `testNodeItems1()` , `testGangIdAnnotation` etc, they just add extra information as needed for these tests.  I guess we could add helper methods to testFixtures to cover these cases, but I'm not convinced they're sufficiently generic yet.

PR #2088, opened 2023-02-06
Executor fixes to handle new Executor API

JamesMurkin:
Fixes for a variety of issues found during integration with the executor api


internal/executor/reporter/event_sender.go
```diff
@@ -43,23 +45,93 @@ func (eventSender *ExecutorApiEventSender) SendEvents(events []EventMessage) err
...
+		_, err = eventSender.eventClient.ReportEvents(context.Background(), eventList)
+		if err != nil {
+			return err
+		}
+	}
+
 	return err
 }
 
+func splitIntoEventListWithByteLimit(sequences []*armadaevents.EventSequence, maxEventListSizeBytes int) ([]*executorapi.EventList, error) {
```
d80tb7, line 75: this function seems a bit weird- we limit the sequences by size by then limit them again when we put them into the event list?
I'm guessing you do this because limitSequencesByByteSize won't account for the overhead of the EventList wrapper.  If it were me I'd just limit the sequences to 2Mb (half of the 4MB grpc limit) and assume that the overhead won't add another 2MB.

Long term we should write a generic function that can take a slice of proto objects and split them appropriately, then we can use the same function for EventSequence and EventList

internal/executor/reporter/event_sender.go
```diff
@@ -43,23 +45,93 @@ func (eventSender *ExecutorApiEventSender) SendEvents(events []EventMessage) err
...
+
+func filterJobRunEvents(events []*armadaevents.EventSequence_Event) []*armadaevents.EventSequence_Event {
+	result := make([]*armadaevents.EventSequence_Event, 0, len(events))
+	for _, event := range events {
+		switch typed := event.GetEvent().(type) {
+		case *armadaevents.EventSequence_Event_JobSucceeded:
+			continue
+		case *armadaevents.EventSequence_Event_JobErrors:
+			continue
+		case *armadaevents.EventSequence_Event_JobRunAssigned:
```
d80tb7, line 114: you could just comma separate these and do the append once

d80tb7 (dismissed):

d80tb7 (approved):

PR #2105, opened 2023-02-13
feat: added extraInitCommand to lookout migration jobs.

valkenburg-prevue-ch:
When deploying on a cluster with istio injection enabled, the istio sidecar does not detect that a job has finished, and keeps the pod alive and unready forever. Running with 
```yaml
extraInitCommand: "curl -fsI -X POST http://localhost:15020/quitquitquit ;"
``` 
the job completes cleanly.


kannon92 (commented):
Thank you for the PR. Are you evaluating Armada for any potential purpose?

suprjinx:
It might be better to separate this into one or more InitContainers

valkenburg-prevue-ch:
@suprjinx Thanks for the suggestion and the edit. Indeed, this runs after the job. Before I saw your edit, I looked up these two examples, which suggest that this method is common practice (if it is not the only solution):
https://github.com/minio/minio/blob/master/helm/minio/templates/post-job.yaml#L128
https://github.com/apache/pulsar-helm-chart/blob/master/charts/pulsar/templates/pulsar-cluster-initialize.yaml#L113

Either way, this PR is just a suggestion, I figured it's faster to just submit the PR so  the code can be discussed, rather than creating an issue and spending time on explaining in words what is clearer in code.

@kannon92 Thanks for asking, on of your colleagues reached out to me on LinkedIn, I'll try to have a call with them to exchange ideas. The brief answer is that I plan to use Armada in the way it is intended to use, for large ML parameter searches.

kannon92:
> 

Yep! I’m not as familiar with Istio so I wonder if there are other issues if you try to use it with Armada.

And great! Please don’t be a stranger on your experience using Armada. We are on the Armada channel under CNCF if you have any questions.

c-rindi:
Here's the link to the CNCF Slack #armada channel - https://cloud-native.slack.com/archives/C03T9CBCEMC

valkenburg-prevue-ch:
Just to report on a related topic: pulsar can run fine in an istio mesh, and the easiest way is to move to the streamnative pulsar operator for deploying a pulsar cluster. See [this discussion](https://github.com/apache/pulsar/discussions/19703#discussioncomment-5250906).

kannon92:
> Just to report on a related topic: pulsar can run fine in an istio mesh, and the easiest way is to move to the streamnative pulsar operator for deploying a pulsar cluster. See [this discussion](https://github.com/apache/pulsar/discussions/19703#discussioncomment-5250906).

Interesting.  Thank you for following up with the pulsar community.  Would you mind creating an issue on the armada board to suggest a migration to the streamnative pulsar operator?

PR #2111, opened 2023-02-14
Quickly populate lookout db

### LINKED ISSUE #1742: Create tool to quickly populate Lookout database with sample for API/UI testing
Write a tool that would quickly populate a set of data in the Lookout db, so that the API and UI can be tested without the need for creating events and waiting for them to be ingested.

Mo-Fatah:
#### What type of PR is this?
Feature
This Go tool directly inserts jobs into the Lookout database and simulates each job's state change ( from queued -> pending -> running, after that some of them randomly being failed, cancelled or succeeded ) without the need of submitting real jobs into armada which requires the whole system to be up and running. 
This will be helpful in UI/API testing by simulating any desired number of submitted jobs.

Fixes #1742


internal/lookout/db-gen/main.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+}
+
+/* Some helper functions */
+func setupDB() *sql.DB {
+	host := flag.String("host", "localhost", "database host")
+	user := flag.String("user", "postgres", "database user")
+	password := flag.String("password", "psw", "database password")
+	flag.Parse()
+	connectionString := fmt.Sprintf("host=%s port=5432 user=%s password=%s sslmode=disable", *host, *user, *password)
```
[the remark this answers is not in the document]
Mo-Fatah: done

suprjinx:
> Assuming you have a postgres db can you write some test cases that verify this functionality?

An alternative might be to write some pure unit tests which setup a sqlite db

Mo-Fatah:
> I think we would like to see some integration tests with the CLI.
> 
> Assuming you have a postgres db can you write some test cases that verify this functionality?
@kannon92 I am confused a little bit, can you elaborate which tests you are talking about here?

kannon92:
I’d like a test case that asserts that this tool works. Integration test where you insert data into Postgres and verify its there

Mo-Fatah:
@kannon92 I feel it is a bit messy, also there are many assumption with the test, it assumes postgres is running on localhost for example. what do you think ?

kannon92:
> @kannon92 I feel it is a bit messy, also there are many assumption with the test, it assumes postgres is running on localhost for example. what do you think ?

Its coming along. I think this won't work if you tried to run this in the CI.  You are assuming that postgres must be started so if I tried to run this test right now it would fail unless I started postgres. 

It would be ideal to have a make target that runs this test.  

Ideally I'd like to do something like `make populate-lookout-test` and that would start postgres and then run your test cases.  And then the test would teardown postgres.

Mo-Fatah:
I think that preparing the postgres db will require also to setup the schema as well. I was thinking, wouldn't that be an overkill a bit ? after all, the code itself is a test utility to test out the UI.

kannon92:
> I think that preparing the postgres db will require also to setup the schema as well. I was thinking, wouldn't that be an overkill a bit ? after all, the code itself is a test utility to test out the UI.

I found that the best approach in software engineering is to always think of tests in your definition of done.  How can you guarantee that this code will always work?  The point of tests is just make sure that your code is actually working as intended.  And you want the tests because you want to be notified that this code no longer works.

Mo-Fatah:
> > I think that preparing the postgres db will require also to setup the schema as well. I was thinking, wouldn't that be an overkill a bit ? after all, the code itself is a test utility to test out the UI.
> 
> I found that the best approach in software engineering is to always think of tests in your definition of done. How can you guarantee that this code will always work? The point of tests is just make sure that your code is actually working as intended. And you want the tests because you want to be notified that this code no longer works.

makes sense, Thank you! 
did you mean at first the root Makefile or create a Makefile under `db-gen`

kannon92:
> > > I think that preparing the postgres db will require also to setup the schema as well. I was thinking, wouldn't that be an overkill a bit ? after all, the code itself is a test utility to test out the UI.
> > 
> > 
> > I found that the best approach in software engineering is to always think of tests in your definition of done. How can you guarantee that this code will always work? The point of tests is just make sure that your code is actually working as intended. And you want the tests because you want to be notified that this code no longer works.
> 
> makes sense, Thank you! did you mean at first the root Makefile or create a Makefile under `db-gen`

Makefile at the root.

Mo-Fatah:
well, the current make target depends on the `localdev/docker-compose` , I tried to run the `lookout` `lookout-ingester` containers manually but I couldn't figure out how. I just pushed the code so I can get some help.

richscott:
> well, the current make target depends on the `localdev/docker-compose` , I tried to run the `lookout` `lookout-ingester` containers manually but I couldn't figure out how. I just pushed the code so I can get some help.

I would just add a new target that runs Postgres in Docker, much like the `tests` target in the Makefile, where it does:
```
docker run -d --name=postgres $(DOCKER_NET) -p 5432:5432 -e POSTGRES_PASSWORD=psw postgres:14.2
  sleep 3
  function tearDown { docker rm -f redis postgres; }; trap tearDown EXIT
  ```
You could add something similar(except leave out the Redis mentions) in your Makefile target, then run your unit tests that need Postgres.

**Comment by Mo-Fatah** [comment]
Yes, but the problem I faced isn't with Postgres container, the problem is with the database migration so the container becomes ready for direct insertions, otherwise it will be just a postgres container running without any tables. I couldn't trace back which service in the docker-compose file is responsible for the database migration.

**Comment by richscott** [comment]
> Yes, but the problem I faced isn't with Postgres container, the problem is with the database migration so the container becomes ready for direct insertions, otherwise it will be just a postgres container running without any tables. I couldn't trace back which service in the docker-compose file is responsible for the database migration and how to build it manually

Aha - yeah. Ugh. I really really really really wish our database migrations were a completely separate job from Lookout service startup.  I will look at the code and see if there's an non-messy way to run the migration logic separately.

**Comment by Mo-Fatah** [comment]
wow, first time to do git squash, it is so satisfying

File: internal/lookout/db-gen/db_gen_test.go [hunk 2]
```diff
@@ -0,0 +1,64 @@
+package main
+
+import (
+	"database/sql"
+	"os/exec"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/armadaproject/armada/internal/lookout/repository/schema"
```
kannon92, line 10: This formatting looks off.  Can you indent import and remove a line?

Mo-Fatah: The linting check failed in the CI before. I ran `golangci-lint run --fix` and it introduced this new line.

internal/lookout/db-gen/main.go
```diff
@@ -0,0 +1,231 @@
...
+	case 0:
+		fail(job_id, db)
+	case 1:
+		cancel(job_id, db)
+	default:
+		success(job_id, db)
+	}
+}
+
+func pending(job_id *string, db *sql.DB) {
```
kannon92, line 120: I think you should combine these functions into a single function..

func updateAndSetByState(state int, job_id *string, db *sql.DB) {
...
	updateState := `UPDATE job SET state = $1, job_updated = $2 WHERE job_id = $3;`
	_, err := db.Exec(updateState, state, updateTime, job_id)
}

and then call `updateAndSetByState(PENDING, ..)

Most of these functions only differ by State so I'd be nice to combine these.

Mo-Fatah: yes, the query for different differs slightly. I refactored it now.

internal/lookout/db-gen/db_gen_test.go
Mo-Fatah: The linter separates local imports from third party imports I believe. like [here](https://github.com/armadaproject/armada/blob/5e5f55c65f0e99faf3b8d319de849c480652ca56/internal/armadactl/queue.go#L9)

kannon92: It looks good now. It was the indenting but maybe it was a big in UI

kannon92 (approved):
Nice work!!

PR #2117, opened 2023-02-14
[Pulsar Scheduler] Remember Scheduled Jobs

d80tb7:
Review this after #2116 is merged as it also has those changes in it!

This change introduces a few changes:

* Pulsar scheduler now remembers which jobs it has scheduled between executor updates so that it doesn't overschedule.
* The Node the job is running on is now persisted in the scheduler db (this is needed for the above)
* `non_armada_allocated_resources` is now present on `SchedulerObjects.Node` and is populated from `api.NodeInfo`

Things that aren't done as part of this PR but which will be done as subsequent PRs:
* Pulsar scheduler needs to discard cluster if the number of jobs/amount of resource which has been scheduled but hasn't yet been reported by the executor exceeds some config limit.
* `non_armada_allocated_resources`on SchedulerObjects.Node` isn't actually used by the scheduler.


internal/scheduler/scheduling_algo.go
```diff
@@ -202,13 +220,46 @@ func (l *LegacySchedulingAlgo) scheduleOnExecutor(
 	updatedJobs := make([]*jobdb.Job, len(jobs))
 	for i, report := range legacyScheduler.SchedulingRoundReport.SuccessfulJobSchedulingReports() {
 		job := report.Job.(*jobdb.Job)
-		job = job.WithQueued(false).WithNewRun(executor.Id)
+		nodeName := ""
+		if len(report.PodSchedulingReports) > 0 {
+			nodeName = report.PodSchedulingReports[0].Node.Name
+		} else {
+			log.Warnf("Could not resolve node for Job %s as no PodSchedulingReports were present", job.Id())
```
JamesMurkin, line 227: Should we still lease the job here? Instead of just skip the job

My concern is we'll be leasing it with nodeName as empty
 - Does this not cause issues?

internal/scheduler/scheduling_algo.go
```diff
@@ -202,13 +220,46 @@ func (l *LegacySchedulingAlgo) scheduleOnExecutor(
...
 	return updatedJobs, nil
 }
 
-func (l *LegacySchedulingAlgo) constructNodeDb(nodes []*schedulerobjects.Node, priorityClasses map[string]configuration.PriorityClass) (*NodeDb, error) {
+// constructNodeDb constructs a node db with all jobs bound to it
+func (l *LegacySchedulingAlgo) constructNodeDb(nodes []*schedulerobjects.Node, jobs []*jobdb.Job, priorityClasses map[string]configuration.PriorityClass) (*NodeDb, error) {
+	nodesByName := make(map[string]*schedulerobjects.Node, len(nodes))
+	for _, node := range nodes {
+		// Clear out node
+		node.AllocatableByPriorityAndResource = schedulerobjects.NewAllocatableByPriorityAndResourceType(
```
JamesMurkin, line 240: When is nonArmadaAllocatedResources actually used, should it be used here?

JamesMurkin: Nevermind PR comment said it isn't used yet

JamesMurkin (approved):

JamesMurkin (approved):

internal/scheduler/scheduling_algo.go
d80tb7: I think the right thing to do is to return an error here, which will have the effect of cancelling the entire leasing cycle. I think this is correct because:
- If the legacyScheduler has "leased" the job but hasn't assigned a node then something has gone very wrong and the rest of the leases probably can't be trusted.
- Simply not leasing that job might cause issues- e.g. if that job was part of a gang.

I think we basically need to confirm that we can resolve a valid `nodeName`  and return error if we can't.  This could be because `report.PodSchedulingReports` is empty (already checked) or because `report.PodSchedulingReports` is valid but `report.PodSchedulingReports[0].Node.Name == ""` (not currently checked).

Issue #2118, opened 2023-02-14
Integration tests are consistency failing


kannon92:
**Describe the bug**
Integration tests are failing.

**Files Or Markdown that can reproduce the issue**

**Expected behavior**
Integration tests pass.

**Actual behavior**

**Additional context**

PR #2119, opened 2023-02-14
WIP: Fix integration

kannon92:
I noticed some weirdness with running server and executor in same docker-compose.  Moving server to after executor seems to allow the queue creation to work.  

Our tests were failing to create queues which means that the server wasn't actually up.


kannon92:
Fixes #2118

suprjinx (approved):

PR #2125, opened 2023-02-15
cancel test

### LINKED ISSUE #2124: Testsuite cancel jobs are consistently failing
@JamesMurkin did a bit of research into why the cancel jobs are failing.  

So it isn't totally obvious
Locally if you just run that test on its own, it works 100% of the time
If you run the whole suite, it regularly doesn't work
Seems to be some race when they are all submitted
the 1x1 cancelling failure
I think the cancel happens before the job is even processed by pulsar
The 1x1 jobset cancel
Probably the same issue
Probably putting a 5s sleep (or even better, watch the stream to check the jobs exist) before cancelling would fix it

This ticket should fix these tests so our e2e tests can pass in the CI.

kannon92:
Fixes #2124 

This could be a fast fix so we can enable tests.  Long term I think there is a better discussion around how pulsar handles this.


severinson (approved):

suprjinx (commented):
can we get away with less time? +9secs is not trivial

PR #2129, opened 2023-02-16
Set executor timeout in a single place

JamesMurkin:
We currently have executor timeout configured in 2 places. Only one of which get set in config - making places that use the other location always consider every executor as timed out

Reducing this to a single place and making all places use this location

### DIFF
--- a/config/scheduler/config.yaml
+++ b/config/scheduler/config.yaml
@@ -1,5 +1,4 @@
 cyclePeriod: 10s
-executorTimeout: 1h
 databaseFetchSize: 1000
 pulsarSendTimeout: 5s
 internedStringsCacheSize: 100000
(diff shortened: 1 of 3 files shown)


carlocamurri (approved):

d80tb7:
So I think the original intention here was that there would be two separate timeouts- although we shouldn't call them both `executorTimeout`.

* A timeout to determine when we should stop scheduling jobs on an executor (and other related things like considering the executor for the submit check)
* A timeout to determine when we should consider an executor as lost and we should cancel all jobs on an executor.

I think we need 2 timeouts because the the first should be relatively short (e.g. 10 mins) while the second should be long (e.g. 1 hour)

JamesMurkin:
That is fair, I'll re-raise with the minimum change (Actually setting scheduling.executorTimeout + possibly renaming)

PR #2153, opened 2023-02-21
[EventIngester] Fix Issue With Retrying On Redis Down

d80tb7:
When Ingesting into Redis fails the event ingester will retry.  Unfortunately this wasn't working in some cases because we reuse the redis pipe.  If the pipe had been cleared by the command, the next time the retry was attempted we would be doing a no-op and so no data would be inserted.

This change makes it so that every retry creates a new pipe.

### DIFF
--- a/internal/eventingester/store/eventstore.go
+++ b/internal/eventingester/store/eventstore.go
@@ -87,24 +87,25 @@ func (repo *RedisEventStore) doStore(update []*model.Event) error {
 		uniqueJobSets[key] = true
 	}
 
-	pipe := repo.db.Pipeline()
-	for _, e := range data {
-		pipe.XAdd(&redis.XAddArgs{
-			Stream: e.key,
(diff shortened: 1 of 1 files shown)


internal/eventingester/store/eventstore.go
```diff
@@ -87,24 +87,25 @@ func (repo *RedisEventStore) doStore(update []*model.Event) error {
...
-	pipe := repo.db.Pipeline()
-	for _, e := range data {
-		pipe.XAdd(&redis.XAddArgs{
-			Stream: e.key,
-			Values: map[string]interface{}{
-				dataKey: e.data,
-			},
-		})
-	}
+	return ingest.WithRetry(func() (bool, error) {
```
kannon92, line 90: Is there anyone way to write a test case that verifies this functionality?

JamesMurkin (approved):

PR #2157, opened 2023-02-22
feat: add .editorconfig

dejanzele:
Add .editorconfig


.editorconfig
```diff
@@ -0,0 +1,27 @@
...
+trim_trailing_whitespace = true
+indent_style             = space
+indent_size              = 2
+
+[*.{yml, yaml}]
+indent_style = space
+indent_size  = 2
+
+[{Makefile,go.mod,go.sum,*.go,.gitmodules}]
+indent_style = tab
```
ClifHouck, line 17: Tab makes sense for files that require it, but I'm against it otherwise. Spaces always render consistently across editors with monospace fonts, even ones that don't support `.editorconfig`. 

For `*.go` files I'd especially prefer them to be space indented.

.editorconfig
```diff
@@ -0,0 +1,27 @@
+; https://editorconfig.org/
+
+root = true
+
+[*]
+insert_final_newline     = true
+charset                  = utf-8
+trim_trailing_whitespace = true
+indent_style             = space
+indent_size              = 2
```
ClifHouck, line 10: I'd think 4 is a better value for `indent_size`. Not too much and not too little. Easier to distinguish different indent levels.

ClifHouck (changes requested):
I'm not going to gate-keep this if many other contributors believe `tabs` are a better indent style for `*.go` files, but I'm `-1`ing for now because I'm strongly against it.

.editorconfig
suprjinx: gofmt apparently will turn them back to tabs anyway!

ClifHouck: TIL, oops.

kannon92 (approved):

PR #2159, opened 2023-02-22
Add cache to submitchecks

JamesMurkin:
On submission we check if jobs are schedulable on any known node
 - If they are not, submission is rejected

Often submissions contain many (100-500) jobs with the same scheduling requirements

So now instead of checking if each are schedulable individually we:
 - Hash the job scheduling requirements
 - Cache the result for scheduling a scheduling requirement
 - Use the cache value when it is present

I have also made it so GangIdAnnotation and GangCardinalityAnnotation are no longer configurable and defined as constants
 - These likely should never be changed after initial installation - so being constant makes more sense
 - It means we don't need to pass the config around everywhere - and helps us filter just those annotations in PodRequirementFromLegacySchedulerJob
 - I have done the minimal change for now - and some structs still take these constants as arguments

### DIFF
--- a/internal/common/util/hash.go
+++ b/internal/common/util/hash.go
@@ -79,29 +79,28 @@ const (
 //
 // Notes on the value:
 //
-//   * Unexported fields on structs are ignored and do not affect the
+//   - Unexported fields on structs are ignored and do not affect the
 //     hash value.
 //
-//   * Adding an exported field to a struct with the zero value will change
+//   - Adding an exported field to a struct with the zero value will change
 //     the hash value.
 //
 // For structs, the hashing can be controlled using tags. For example:
 //
-//    struct {
-//        Name string
-//        UUID string `hash:"ignore"`
-//    }
+//	struct {
+//	    Name string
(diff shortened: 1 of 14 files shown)


severinson (approved):

severinson (approved):

internal/common/util/hash.go
```diff
@@ -0,0 +1,502 @@
+// This code is largely a copy from https://github.com/mitchellh/hashstructure/blob/master/hashstructure.go
```
kannon92, line 1: Why can't we just reuse this function?  Esp since you are pulling it in for go.mod and go.sum?

JamesMurkin: Reuse which function? 

We had to modify the Hash function to special case resource.Quantity as there is no way to extend it
 - Without it - hashing doesn't work for resource.Quantity 

We plan to ask the library if they can allow extension in some way - so we don't have to copy their code to make a small tweak

kannon92: I see.

PR #2169, opened 2023-02-23
test: add domain_test

Mo-Fatah:
#### What type of PR is this?
test
add test cases for `internal/armada/metrics/domain.go` for improving test coverage
The test cases are straightforward as the logic of `domain.go` itself, if you think of some corner cases to test I would love to add them.


internal/armada/metrics/domain_test.go
```diff
@@ -0,0 +1,95 @@
...
+	armadaresource "github.com/armadaproject/armada/internal/common/resource"
+)
+
+func TestRecordJobRuntime_ShouldReportRuntimeDurations(t *testing.T) {
+	jobMetricsRecorder := NewJobMetricsRecorder()
+	jobMetricsRecorder.RecordJobRuntime("cpu", "", time.Second)
+	jobMetricsRecorder.RecordJobRuntime("cpu", "", 2*time.Second)
+	jobMetricsRecorder.RecordJobRuntime("cpu", "", 3*time.Second)
+
+	jobMetricsRecorder.RecordJobRuntime("gpu", "", 2*time.Second)
```
kannon92, line 18: I don't understand the labels "cpu" with duration.  It seems weird to me..

Mo-Fatah: yes they can serve as arbitrary pool tag names, but I think when I wrote them I misunderstood the context here. 

> I wonder if this test should also verify with having a couple different priority classes for the same pool?

I believe it will create just another `QueueMetricsRecorder`, for example if we recorded `pool1` with priority classes `1` and `2`, we will have 2 `QueueMetricsRecorder` with different `Metrics`
https://github.com/armadaproject/armada/blob/371fd974cb02ae70fb5c4a54cdedfe8ed487b780/internal/armada/metrics/domain.go#L55-L67

you think we should verify that both metrics are calculated separately ?

internal/armada/metrics/domain_test.go
```diff
@@ -0,0 +1,95 @@
...
+		case "disk":
+			assert.Equal(t, float64(8), queueMetric.Durations.GetMin())
+			assert.Equal(t, float64(10), queueMetric.Durations.GetMax())
+			assert.Equal(t, float64(18), queueMetric.Durations.GetSum())
+			assert.Equal(t, uint64(2), queueMetric.Durations.GetCount())
+		}
+	}
+}
+
+func TestRecordResources_ShouldReportRuntimResources(t *testing.T) {
```
kannon92, line 52:
```suggestion
func TestRecordResources_ShouldReportRuntimeResources(t *testing.T) {
```

internal/armada/metrics/domain_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	armadaresource "github.com/armadaproject/armada/internal/common/resource"
+)
+
+func TestRecordJobRuntime_ShouldReportRuntimeDurations(t *testing.T) {
+	jobMetricsRecorder := NewJobMetricsRecorder()
+	jobMetricsRecorder.RecordJobRuntime("cpu", "", time.Second)
```
kannon92, line 14: Reading this test, I think you are testing three cases in one function. Could you refactor this to be a table driven test?

Mo-Fatah: can you see now if this what you meant?

Mo-Fatah:
there is a bug actually in the second function, I will fix it now

internal/armada/metrics/domain_test.go
kannon92: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

I think you are just missing a few items but this looks good to me!

Mo-Fatah: Cool. I tried to make something similar to the article, but I found it a bit tricky in this context.

richscott (dismissed):

kannon92:
@Mo-Fatah looks like something went wrong with this since we last updated.  Can you fix this?  Otherwise it looks good to me!

Mo-Fatah:
> internal/armada/metrics/domain.go

the file I am testing, `internal/armada/metrics/domain.go`, no longer exists. I will see if it is deleted entirely in a previous PR or it was moved elsewhere

kannon92:
> @kannon92 the file I am testing, `internal/armada/metrics/domain.go`, no longer exists in the `metrics` package. I will see if it was deleted entirely in a previous PR or it was moved elsewhere

https://github.com/armadaproject/armada/tree/master/internal/common/metrics maybe?

Mo-Fatah:
> https://github.com/armadaproject/armada/tree/master/internal/common/metrics maybe?

yes it is, pushing the changes now

kannon92 (approved):

PR #2189, opened 2023-02-26
[Pulsar Scheduler] Load errors in chunks

d80tb7:
Change the pulsar scheduler so that we load errors in chunks.  This means that we don;t try and load an unbounded number of errors from the db in a single query.

### DIFF
--- a/internal/scheduler/database/job_repository.go
+++ b/internal/scheduler/database/job_repository.go
@@ -3,15 +3,17 @@ package database
 import (
 	"context"
 	"fmt"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4"
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/pkg/errors"
+
 	"github.com/armadaproject/armada/internal/common/compress"
 	"github.com/armadaproject/armada/internal/common/database"
 	protoutil "github.com/armadaproject/armada/internal/common/proto"
 	armadaslices "github.com/armadaproject/armada/internal/common/slices"
 	"github.com/armadaproject/armada/pkg/armadaevents"
-	"github.com/google/uuid"
-	"github.com/jackc/pgx/v4"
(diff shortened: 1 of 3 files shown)


internal/scheduler/database/job_repository.go
```diff
@@ -72,20 +73,48 @@ func NewPostgresJobRepository(db *pgxpool.Pool, batchSize int32) *PostgresJobRep
...
-		if err != nil {
-			return nil, errors.WithStack(err)
+
+	err := r.db.BeginTxFunc(ctx, pgx.TxOptions{
+		IsoLevel:       pgx.ReadCommitted,
+		AccessMode:     pgx.ReadWrite,
+		DeferrableMode: pgx.Deferrable,
+	}, func(tx pgx.Tx) error {
+		for _, chunk := range chunks {
+			tmpTable, err := insertRunIdsToTmpTable(ctx, tx, chunk)
```
JamesMurkin, line 87: What is the advantage of doing it this was vs just calling `SelectRunErrorsById` in chunks?

d80tb7: `SelectRunErrorsById` is using an `IN` clause.  Really large`IN` Clauses (think tens of thousands of values) tend not to be very efficient and extremely large `IN` Clauses  don't work at all.  Using a tmp table means you can have larger chunk sizes and should always get decent performance.

There is a question as to whether we could just cap the chunk size at 10K and test to see if postgres `IN` is just as fast (or faster) than the tmp table approach at this scale.  If that's the case then we could probably get away  with the simpler approach.

JamesMurkin: Sounds like a reasonable explanation - was just trying to avoid code complicated if we didn't need it

But sounds fair - especially if we want to handle large number of ids (and hence large chunk sizes)

I'll approve for now - if you want to test just having a 10k chunk size and doing an `IN` then that is up to you

JamesMurkin (approved):

PR #2209, opened 2023-03-01
Lookout v2 - create separate state for preempted jobs

carlocamurri:
This might change in the future, as Armada might want to re-enqueue preempted jobs. For now though, we have to treat is as a terminal state for visibility, and have to deal with potentially out of order events.


severinson (approved):

internal/lookoutingesterv2/lookoutdb/insertion.go
```diff
@@ -816,22 +836,43 @@ func (l *LookoutDb) filterEventsForTerminalJobs(
...
 
-	cancelledJobs := make(map[string]bool)
+	terminalJobs := make(map[string]int)
 	for rows.Next() {
 		jobId := ""
-		err := rows.Scan(&jobId)
+		var state int16
+		err := rows.Scan(&jobId, &state)
 		if err != nil {
 			log.WithError(err).Warnf("Cannot retrieve jobId from row. Cancelled job will not be filtered out")
```
d80tb7, line 845:
```suggestion
			log.WithError(err).Warnf("Cannot retrieve jobId from row. Terminal jobs will not be filtered out")
```

d80tb7 (dismissed):

d80tb7 (approved):

PR #2210, opened 2023-03-01
add codecov

Mo-Fatah:
trying to exclude generated code from code coverage report (eg `*.pb.go`) since these file affects the the total code coverage percentage significantly


codecov.yml
```diff
@@ -0,0 +1,2 @@
+ignore:
```
kannon92, line 1: I feel so stupid for not realizing I could do this..  I spent so long trying to see how to exclude from `go test`.

Mo-Fatah: I also spent a few hours trying also to change the test scripts in `makefile` to exclude those files :sweat_smile: 
then I found it in their [documentation](https://docs.codecov.com/docs/ignoring-paths)

Mo-Fatah: Is there a way to see the new coverage stats for this PR ? I looked into the coverage CI logs  and followed the link of the success status : https://app.codecov.io/github/armadaproject/armada/commit/216017c33006d880ed658cd14299703eeb3f0673
but it produces and error.

richscott: I get the same error when trying to view that results link as well. I'll try again in an hour, and see if it works.

richscott: There's a JS error on that page - hopefully they are fixing it right now.

richscott: I just checked another PR (one for Lookout V2, by Carlo Camurri) in this repo, and the CodeCov results link is broken for that one as well, so I think this is very much a CodeCov issue for multiple PRs/repos.  I also found another armada PR where the upload of the CodeCov results returns a "Server 500" error from CodeCov, so they are having multiple problems.

kannon92: https://app.codecov.io/github/armadaproject/armada/tree/exclude-generated-code-cov

Yea this tool is not the best..  You get what you pay for!

kannon92: Looks like this PR did not work though.

kannon92: Screenshot 2023-03-01 at 3 08 29 PM

Mo-Fatah: hmmm, I will try something. may be I will exclude a specific file or a directory without regex to see the if this `codecov.yml` is working at all.

richscott: Screen Shot 2023-03-01 at 1 14 41 PM

Does this need to be merged in the `armada` repo, for it to actually take effect?

Mo-Fatah: I have no clue tbh. but I checked this [PR's](https://github.com/armadaproject/armada/pull/2169) branch [here](https://app.codecov.io/github/armadaproject/armada/tree/improve-internal-test-cov) and the effect appears without merging it (from 46.71% to 46.71% )

Mo-Fatah: I just excluded the `cmd` directory and  it worked 
Screenshot from 2023-03-01 22-41-50

So I believe the problem was with the regex

Mo-Fatah:
problem explained here #2211

kannon92:
I was finding this PR to say I think w/e we did worked.  

I enabled CodeCov app and I notice more things happening.  I also saw this:

Screenshot 2023-03-02 at 3 19 20 PM

Mo-Fatah:
Great! I closed it because it is so messy and I thought I should submit another PR when the App is enabled

Mo-Fatah:
I believe currently we want to ignore `*.pb.go` and `.sql.go` files, something else ?

kannon92:
*_moq.go files are one I can think of.

PR #2214, opened 2023-03-03
feat: first version of new workflows

dejanzele:
Fixes #

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-121) by [Unito](https://www.unito.io)


.github/workflows/test.yml
```diff
@@ -0,0 +1,204 @@
...
+        uses: actions/checkout@v3
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+
+      - name: Setup dependencies
+        shell: bash
+        run: make download
```
severinson, line 29: Do this instead https://github.com/armadaproject/armada/blob/master/makefile#L576
```
go run github.com/magefile/mage@v1.14.0 BootstrapTools
go mod download
go mod tidy
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,204 @@
+name: Code Build and Tests
```
severinson, line 1: This means unit and integration tests will run sequentially? Can we run them in paralle?

.github/workflows/test.yml
severinson: So we can start removing the need for the makefile.

severinson (approved):

PR #2215, opened 2023-03-03
Scheduler refactoring

severinson:


internal/scheduler/legacyscheduler.go
```diff
@@ -95,595 +97,579 @@ func SchedulingConstraintsFromSchedulingConfig(
...
-// QueueCandidateGangIterator is an iterator over gangs in a queue that could be scheduled
-// without exceeding per-queue limits.
-type QueueCandidateGangIterator struct {
-	ctx context.Context
-	SchedulingConstraints
-	QueueSchedulingRoundReport *QueueSchedulingRoundReport
-	queuedGangIterator         *QueuedGangIterator
+	preemptedJobs := maps.Values(preemptedJobsById)
+	scheduledJobs := maps.Values(scheduledJobsById)
+	if s := JobsSummary(preemptedJobs); s != "" {
```
d80tb7, line 262: do  you want to log all this  at info level?  There could be  thousands of ids here I think?

internal/scheduler/legacyscheduler.go
```diff
@@ -95,595 +97,579 @@ func SchedulingConstraintsFromSchedulingConfig(
...
 	if err != nil {
-		return nil, false, err
+		return nil, err
 	}
-	unschedulableReason := ""
-	for _, report := range reports {
-		if report.UnschedulableReason != "" {
-			unschedulableReason = report.UnschedulableReason
-			break
+	for jobId := range preemptedJobsById {
```
d80tb7, line 382: This check (and  the one below) are a bit weird?  Are we not confident we're scheduling consistently here?

I guess to expand, my rule would be that you only run checks when you can't be sure of something(e.g.input has come from by a user  or you're sending data over a flaky connection). In this case I think all scheduling being done here is (or at least should) be done using pre-validated input using a  deterministic algo we control.

d80tb7: Just noticed that this entire function is just runtime checks.  I'd remove the whole thing.

internal/scheduler/legacyscheduler.go
severinson: It does put a lot of stuff into the logs, but I don't think it's a large enough number to be an issue. The executor logs the ids of all jobs it renews leases for and this is typically a much smaller number of ids.

internal/scheduler/legacyscheduler.go
severinson: Yes, we currently only need it to figure out where jobs ended up (because `Scheduler` doesn't track nodes). I'm changing `Scheduler` to track that explicitly, after which we won't need to call this function. It has been useful to find bugs in staging. But maybe the right trade-off is to put it behind a flag that we leave disabled in prod?

internal/scheduler/legacyscheduler.go
```diff
@@ -95,595 +97,579 @@ func SchedulingConstraintsFromSchedulingConfig(
...
-		return nil, nil
-	}
+	// NodeDb snapshot prior to making any changes.
+	// We compare against this snapshot after scheduling to detect changes.
+	snapshot := sch.nodeDb.Txn(false)
 
-	if it.next != nil {
-		return it.next, nil
+	// Evict preemptible jobs.
+	evictorResult, inMemoryJobRepo, err := sch.evict(
```
d80tb7, line 168: the following ~20  lines are duplicated again just below- we could probably extract this to it's own function.

internal/scheduler/legacyscheduler.go
d80tb7: I guess if we view ELK as the only viable log store then it's not completely unreasonable.  It's more just that generally I'd view this sort of verbose logging as something you'd put at debug level and specifically turn on to troubleshoot issues, rather than something you'd log at info proactively.

d80tb7: > The executor logs the ids of all jobs it renews leases for and this is typically a much smaller number of ids.

This is also wrong!

internal/scheduler/legacyscheduler.go
d80tb7: I think for now- put it behind a flag and we can turn it on in  staging.  I think we should aim to get rid of it sooner rather than later though because it's basically a band aid.

internal/scheduler/legacyscheduler.go
severinson: Note that the next eviction uses a different evictor.

d80tb7 (approved):

PR #2218, opened 2023-03-06
Add Rescheduler to new scheduler

severinson:
To enable preemption to fair share when using the new scheduler.


internal/scheduler/scheduling_algo.go
```diff
@@ -190,42 +189,67 @@ func (l *LegacySchedulingAlgo) scheduleOnExecutor(
...
-	return updatedJobs, nil
+	return result, nil
+}
+
+type schedulerJobRepositoryAdapter struct {
+	db  *jobdb.JobDb
+	txn *jobdb.Txn
+}
+
+func (repo *schedulerJobRepositoryAdapter) GetQueueJobIds(queue string) ([]string, error) {
```
d80tb7, line 236: This f

internal/scheduler/scheduling_algo.go
```diff
@@ -190,42 +189,67 @@ func (l *LegacySchedulingAlgo) scheduleOnExecutor(
...
-	if err != nil {
-		return nil, err
+
+	for i, job := range result.PreemptedJobs {
+		jobDbJob := job.(*jobdb.Job)
+		node := result.NodeByJobId[jobDbJob.GetId()]
+		if node == nil {
+			return nil, errors.Errorf("job %s not mapped to any node", jobDbJob.GetId())
+		}
+		if run := jobDbJob.LatestRun(); run != nil {
```
d80tb7, line 215: I think in scheduler.go you error if a job is preempted but doesn't have a run. I think we should  be consistent here and either ignore in both case or error in both cases.

internal/scheduler/scheduler_test.go
```diff
@@ -665,7 +665,7 @@ type testSchedulingAlgo struct {
 	shouldError    bool
 }
 
-func (t *testSchedulingAlgo) Schedule(ctx context.Context, txn *jobdb.Txn, jobDb *jobdb.JobDb) ([]*jobdb.Job, error) {
```
d80tb7, line 668: shouldn't there be a test here that we do indeed create the correct prempted messages?

internal/scheduler/scheduling_algo.go
d80tb7: If we do need these functions I'd suggest writing a bit of doc and renaming this one  to `GetQueuedJobIds` as it does fetches all queued jobs

d80tb7:
So One meta question here is how the legacyscheduler determines what to evict.  Does it consider jobs that the executor considers running  or does it consider jobs that jobdb considers running.  If, as I suspect, it is the former, then I think you'll run into problems with cancelled/expired jobs where:
a) You can try to preempt a job that doesn't exist in jobdb because it's been cancelled /expired (but the executor doesn't know it yet).  This might lead to panics/errors as we'll be looking up jobs that don't exist.
b) You can try and preempt a job that does exist in JobDB but you generated a nonsensical sequence of messages:
```
JobRunFailed[Cancelled]->JobFailed[Cancelled]->JobRunPreempted->JobRunFailed[Preempted]->JobFailed[Preempted]
```

internal/scheduler/scheduling_algo.go
severinson: We need it to implement the `JobRepository interface`. I suggest leaving these for now and separately improving `Rescheduler` to take an iterator over queued job ids.

internal/scheduler/scheduling_algo.go
severinson: Agreed. Fixed.

severinson:
> So One meta question here is how the legacyscheduler determines what to evict. Does it consider jobs that the executor considers running or does it consider jobs that jobdb considers running. If, as I suspect, it is the former, then I think you'll run into problems with cancelled/expired jobs where: a) You can try to preempt a job that doesn't exist in jobdb because it's been cancelled /expired (but the executor doesn't know it yet). This might lead to panics/errors as we'll be looking up jobs that don't exist. b) You can try and preempt a job that does exist in JobDB but you generated a nonsensical sequence of messages:
> 
> ```
> JobRunFailed[Cancelled]->JobFailed[Cancelled]->JobRunPreempted->JobRunFailed[Preempted]->JobFailed[Preempted]
> ```

It's the jobs that JobDb considers to be running;:
https://github.com/armadaproject/armada/blob/c39801a899c36021225e14808279284d44a4d6d9/internal/scheduler/scheduling_algo.go#L268

d80tb7 (approved):

Issue #2221, opened 2023-03-06
[MageFiles] - Build Lookout UI Kind Cluster and Armada Dependency containers from mage


Sharpz7:
All dependencies (Pulsar, Redis and Postgres) are to be built from mage.

PR #2222, opened 2023-03-06
Magefile and Localdev merges - first run

Sharpz7:
# Summary

Closes https://github.com/armadaproject/armada/issues/2221
Closes https://github.com/armadaproject/armada/issues/2235
Closes https://github.com/armadaproject/armada/issues/2246
Closes https://github.com/armadaproject/armada/issues/2247
Closes https://github.com/armadaproject/armada/issues/2248

# The New Process

Instead of compiling from docker-compose, localdev now builds the CI docker images locally. This doesn't slow us down (We still have to compile the services, lookoutUI anyway), and instead encourages users to use good, supported debugging methods (Delve, VSCode, Goland)

# How to Use

For quick testing, run `developer/init.sh` for a quick build, but you can also still use the individual mage targets and docker-compose commands.

Please refer to: 
- https://github.com/armadaproject/armada/blob/d53cff9f9e33026a64372c2539a841f57dab4402/docs/developer.md
- https://github.com/armadaproject/armada/blob/d53cff9f9e33026a64372c2539a841f57dab4402/docs/development_guide.md

Also note that `localdev/run.sh` still works :))

# Problems

## Delay before Armada Runs

There seems to be a slight delay between Armada Booting, and the testsuite successfully running. ~(30seconds)
Before this time, you get:

**Update**
This seems to be a known design problem: https://github.com/armadaproject/armada/blob/master/magefiles/ci.go#L30. I have therefore added a 15 second pause.

# Breaking Changes

Delve debugging has not been tested, extensively, so expect problems until https://github.com/armadaproject/armada/issues/2249

`go-path` also seemed to cause problems in the debug docker-compose `dlv` would fail to be found.


developer/README.md
```diff
@@ -0,0 +1,21 @@
+# Dependencies (TEMPORARY)
```
kannon92, line 1: Why Temporary?

localdev/config/armada/config.yaml
```diff
@@ -1,17 +1,13 @@
...
-priorityHalfTime: 20m
-cancelJobsBatchSize: 1000
 grpc:
   keepaliveParams:
     maxConnectionIdle: 5m
     time: 2h
     timeout: 20s
   keepaliveEnforcementPolicy:
     minTime: 10s
+    # ?
```
kannon92, line 8: Why is this here?

localdev/config/executor/config.yaml
```diff
@@ -1,42 +1,27 @@
...
+  # This line is different
   impersonateUsers: true
   etcd:
     fractionOfStorageInUseSoftLimit: 0.8
     fractionOfStorageInUseHardLimit: 0.9
     minimumAvailable: 2
+  # These lines are different
   QPS: 100
   Burst: 100
+
```
kannon92, line 17:
```suggestion
```

kannon92 (dismissed):
Nits but it lgtm

developer/README.md
Sharpz7: Just leaving this here for reference - I will need to re-write the dev docs.

config/executor/config.yaml
```diff
@@ -19,6 +19,7 @@ task:
   utilisationEventReportingInterval: 5m
 apiConnection:
   armadaUrl: "server:50051"
+  forceNoTls: false
```
Sharpz7, line 22: I am assuming this will not break anything?

It is required to allow the env var to work I think, it can only replace, not create.

developer/env/armada.env
```diff
@@ -0,0 +1 @@
+ARMADA_SCHEDULING_PREEMPTION_DEFAULTPRIORITYCLASS=""
```
Sharpz7, line 1: Created seperate env files to stop future potential conflicts.

i.e `ARMADA_POSTGRES_CONNECTION_DBNAME` for lookout or lookoutv2? there is no distinction.

magefiles/developer.go
```diff
@@ -0,0 +1,90 @@
...
+	mg.Deps(StopComponents)
+	mg.Deps(StopDependencies)
+	mg.Deps(KindTeardown)
+}
+
+// Dependencies include pulsar, postgres (v1 and v2) as well as redis
+func StartDependencies() error {
+	// If user is on arm, export to env "PULSAR_IMAGE=kezhenxu94/pulsar"
+	if onArm() {
+		os.Setenv("PULSAR_IMAGE", "kezhenxu94/pulsar")
```
kannon92, line 40: Prob can use Rich’s image

Sharpz7: Good point! Thank you.

kannon92 (approved):

PR #2245, opened 2023-03-09
Gang preemption

severinson:
All-or-nothing preemption for gang jobs.

Builds on https://github.com/armadaproject/armada/pull/2243; merge that before reviewing this PR.


internal/scheduler/scheduling_algo.go
```diff
@@ -199,6 +199,10 @@ func (l *LegacySchedulingAlgo) scheduleOnExecutor(
 		nodeDb,
 		priorityFactorByQueue,
 		allocatedByQueueAndPriority,
+		// TODO: Add missing maps to enable gang preemption.
```
severinson, line 202: I wanted to test that the gang preemption design works out before adding it also to the new scheduler.

JamesMurkin (approved):

PR #2261, opened 2023-03-13
[WIP] Split lease request and job submission

JamesMurkin:
This PR does several things (unfortunately)
 - Splits the lease request and job submissions
   - This used to be all done in a single step. Now we have 2 steps:
     - 1. Request leases and save them into the RunStateStore
     - 2. Read all unprocessed runs from the RunStateStore and submit them to kubernetes 
 - Preempt jobs based on Ids given to us by the server
 - Adds unit tests


internal/executor/job/job_run_state_store.go
```diff
@@ -127,7 +193,11 @@ func (stateStore *JobRunStateStore) Get(runId string) *RunState {
 	stateStore.lock.Lock()
 	defer stateStore.lock.Unlock()
 
-	return stateStore.jobRunState[runId].DeepCopy()
+	run, exists := stateStore.jobRunState[runId]
```
d80tb7, line 196: Don't we need to mark the state as deleted (or actually delete it?). If we don't we should add a comment here as it's a bit weird to have a function called "Delete" that doesn't seem to delete anything!

internal/executor/job/processors/preempt_runs.go
```diff
@@ -0,0 +1,98 @@
...
+
+			if util.IsInTerminalState(pod) {
+				// If pod is already finished, nothing to preempt
+				return
+			}
+
+			if !util.IsReportedPreempted(pod) {
+				err := j.reportPodPreempted(runInfo.Run, runInfo.Pod)
+				if err != nil {
+					log.Error(err)
```
d80tb7, line 65: will this log something so we know the error has occurred as part of a preemption?

internal/executor/job/processors/remove_runs.go
```diff
@@ -0,0 +1,69 @@
...
+
+	util.ProcessItemsWithThreadPool(context.Background(), 20, runPodInfos,
+		func(runInfo *runPodInfo) {
+			pod := runInfo.Pod
+			if pod == nil {
+				j.jobRunStateStore.Delete(runInfo.Run.Meta.RunId)
+				return
+			}
+
+			if util.IsInTerminalState(pod) && util.HasCurrentStateBeenReported(pod) {
```
d80tb7, line 48: Ho good is the state model here?  e.g  if a pod is simultaneously CancelRequested and PreemptionRequested, what sort of guarantees do we get about what will happen?

internal/executor/job/processors/remove_runs.go
```diff
@@ -0,0 +1,69 @@
...
+				// If pod is already finished, don't delete it so users can view pods
```
d80tb7, line 49:
```suggestion
				// If pod is already finished, don't delete it as users may want to view pods
```

internal/executor/job/processors/remove_runs.go
```diff
@@ -0,0 +1,69 @@
+package processors
+
+import (
```
d80tb7, line 3: My OO brain says that the terminal event processors are 90% the same and we could therefore exrtract some of this out.

internal/executor/job/testfixtures.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/armadaproject/armada/pkg/api"
+)
+
+// TestJobRunStateStore Just wraps JobRunStateStore but allows tests to set the initial state
+type TestJobRunStateStore struct {
+	jobRunStateStore *JobRunStateStore
+}
+
+func NewTestJobRunStateStore(initialJobRuns []*RunState) *TestJobRunStateStore {
```
d80tb7, line 14: Maybe I'm being dim but why do we need a whole new struct here  Can't we just have a `NewJobRunStateStoreWithInitialState` function in `job_run_state_store.go` that returns a a `JobRunStateStore` with the state set?

internal/executor/job/testfixtures.go
```diff
@@ -0,0 +1,84 @@
+package job
```
d80tb7, line 1: I'd either name this `testfixtures_test.go` or move it into a special `testfixtures` package.  As it stands right now, I think this will be compiled into our prod binary.

internal/executor/reporter/testfixtures.go
```diff
@@ -0,0 +1,25 @@
+package reporter
```
d80tb7, line 1: As above, you probably want to name this testfixtures_test.go or move into a dedicated fixtures package.

internal/executor/service/job_requester.go
```diff
@@ -2,99 +2,196 @@ package service
...
+	}
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
-	stream, err := requester.executorApiClient.LeaseJobRuns(ctx, grpcretry.Disable(), grpc.UseCompressor(gzip.Name))
+	leaseResponse, err := r.leaseRequester.LeaseJobRuns(ctx, leaseRequest)
+	if err != nil {
+		log.Errorf("Failed to request new jobs leases as because %s", err)
+		return
+	}
+	logAvailableResources(leaseRequest.AvailableResource, len(leaseResponse.LeasedRuns))
```
d80tb7, line 63: I'd log  a summary here of what you got- e.g. how many new jobs, how many jobs to preempt etc.

internal/executor/service/job_requester.go
```diff
@@ -2,99 +2,196 @@ package service
...
-		return nil, nil, errors.WithStack(err)
+
+	// We make the assumption here that JobRunStateStore knows about all job runs and don't reconcile again against kubernetes
+	// This should be a safe assumption - and would be a bug if it was ever not true
+	allJobRuns := r.jobRunStateStore.GetAll()
+	allJobRunIds = append(allJobRunIds, util2.Map(allJobRuns, func(val *job.RunState) string {
+		return val.Meta.RunId
+	})...)
+
+	unassignedIds := slices.Subtract(allJobRunIds, allAssignedRunIds)
```
d80tb7, line 111: Why is is `slices.Subtract` but `util2.Filter`?!

internal/executor/service/lease_requester.go
```diff
@@ -0,0 +1,112 @@
...
+		MinimumJobSize:      requester.minimumJobSize,
+		Resources:           request.AvailableResource,
+		Nodes:               request.Nodes,
+		UnassignedJobRunIds: request.UnassignedJobRunIds,
+	}
+	if err := stream.Send(leaseRequest); err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	leaseRuns := make([]*executorapi.JobRunLease, 0, 10)
```
d80tb7, line 70: why an initial size of 10 here?  I guess my view here would be that if we're only getting 10 jobs then the perf impact of array resizing doesn't matter, while if we're getting thousands of jobs, then the an initial size of 10 doesn't help much.

internal/executor/service/lease_requester.go
```diff
@@ -0,0 +1,112 @@
...
+			switch typed := res.GetEvent().(type) {
+			case *executorapi.LeaseStreamMessage_Lease:
+				leaseRuns = append(leaseRuns, typed.Lease)
+			case *executorapi.LeaseStreamMessage_PreemptRuns:
+				runIdsToPreempt = append(runIdsToPreempt, typed.PreemptRuns.JobRunIdsToPreempt...)
+			case *executorapi.LeaseStreamMessage_CancelRuns:
+				runIdsToCancel = append(runIdsToCancel, typed.CancelRuns.JobRunIdsToCancel...)
+			case *executorapi.LeaseStreamMessage_End:
+				shouldEndStreamCall = true
+			}
```
d80tb7, line 99: I'd put a default here and log an error.  It'll help us if we update the protocol and and up with mismatched versions.

internal/executor/service/testfixtures.go
```diff
@@ -0,0 +1,14 @@
+package service
```
d80tb7, line 1: as above- let's rename ot `testfixtures_test` or put in a dedicated package.

internal/executor/utilisation/testfixtures.go
```diff
@@ -0,0 +1,16 @@
+package utilisation
```
d80tb7, line 1: rename!

d80tb7: or actually I'd remove this and just use mocks

internal/executor/job/job_run_state_store.go
```diff
@@ -81,25 +95,59 @@ func (stateStore *JobRunStateStore) reportRunActive(pod *v1.Pod) {
...
+	stateStore.lock.Lock()
+	defer stateStore.lock.Unlock()
+
+	currentState, present := stateStore.jobRunState[runMeta.RunId]
+	if !present {
+		log.Warnf("run unexpected reported as failed submission (runId=%s, jobId=%s), no current state exists", runMeta.RunId, runMeta.JobId)
+		currentState = &RunState{
+			Meta: runMeta,
+		}
+		stateStore.jobRunState[runMeta.RunId] = currentState
```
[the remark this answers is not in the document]
JamesMurkin: This should never be possible and should just warn + exit. Let me fix it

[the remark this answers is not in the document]
JamesMurkin: Actually looking at it - it would be completely fine. However it is handling a case that should never happen - so probably fine to just do as I said above

internal/executor/job/processors/preempt_runs.go
JamesMurkin: Yes but I'll make it more explicit at this level

internal/executor/job/processors/remove_runs.go
JamesMurkin: Still relatively weak - it isn't a unified state yet

So no guarantee - beyond the pod will get deleted 

It is a fair point - we could run these sequentially rather in parallel (or combine them)

internal/executor/job/testfixtures.go
JamesMurkin: Yes I didn't start like this - but this was the end result.

I'll just do that

PR #2264, opened 2023-03-14
Preemption fixes and improvements

severinson:
This pr changes how resources are accounted for in lease.go. We currently send resources allocated per queue separately. With this pr, we change it so that the server aggregates the resource usage per queue based on the jobs the executor has reported exists in the cluster. This is necessary to ensure resource accounting is correct now that we send all jobs to the executor with the same priority class; without this change, all resource usage is reported at the armada-default priority class (the class all jobs are marked as when sent to the executor now).

Removes the job deletion grace period. Jobs are now deleted immediately when calling delete instead of setting an expiry on the key. This is necessary to ensure preempted jobs don't factor into the resources used by a queue.


internal/armada/repository/job_test.go
```diff
@@ -189,59 +188,6 @@ func TestDeleteQueuedJob(t *testing.T) {
 	})
 }
 
-func TestDeleteJobShouldSetJobObjectToExpire(t *testing.T) {
```
severinson, line 192: Do we rely on being able to return leases for deleted jobs?

internal/armada/repository/job_test.go
```diff
@@ -259,24 +205,6 @@ func TestDeleteWithSomeMissingJobs(t *testing.T) {
 	})
 }
 
-func TestReturnLeaseForDeletedJobShouldKeepJobDeleted(t *testing.T) {
```
JamesMurkin, line 262: Does this Test still pass with your changes?

Most likely this was added as previously I imagine that wasn't the case

internal/armada/repository/job_test.go
JamesMurkin: Not conceptually but possibly inadvertently - I can think of edge cases where it could happen - however as long as the server treats it as a noop it should be fine.

internal/armada/repository/job_test.go
severinson: Since making returning lease for deleted jobs a no-op, this passes.

internal/armada/repository/job_test.go
severinson: Done

internal/armada/server/lease.go
```diff
@@ -600,32 +620,128 @@ func (q *AggregatedQueueServer) getJobs(ctx context.Context, req *api.StreamingL
...
+			deleteErrorByJobId := armadamaps.MapKeys(deletionResult, func(job *api.Job) string { return job.Id })
+			for jobId := range preemptedApiJobsById {
+				if err, ok := deleteErrorByJobId[jobId]; !ok {
+					log.Errorf("deletion result missing for preempted job %s", jobId)
+				} else if err != nil {
+					log.Errorf("failed to delete preempted job %s: %s", jobId, err.Error())
+				}
+			}
+		}
+		if jobs, err := q.jobRepository.GetExistingJobsByIds(jobIdsToDelete); err != nil {
```
JamesMurkin, line 675: Is this section actually needed?

JamesMurkin (approved):

internal/armada/server/lease.go
severinson: No. I've removed it.

JamesMurkin (approved):

PR #2266, opened 2023-03-14
Better Error Handling when rejecting Jobs

### LINKED ISSUE #1748: Better Error Message When Rejecting Jobs
When we reject a job (for example because it can't be scheduled on any node) the whole grpc call fails.  There is an error message that is returned, but users have said that this doesn't give enough infomation about what was valid. Instead they ask:

```
is it possible for these errors to be returned on the Error field on the JobResponseItem that we're given?
```

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-80) by [Unito](https://www.unito.io)

Mo-Fatah:
#### What type of PR is this?
Feature

**Old behavior** : When a job cannot be scheduled for any reason, the whole gRpc call will fail, and an error will be returned telling the client only the order of the job that couldn't be scheduled

**Suggested Changes** :  
- Define a new Protobuf message `JobFailedToSchedule` that include the `PodSpecs` of this job, and the relevant errors
- When a submission request comes in, filter out the "Can not be scheduled jobs"
- Continue the submission process with all remaining eligible jobs
- Return the `JobSubmitResponse` and include with it all the jobs that failed to be scheduled

This will give the client more information of what exactly went wrong instead of rejecting all jobs. 

Fixes #1748


kannon92:
Maybe @severinson or @d80tb7 should weigh in on this PR?

I normally don't suggest changing the submit.proto definitions without any consulation.

Mo-Fatah:
> I normally don't suggest changing the submit.proto definitions without any consulation.

yeah I get your point. I thought if this can fix the issue a bit and things doesn't break terribly, then maybe go this way and see the team opinion on this.

PR #2268, opened 2023-03-15
Fix gang preemption

severinson:
Specifically:
- Make sure `nodeIdByJobId` is populated in `lease.go`, so that when evicting we can figure out which node each gang job is on.
- If some jobs in a gang have terminated, ensure that the cardinality of the gang is updated before rescheduling, so that the remaining jobs can be re-scheduled.

In addition, the eviction logic has been made more robust, by adding checks where it makes sense, and better tested.

### DIFF
--- a/internal/armada/server/lease.go
+++ b/internal/armada/server/lease.go
@@ -282,6 +282,7 @@ func (q *AggregatedQueueServer) getJobs(ctx context.Context, req *api.StreamingL
 	allocatedByQueueForCluster := make(map[string]schedulerobjects.QuantityByPriorityAndResourceType)
 	jobIdsByGangId := make(map[string]map[string]bool)
 	gangIdByJobId := make(map[string]string)
+	nodeIdByJobId := make(map[string]string)
 	for _, nodeInfo := range req.Nodes {
 		node, err := api.NewNodeFromNodeInfo(
 			&nodeInfo,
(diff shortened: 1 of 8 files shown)


internal/common/validation/podspec.go
severinson, line 1: Changes to this file are unrelated to preemption. Just cleaned it up a bit when I was going over the config.

internal/armada/server/lease.go
```diff
@@ -386,6 +371,12 @@ func (q *AggregatedQueueServer) getJobs(ctx context.Context, req *api.StreamingL
 		if skipNode {
 			continue
 		}
+
+		// Record the which node each job is scheduled on. Necessary for gang preemption.
+		for _, job := range jobs {
+			nodeIdByJobId[job.Id] = node.Id
```
severinson, line 377: Not populating this was the reason gang preemption didn't work correctly previously.

internal/scheduler/legacyscheduler_test.go
```diff
@@ -1566,6 +1596,54 @@ func TestReschedule(t *testing.T) {
 			// To test the gang evictor, we need to disable stochastic eviction.
 			NodeEvictionProbability: pointer.Pointer(0.0),
 		},
+		"gang preemption avoid cascading preemptions": {
```
severinson, line 1599: Test to ensure evicting the rest of a gang job doesn't cause more partial gang evictions.

JamesMurkin (approved):

PR #2271, opened 2023-03-16
Improve handling of Pending pods

JamesMurkin:
- Force kill the stuck pod if it reaches the kill timeout - like we do with other forms of pod deletion
 - Don't forget about issues if the pod has gone into Running - if the pod has already been marked for deletion
   - This is an edge case. It means we detected it was stuck, tried to delete it and then it started running.
   - This gets tricky without a state model to handle well. Ideally we need to see if it completes or deletes first and handle it appropriately
 - Consider an issue resolved if a pod we thought was stuck has since completed.


internal/executor/context/cluster_context.go
```diff
@@ -323,20 +323,26 @@ func (c *KubernetesClusterContext) AddClusterEventAnnotation(event *v1.Event, an
...
-	if !util.IsMarkedForDeletion(pod) {
-		_, err := c.markForDeletion(pod)
+	if !util.IsMarkedForDeletion(currentPod) {
+		_, err := c.markForDeletion(currentPod)
 		if err != nil {
 			return err
 		}
-		currentPod, err = c.waitForPodUpdateInLocalCache(pod, time.Second*5)
+		// Get latest pod state - bypassing cache
+		timeout, cancel := context.WithTimeout(context.Background(), time.Second*10)
```
kannon92, line 337: Should you just make this configurable?

d80tb7 (approved):

PR #2283, opened 2023-03-17
Fixes to scheduler logger + leader election

JamesMurkin:
- Set the token to Invalid in the constructor
   - to avoid nil token
 - Set logger on context
   - This was not being set and causing many places using the logger in the context to log nothing
 - Swap "cluster" leader election mode to "kubernetes"
   - This now tries to use in cluster config and if not, falls back to looking for a kubeconfig. This makes it easy to test the leader election locally

### DIFF
--- a/internal/scheduler/schedulerapp.go
+++ b/internal/scheduler/schedulerapp.go
@@ -194,7 +194,7 @@ func createLeaderController(config LeaderConfig) (LeaderController, error) {
 		return NewStandaloneLeaderController(), nil
 	case "kubernetes":
 		log.Infof("Scheduler will run kubernetes mode")
-		clusterConfig, err := loadConfig()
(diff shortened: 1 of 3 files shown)


internal/scheduler/schedulerapp.go
```diff
@@ -203,3 +207,15 @@ func createLeaderController(config LeaderConfig) (LeaderController, error) {
 		return nil, errors.Errorf("%s is not a value leader mode", config.Mode)
 	}
 }
+
+func loadConfig() (*rest.Config, error) {
```
d80tb7, line 211: I'd be tempted to call this `loadClusterConfig` as there's lots of other config floating around in this package.

d80tb7 (dismissed):

d80tb7 (approved):

PR #2287, opened 2023-03-21
Scheduler reports/contexts separation and improved reports

severinson:
There are now three different reports, with output as exemplified below.
The examples show just a single cluster (´Cluster1´). If there are several clusters, data is shown for all available clusters.
Reports are still strings created from scheduling contexts. We'll go further to create specific proto objects for reports in a follow-up PR that can be requested and parsed programmatically.

```
> go run cmd/armadactl/main.go scheduling-report
Cluster1:
 Most recent attempt:
  Started:                                 2023-03-22 16:29:39.5035361 +0000 GMT m=+15.269667301
  Finished:                                2023-03-22 16:29:39.5045666 +0000 GMT m=+15.270697801
  Duration:                                1.0305ms
  Total capacity:                          {cpu: 19900m, ephemeral-storage: 1055762868Ki, hugepages-1Gi: 0, hugepages-2Mi: 0, memory: 16159044Ki, pods: 110}
  Jobs scheduled:                          0
  Total resources scheduled:               {}
  Total resources scheduled (by priority): {}
  Scheduled queues:                        []
  Termination reason:                      no remaining candidate jobs
 Most recent successful attempt:
  Started:                                 2023-03-22 16:29:34.4164061 +0000 GMT m=+10.182537301
  Finished:                                2023-03-22 16:29:34.4217851 +0000 GMT m=+10.187916301
  Duration:                                5.379ms
  Total capacity:                          {pods: 110, cpu: 19900m, ephemeral-storage: 1055762868Ki, hugepages-1Gi: 0, hugepages-2Mi: 0, memory: 16159044Ki}
  Jobs scheduled:                          19
  Total resources scheduled:               {cpu: 19, ephemeral-storage: 152Gi, memory: 190Mi}
  Total resources scheduled (by priority): {1000: {ephemeral-storage: 152Gi, memory: 190Mi, cpu: 19}}
  Scheduled queues:                        [a]
  Termination reason:                      no remaining candidate jobs
```

```
> go run cmd/armadactl/main.go queue-report --queue a
Cluster1:
 Most recent attempt:
  Time:                                                     2023-03-22 16:30:29.7829922 +0000 GMT m=+65.549123401
  Queue:                                                    a
  Total allocated resources after scheduling:               {cpu: 19, ephemeral-storage: 152Gi, memory: 190Mi}
  Total allocated resources after scheduling (by priority): {1000: {ephemeral-storage: 152Gi, memory: 190Mi, cpu: 19}}
  Scheduled resources:                                      {}
  Scheduled resources (by priority):                        {}
  Number of jobs scheduled:                                 0
  Number of jobs that could not be scheduled:               2
  Unschedulable jobs:
   2: scheduling would exceed cpu quota (total limit for this queue) jobs [01gw52ezw916ee8wzyswhwg41w] (and 1 others not shown)
 Most recent successful attempt:
  Time:                                                     2023-03-22 16:30:24.749787 +0000 GMT m=+60.515918201
  Queue:                                                    a
  Total allocated resources after scheduling:               {cpu: 19, ephemeral-storage: 152Gi, memory: 190Mi}
  Total allocated resources after scheduling (by priority): {1000: {memory: 190Mi, cpu: 19, ephemeral-storage: 152Gi}}
  Scheduled resources:                                      {memory: 20Mi, cpu: 2, ephemeral-storage: 16Gi}
  Scheduled resources (by priority):                        {1000: {memory: 20Mi, cpu: 2, ephemeral-storage: 16Gi}}
  Number of jobs scheduled:                                 2
  Number of jobs that could not be scheduled:               2
  Scheduled jobs:                                           [01gw52ezw916ee8wzysn1kwa5m] (and 1 others not shown)
  Unschedulable jobs:
   2: scheduling would exceed cpu quota (total limit for this queue) jobs [01gw52ezw916ee8wzysxmk2b8d] (and 1 others not shown)
```

```
> go run cmd/armadactl/main.go job-report --jobId 01gw52ezw916ee8wzysxmk2b8d
Cluster1:
 Time:                       2023-03-22 16:30:44.8650321 +0000 GMT m=+80.631163301
 Job id:                     01gw52ezw916ee8wzysxmk2b8d
 Number of nodes in cluster: 1
 UnschedulableReason:        none
 ---
 Assigned node:                Cluster1-armada-test-worker
 Number of matched node types: 1
 Excluded node types:          none
 Node filter:                  considering nodes with 1 cpu allocatable at priority 1000
 Excluded nodes:
  1: pod requires 8Gi ephemeral-storage, but only 0 is available
```


internal/scheduler/reports.go
```diff
@@ -6,441 +6,314 @@ import (
...
-	// Scheduling reports for the jobs that were most recently attempted to be scheduled.
-	MostRecentJobSchedulingReports *lru.Cache
-	// Scheduling reports for the most recently seen queues.
-	MostRecentQueueSchedulingReports *lru.Cache
+// SchedulingContextRepository stores scheduling contexts associated with recent scheduling attempts.
+type SchedulingContextRepository struct {
+	// Maps executor id to *SchedulingContext.
+	// The most recent attempt.
+	mostRecentSchedulingContextByExecutor SchedulingContextByExecutor
+	// The most recent attempt where a non-zero amount of resources where scheduled.
```
d80tb7, line 26:
```suggestion
	// The most recent attempt where a non-zero amount of resources were scheduled.
```

internal/scheduler/reports.go
```diff
@@ -6,441 +6,314 @@ import (
...
+	w.Flush()
+	return sb.String()
 }
 
-func (repo *SchedulingReportsRepository) AddMany(queueName string, reports []*JobSchedulingReport) {
-	for _, report := range reports {
-		repo.Add(queueName, report)
-	}
+func (repo *SchedulingContextRepository) GetQueueReport(_ context.Context, queue *schedulerobjects.Queue) (*schedulerobjects.QueueReport, error) {
+	queueName := strings.TrimSpace(queue.Name)
```
d80tb7, line 201: why do we need to do this?  If we're getting this from an external system (e.g. it;'s coming over grpc) then I think we should do and validation/sanitisation in the service layer.

internal/scheduler/reports.go
```diff
@@ -6,441 +6,314 @@ import (
...
 	}
+	w.Flush()
+	return sb.String()
 }
 
-func (repo *SchedulingReportsRepository) AddMany(queueName string, reports []*JobSchedulingReport) {
-	for _, report := range reports {
-		repo.Add(queueName, report)
-	}
+func (repo *SchedulingContextRepository) GetQueueReport(_ context.Context, queue *schedulerobjects.Queue) (*schedulerobjects.QueueReport, error) {
```
d80tb7, line 200: Shouldn't there be a read lock around this (and also GetSchedulingReport)?

More generally, even if the logic here is correct, I think the mutable state protected by mutexes here is going to bight us sooner or later as this stuff is inherantly difficult to reason about and test.  I'd consider using another pattern here to try and make it safer. If it were me I'd probably try and create a state struct which held all the mutable state and then store that in an atomicReference. Writers would  deep copy the struct, update the copy and then store a new state.  Readers would just grab the struct.  It might also be possible to use iummutable maps int he struct which then might event remove the need to make a deep copy.  You'd still want a write mutex, but this would mean that any new read methods wouldn't need to worry about threading issues

internal/scheduler/schedulercontexts.go
```diff
@@ -0,0 +1,364 @@
...
+	QueueSchedulingContexts map[string]*QueueSchedulingContext
+	// Total resources across all clusters available at the start of the scheduling cycle.
+	TotalResources schedulerobjects.ResourceList
+	// Resources assigned across all queues during this scheduling cycle.
+	ScheduledResourcesByPriority schedulerobjects.QuantityByPriorityAndResourceType
+	// Total number of jobs successfully scheduled in this round.
+	NumScheduledJobs int
+	// Reason for why the scheduling round finished.
+	TerminationReason string
+	// Protects everything in this struct.
```
d80tb7, line 36: Why does this need to be threadsafe?  I would have assumed that this is a simple record type that we pass round the single-threaded scheduling cycle.

internal/scheduler/schedulercontexts.go
```diff
@@ -0,0 +1,364 @@
...
+		)
+		sctx.NumScheduledJobs++
+	}
+	if qctx := sctx.QueueSchedulingContexts[jctx.Job.GetQueue()]; qctx != nil {
+		qctx.AddJobSchedulingContext(jctx, isEvictedJob)
+	}
+}
+
+// ClearJobSpecs zeroes out job specs to reduce memory usage.
+func (sctx *SchedulingContext) ClearJobSpecs() {
```
d80tb7, line 108: To me this is a bit of a code smell.  I think the issue here is that the scheduling context provides two main functions.  A) to act as a store of state during the scheduler cycle and B) to provide a record of what we did.  Currently we're conflating these two functions into a single struct which means we have to do things like this.  I think ideally we have a separate set of object that represent reports and transform into those at the at the end of the scheduling cycle.  Obviously there would be a fair amount of overlap in the fields each would store, but it would means that we can have specialised structs for each which would avoid the need to do stuff like this and have other benefits (e.g. reports really should be immutable, while the context probably shouldnt).

  In the medium term I think we're going to need to do the above regardless, because  we probably want to expose the report struct over grpc and in this case we almost certainly don't want to expose our internal scheduling context.

internal/scheduler/schedulercontexts.go
d80tb7: Also all the fields in this struct are exported but they need to be protected by a mutex.  I don't think that will be obvious to callers.

d80tb7: (above 2 comments apply to the other contexts below)

internal/scheduler/schedulercontexts.go
```diff
@@ -0,0 +1,364 @@
...
+		Started:                      time.Now(),
+		ExecutorId:                   executorId,
+		QueueSchedulingContexts:      queueSchedulingContexts,
+		TotalResources:               totalResources.DeepCopy(),
+		ScheduledResourcesByPriority: make(schedulerobjects.QuantityByPriorityAndResourceType),
+	}
+}
+
+func (sctx *SchedulingContext) String() string {
+	var sb strings.Builder
```
d80tb7, line 65: doesn't this need some locking?

internal/scheduler/schedulercontexts.go
```diff
@@ -0,0 +1,364 @@
...
+		ScheduledResourcesByPriority:      make(schedulerobjects.QuantityByPriorityAndResourceType),
+		SuccessfulJobSchedulingContexts:   make(map[string]*JobSchedulingContext),
+		UnsuccessfulJobSchedulingContexts: make(map[string]*JobSchedulingContext),
+	}
+}
+
+const maxPrintedJobIdsByReason = 1
+
+func (qctx *QueueSchedulingContext) String() string {
+	var sb strings.Builder
```
d80tb7, line 173: does this need a lock?

d80tb7 (changes requested):
I really like the concept of the PR and the scheduling reports look great, I've just got some concerns around the use of locks here, which we probably should discuss.

internal/scheduler/reports.go
severinson: Agreed. I've changed it to use `atomic.Pointer` + deep-copy.

internal/scheduler/schedulercontexts.go
severinson: Those mutexes are left over from when I was planning out how to parallelise the scheduler (which I never ended up doing). I've removed these mutexes.

internal/scheduler/schedulercontexts.go
severinson: Agreed. Would prefer to do that in a follow-up PR.

internal/scheduler/schedulercontexts.go
severinson: No, the contexts themselves shouldn't be mutated while this function is called.

internal/scheduler/reports.go
d80tb7: Ok great- yeah we should separate storage from service but we can do that later.  I'd consider adding a comment to the function saying it's a user-facing RPC call, as I don't think this is obvious right now.

internal/scheduler/reports.go
```diff
@@ -5,442 +5,394 @@ import (
...
 	}
+	rv := &SchedulingContextRepository{
+		mostRecentJobSchedulingContextByExecutorByJobId: jobSchedulingContextByExecutorByJobId,
+		executorIds: make(map[string]bool),
+	}
+	mostRecentSchedulingContextByExecutor := make(SchedulingContextByExecutor)
+	mostRecentSuccessfulSchedulingContextByExecutor := make(SchedulingContextByExecutor)
+	mostRecentQueueSchedulingContextByExecutorByQueue := make(map[string]QueueSchedulingContextByExecutor)
+	mostRecentSuccessfulQueueSchedulingContextByExecutorByQueue := make(map[string]QueueSchedulingContextByExecutor)
+	sortedExecutorIds := make([]string, 0)
```
d80tb7, line 69: I don't think this is correct.  You need the state updae to be appplied atomically otherwise e.g. the reader may get a list of executorIds that don't correspond to the map keys.  I think what you want is a struct that contains all of the state and a single atomic reference which holds it.

severinson: That should be safe. Only `sortedExecutorIds` is used by the readers so it's fine for it to be briefly inconsistent with the map.

internal/scheduler/reports.go
severinson: Done

PR #2289, opened 2023-03-22
Lookout v2 - display Job Logs in sidebar

carlocamurri:
Similar to how it worked in V1:
* Load logs every second
* Optionally display timestamps
* Container and run selection
* Defaults to first non-init container

https://user-images.githubusercontent.com/14963658/227289313-dd1f312e-ffd0-4545-bc9d-0ef0e5318270.mov


samclark:
Looks great. One comment would be do you think we should have an option to not "auto reload logs". Maybe we could default to not auto-reloading the log and then option to "follow logs"?

carlocamurri:
> Looks great. One comment would be do you think we should have an option to not "auto reload logs". Maybe we could default to not auto-reloading the log and then option to "follow logs"?

Don't think this is necessary for three reasons:
* The updated logs should always be displayed, requiring users to manually reload is awkward (and a complaint of the old design)
* If an error does occur, no further requests are made automatically, and a reload button is shown to manually retry. This is to avoid spamming the UI with errors.
* Scrolling up ensures that the view doesn't get automatically scrolled down on new reloads, so you can investigate older logs without your view changing

JamesMurkin (approved):

PR #2298, opened 2023-03-29
Job Service: add Postgres as alternate backend db

### LINKED ISSUE #2163: JobService - Allow jobservice to use other SQL databases
Our jobservice is used for caching of events for jobs.  We currently use SQLLITE but we have interest in using postgres.  

We should allow our database access to be pluggable.  

We should not break the existing usage of sqllite as that is currently used in production.

Jobservice code in question is here: https://github.com/armadaproject/armada/blob/master/internal/jobservice/repository/sql_job_service.go

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/AR-114) by [Unito](https://www.unito.io)

richscott:
These changes add support for using Postgres as a backend database for the Job Service. Selection of using either (the traditional) `sqlite` or `postgres` is done in `config/jobservice/config.yaml` by changing the `databaseType` setting. Note that if you select `postgres` you will probably have to also modify the entries in the `postgres` block of that file.

Note that the `makefile` has been modified slightly to run the Job Service db repository package tests separately from the other tests, and it runs it twice, once for SQLite and once for Postgres. Running `make tests` will exercise both, using a container for Postgres, and a local file for SQLite. The configuration structures and YAML config content was borrowed from the Lookout database code.

Fixes #2163

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/AR-29) by [Unito](https://www.unito.io)


kannon92:
Nice work! I think your airflow tests are failing because maybe it’s assuming Postgres but those containers aren’t started?

kannon92 (approved):

PR #2305, opened 2023-03-31
Airflow deferrable operator

### LINKED ISSUE #1992: Build an Async python client
To improve Airflow scalability a useful technique is to use a deferred operator.

https://airflow.apache.org/docs/apache-airflow/stable/concepts/deferring.html

"Standard [Operators](https://airflow.apache.org/docs/apache-airflow/stable/concepts/operators.html) and [Sensors](https://airflow.apache.org/docs/apache-airflow/stable/concepts/sensors.html) take up a full worker slot for the entire time they are running, even if they are idle; for example, if you only have 100 worker slots available to run Tasks, and you have 100 DAGs waiting on a Sensor that’s currently running but idle, then you cannot run anything else - even though your entire Airflow cluster is essentially idle. reschedule mode for Sensors solves some of this, allowing Sensors to only run at fixed intervals, but it is inflexible and only allows using time as the reason to resume, not anything else.

This is where Deferrable Operators come in. A deferrable operator is one that is written with the ability to suspend itself and free up the worker when it knows it has to wait, and hand off the job of resuming it to something called a Trigger. As a result, while it is suspended (deferred), it is not taking up a worker slot and your cluster will have a lot less resources wasted on idle Operators or Sensors."

An example of this is the standard KubernetesPodOperator at
https://github.com/apache/airflow/blob/main/airflow/providers/cncf/kubernetes/operators/kubernetes_pod.py#L129

and its async counterpart at:
https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py#L22

They can practically be swapped for one another in code.

Can we start a conversation to see if it makes sense to build an async version of armadaoperator too?

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-100) by [Unito](https://www.unito.io)

ClifHouck:
Closes  #1992

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-24) by [Unito](https://www.unito.io)


ClifHouck:
rebased

suprjinx (approved):

client/python/armada_client/client.py
```diff
@@ -32,14 +30,10 @@ class ArmadaClient:
...
                     for more information.
-    :param max_workers: number of cores for thread pools, if unset, defaults
-                        to number of CPUs
     :return: an Armada client instance
     """
 
-    def __init__(self, channel, max_workers: Optional[int] = None):
-        self.executor = ThreadPoolExecutor(max_workers=max_workers or os.cpu_count())
-
+    def __init__(self, channel):
```
kannon92, line 36: Why did we make this change?

kannon92: I don't really know why it was there in the first place but curious why it was removed.

client/python/tests/unit/test_asyncio_client.py
```diff
@@ -0,0 +1,324 @@
...
+        "test": "test",
+    }
+    required_node_labels = {
+        "test": "test",
+    }
+
+    ingress = submit_pb2.IngressConfig()
+    services = submit_pb2.ServiceConfig()
+
+    print(type(aio_client))
```
kannon92, line 81: We should remove this.

third_party/airflow/armada/operators/armada_deferrable.py
```diff
@@ -0,0 +1,315 @@
...
+    :return: A deferrable armada operator instance.
+    """  # noqa
+
+    def __init__(
+        self,
+        name: str,
+        armada_channel_args: dict,
+        job_service_channel_args: dict,
+        armada_queue: str,
+        job_request_items,
```
kannon92, line 77: Can you add a type here?  

It would also be ideal to have the shape of the armada_channel and job_service_channel_args.  dict is basically a nontype.

kannon92 (approved):
Mostly nits.  You can do them in a followup if you want to get this released.

client/python/armada_client/client.py
ClifHouck: `self.executor` is not/never used, therefore neither is `max_workers`.

PR #2308, opened 2023-03-31
[Pulsar Scheduler] Don't Schedule Jobs When Executors Are Lagging Behind

d80tb7:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-21) by [Unito](https://www.unito.io)


JamesMurkin:
@d80tb7 

Is this good to review? I made a minor change to the unit test as it was failing - but looks like it was just a copy paste error

d80tb7:
@JamesMurkin yup I think this one is god to review

config/scheduler/config.yaml
```diff
@@ -75,6 +75,7 @@ scheduling:
     memory: 1.0
     cpu: 1.0
   maximumJobsToSchedule: 5000
+  maxUnacknowledgedJobsPerExecutor: 2500
```
JamesMurkin, line 78: Possibly we should re-adjust numbers in config.

We allow scheduling 5000 per round, but only allow 2500 lag - which can easily happen in a single round

Possibly maximumJobsToSchedule should be 1000 instead

JamesMurkin (approved):

PR #2321, opened 2023-04-03
Prototype Workflow API

severinson:
Prototype API for expressing heterogeneous batch workflows that require coordination between multiple applications, with mixed scheduling constraints (gang and otherwise) and elasticity.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-16) by [Unito](https://www.unito.io)


kannon92 (commented):
Could we utilize an open source definition for workflows rather than our own API?

argo workflows? Cwl? We could use something like Couler to generate airflow or Argo workflow dags? 

is there a proposal for this work?

severinson:
> Could we utilize an open source definition for workflows rather than our own API?
> 
> argo workflows? Cwl? We could use something like Couler to generate airflow or Argo workflow dags?
> 
> is there a proposal for this work?

The question is indeed whether there already exists something that could work for us. Answering that question requires finding a minimal solution to the problem at hand and then determining whether that solution can already be expressed in some other way.

michelsumbul:
Hi @severinson ,
Prototyping stuff is great to have a good understand of what good look like.
Before we merge this to master can we have a chat and complete description of what we try to do here? Prototyping should not require merge to master.
If its enabling "flat" gang scheduling fine but if its allow user to have workflow, like Dag in airflow or dollar$ with task being started only when previous one is finish, this look like a bad idea as it will turn armada to an orchestrator and not anymore only as a scheduler. If user want dependency between tasks (t1 do this, t2 do that, etc....) we should look at other tool.

Michel

PR #2326, opened 2023-04-03
Adding latest images

### LINKED ISSUE #2242: Armada and armada-operator should push a `latest` tag to dockerhub
currently we only push particular SHA1s, but it would be good to have support for 'latest' tag.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-43) by [Unito](https://www.unito.io)

Sharpz7:
Closes #2242


kannon92:
https://www.howtogeek.com/devops/understanding-dockers-latest-tag/

I don’t think we should do this PR. Sounds like latest is a bit confusing and I’m not sure I like the idea of overwriting a docker image on every push.

PR #2334, opened 2023-04-04
feat: refactor armada server chart and regenerate docs

dejanzele:
Refactor Armada Server chart
Regenerate docs using helm-docs

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-10) by [Unito](https://www.unito.io)


suprjinx (approved):

deployment/armada/values.yaml
```diff
@@ -1,34 +1,126 @@
...
+# -- If a port number is specified, REST server will be exposed on that node port
+#httpNodePort: 50051
+
+# -- If specified, custom ServiceAccount name will be attached to Armada Server Deployment resource and the default ServiceAccount will not be created
+customServiceAccount: ""
+# -- More info here https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
+# -- Additional ServiceAccount properties (e.g. automountServiceAccountToken, imagePullSecrets, etc.)
+serviceAccount: {}
 
 applicationConfig:
```
severinson, line 86: Do we need any config here? Check if we do anything different from the default config.

deployment/armada/values.yaml
```diff
@@ -1,34 +1,126 @@
+nameOverride: ""
+fullnameOverride: ""
+
 image:
+  # image repository
   repository: gresearchdev/armada-server
-  tag: LATEST
+  # image tag
+  tag: v0.3.39
```
severinson, line 8: Usa a variable? All components should normally be set to the same version.

deployment/armada/values.yaml
dejanzele: We do the following thing, which I do not think is very safe and intuitive: https://github.com/armadaproject/armada/blob/master/deployment/armada/templates/serviceaccount.yaml#L3-L11

dejanzele: Do you think it is better to omit any config in the values.yaml and have docs on what is the minimal app configuration?

deployment/armada/values.yaml
dejanzele: What did we agree for the tag?
Should we keep LATEST here and in the Chart YAML like it was before, with the note that LATEST isn't a valid tag for Helm charts and the Chart is not well-formed?

When we release a versioned chart, we do version substitution there

deployment/armada/values.yaml
severinson: I think if we already have defaults in config.yaml that make sense, we shouldn't re-specify those values here. Only add to the helm files necessary overrides.

deployment/armada/values.yaml
severinson: Could we have the tag in the helm files be a variable and then set the values in the values.yaml to the latest tag when we release the charts?

dejanzele: Helm doesn't allow templating in the Chart.yaml sadly

severinson (dismissed):

severinson (dismissed):

deployment/armada/templates/_helpers.tpl
```diff
@@ -19,6 +19,28 @@
 {{- printf "%s%s" ( include "armada.users.name" .) ".yaml" -}}
 {{- end }}
 
+{{- define "armada.resource-quantity" -}}
+    {{- $value := . -}}
+    {{- $unit := 1.0 -}}
+    {{- if typeIs "string" . -}}
+        {{- $base2 := dict "Ki" 0x1p10 "Mi" 0x1p20 "Gi" 0x1p30 "Ti" 0x1p40 "Pi" 0x1p50 "Ei" 0x1p60 -}}
```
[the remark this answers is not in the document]
dejanzele: Definitely more easy in Go than Helm, Operator handles this more gracefully

severinson (approved):

PR #2343, opened 2023-04-06
[armada-scheduler] Avoid retrying jobs on the same node

JamesMurkin:
This PR adds the functionality to avoid nodes on retrying jobs

It will only avoid a node if:

* The lease was returned
* The runAttempted flag was set on the lease return

This is so if the lease returned for reasons that didn't involve then node - we don't avoid the node

The node avoidance is implemented using affinities and setting a node anti-affinity

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-142) by [Unito](https://www.unito.io)


internal/scheduler/kubernetesobjects/affinity/affinity.go
```diff
@@ -0,0 +1,64 @@
+package affinity
+
+import (
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/armadaproject/armada/internal/common/util"
+)
+
+func AddNodeAntiAffinity(affinity *v1.Affinity, labelName string, labelValue string) {
+	if affinity == nil {
```
d80tb7, line 10: why is it ok to silently ignore a nil here?

internal/scheduler/scheduler.go
```diff
@@ -273,13 +285,39 @@ func (s *Scheduler) syncState(ctx context.Context) ([]*jobdb.Job, error) {
 		jobsToUpdateById[jobId] = job
 	}
 
+	// Ensure queued jobs being retried have node anti affinities for any nodes they have already run on
+	for _, job := range jobsToUpdateById {
```
d80tb7, line 289: can this be simplified?  We seem to iterate over `jobsToUpdateById` twice and do much the same thing in each case?

It feels like we could iterate over once and apply the anti affinity in one place.

internal/scheduler/scheduler.go
```diff
@@ -273,13 +285,39 @@ func (s *Scheduler) syncState(ctx context.Context) ([]*jobdb.Job, error) {
...
+
 	// Jobs with no active runs are queued.
 	for _, job := range jobsToUpdateById {
 		// Determine if the job needs to be re-queued.
 		// TODO: If have an explicit queued message, we can remove this code.
 		run := job.LatestRun()
 		desiredQueueState := false
-		requeueJob := run != nil && run.Returned() && job.NumReturned() <= s.maxLeaseReturns
+		requeueJob := run != nil && run.Returned() && job.NumAttempts() < s.maxAttemptedRuns
+		if requeueJob && !job.Queued() && run.Returned() && run.RunAttempted() {
```
d80tb7, line 308: can we simplify this?  To work out if the job is going to fall into this, we've got 6 bools to check!

To tie this in with my previous comment- could we not have a single loop and do something like:
```
if job.Queued || requeueJob {
  // Ensure affinity logic is set
}
```

internal/scheduler/scheduler.go
```diff
@@ -487,6 +558,20 @@ func (s *Scheduler) generateUpdateMessagesFromJob(job *jobdb.Job, jobRunErrors m
 		} else if lastRun.Failed() {
 			job = job.WithFailed(true).WithQueued(false)
 			runError := jobRunErrors[lastRun.Id()]
+			if lastRun.Returned() {
+				errorMessage := fmt.Sprintf("Maximum number of attempts (%d) reached - this job will no longer be retried", s.maxAttemptedRuns)
+				if job.NumAttempts() < s.maxAttemptedRuns {
+					errorMessage = fmt.Sprintf("Job was attempeted %d times, and has been tried once on all nodes it can run on - this job will no longer be retried", job.NumAttempts())
```
d80tb7, line 564:
```suggestion
					errorMessage = fmt.Sprintf("Job was attempted %d times, and has been tried once on all nodes it can run on - this job will no longer be retried", job.NumAttempts())
```

internal/scheduler/scheduler.go
```diff
@@ -310,6 +348,39 @@ func (s *Scheduler) syncState(ctx context.Context) ([]*jobdb.Job, error) {
...
+func (s *Scheduler) addNodeAntiAffinitiesForAttemptedRunsIfSchedulable(job *jobdb.Job) (*jobdb.Job, bool, error) {
+	schedulingInfoWithNodeAntiAffinity, err := s.createSchedulingInfoWithNodeAntiAffinityForAttemptedRuns(job)
+	if err != nil {
+		return nil, false, err
+	}
+	podSchedulingRequirement := PodRequirementFromJobSchedulingInfo(schedulingInfoWithNodeAntiAffinity)
+	if podSchedulingRequirement == nil {
+		return nil, false, fmt.Errorf("no pod scheduling requirement found for job %s", job.GetId())
+	}
+	isSchedulable, _ := s.submitChecker.CheckPodRequirements(podSchedulingRequirement)
```
d80tb7, line 380: This is a bit dodgy- I think this function is being called from `syncState` which means that all replicas of the scheduler are executing this regardless of whether they are leader.  Previously all the code called by this function was deterministic so we could guarantee that all replicas would end up in the same state.  This code breaks that assumption which means we could lose a job.  For example:
- Leader executes this code and determines job is ok to requeue
- Follower executes this code and determines job should fail (and marks it as such)
- Leader leases the job to the executor
- Follower Becomes Leader and fails the job
- Job run is now orphaned

TBF this whole bit of code is awkward because the point of `syncState` is to make the internal in-memory model look like the database.  The re-queuing of jobs doesn't fit nicely into that because we're not just taking the state from the db, but rather using scheduler-specific business logic to modify the state and at that point you run the risk that different scheduler instances may make different decisions. I've gone over a couple of different ways of sorting this and the best I can come up with is:
- Always requeue  in `syncstate`, but leave the job run as failed
- Move the requeing logic into `generateUpdateMessages` which can fail the job if it determines it shouldn't be requeued.

I think this works, because it means that only the leader will be checking whether something can be requeued.

JamesMurkin: I agree with this and we can move it around

Is there a reason we don't send a requeue message in generateUpdateMessages - and remove all the business logic from syncstate.

Then only leader can decide if we requeue or fail the job and the others just get their state from the db.

d80tb7: I think a `requeue` message is probably the best solution, it would also make life easier for e.g. lookout which could then mark the state correctly without having to gues what the scheduler has decided. 

The only slightly tricky bit is what to store in the schedulerDb as a bool won't be enough.  I think an int giving the number times it has been requeued should be enough, but I haven't thought through this enough to be sure.

d80tb7: I do wonder if the requeue message could include the updated scheduling requirements.  That would be nice from the Pov of ensuing that all schedulers agree what nodes are being excluded.

JamesMurkin: Are you envisioning that we'd conditionally exclude nodes?

Currently its either exclude all nodes we've tried on or fail - so it shouldn't be ambiguous  - however I'm happy to be proven wrong.

Although including the scheduling requirements on the requeue message would work and be pretty flexible

d80tb7: So you're right- at the moment the logic is unambiguous so each scheduler can determine which nodes to exclude purely from the runs.  My thinking behind suggesting an explicit update to the scheduling requirements is mainly based around the feeling that the fewer decisions non-leader schedulers make, the less chance there is of them getting out of sync.  There's a few secondary benefits of this approach too- e.g. something like lookout could now show excluded nodes, but I'm not sure these are compelling enough to make a change.

The other thing I'm thinking of is how we integrate this with the hash-based scheduling optimization, which should go a long way toward mitigating the performance of `scheduleMany` as discussed in #2319.  Ideally we don't always compute hashes in `syncState` because this will involve computing `len(queue)` hashes every time the leader fails over.  Instead, I was hoping to calculate them in the ingester and a `requeued` message with an updated schedulingRequirements would be useful for this I think.

JamesMurkin: I think that is all reasonable and won't be a lot more work.

In terms of implementation - would the plan be to just updated the scheduling_info column on the job table - on each `requeue` message that changes the scheduling  requirements?
 - Or would we want to record these updates elsewhere?
 - One pain point here is working out when the scheduling_info has changed in a cheap way - maybe this would be a reason to implement the hash in this PR

Would you be happy if I updated this PR to:
 - Add `requeue` event
 - Move all requeue logic to generateUpdateMessages (i.e send requeue event) and simplify `syncState`

Unsure if I want to also add a scheduling info hash in this PR - it depends the answer to the above

d80tb7: > Add requeue event
> Move all requeue logic to generateUpdateMessages (i.e send requeue event) and simplify syncState

That seems good.  My suggestion for working out whether scheduling info has changed would be to add a seq no to scheduling info (both the table and the in-mem representation).  Every time you generate a new scheduling infor, increment the seq no and include it in the  `requeue` event.  The ingester should only overwrite the scheduling info if the seq num is higher (I think you can do this relatively cheaply with an ON CONFLICT), likewise the in memory representation should only be overwritten if the seq num is higher.

With regard for hashing- I think a tentative plan would be:
- Have a hash col that lives alongside the scheduling info and is versioned with it
- Scheduler ingester calculates and inserts the initial hash on the submit message
- Requeue message needs to include a hash (the leader scheduler will have to calculate the has when it requeues). The logic for updating the hash (both for the ingester and for follower schedulers) is identical to updating the scheduling info.

If you can get the hashing in that would be great, but if not it shouldn't be too hard to get it in later.

d80tb7: > Add requeue event
> Move all requeue logic to generateUpdateMessages (i.e send requeue event) and simplify syncState

That seems good.  My suggestion for working out whether scheduling info has changed would be to add a seq no to scheduling info (both the table and the in-mem representation).  Every time you generate a new scheduling infor, increment the seq no and include it in the  `requeue` event.  The ingester should only overwrite the scheduling info if the seq num is higher (I think you can do this relatively cheaply with an ON CONFLICT), likewise the in memory representation should only be overwritten if the seq num is higher.

With regard for hashing- I think a tentative plan would be:
- Have a hash col that lives alongside the scheduling info and is versioned with it
- Scheduler ingester calculates and inserts the initial hash on the submit message
- Requeue message needs to include a hash (the leader scheduler will have to calculate the has when it requeues). The logic for updating the hash (both for the ingester and for follower schedulers) is identical to updating the scheduling info.

If you can get the hashing in that would be great, but if not it shouldn't be too hard to get it in later.

JamesMurkin:
Hi @d80tb7 

So I've made the changes:
 - Requeue jobs with an explicit event (which can also change scheduling info and is versioned)
 - Have an explicit queued state in the jobs table - so the scheduler doesn't need to work out the queued state each time
 
Functionally this works, I know there are some points that need tidying:
 - Naming could be improved
 - The update queries are not efficient. sqlc isn't helping me out here - but I can just write an efficient native query relatively easily (luckily postgres has a nice support for this)
 - A few other minor points
 
However before I finished off - I just wanted your overall opinion on if I'm butchering things too much

internal/scheduleringester/dbops.go
```diff
@@ -254,6 +303,14 @@ func (a MarkJobsCancelled) CanBeAppliedBefore(b DbOperation) bool {
 	return !definesJob(a, b)
 }
 
+func (a UpdateJobSchedulingInfo) CanBeAppliedBefore(b DbOperation) bool {
+	return !definesJob(a, b)
+}
+
+func (a UpdateJobQueuedState) CanBeAppliedBefore(b DbOperation) bool {
```
d80tb7, line 310: this feels a bit wrong in that moving a `jobrequeued` event before the job failure (or even the job lease) seems incorrect.  In practice I think this will work because of our versioning- but it might be worth speaking to Albin to see how we could tighten this up.

internal/scheduler/scheduler.go
```diff
@@ -485,18 +524,68 @@ func (s *Scheduler) generateUpdateMessagesFromJob(job *jobdb.Job, jobRunErrors m
...
-			runError := jobRunErrors[lastRun.Id()]
-			jobErrors := &armadaevents.EventSequence_Event{
-				Created: s.now(),
-				Event: &armadaevents.EventSequence_Event_JobErrors{
-					JobErrors: &armadaevents.JobErrors{
-						JobId:  jobId,
-						Errors: []*armadaevents.Error{runError},
+			requeueJob := lastRun.Returned() && job.NumAttempts() < s.maxAttemptedRuns
+
+			if requeueJob && lastRun.RunAttempted() {
```
d80tb7, line 529: When we become leader (either on startup or through failover) we pass all jobs to this function.  In this case we will get jobs that have already been requeued fall into this code branch.  In that case I *think* the code will do the right thing- i.e. it'll end up with the jobs queued with the correct anti affinity set but it might be either writing a test to be sure, or seeing if we can already detect the job is in the correct state and short circuit.

internal/scheduleringester/instructions.go
```diff
@@ -216,7 +248,14 @@ func (c *InstructionConverter) handleJobRunErrors(jobRunErrors *armadaevents.Job
 				JobID: jobId,
 				Error: bytes,
 			}
-			markRunsFailed[runId] = &JobRunFailed{LeaseReturned: runError.GetPodLeaseReturned() != nil}
+			runAttempted := false
```
d80tb7, line 251: should this default to true?

d80tb7:
> Hi @d80tb7
> 
> So I've made the changes:
> 
>     * Requeue jobs with an explicit event (which can also change scheduling info and is versioned)
> 
>     * Have an explicit queued state in the jobs table - so the scheduler doesn't need to work out the queued state each time
> 
> 
> Functionally this works, I know there are some points that need tidying:
> 
>     * Naming could be improved
> 
>     * The update queries are not efficient. sqlc isn't helping me out here - but I can just write an efficient native query relatively easily (luckily postgres has a nice support for this)
> 
>     * A few other minor points
> 
> 
> However before I finished off - I just wanted your overall opinion on if I'm butchering things too much
> 
> Also I've included JobSchedulingInfo in the JobRequeued event - however I've for now imported it from our internal schedulerobjects proto
> 
> We could make duplicate proto and convert them? Or move JobSchedulingInfo. It is a bit awkward

Yeah this all looks good- I made a few coments- none of them serious I don't think.  Even the comment about the reordering of events in the ingester I don't think will cause any issues.

Re the proto- I think for now that's fine.  My personal view, which  admittedly isn't quite crystallized yet, is something like:
- all the proto types in schedulerObjects should be treated as types used for serialisation only- they shouldn't be used in side the scheduler itself.
- as internal serialised types they should live alongside the proto used for armadamessages which means importing one from the other should be fine.
- The Scheduler should use it's own internal types which are probably largely mirror the proto types, but have the advantage that we can control their code so as to store things as efficiently as possible for the scheduler.
-

internal/scheduleringester/dbops.go
d80tb7: actually I just had a better look at the code- I think what you've done here is fine and the real issue is with the ingester in that we don't perform all updates for a given batch inside a transaction.  I'll make a separate PR for that.

internal/scheduleringester/instructions.go
JamesMurkin: Probably yes

It is a bit awkward because the value only matters if lease was returned - so its value is somewhat irrelevant for other cases

internal/scheduleringester/dbops.go
d80tb7: If we put everything inside a transaction then I think it will work as expected. I think what is happening right now is that the state after all operations have been inserted is correct, but intermediate states are not correct.  If we move everything into a single transaction then this will prevent readers from seeing these intermediate states and we should be fine.  Note that this issue applies to various operations, not just the new new one you've added here.

The alternative would be to add some more complex logic around how operations may be re-ordered.  I'm not a big fan of this because the current mechanism doesn't give us enough flexibility to be able to easily represent this.  E.g. right now we can easily prevent this operation being moved backwards, but we can't easily prevent it being moved forwards (i.e. other operations being moved backwards, passing this one).  Changing this would be something of an effort and I think not necessary, as long as we ensure the final database state is correct (which we can if we use a single transaction).

tagging @severinson here in case he has an opinion.

internal/scheduleringester/instructions.go
d80tb7: In practice it probably doesn't matter too much- although I'd argue that in general it's safer to assume the run was attempted if we can't determine whether it was.

internal/scheduler/scheduler.go
JamesMurkin: I'm relatively confident it will be sufficient to check it is queued - however we could also validate the affinity is as we expect

However this would only handle if we managed to get into an unexpected state - but maybe that isn't a bad thing

internal/scheduler/database/migrations/001_initialize_schema.up.sql
```diff
@@ -26,6 +30,8 @@ CREATE TABLE jobs (
     submit_message bytea NOT NULL,
     -- JobSchedulingInfo message stored as a proto buffer.
     scheduling_info bytea NOT NULL,
+    -- The current version of the JobSchedulingInfo, used to prevent JobSchedulingInfo state going backwards on even replay
```
d80tb7, line 33:
```suggestion
    -- The current version of the JobSchedulingInfo, used to prevent JobSchedulingInfo state going backwards even on replay
```

internal/scheduler/kubernetesobjects/affinity/affinity.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/armadaproject/armada/internal/common/util"
+)
+
+func AddNodeAntiAffinity(affinity *v1.Affinity, labelName string, labelValue string) error {
+	if affinity == nil {
+		return fmt.Errorf("failed to add not anti affinity, as provided affinity is nil")
```
d80tb7, line 13:
```suggestion
		return errors.Errorf("failed to add node anti affinity, as provided affinity is nil")
```

internal/scheduler/scheduler.go
```diff
@@ -310,8 +307,45 @@ func (s *Scheduler) syncState(ctx context.Context) ([]*jobdb.Job, error) {
 	return jobsToUpdate, nil
 }
 
+func (s *Scheduler) createSchedulingInfoWithNodeAntiAffinityForAttemptedRuns(job *jobdb.Job) (*schedulerobjects.JobSchedulingInfo, error) {
+	newSchedulingInfo := proto.Clone(job.JobSchedulingInfo()).(*schedulerobjects.JobSchedulingInfo)
+	newSchedulingInfo.Version = job.JobSchedulingInfo().Version + 1
+	podSchedulingRequirement := PodRequirementFromJobSchedulingInfo(newSchedulingInfo)
+	if podSchedulingRequirement == nil {
+		return nil, fmt.Errorf("no pod scheduling requirement found for job %s", job.GetId())
```
d80tb7, line 315:
```suggestion
		return nil, errors.Errorf("no pod scheduling requirement found for job %s", job.GetId())
```

internal/scheduler/scheduler.go
```diff
@@ -310,8 +307,45 @@ func (s *Scheduler) syncState(ctx context.Context) ([]*jobdb.Job, error) {
...
+}
+
+func (s *Scheduler) addNodeAntiAffinitiesForAttemptedRunsIfSchedulable(job *jobdb.Job) (*jobdb.Job, bool, error) {
+	schedulingInfoWithNodeAntiAffinity, err := s.createSchedulingInfoWithNodeAntiAffinityForAttemptedRuns(job)
+	if err != nil {
+		return nil, false, err
+	}
+	podSchedulingRequirement := PodRequirementFromJobSchedulingInfo(schedulingInfoWithNodeAntiAffinity)
+	if podSchedulingRequirement == nil {
+		return nil, false, fmt.Errorf("no pod scheduling requirement found for job %s", job.GetId())
```
d80tb7, line 341:
```suggestion
		return nil, false, errors.Errorf("no pod scheduling requirement found for job %s", job.GetId())
```

internal/scheduler/scheduler.go
```diff
@@ -484,19 +526,69 @@ func (s *Scheduler) generateUpdateMessagesFromJob(job *jobdb.Job, jobRunErrors m
...
+						job = jobWithAntiAffinity
+					} else {
+						// If job is not schedulable with anti-affinity added. Do not requeue it and let it fail
+						requeueJob = false
+					}
+				}
+			}
+
+			if requeueJob {
+				job = job.WithUpdatedRun(lastRun.WithFailed(false))
```
d80tb7, line 547: do we need to do this? I previously had to set `lastRun.Failed = false` because I was doing this in `syncState` and I needed to prevent the job from being failed later.  I think now we are handling the requeing in the correct place so we can indeed treat the job as failed.

internal/scheduler/scheduler.go
```diff
@@ -484,19 +526,69 @@ func (s *Scheduler) generateUpdateMessagesFromJob(job *jobdb.Job, jobRunErrors m
...
 					},
-				},
+				}
+
+				events = append(events, requeueJobEvent)
+			} else {
+				runError := jobRunErrors[lastRun.Id()]
+				job = job.WithFailed(true).WithQueued(false)
+				if lastRun.Returned() {
+					errorMessage := fmt.Sprintf("Maximum number of attempts (%d) reached - this job will no longer be retried", s.maxAttemptedRuns)
```
d80tb7, line 567:
```suggestion
					errorMessage := errors.Sprintf("Maximum number of attempts (%d) reached - this job will no longer be retried", s.maxAttemptedRuns)
```

internal/scheduler/scheduler.go
```diff
@@ -484,19 +526,69 @@ func (s *Scheduler) generateUpdateMessagesFromJob(job *jobdb.Job, jobRunErrors m
...
+					if job.NumAttempts() < s.maxAttemptedRuns {
+						errorMessage = fmt.Sprintf("Job was attempeted %d times, and has been tried once on all nodes it can run on - this job will no longer be retried", job.NumAttempts())
```
d80tb7, line 569:
```suggestion
						errorMessage = errors.Sprintf("Job was attempeted %d times, and has been tried once on all nodes it can run on - this job will no longer be retried", job.NumAttempts())
```

internal/scheduleringester/schedulerdb.go
```diff
@@ -81,6 +83,52 @@ func (s *SchedulerDb) WriteDbOp(ctx context.Context, op DbOperation) error {
 				return errors.WithStack(err)
 			}
 		}
+	case UpdateJobSchedulingInfo:
+		args := make([]interface{}, 0, len(o)*3)
+		argMarkers := make([]string, 0, len(o))
+
+		currentIndex := 1
+		for key, value := range o {
```
d80tb7, line 91: this might be simpler as a `pgx.batch` query (https://github.com/jackc/pgx/blob/master/batch_test.go#L30). I suspect the performance won't be much worse either.

internal/scheduleringester/schedulerdb.go
```diff
@@ -81,6 +83,52 @@ func (s *SchedulerDb) WriteDbOp(ctx context.Context, op DbOperation) error {
...
+			`update jobs as j set  scheduling_info = updated.scheduling_info, scheduling_info_version = updated.scheduling_info_version
+				 from (values %s) as updated(job_id, scheduling_info, scheduling_info_version)
+                 where j.job_id = updated.job_id and updated.scheduling_info_version > j.scheduling_info_version`, argMarkersString)
+
+		_, err := s.db.Exec(ctx, updateJobInfoSqlStatement, args...)
+		if err != nil {
+			return errors.WithStack(err)
+		}
+	case UpdateJobQueuedState:
+		args := make([]interface{}, 0, len(o)*3)
```
d80tb7, line 110: as above- I think this could be a `pgx.batch` query.

d80tb7 (approved):
Raised some comments- but nothing super important.

PR #2367, opened 2023-04-13
Add scheduler to go-releaser and github action to publish images

suprjinx:
Fixes #

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-171) by [Unito](https://www.unito.io)


.github/workflows/build-release-images.yml
```diff
@@ -42,6 +42,8 @@ jobs:
           docker save armada-lookout-v2 | gzip > docker-images/armada-lookout-v2.tar.gz
           docker save armada-lookout-ingester-v2 | gzip > docker-images/armada-lookout-ingester-v2.tar.gz
           docker save armada-event-ingester | gzip > docker-images/armada-event-ingester.tar.gz
+          docker save armada-scheduler | gzip > docker-images/armada-scheduler.tar.gz
+          docker save armada-scheduler-ingester | gzip > docker-images/armada-scheduler-ingester.tar.gz
```
richscott, line 46: We don't have to do this now for this PR, but given that the image names here are consistent substrings of the constructed tarfile, we should maybe move these into a single Bash array, and just loop over it, to make it more concise and avoid any future artifact-name exceptions.

richscott (approved):

kannon92 (dismissed):

richscott (approved):

PR #2371, opened 2023-04-14
Fix flakey leader election test

JamesMurkin:
Make the leader election state changes slightly slower so test is more reliable

It seems the test is reliable on its own - but when you run many at once sometimes a state transition happens too quickly and is missed by the test.

I haven't looked super deeply - probably just lack of cpu causing the polling to happen less often than we think.

Slowing down the state transitions seems to make the test reliable - at least I can't make it fail locally anymore

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-174) by [Unito](https://www.unito.io)


JamesMurkin:
#2371

d80tb7 (approved):

PR #2373, opened 2023-04-14
Use inline cache for DOCKER_BUILDX_CACHE_FROM by default

ClifHouck:
#### What type of PR is this?

Build Fix

Goreleaser build is broken for some because `registry` is used as the default for `DOCKER_BUILDX_CACHE_FROM` but no registry is specified by default. Change this to `inline` so whatever is built locally is used by default.

Fixes #

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-176) by [Unito](https://www.unito.io)


kannon92 (approved):
Nice find!

richscott (approved):

PR #2374, opened 2023-04-14
Refactor the scheduler

severinson:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-177) by [Unito](https://www.unito.io)


internal/armada/server/lease.go
```diff
@@ -463,102 +467,56 @@ func (q *AggregatedQueueServer) getJobs(ctx context.Context, req *api.StreamingL
...
+		&SchedulerJobRepositoryAdapter{
+			r: q.jobRepository,
+		},
+		nodeDb,
+		nodeIdByJobId,
+		jobIdsByGangId,
+		gangIdByJobId,
+	)
+	result, err := sch.Schedule(
+		ctxlogrus.ToContext(
```
d80tb7, line 500: is it right to use the background context here?  I though previously we used the grpc context which meant a) we would timeout, b) we would automatically stop scheduling if the grpc channel died.

internal/armada/server.go
```diff
@@ -344,64 +338,13 @@ func validateCancelJobsBatchSizeConfig(config *configuration.ArmadaConfig) error
 }
 
 func validatePreemptionConfig(config configuration.PreemptionConfig) error {
-	if !config.Enabled {
-		return nil
-	}
-
-	// validate that the default priority class is in the priority class map
+	// Chech that the default priority class is in the priority class map.
```
d80tb7, line 341:
```suggestion
	// Check that the default priority class is in the priority class map.
```

internal/common/slices/slices.go
```diff
@@ -60,6 +61,32 @@ func Flatten[S ~[]E, E any](s []S) S {
...
+func Concatenate[S ~[]E, E any](s ...S) S {
+	return Flatten(s)
+}
+
+// Shuffle shuffles s.
+func Shuffle[S ~[]E, E any](s ...S) {
+	rand.Shuffle(len(s), func(i, j int) { s[i], s[j] = s[j], s[i] })
+}
+
+// Unique returns a copy of s with duplicate elements removed, keeping only the first occurence.
```
d80tb7, line 74:
```suggestion
// Unique returns a copy of s with duplicate elements removed, keeping only the first occurrence.
```

internal/scheduler/interfaces/interfaces.go
```diff
@@ -0,0 +1,30 @@
+package interfaces
+
+import (
+	"github.com/armadaproject/armada/internal/armada/configuration"
+	"github.com/armadaproject/armada/internal/scheduler/schedulerobjects"
+)
+
+// LegacySchedulerJob is the job interface used throughout the scheduler.
+//
+// TODO: Rename to SchedulerJob.
```
d80tb7, line 10: Do we need to rename this?  I was hoping that once we've deprecated the old lease.go, we can just move to a single concrete, job impl.

internal/scheduler/nodedb/nodedb.go
```diff
@@ -155,6 +160,18 @@ func (nodeDb *NodeDb) String() string {
 	return sb.String()
 }
 
+func (nodeDb *NodeDb) NumNodes() int {
```
d80tb7, line 163: These functions are a bit weird because they have mutexes, which implies that the dat can be modified by another thread, but if so then they don't seem that useful?
e.g. if I do:
```
if nodeDb.NumNodes() > 0 {
   // Do something that requires at least one node
}
```
It's not guaranteed to work because NumNodes could have changed between my check and the actual code. I think in practise this doesn't matter because right now nodeDb is only accessed from a single thread, but I think it makes the mutex unnecessary.  More generally, to make this work in a multi thread env, we'd need to change the nbodedb so that all functions operate off a snapshot of data, so you can guarantee repeatable read.

internal/scheduler/nodedb/nodeiteration.go
```diff
@@ -0,0 +1,413 @@
...
+func (it *NodePairIterator) WatchCh() <-chan struct{} {
+	panic("not implemented")
+}
+
+func (it *NodePairIterator) NextItem() (rv *NodePairIteratorItem) {
+	defer func() {
+		if rv == nil {
+			return
+		}
+		if rv.NodeA != nil {
```
d80tb7, line 93: I'm not sure I like the use of a deferred function here to  set these values to nil, nor really the use of nil to indicate that there might be more data available.  The general convention is that if a value is nil, it indicates that no more data is available.

internal/armada/server/lease.go
severinson: You're right. That's an error. Fixed.

internal/scheduler/interfaces/interfaces.go
severinson: Sure, I'll leave it.

internal/scheduler/nodedb/nodedb.go
severinson: You're right. The NodeDb isn't thread-safe anyway. I'll leave the lock anyway, since it at least protects against races during initialisation.

internal/scheduler/nodedb/nodeiteration.go
severinson: I interpret `nil` for either `A` or `B` here to mean that there's no entry for the

internal/scheduler/common.go
```diff
@@ -0,0 +1,278 @@
...
+	armadaresource "github.com/armadaproject/armada/internal/common/resource"
+	armadaslices "github.com/armadaproject/armada/internal/common/slices"
+	schedulerconfig "github.com/armadaproject/armada/internal/scheduler/configuration"
+	schedulercontext "github.com/armadaproject/armada/internal/scheduler/context"
+	"github.com/armadaproject/armada/internal/scheduler/interfaces"
+	"github.com/armadaproject/armada/internal/scheduler/schedulerobjects"
+)
+
+// SchedulerResult is returned by Rescheduler.Schedule().
+type SchedulerResult struct {
```
d80tb7, line 22: would it be possible to define this as `type SchedulerResult[T interfaces.LegacySchedulerJob] struct`? That would mean:
- Functions like `NewSchedulerResult` would be simpler because you wouldn't need to do the casting
- You could remove PreemptedJobsFromSchedulerResult and the associated lack of compile-time safety
- Performance would be better because we wouldn't needs to reallocate arrays, just to perform casts.

internal/scheduler/common.go
```diff
@@ -0,0 +1,278 @@
...
+// cast to type T.
+func ScheduledJobsFromSchedulerResult[T interfaces.LegacySchedulerJob](sr *SchedulerResult) []T {
+	rv := make([]T, len(sr.ScheduledJobs))
+	for i, job := range sr.ScheduledJobs {
+		rv[i] = job.(T)
+	}
+	return rv
+}
+
+func JobsSummary(jobs []interfaces.LegacySchedulerJob) string {
```
d80tb7, line 72: This function probably warrents a comment about what it does, along with a perfromance warning as I suspect it could be quite slow if `jobs` is large.

d80tb7: I'd also ague that this function sounds incredible generic ("scheduler.JobsSummary") but to me looks like it's producing a string fromatted for relatively specific use case.  E.g. it straight away makes the assuption that the jobs passed in have all been evicted.

internal/scheduler/common.go
```diff
@@ -0,0 +1,278 @@
...
+	return jctxs
+}
+
+func isEvictedJob(job interfaces.LegacySchedulerJob) bool {
+	return job.GetAnnotations()[schedulerconfig.IsEvictedAnnotation] == "true"
+}
+
+func targetNodeIdFromLegacySchedulerJob(job interfaces.LegacySchedulerJob) (string, bool) {
+	req := PodRequirementFromLegacySchedulerJob(job, nil)
+	nodeId, ok := req.NodeSelector[schedulerconfig.NodeIdLabel]
```
d80tb7, line 185: can `req.NodeSelector` be nil here?

internal/scheduler/common.go
```diff
@@ -0,0 +1,278 @@
...
+	return nodeId, ok
+}
+
+func GangIdAndCardinalityFromLegacySchedulerJob(job interfaces.LegacySchedulerJob, priorityClasses map[string]configuration.PriorityClass) (string, int, bool, error) {
```
d80tb7, line 189: I'd document this function to indicate what each of the return values mean.  Really though I'd do the following:

- Add `GetGangId` and `GetGangCardinality` (and possibly `isGangJob`) functions to the LegacySchedulerJob interface
- For `SchedulerJob` model these as first class fields with no annotation lookups necessary
- For api.job use the logic here.

internal/scheduler/common.go
severinson: Yes, but it would mean adding type parameters to large parts of the scheduler. If we're anyway moving away from using this interface later, I'd prefer to not add a type parameter to `SchedulerResult` now.

internal/scheduler/common.go
```diff
@@ -0,0 +1,278 @@
...
+			JobId:      job.GetId(),
+			Job:        job,
+			Req:        PodRequirementFromJobSchedulingInfo(job.GetRequirements(priorityClasses)),
+		}
+	}
+	return jctxs
+}
+
+func isEvictedJob(job interfaces.LegacySchedulerJob) bool {
+	return job.GetAnnotations()[schedulerconfig.IsEvictedAnnotation] == "true"
```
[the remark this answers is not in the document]
severinson: Lookups on nil maps is safe and returns the zero value of the map. So it does return false if `job.GetAnnotations()` returns `nil`.

internal/scheduler/common.go
severinson: Yes. Let's add a `nil` check.

internal/armada/configuration/types.go
```diff
@@ -102,6 +102,8 @@ type SchedulingConfig struct {
 	MaximalResourceFractionPerQueue map[string]float64
 	// Max number of jobs to scheduler per lease jobs call.
 	MaximumJobsToSchedule uint
+	// Max number of gangs to scheduler per lease jobs call.
+	MaximumGangsToSchedule uint
```
d80tb7, line 106: Why do we have this?  Is it because we don't want to schedule too may pods at once?

internal/scheduler/constraints/constraints.go
```diff
@@ -0,0 +1,167 @@
...
+	return true, "", nil
+}
+
+func (constraints *SchedulingConstraints) CheckPerQueueAndPriorityClassConstraints(
+	ctx context.Context,
+	sctx *schedulercontext.SchedulingContext,
+	queue string,
+	priorityClassName string,
+) (bool, string, error) {
+	// We assume that all jobs in a gang have the same priority class (which we enforce at job submission).
```
d80tb7, line 108: It's not clear to me why this comment is here as the check immediately following this doesn't seem to care about gangs or priority.  Should it be moved to line 114?  (Or even removed completely)

internal/scheduler/constraints/constraints.go
```diff
@@ -0,0 +1,167 @@
...
+			return false, unschedulableReason, nil
+		}
+	}
+
+	return true, "", nil
+}
+
+// exceedsResourceLimits returns true if used/total > limits for some resource t,
+// and, if that is the case, a string indicating which resource limit was exceeded.
+func exceedsResourceLimits(_ context.Context, used, total schedulerobjects.ResourceList, limits map[string]float64) (bool, string) {
```
d80tb7, line 158: why are we passing the context through here if it isn't used?

internal/scheduler/context/context.go
```diff
@@ -241,6 +372,56 @@ func (qctx *QueueSchedulingContext) ClearJobSpecs() {
...
+	Created               time.Time
+	Queue                 string
+	PriorityClassName     string
+	JobSchedulingContexts []*JobSchedulingContext
+	TotalResourceRequests schedulerobjects.ResourceList
+}
+
+func NewGangSchedulingContext(jctxs []*JobSchedulingContext) *GangSchedulingContext {
+	queue := ""
+	for _, jctx := range jctxs {
```
d80tb7, line 385: What does it mean if some of the job contexts don't have a queue?!   Surely we can just pick the first one.

internal/scheduler/gang_scheduler.go
```diff
@@ -0,0 +1,121 @@
...
+		schedulingContext: sctx,
+		nodeDb:            nodeDb,
+	}, nil
+}
+
+func (sch *GangScheduler) Schedule(ctx context.Context, gctx *schedulercontext.GangSchedulingContext) (ok bool, unschedulableReason string, err error) {
+	defer func() {
+		// If any job in a gang fails to schedule, set the unschedulableReason for all jobs in the gang,
+		// and remove it from the scheduling context.
+		if unschedulableReason == "" {
```
d80tb7, line 38: surely the check here should be against `ok` not `unschedulableReason` else the jobs won't be evicted if `ok` is false but `unschedulableReason` is the empty string?

d80tb7: More generally it's not immediately clear to me what comination of values is ok here.  What does it mean if err is non nill but ok is true?!  Generally this isn't a concern for callers because err trumps all, but here you've got a deferred function that is taking action without considering err.  It seems dangerous to mutate state in the case that some unknown error may or may not have occurred?

internal/scheduler/nodedb/nodedb.go
```diff
@@ -566,6 +529,17 @@ func EvictPodFromNode(req *schedulerobjects.PodRequirements, node *schedulerobje
...
 
+	// Ensure we track allocated resources at evictedPriority.
+	if _, ok := node.AllocatableByPriorityAndResource[evictedPriority]; !ok {
+		pMin := int32(math.MaxInt32)
+		for p := range node.AllocatableByPriorityAndResource {
+			if p < pMin {
+				pMin = p
+			}
+		}
+		node.AllocatableByPriorityAndResource[evictedPriority] = node.AllocatableByPriorityAndResource[pMin].DeepCopy()
```
d80tb7, line 540: Shouldn't we error if `PMin` is still `math.MaxInt32` at this point?

internal/scheduler/nodedb/nodeiteration.go
```diff
@@ -0,0 +1,409 @@
...
+	"fmt"
+
+	"github.com/hashicorp/go-memdb"
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+
+	"github.com/armadaproject/armada/internal/scheduler/schedulerobjects"
+)
+
+const NodeDominantQueueWildcard = "*"
```
d80tb7, line 16: Do we still need this now we no longer consider dominantQueues?

internal/scheduler/preempting_queue_scheduler.go
```diff
@@ -0,0 +1,843 @@
...
+		// This refers to resources available across all clusters, i.e.,
+		// it may include resources not currently considered for scheduling.
+		log.Infof(
+			"no resources with non-zero weight available for scheduling on any cluster: resource scarcity %v, total resources %v",
+			sch.schedulingContext.ResourceScarcity, sch.schedulingContext.TotalResources,
+		)
+		return &SchedulerResult{}, nil
+	}
+	if ResourceListAsWeightedApproximateFloat64(sch.schedulingContext.ResourceScarcity, sch.nodeDb.TotalResources()) == 0 {
+		// This refers to the resources currently considered for schedling.
```
d80tb7, line 106:
```suggestion
		// This refers to the resources currently considered for scheduling.
```

internal/scheduler/preempting_queue_scheduler.go
```diff
@@ -0,0 +1,843 @@
...
+					continue
+				}
+				for _, q := range rl.Resources {
+					if q.Cmp(resource.Quantity{}) == -1 {
+						overSubscribedPriorities[p] = true
+						break
+					}
+				}
+			}
+			return len(overSubscribedPriorities) > 0 && rand.Float64() < perNodeEvictionProbability
```
d80tb7, line 753: lets get a rand we can control in here so that we can test this

internal/scheduler/schedulerobjects/resourcelist.go
```diff
@@ -216,7 +216,7 @@ func (a ResourceList) Equal(b ResourceList) bool {
 	return true
 }
 
-// IsStrictlyNonNegative returns true if there are no quantities in a with value less than zero.
+// IsStrictlyNonNegative returns true if there is no quantity in a less than zero.
```
d80tb7, line 219:
```suggestion
// IsStrictlyNonNegative returns true if there is no quantity less than zero.
```

internal/scheduler/constraints/constraints.go
severinson: Fixed

internal/scheduler/constraints/constraints.go
severinson: Removed the conttext.

internal/scheduler/context/context.go
severinson: Agreed. I've changed it,

internal/scheduler/gang_scheduler.go
severinson: I've changed it to be
```
	defer func() {
		// If any job in a gang fails to schedule, set the unschedulableReason for all jobs in the gang,
		// and remove it from the scheduling context.
		if err == nil && !ok {
			for _, jctx := range gctx.JobSchedulingContexts {
				sch.schedulingContext.EvictJob(jctx.Job)
				jctx.UnschedulableReason = unschedulableReason
			}
			sch.schedulingContext.AddGangSchedulingContext(gctx)
		}
	}()
```

internal/scheduler/gang_scheduler.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return
+}
+
+func (sch *GangScheduler) trySchedule(ctx context.Context, gctx *schedulercontext.GangSchedulingContext) (bool, string, error) {
+	pctxs, ok, err := sch.nodeDb.ScheduleMany(gctx.PodRequirements())
+	if err != nil {
+		return false, "", err
+	}
+	if len(pctxs) > len(gctx.JobSchedulingContexts) {
```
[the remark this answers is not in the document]
severinson: It may return fewer pctxs than there are jctxs. Since NodeDb short-circuits if a pod fails to schedule.

internal/scheduler/nodedb/nodeiteration.go
severinson: Good catch. Deleted.

internal/scheduler/preempting_queue_scheduler.go
```diff
@@ -0,0 +1,843 @@
...
+) *Evictor {
+	if perNodeEvictionProbability <= 0 {
+		return nil
+	}
+	return NewPreemptibleEvictor(
+		jobRepo,
+		priorityClasses,
+		defaultPriorityClass,
+		func(_ context.Context, node *schedulerobjects.Node) bool {
+			return len(node.AllocatedByJobId) > 0 && rand.Float64() < perNodeEvictionProbability
```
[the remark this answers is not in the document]
severinson: All the tests use either 1 or 0 for the probability.

internal/scheduler/preempting_queue_scheduler.go
severinson: Done

severinson: Changed to
```
func NewOversubscribedEvictor(
	jobRepo JobRepository,
	priorityClasses map[string]configuration.PriorityClass,
	perNodeEvictionProbability float64,
	random *rand.Rand,
) *Evictor {
	if perNodeEvictionProbability <= 0 {
		return nil
	}
	if random == nil {
		random = rand.New(rand.NewSource(int64(time.Now().Nanosecond())))
	}
```

internal/scheduler/preempting_queue_scheduler.go
```diff
@@ -0,0 +1,843 @@
...
+	}
+	return allGangJobIds, gangNodeIds, nil
+}
+
+// Some jobs in a gang may have terminated since the gang was scheduled.
+// For these gangs, we need to set the gang cardinality to the number of jobs in the gang yet to terminate.
+// Otherwise the evicted gang jobs will not be schedulable, since some gang jobs will be considered missing.
+func (sch *PreemptingQueueScheduler) setEvictedGangCardinality(evictedJobsById map[string]interfaces.LegacySchedulerJob) error {
+	for _, job := range evictedJobsById {
+		gangId, ok := sch.gangIdByJobId[job.GetId()]
```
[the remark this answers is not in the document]
d80tb7: ok yes agreed- I think a lot of this code will get simpler once we ditch the annotation.s

d80tb7 (approved):

PR #2376, opened 2023-04-16
Add PodRequirementsHash to JobSchedulingInfo

JamesMurkin:
This field is:
 - Set by the scheduler ingester on initial job insert
 - Updated by the scheduler when requirements are changed via JobRequeued event


internal/scheduler/scheduler.go
```diff
@@ -328,6 +329,11 @@ func (s *Scheduler) createSchedulingInfoWithNodeAntiAffinityForAttemptedRuns(job
 		}
 	}
 	podSchedulingRequirement.Affinity = newAffinity
+	newRequirementsHash, err := hash.CalculatePodRequirementsHash(podSchedulingRequirement)
```
d80tb7, line 332: So my initial assumption was that we'd calculate the hash in the ingester, mainly because if the queue is very large (i.e. millions), you might spend a non-trivial amount of time calculating hashes at startup.

d80tb7: I should say that we can definitely start like this as it's a simpler implementation and then move  this to the ingester as and when we need it.

d80tb7: AH ok I've got this wrong! You *are* calculating this in the schedulerIngester's `instructions.go`, this is just for the case where we have excluded nodes.   That being the case you can safely ignore me!

d80tb7 (approved):

PR #2381, opened 2023-04-19
Attempted CI Fixes

### LINKED ISSUE #2322: Fix flaky Executor Connection in CI
The server takes a variable amount of time to find executor clusters. This creates issues in CI.

By starting the executor in advance of the server, we should be able to avoid this.

A function could also be wrote that runs a single job to check for it working correctly.

Sharpz7:
Closes #2322

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-184) by [Unito](https://www.unito.io)


magefiles/ci.go
```diff
@@ -18,7 +21,14 @@ func ciSetup() error {
 
 	mg.Deps(CheckForPulsarRunning)
 
-	err = dockerComposeRun("up", "-d", "server", "executor")
+	// By starting the executor first,
+	// we can ensure that the server will be able to connect to it
```
suprjinx, line 25: the server does not connect to the executor -- but maybe this buys a little time for pulsar init?

suprjinx: oh i see, it needs to receive poll from executor or job submit fails.

suprjinx (approved):

magefiles/ci.go
Sharpz7: I should probably make this comment more clear.

Essentially there is a "executor check" that the server runs, to see what executor clusters are available. The executor can be sending information to the server, but this check can still fail.

This happens on a poll everything 1 minute I think.

suprjinx (approved):

PR #2386, opened 2023-04-21
Add configurable prefix to all Job Service table names (BATCH-185)

richscott:
These changes add a new configuration setting for Job Service, called `tablesPrefix`, which makes the creation and usage of tables in Job Service's database to have a common prefix. For example, if `tablesPrefix: "test_"` is set in the configuration file, the `jobservice` table will then be named `test_jobservice`. All table creation and queries will use the prefix.

Other changes:
- add new Job Service config setting `tablesPrefix`
- change all SQL strings that reference Job Service tables to be constructed
dynamically, based on the `tablesPrefix` configuration value (default is an empty string).
- various data/function renaming per suggestions from Geoff Wilson:  'sqlJobRepo' to just 'jobRepo',
'JSRepoPostgres' is now 'RepositoryPostgres', WithSqlServiceRepo() is now WithRepository().
- shorten repository Setup() function by coalescing all table and index creation/manipulation
to a slice of strings, which we loop through to execute
- rename 'log' object to 'logger', to remove warning about name collision with the Go 'log' package.

Fixes https://gr-oss.atlassian.net/browse/BATCH-185

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-194) by [Unito](https://www.unito.io)


richscott:
Closing this PR - we will be using separate databases within the Postgres instance to do isolation of tables between JS and Airflow and/or other apps in the same db instance.

PR #2391, opened 2023-04-24
[Scheduler Ingester] Perform All Updates Inside A Single Transaction

d80tb7:
Scheduler Ingester now applies a batch of operations inside a single transaction.  This has a couple of benefits:
- The operations may be reordered to improve ingestion performance.  By using a single transaction, we ensure that intermediate states are not visible to readers and so reordering only needs to ensure the final state is correct.
- If there is a failure, the entire batch will fail, therefore we won't end up in some sort of intermediate state.

As part of this PR I also implemented a lock around the actual database operations (implemented using a pg_advisory lock). This is needed because we rely on the scheduler seeing monotonically increasing sequence numbers and this will not be the case if two ingesters are writing to the dtaabase at the same time.  This was an issue previously, but would be much worse now that the individual transactions are much bigger.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-192) by [Unito](https://www.unito.io)


internal/scheduleringester/schedulerdb.go
```diff
@@ -26,27 +26,70 @@ type SchedulerDb struct {
...
+		metrics:        metrics,
+		initialBackOff: initialBackOff,
+		maxBackOff:     maxBackOff,
+		lockTimeout:    lockTimeout,
+	}
 }
 
+// Store persists all operations in the database.  Note that:
+//   - this function will retry until it either succeeds or a terminal error is encountered
+//   - this function weill take out a postgres lock to ensure that other ingesters are not writing to the database
```
JamesMurkin, line 50:
```suggestion
//   - this function will take out a postgres lock to ensure that other ingesters are not writing to the database
```

JamesMurkin (approved):

PR #2417, opened 2023-05-02
Fix Pulsar-scheduler accounting

severinson:
In addition to tidying up:

* Enables gang preemption.
* Changes resource allocation accounting to be based on jobDb instead of on executor reports.
* Adds priority class name overrides.


internal/scheduler/scheduling_algo.go
```diff
@@ -78,86 +78,56 @@ func NewLegacySchedulingAlgo(
...
+			if err := l.schedulingContextRepository.AddSchedulingContext(sctx); err != nil {
+				logging.WithStacktrace(log, err).Error("failed to add scheduling context")
+			}
+		}
+
+		// Update jobDb.
 		preemptedJobs := PreemptedJobsFromSchedulerResult[*jobdb.Job](schedulerResult)
 		scheduledJobs := ScheduledJobsFromSchedulerResult[*jobdb.Job](schedulerResult)
 		if err := jobDb.Upsert(txn, preemptedJobs); err != nil {
+			// TODO: We need to do something here to mark the jobs as preempted in the jobDb.
```
d80tb7, line 116: We'll Need to fix this asap otherwise we'll get into he wrong state won't we?

d80tb7 (approved):

PR #2423, opened 2023-05-04
fix hashing function

d80tb7:
protoutil.hash was (embarrassingly) returning the bytes of the json serialized message rather than the hash  This PR fixes that and adds some more testing to ensure this never happens again.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-228) by [Unito](https://www.unito.io)

### DIFF
--- a/internal/common/proto/protoutil_test.go
+++ b/internal/common/proto/protoutil_test.go
@@ -89,7 +89,7 @@ func TestHashMany(t *testing.T) {
 
 func TestHash_NoCollisions(t *testing.T) {
 	hashes := make(map[string]bool)
-	for i := 0; i < 1000000; i++ {
+	for i := 0; i < 100000; i++ {
 		intialHash, err := Hash(&schedulerobjects.JobSchedulingInfo{Lifetime: uint32(i)})
 		require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


JamesMurkin (approved):

internal/common/proto/protoutil_test.go
```diff
@@ -70,6 +70,35 @@ func TestMustMarshallAndCompress(t *testing.T) {
...
+	hash3, err := HashMany([]*schedulerobjects.JobSchedulingInfo{input1, input3})
+	require.NoError(t, err)
+	require.Equal(t, 20, len(hash3))
+	assert.Equal(t, hash1, hash2)
+	assert.NotEqual(t, hash1, hash3)
+}
+
+func TestHash_NoCollisions(t *testing.T) {
+	hashes := make(map[string]bool)
+	for i := 0; i < 1000000; i++ {
```
dave-gantenbein, line 92: Do we really want to run one million assertions for this on every test invocation?

d80tb7: it's a fair point- although it takes about 2 seconds on my laptop so shouldn't be too bad.

PR #2440, opened 2023-05-09
Fix use of sync.Map.LoadOrStore result

ClifHouck:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-255) by [Unito](https://www.unito.io)

### DIFF
--- a/internal/jobservice/application.go
+++ b/internal/jobservice/application.go
@@ -125,8 +125,8 @@ func (a *App) StartUp(ctx context.Context, config *configuration.JobServiceConfi
 			}
 			for _, value := range jobSets {
 				queueJobSet := value.Queue + value.JobSet
-				_, ok := subscribeMap.LoadOrStore(queueJobSet, true)
-				if !ok {
+				_, loaded := subscribeMap.LoadOrStore(queueJobSet, true)
+				if loaded {
 					eventJob := eventstojobs.NewEventsToJobService(value.Queue, value.JobSet, eventClient, sqlJobRepo)
 					go func(value repository.SubscribedTuple) {
 						err := eventJob.SubscribeToJobSetId(context.Background(), config.SubscribeJobSetTime, value.FromMessageId)


internal/jobservice/application.go
```diff
@@ -125,8 +125,8 @@ func (a *App) StartUp(ctx context.Context, config *configuration.JobServiceConfi
 			}
 			for _, value := range jobSets {
 				queueJobSet := value.Queue + value.JobSet
-				_, ok := subscribeMap.LoadOrStore(queueJobSet, true)
-				if !ok {
+				_, loaded := subscribeMap.LoadOrStore(queueJobSet, true)
+				if loaded {
```
ClifHouck, line 129: See https://pkg.go.dev/sync#Map.LoadOrStore

We actually only wanted to enter the subscribe logic if the value was *loaded*, otherwise we'd be running the same stream again.

internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -87,7 +87,11 @@ func (eventToJobService *EventsToJobService) streamCommon(ctx context.Context, t
 			select {
 			case <-ctx.Done():
 				log.Errorf("context is done on %s/%s", eventToJobService.queue, eventToJobService.jobSetId)
-				return nil
+				_, err := eventToJobService.jobServiceRepository.UnsubscribeJobSet(
```
ClifHouck, line 90: We need to actually clean up the the job set sub from the DB when we bail out. So do it.

kannon92, line 90: I think this is not what we want actually.  I went back and forth but we still want the list of jobsets

kannon92: Otherwise once we unsubscribe we no longer have the list and what we really want is to reconnect to the existing jobset.

ClifHouck: Then I don't know how to solve this, because with the current logic we are never going to clear a subscription.

kannon92 (approved):

internal/jobservice/eventstojobs/eventstojobs.go
kannon92: I think when we unsubscribe we close the event connection.  Isn't that enough?

internal/jobservice/application.go
suprjinx: but 'loaded' will reflect true if it was already in the map

PR #2442, opened 2023-05-09
Only call GetJobSetEvents once

kannon92:
When using a GRPC stream you should get the stream and then start looping over that.  Calling it a bunch is not ideal.  

This is a POC for right now to test out if this actually works.

Thanks @d80tb7

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-257) by [Unito](https://www.unito.io)


internal/jobservice/eventstojobs/eventstojobs.go
```diff
@@ -82,6 +82,17 @@ func (eventToJobService *EventsToJobService) streamCommon(ctx context.Context, t
 			}
 		}()
 
+		log.Infof("GetJobEventMessage for %s/%s with id %s", eventToJobService.queue, eventToJobService.jobSetId, fromMessageId)
```
richscott, line 85: Are Daniele and/or others comfortable with info-level logging on this, or would they prefer it to be debug-level?

richscott (commented):
Just a quick question about logging in one place, otherwise I think it looks good.

ClifHouck (approved):
LGTM, though I echo Rich's comment on possibly lowering the log-level for every message to debug.

richscott (approved):

internal/jobservice/eventstojobs/eventstojobs.go
kannon92: I think it would be fine as this should only be once every subscribe.  But we can always mess around with this.

PR #2453, opened 2023-05-11
Add index on Job Service tables Timestamp column.

richscott:
The Job Service database uses a trigger to invoke a small stored procedure to delete all entries in the `jobservice` and `jobsets` tables that have their `timestamp` column older than the `purgeTime` configuration value (value is in seconds).  This trigger is fired upon each insert to the `jobsets` table, so the deletes in this proc must be as fast as possible. This change adds an index to the `timestamp` column of both tables. Manual testing shows a considerable speedup when the tables have 50k or more rows, using `EXPLAIN ANALYZE DELETE from .. WHERE...`.

Also, refactor the table creation logic to make more compact and eliminate repetition of db.Exec(), etc,.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-266) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/release-rc.yml
+++ b/.github/workflows/release-rc.yml
@@ -62,7 +62,7 @@ jobs:
         id: "tag_create"
         with:
           tag: "v${{ steps.set_next_version.outputs.next_version }}"
-          tag_exists_error: true
+          force_push_tag: true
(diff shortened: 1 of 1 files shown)


kannon92 (approved):

ClifHouck (approved):
LGTM

PR #2460, opened 2023-05-12
Lookout - configurable title

carlocamurri:
Allows a configurable title to be specified in the UI config. This is to more easily differentiate between Lookout instances.
Note that this also changes the browser tab title.
Noop if the config field is unset.

image

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-273) by [Unito](https://www.unito.io)


kannon92 (commented):
Is there a default if custom not specified?

I see you override or prefix but not clear if custom title is not specified.

JamesMurkin (approved):

Issue #2255, opened 2023-03-10
Create an airflow instance similar to GR's setup


kannon92:

kannon92:
https://github.com/kannon92/armada-demo

I’ve created a demo repo where I run airflow locally and jobservice and point to our eks demo cluster.  I’d love to get a helm chart install of airflow that works on kind.

VasuDevrani:
Hey @kannon92, I'm kinda beginner in this, armadaproject
I would appreciate if you could break this task down for me. Like exactly what I am supposed to do here.
And, by the way I'm well versed with helm charts and kubernetes.

thanks

kannon92:
There is a ticket around code coverage increase if you want more familiarity with the code base

MdSahil-oss:
@kannon92 Assign this issue to me, I would like to work on it.

kannon92:
https://airflow.apache.org/docs/helm-chart/stable/quick-start.html Could be a useful guide.

Ideally I’d love a script or something that can automate installing airflow on kind and demonstrate it working with our local armada setup.

kannon92:
https://marclamberti.com/blog/airflow-on-kubernetes-get-started-in-10-mins/ for an example on how to setup airflow on Kubernetes.  It would be nice to have a way to also include our example dags and have this work for Armada.

enesonus:
Hi @kannon92 I would like to work on this issue,

Do you have any suggestions about the path I should follow, what I should pay attention to before I start to work on this issue or are the links you provided sufficient?
Any help would be appreciated.

akagami-harsh:
@kannon92 HI, is this issue still relevant?

kannon92:
I think so but @Sharpz7 would know better.

ambuj-1211:
hi @kannon92 and @Sharpz7 I am new to the open-source world and would love to contribute to this project. What should I know about this project related to this issue as I have already done research on airflow and creating airflow instances on Kubernetes?

PR #2468, opened 2023-05-15
Localdev airflow environment with mage

ClifHouck:
Supports `mage airflow {start,stop,restart}` to build/teardown/restart an airflow instance with docker compose. DAGs can communicate with armada service with docker dns names. 

For instance, `server:50051` for the armada-server grpc channel, and `jobservice:60003` for the jobservice grpc channel.

TODO:
- [ ] A little documentation on use. 
- [ ] Figure out a way to deploy changes to armada python packages within airflow docker containers without the need for a complete rebuild/restart.

*Note*: I could not get an "editable" package installation working with `armada-airflow` without changing the base import name from `armada` to `armada-airflow`. A breaking change so one that needs to be considered before this merges.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-281) by [Unito](https://www.unito.io)


kannon92:
Please solve the merge conflicts.

localdev/airflow/docker-compose.yaml
```diff
@@ -0,0 +1,309 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
kannon92, line 1: This seems like a lot of tech debt to inherit.  Maybe we should just look into https://github.com/armadaproject/armada/issues/2255 and see how we can create an airflow instance via helm.  

I really don't think we want to support a third-party compose chart for airflow.  I'd rather use an official helm chart.

third_party/airflow/pyproject.toml
```diff
@@ -24,5 +24,8 @@ test = ["pytest==7.3.1", "coverage>=6.5.0", "pytest-asyncio==0.21.0"]
 docs = ["sphinx", "sphinx-jekyll-builder"]
 
 [build-system]
-requires = ["setuptools"]
-build-backend = "setuptools.build_meta"
+requires = ["flit_core>=3.8.0,<4"]
```
kannon92, line 27: What is `flit_core`?

ClifHouck: https://github.com/pypa/flit

I couldn't get `pip install -e` to work with `armada-airflow` on setuptools so I tried another build backend.

localdev/airflow/docker-compose.yaml
ClifHouck: I can look at using a helm chart instead.

suprjinx: we already have an airflow service in the root docker-compose.yaml -- why not use that?

ClifHouck: The official airflow docker-compose.yaml gets us a better dag executor out of the box, as well as running the triggerer process (needed by deferrable operators) by default. It also gets us each airflow executable living in its own container instead of all of them being executed by the same container. 

You could argue we can adapt this script to do it as well, but the official docker-compose.yaml seemed easier to configure and work with.

kannon92 (commented):
I think we should float this by GR to see if they are okay with a name change for the module.  This does seem to be a breaking change and would break existing workflows.  So we should communicate this out if we want to go this route.

ClifHouck:
I reverted the change to the module import name of `armada-airflow` and it still works. I'm not sure what changed, or if I was just hallucinating the error, but in any case this seems to work for me locally.

kannon92 (approved):

PR #2469, opened 2023-05-15
Scheduling limit improvements

severinson:
Specifically:
- Remove per queue and round limits. Since these limits may result in scheduling a queue over its fair share instead of one below it, thus causing unnecessary preemptions in subsequent rounds.
- Remove overall resource limits. We don't need them now that we have per-PC limits.
- Fix a bug where reaching the max scheduled jobs may cause evicted jobs to not be-rescheduled.
- Add per-pool scheduling limits.
- Rename various limits for clarity.

### DIFF
--- a/internal/scheduler/gang_scheduler.go
+++ b/internal/scheduler/gang_scheduler.go
@@ -25,6 +25,9 @@ type GangScheduler struct {
 	unsuccessfulSchedulingKeys map[schedulerobjects.SchedulingKey]*schedulercontext.JobSchedulingContext
 	// If true, the unsuccessfulSchedulingKeys check is omitted.
 	skipUnsuccessfulSchedulingKeyCheck bool
+	// Name of the queue that was scheduled in the previous call to Schedule().
+	// Empty is no job was scheduled in the previous call to Schedule(), or if this is the first call.
+	queueScheduledInPreviousCall string
 }
 
 func NewGangScheduler(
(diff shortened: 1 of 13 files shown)


internal/scheduler/gang_scheduler.go
```diff
@@ -25,6 +25,9 @@ type GangScheduler struct {
 	unsuccessfulSchedulingKeys map[schedulerobjects.SchedulingKey]*schedulercontext.JobSchedulingContext
 	// If true, the unsuccessfulSchedulingKeys check is omitted.
 	skipUnsuccessfulSchedulingKeyCheck bool
+	// Name of the queue that was scheduled in the previous call to Schedule().
+	// Empty is no job was scheduled in the previous call to Schedule(), or if this is the first call.
```
severinson, line 29:
```suggestion
	// Empty if no job was scheduled in the previous call to Schedule(), or if this is the first call.
```

internal/scheduler/gang_scheduler.go
```diff
@@ -45,25 +45,38 @@ func (sch *GangScheduler) SkipUnsuccessfulSchedulingKeyCheck() {
 }
 
 func (sch *GangScheduler) Schedule(ctx context.Context, gctx *schedulercontext.GangSchedulingContext) (ok bool, unschedulableReason string, err error) {
+	// Exit immediately if this is a new gang we've hit any round limits. Otherwise, try scheduling the gang.
```
JamesMurkin, line 48: Could this be reworded?

Possibly just needs an `and` between gang and we've

PR #2471, opened 2023-05-16
Use AsApproximateFloat64 to improve scheduler performance

JamesMurkin:
Swapping this to use `AsApproximateFloat64` gives a 5-10% performance improvement to scheduling

The reason it should be safe is because the behaviour is only different if the float exceeds max int and we don't expect any pod resource requests to exceed max int

Longer term we could get around this by computing this weighted resource list on the job upfront, rather than on the fly in the scheduler (as the weightings are static). This would mean the performance wouldn't matter - and we could go back to the "safe" call if we wanted

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-284) by [Unito](https://www.unito.io)

### DIFF
--- a/internal/scheduler/common.go
+++ b/internal/scheduler/common.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/armadaproject/armada/internal/armada/configuration"
 	armadamaps "github.com/armadaproject/armada/internal/common/maps"
-	armadaresource "github.com/armadaproject/armada/internal/common/resource"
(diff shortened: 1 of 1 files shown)


severinson:
I tried this before and it would yield incorrect results even for relatively small numbers. Might be worth double checking. In any case, maybe we can' entirely remove the need for using this function in the inner loop instead.

PR #2474, opened 2023-05-16
Only apply dynamic checks for evicted jobs

severinson:
To avoid preempting jobs if, e.g., a taint is added to a node.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-287) by [Unito](https://www.unito.io)

### DIFF
--- a/config/executor/config.yaml
+++ b/config/executor/config.yaml
@@ -47,8 +47,8 @@ kubernetes:
   stuckTerminatingPodExpiry: 1m
   podKillTimeout: 5m
   minimumResourcesMarkedAllocatedToNonArmadaPodsPerNode:
-    cpu: 100m
-    memory: 50Mi
+    cpu: 1
+    memory: 200Mi
   minimumResourcesMarkedAllocatedToNonArmadaPodsPerNodePriority: 2000001000 # same priority as system-node-critical
   podDefaults:
     ingress:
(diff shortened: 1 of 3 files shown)


config/executor/config.yaml
```diff
@@ -47,8 +47,8 @@ kubernetes:
   stuckTerminatingPodExpiry: 1m
   podKillTimeout: 5m
   minimumResourcesMarkedAllocatedToNonArmadaPodsPerNode:
-    cpu: 100m
-    memory: 50Mi
+    cpu: 1
```
severinson, line 50: I was seeing preemptions locally due to overcommitting the kind node without this.

internal/scheduler/preempting_queue_scheduler.go
```diff
@@ -861,6 +861,9 @@ func defaultPostEvictFunc(ctx context.Context, job interfaces.LegacySchedulerJob
 	}
 
 	// Add a toleration to allow the job to be re-scheduled even if node is unschedulable.
+	//
+	// TODO: Because req is created with a new tolerations slice above, this toleration doesn't persist.
```
severinson, line 865: This is a bit unfortunate. But I don't see a good solution until we have a common job notion that doesn't create new objects every time.

JamesMurkin (approved):

PR #2477, opened 2023-05-17
[scheduler-ingester] Improvements to jobset operations

JamesMurkin:
- Now CancelJobSet and ReprioritiseJobSet correctly only operate on a specific queue. They were incorrectly updating all jobsets with a given name, ignoring queue
 - Made CancelJobSet so it handles the JobState filter
   - If Queued -  we cancel all queued
   - If Pending or Running - we cancel all running
   - No states means cancel the whole jobset
   - This matches the existing functionality. In future we may want better support for Pending vs Running - or change the API to only support Queued vs Leased

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-291) by [Unito](https://www.unito.io)

### DIFF
--- a/internal/scheduleringester/instructions.go
+++ b/internal/scheduleringester/instructions.go
@@ -319,27 +319,8 @@ func (c *InstructionConverter) handleCancelJob(cancelJob *armadaevents.CancelJob
 }
 
 func (c *InstructionConverter) handleCancelJobSet(cancelJobSet *armadaevents.CancelJobSet, meta eventSequenceCommon) ([]DbOperation, error) {
-	cancelQueued := false
-	cancelLeased := false
-	if len(cancelJobSet.States) == 0 {
-		cancelQueued = true
-		cancelLeased = true
-	} else {
-		cancelQueued = slices.Contains(cancelJobSet.States, armadaevents.JobState_QUEUED)
-		containsPending := slices.Contains(cancelJobSet.States, armadaevents.JobState_PENDING)
-		containsRunning := slices.Contains(cancelJobSet.States, armadaevents.JobState_RUNNING)
-
-		// For now Armada only supports cancelling both pending and running together, so either both must be present or neither
-		if containsPending != containsRunning {
-			log.Warnf("skipping cancel jobset %s %s because an invalid combination of states was provided.",
-				meta.queue, meta.jobset)
-			return []DbOperation{}, nil
-		}
-
-		if containsPending && containsRunning {
-			cancelLeased = true
-		}
-	}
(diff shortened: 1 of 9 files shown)


internal/scheduleringester/instructions.go
```diff
@@ -314,9 +318,18 @@ func (c *InstructionConverter) handleCancelJob(cancelJob *armadaevents.CancelJob
 	}}, nil
 }
 
-func (c *InstructionConverter) handleCancelJobSet(jobset string) ([]DbOperation, error) {
+func (c *InstructionConverter) handleCancelJobSet(cancelJobSet *armadaevents.CancelJobSet, meta eventSequenceCommon) ([]DbOperation, error) {
+	cancelQueued := len(cancelJobSet.States) == 0 || slices.Contains(cancelJobSet.States, armadaevents.JobState_QUEUED)
+	cancelLeased := len(cancelJobSet.States) == 0 || slices.Contains(cancelJobSet.States, armadaevents.JobState_PENDING) || slices.Contains(cancelJobSet.States, armadaevents.JobState_RUNNING)
```
d80tb7, line 323: shouldn't it be AND rather than OR here a otherwise sending a an instruction to cancel the pending will cancel the running too?  Or is this just for compatibility with the old system?

internal/scheduleringester/schedulerdb.go
```diff
@@ -148,10 +149,36 @@ func (s *SchedulerDb) WriteDbOp(ctx context.Context, tx pgx.Tx, op DbOperation)
...
-		err := queries.MarkJobsCancelRequestedBySets(ctx, jobSets)
-		if err != nil {
-			return errors.WithStack(err)
+		for jobSetInfo, cancelDetails := range o {
+			if !cancelDetails.cancelQueued && !cancelDetails.cancelLeased {
+				continue
+			}
+			if cancelDetails.cancelQueued && cancelDetails.cancelLeased {
+				err := queries.MarkJobsCancelRequestedBySet(
+					ctx,
```
d80tb7, line 158: I think this could be done in one query because it's effectively "update job set state = CANCELLED where queue = x and jobset = y and state in (x, y)"?

I wouldn't hold up this PR to change it but I'd definitely consider making the change as it'll be more efficient and (probably) less code.

d80tb7 (approved):

internal/scheduleringester/instructions.go
JamesMurkin: I've done it this was for compatibility - but I agree AND would be more sensible

However it does leave you in an awkward place if you request just leased and then we do nothing.

Probably the API is wrong and should just be swapped to only allowing cancelling queued/leased.

PR #2478, opened 2023-05-17
Make lookout ingesters process PodUnschedulable errors

JamesMurkin:
The motivation for this change is that we have many runs that never get the node name set - because this information is on the PodUnschedulable error events.

This makes tracking if certain nodes have most of the issues very difficult, as most runs with errors don't get the node set

This PR:
 - No longer skips non-terminal errors, this should be fine as:
    - The only events that are non-terminal are PodUnschedulable and PodTerminated and the ingester handles both of these explicitly
    - Each event will only have 1 error
 - Sets the nodeName on PodUnschedulable error events

Longer term this flow needs more thought:
 - Possibly add a PodNodeAssigned event
 - Possibly add other events to display start up status
 - Add NodeName to other events

However for now this is a minimal change to help with investigation into problem nodes

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-292) by [Unito](https://www.unito.io)

### DIFF
--- a/internal/lookoutingesterv2/instructions/instructions.go
+++ b/internal/lookoutingesterv2/instructions/instructions.go
@@ -427,7 +427,7 @@ func (c *InstructionConverter) handleJobRunErrors(ts time.Time, event *armadaeve
 	}
 
 	for _, e := range event.GetErrors() {
-		// We just interpret the error
+		// We just interpret the first error
 
 		// Certain legacy events mean we don't have a valid run id
 		// In this case we have to invent a fake run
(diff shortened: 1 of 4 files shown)


carlocamurri (dismissed):

internal/lookoutingesterv2/instructions/instructions.go
```diff
@@ -427,66 +427,62 @@ func (c *InstructionConverter) handleJobRunErrors(ts time.Time, event *armadaeve
...
-			isLegacyEvent := runId == eventutil.LEGACY_RUN_ID
-			if isLegacyEvent {
-				jobRun := createFakeJobRun(jobId, ts)
-				runId = jobRun.RunId
-				objectMeta := extractMetaFromError(e)
-				if objectMeta != nil && objectMeta.ExecutorId != "" {
-					jobRun.Cluster = util.Truncate(objectMeta.ExecutorId, maxClusterLen)
-				}
-				update.JobRunsToCreate = append(update.JobRunsToCreate, jobRun)
+		// We just interpret the error
```
JamesMurkin, line 430:
```suggestion
		// We just interpret the first error
```

PR #2481, opened 2023-05-18
[scheduler] Add more config to control scheduling frequency/duration

JamesMurkin:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-295) by [Unito](https://www.unito.io)


internal/scheduler/configuration/configuration.go
```diff
@@ -46,6 +46,12 @@ type Configuration struct {
 	InternedStringsCacheSize uint32 `validate:"required"`
 	// How often the scheduling cycle should run
 	CyclePeriod time.Duration `validate:"required"`
+	// How often the job scheduling should run
+	// This is expected to be a greater value than CyclePeriod as we don't need to schedule every cycle
+	// This keeps the system more responsive as other operations happen in each cycle - such as state changes
+	SchedulePeriod time.Duration `validate:"required"`
```
d80tb7, line 52: not suggesting you add this now, but I think the validation framework is sophisticated to enforce the constraint that `SchedulePeriod` > `CyclePeriod`

internal/scheduler/scheduler.go
```diff
@@ -47,15 +47,19 @@ type Scheduler struct {
 	submitChecker SubmitScheduleChecker
 	// Responsible for publishing messages to Pulsar. Only the leader publishes.
 	publisher Publisher
-	// Minimum duration between scheduling cycles.
+	// Minimum duration between scheduler cycles.
 	cyclePeriod time.Duration
+	// Minimum duration between Schedule() calls - calls that actually schedule new jobs.
+	schedulingPeriod time.Duration
```
d80tb7, line 53: should we rename this to `schedulePeriod` to match the config value?

internal/scheduler/scheduler.go
```diff
@@ -200,17 +207,20 @@ func (s *Scheduler) cycle(ctx context.Context, updateAll bool, leaderToken Leade
 	}
 	events = append(events, expirationEvents...)
 
-	// Schedule jobs.
-	overallSchedulerResult, err := s.schedulingAlgo.Schedule(ctx, txn, s.jobDb)
-	if err != nil {
-		return err
-	}
+	if s.clock.Now().Sub(s.previousSchedulingRoundEnd) > s.schedulingPeriod {
```
d80tb7, line 210: let's log (maybe at debug) about whether we decided to schedule or not and why.  It'll come in useful if we ever end up not scheduling and we can't work out why.

internal/scheduler/scheduling_algo.go
```diff
@@ -91,16 +101,45 @@ func (l *FairSchedulingAlgo) Schedule(
...
+		}
+
+		// We sort clusters lexicographically and schedule them all in order - potentially over multiple scheduling rounds
+		// Skip any that have already been considered
+		if executor.Id <= l.previousScheduleClusterId {
+			continue
+		}
+
 		log.Infof("scheduling on %s", executor.Id)
 		schedulerResult, sctx, err := l.scheduleOnExecutor(
```
d80tb7, line 131: we seem to have ended up with one variable called `sctx` and another called `schedCtx`.   Is it possible to rename at least one of these to distinguish between them!

internal/scheduler/scheduling_algo.go
```diff
@@ -91,16 +101,45 @@ func (l *FairSchedulingAlgo) Schedule(
...
+	slices.SortStableFunc(executorsToSchedule, func(a, b *schedulerobjects.Executor) bool {
+		return strings.Compare(a.Id, b.Id) < 1
+	})
+
+	schedCtx, cancel := context.WithTimeout(ctx, l.maxSchedulingDuration)
+	defer cancel()
+
+	allExecutorsConsidered := false
+	for i, executor := range executorsToSchedule {
+		if schedCtx.Err() != nil {
```
d80tb7, line 115: do we record why we exited- e.g. either because we ran out of time or because we scheduled all executors?  If not, how difficult would it be to add?  Possibly this is something we can do when we do prom metrics

internal/scheduler/scheduling_algo.go
```diff
@@ -91,16 +101,45 @@ func (l *FairSchedulingAlgo) Schedule(
...
+			// We've reached the scheduling time limit, exit gracefully
+			break
+		}
+
+		if i+1 == len(executorsToSchedule) {
```
d80tb7, line 120: this doesn't work quite how I imagined it.  I'd assumed we'd always try and schedule at least `len(executorsToSchedule)`. I.e. assume we have 4 executors, I would asssume:
```
Scheduling round 1: schedules on 1,2,3 (times out)
Scheduling round 2: schedules on 4,1,2,3 (finishes after scheduling on n executors)
```
Whereas I think this code would do:
```
Scheduling round 1: schedules on 1,2,3 (times out)
Scheduling round 2: schedules on 4 (finishes after we reach the end of the slice)
```

d80tb7 (approved):

PR #2498, opened 2023-05-24
Add number of CPUs or GPUs to scheduler test fixtures

zuqq:
I think that this makes the accounting in the table-driven scheduler tests clearer.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-314) by [Unito](https://www.unito.io)


internal/scheduler/testfixtures/testfixtures.go
```diff
@@ -446,31 +395,31 @@ func GpuJob(queue string, priorityClassName string) *jobdb.Job {
...
 	rv := make([]*schedulerobjects.PodRequirements, n)
 	for i := 0; i < n; i++ {
-		rv[i] = TestGpuJob(queue, priority)
+		rv[i] = Test1GpuPodReqs(queue, util.ULID(), priority)
 	}
 	return rv
 }
 
-func TestSmallCpuJob(queue string, priority int32) *schedulerobjects.PodRequirements {
+func Test1CpuPodReqs(queue string, jobId ulid.ULID, priority int32) *schedulerobjects.PodRequirements {
```
zuqq, line 422: The `jobId` argument is of type `ulid.ULID` instead of `string` so that it's harder to confuse it with `queue`.

internal/scheduler/testfixtures/testfixtures.go
```diff
@@ -480,13 +429,14 @@ func TestSmallCpuJob(queue string, priority int32) *schedulerobjects.PodRequirem
 			},
 		},
 		Annotations: map[string]string{
-			schedulerconfig.JobIdAnnotation: util.NewULID(),
+			schedulerconfig.JobIdAnnotation: jobId.String(),
 			schedulerconfig.QueueAnnotation: queue,
 		},
+		NodeSelector: make(map[string]string),
```
zuqq, line 435: This needs to be set for jobs that are evicted; might be a worth a comment.

PR #2499, opened 2023-05-24
Job Service: Improve expired jobsets deletion

Mo-Fatah:
* Remove the database `delete_expired_jobsets` trigger and returned the expired-jobs deletion the the user space. This will minimize the delete operations performed.
* Instead of executing two delete statements after each insertion, the deletion will happen in a persistent goroutine with a time ticker.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-315) by [Unito](https://www.unito.io)


richscott (commented):
I had just a couple comments on the DELETE operations, otherwise it looks good!

internal/jobservice/repository/sqlite.go
```diff
@@ -374,3 +367,35 @@ func (s *JSRepoSQLite) GetSubscribedJobSets(ctx context.Context) ([]SubscribedTu
...
+	ticker := time.NewTicker(time.Duration(s.jobServiceConfig.PurgeJobSetTime) * time.Second)
+	log := log.WithField("JobService", "ExpiredJobSetsPurge")
+	jobsetsStmt := fmt.Sprintf(`DELETE FROM jobsets WHERE Timestamp < (UNIXEPOCH() - %d);`, s.jobServiceConfig.PurgeJobSetTime)
+	jobServiceStmt := fmt.Sprintf(`DELETE FROM jobservice WHERE Timestamp < (UNIXEPOCH() - %d);`, s.jobServiceConfig.PurgeJobSetTime)
+	tableExistsStmt := `SELECT EXISTS (SELECT 1 FROM sqlite_master WHERE type='table' AND name=?);`
+	for range ticker.C {
+		s.lock.Lock()
+		for _, table := range []string{"jobsets", "jobservice"} {
+			var exists bool
+			err := s.db.QueryRow(tableExistsStmt, table).Scan(&exists)
```
richscott, line 381: This is a very robust check, but I think it may not really be necessary - if either of these tables doesn't exist, the DELETE operation will return an error that will loudly say if the table is missing. (And if that's the case, then the user of this code has bigger problems.) I'd just process with running the DELETE directly.

internal/jobservice/repository/sqlite.go
```diff
@@ -374,3 +367,35 @@ func (s *JSRepoSQLite) GetSubscribedJobSets(ctx context.Context) ([]SubscribedTu
...
+				continue
+			}
+			if exists {
+				var stmt string
+				if table == "jobsets" {
+					stmt = jobsetsStmt
+				} else {
+					stmt = jobServiceStmt
+				}
+				_, execErr := s.db.Exec(stmt)
```
richscott, line 393: The first returned value (ignored here) is a `database/sql.Result` value. If there's no error here, maybe it might be useful to use the returned `Result` and call `Result.RowsAffected()`, and then maybe log (maybe at DEBUG level?) the number of rows that it deleted? That might be useful for debugging when they run this in a big production setup. What do you think?

richscott (commented):
I think it looks mostly great - I had just a couple of suggestions - let me know what you think.

internal/jobservice/repository/sqlite.go
Mo-Fatah: I agree, it would be good to log it on debug level.

internal/jobservice/repository/sqlite.go
Mo-Fatah: I added exist-check because I saw some random error logs during testing. But the CI now is not affected by so, it will remove it.

internal/jobservice/repository/postgres.go
```diff
@@ -330,3 +316,26 @@ func (s *JSRepoPostgres) GetSubscribedJobSets(ctx context.Context) ([]Subscribed
...
+
+	ticker := time.NewTicker(time.Duration(s.jobServiceConfig.PurgeJobSetTime) * time.Second)
+	log := log.WithField("JobService", "ExpiredJobSetsPurge")
+
+	for range ticker.C {
+		result, jobsetErr := s.dbpool.Exec(ctx, jobsetsStmt)
+		if jobsetErr != nil {
+			log.Error("error deleting expired jobsets: ", jobsetErr)
+		} else {
+			log.Debugf("Deleted %d expired jobsets", result.RowsAffected())
```
richscott, line 332: I think you'll need to change this to be similar to the logic in sqlite.go. `database/sql.Result.RowsAffected()` returns a `(int 64, error)`.  (Run `go doc dtabase/sql.Result` to see details.) Right now, this code produces (on my Linux box):
```
INFO[2023-06-02T17:01:24Z] Deleted %d expired jobsets0                   JobService=ExpiredJobSetsPurge
INFO[2023-06-02T17:01:24Z] Deleted 0 expired jobs                        JobService=ExpiredJobSetsPurge
```
So if you make a separate call to RowsAffected(), and check for err not being nil, etc, it should produce better output.

richscott (commented):
Just one suggestion in postgres.go to fix a logging output issue.

internal/jobservice/repository/postgres.go
Mo-Fatah: this is weird.
> I think you'll need to change this to be similar to the logic in sqlite.go. database/sql.Result.RowsAffected() returns a (int > 64, error). (Run go doc dtabase/sql.Result to see details.) Right now, this code produces (on my Linux box):

The [RowsAffected](https://pkg.go.dev/github.com/jackc/pgx/v5@v5.3.1/pgconn#CommandTag.RowsAffected) function here returns an `int64` directly, also the [Exec](https://pkg.go.dev/github.com/jackc/pgx/v5@v5.3.1/pgxpool#Conn.Exec) function here in `postgres` is different than the `sqlite` one, here it returns a `CommandTag` not `sql.Result`.

 How did you generate these logs? I will try to reproduce it and see

richscott: I'm so embarrassed - of course you are right! I forgot that for Postgres we are using the `github.com/jackc/pgx/v5` library, which has a slightly different interface than the standard `database/sql` (which the sqlite driver complies with).  My IDE shows that the pgx library has its `Exec` also returning two values: `(CommandTag, error)`.

The way I run it is:
```
$ docker system prune                          # clean out any old containers, etc
$ mage localdev minimal testsuite        # see docs/developer/manual-localdev.md for mage install details
```

That should run successfully - it will take a few minutes. After that, I made the following changes (temporary, I did not commit them), just to exercise and see the logging:
```
diff --git a/e2e/setup/jobservice.yaml b/e2e/setup/jobservice.yaml
index a64851c2..7c82e67a 100644
--- a/e2e/setup/jobservice.yaml
+++ b/e2e/setup/jobservice.yaml
@@ -1,6 +1,6 @@
 subscribeJobSetTime: 60
 subscriberPoolSize: 30
-purgeJobSetTime: 10000
+purgeJobSetTime: 100
 databaseType: "postgres"
 postgresConfig:
   maxOpenConns: 50
diff --git a/internal/jobservice/repository/postgres.go b/internal/jobservice/repository/postgres.go
index e633b6a4..54e4d1c4 100644
--- a/internal/jobservice/repository/postgres.go
+++ b/internal/jobservice/repository/postgres.go
@@ -329,13 +329,13 @@ func (s *JSRepoPostgres) purgeExpired(ctx context.Context) {
                if jobsetErr != nil {
                        log.Error("error deleting expired jobsets: ", jobsetErr)
                } else {
-                       log.Debugf("Deleted %d expired jobsets", result.RowsAffected())
+                       log.Info("Deleted %d expired jobsets", result.RowsAffected())
                }
                result, jobServiceErr := s.dbpool.Exec(ctx, jobServiceStmt)
                if jobServiceErr != nil {
                        log.Error("error deleting expired jobs: ", jobServiceErr)
                } else {
-                       log.Debugf("Deleted %d expired jobs", result.RowsAffected())
+                       log.Infof("Deleted %d expired jobs", result.RowsAffected())
                }
        }
 }
diff --git a/third_party/airflow/tests/integration/test_airflow_operator_logic.py b/third_party/airflow/tests/integration/test_airflow_operator_logic.py
index f65ced29..e523ce57 100644
--- a/third_party/airflow/tests/integration/test_airflow_operator_logic.py
+++ b/third_party/airflow/tests/integration/test_airflow_operator_logic.py
@@ -127,7 +127,7 @@ def success_job(client: ArmadaClient, jobservice: JobServiceClient):
     assert job_message == f"Armada test:{job_id} succeeded"


-@pytest.mark.skip(reason="we should not test performance in the CI.")
+#@pytest.mark.skip(reason="we should not test performance in the CI.")
 def test_parallel_execution(client: ArmadaClient, jobservice: JobServiceClient):
     threads = []
     success_job(client=client, jobservice=jobservice)
@@ -140,7 +140,7 @@ def test_parallel_execution(client: ArmadaClient, jobservice: JobServiceClient):
         thread.join()


-@pytest.mark.skip(reason="we should not test performance in the CI.")
+# @pytest.mark.skip(reason="we should not test performance in the CI.")
 def test_parallel_execution_large(client: ArmadaClient, jobservice: JobServiceClient):
     threads = []
     success_job(client=client, jobservice=jobservice)
```
Then I ran the airflow integration tests (it's misnamed - it doesn't really run Airflow):
```
$ time make tests-e2e-airflow
```
While that is running, in a separate terminal, run:
```
$ docker logs -f jobservice
```
and watch for that logging message to show up (it may take a while).

I believe there is a way to change a running container's overall logging level to Debug from the Info default, but could not quickly find out how to do that, so I just tweaked the level in the code, to verify the message it outputs.

richscott: After all that is done, to shut down all the containers, run
```
$ mage localdevstop
```
then run `docker ps` to verify all are shut down.

richscott: I just ran it again now, here is a sample of the output:
```
INFO[2023-06-02T20:49:37Z] closed the event client connection
ERRO[2023-06-02T20:49:37Z] error on subscribingstream subscription ttl exceeded: 60
INFO[2023-06-02T20:49:42Z] Deleted %d expired jobsets0                   JobService=ExpiredJobSetsPurge
INFO[2023-06-02T20:49:42Z] Deleted 0 expired jobs                        JobService=ExpiredJobSetsPurge
INFO[2023-06-02T20:50:02Z] job service has 2 subscribed job sets         JobService=Startup
INFO[2023-06-02T20:50:02Z] SubscribeToJobSetId start for queue-a/test-c4c80c94-0186-11ee-927c-0242ac13000b with messageId 1685738885791:0:0:1
INFO[2023-06-02T20:50:02Z] SubscribeToJobSetId start for queue-a/test-e897d85c-0186-11ee-927c-0242ac13000b with messageId 1685738953238:0:0:0
```

richscott: This is all on my Linux box running Ubuntu 22.04, Docker 24.0.2, all the other standard versions of our tools.

owenthomas17: The bug you mention here @richscott is in the diff you've posted [here](https://github.com/armadaproject/armada/pull/2499#discussion_r1214826449).
This line:
```
+                       log.Info("Deleted %d expired jobsets", result.RowsAffected())
```

Should be 
```
+                       log.Infof("Deleted %d expired jobsets", result.RowsAffected())
```

Mo-Fatah: Thanks @owenthomas17 !

Yes, as owenthomas17 mentioned, you used `log.Info` instead of `log.Infof` . A similar logs snippet to yours following the same steps and using `log.Infof`:
```shell
INFO[2023-06-03T18:31:59Z] closed the event client connection           
ERRO[2023-06-03T18:31:59Z] error on subscribingstream subscription ttl exceeded: 60 
INFO[2023-06-03T18:32:04Z] Deleted 0 expired jobsets                     JobService=ExpiredJobSetsPurge
INFO[2023-06-03T18:32:04Z] Deleted 0 expired jobs                        JobService=ExpiredJobSetsPurge

```
 I don't like logging zero deletion though, I don't think it is helpful and it will add only noises to the debug logs. I will change it to only log if `result.RowsAffected() > 0`.

internal/jobservice/repository/postgres.go
```diff
@@ -330,3 +316,26 @@ func (s *JSRepoPostgres) GetSubscribedJobSets(ctx context.Context) ([]Subscribed
 	}
 	return tuples, nil
 }
+
+func (s *JSRepoPostgres) purgeExpired(ctx context.Context) {
+	jobsetsStmt := fmt.Sprintf(`DELETE FROM jobsets WHERE Timestamp < (extract(epoch from now()) - %d);`, s.jobServiceConfig.PurgeJobSetTime)
```
owenthomas17, line 321: `jobsetsStmt` and `jobServiceStmt` are assigned outside of the loop below. This means that every `PurgeJobSetTime` * seconds, this routine will re-run the same statements and never increment the time that is looked back on. In practice, this will mean that after the first successful run `purgeExpired` will effectively do nothing.

internal/jobservice/repository/sqlite.go
```diff
@@ -374,3 +367,31 @@ func (s *JSRepoSQLite) GetSubscribedJobSets(ctx context.Context) ([]SubscribedTu
...
+		} else {
+			count, err := result.RowsAffected()
+			if err != nil {
+				log.Errorf("error getting affected rows for expired %s records deletion: %v", table, err)
+			}
+			log.Debugf("Deleted %d expired records in table %s", count, table)
+		}
+	}
+
+	for range ticker.C {
```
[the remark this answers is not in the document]
owenthomas17: I would also say it would be useful to explicitly log that the purge is starting once the loop starts.

internal/jobservice/repository/postgres.go
```diff
@@ -330,3 +316,26 @@ func (s *JSRepoPostgres) GetSubscribedJobSets(ctx context.Context) ([]Subscribed
...
+	jobServiceStmt := fmt.Sprintf(`DELETE FROM jobservice WHERE Timestamp < (extract(epoch from now()) - %d);`, s.jobServiceConfig.PurgeJobSetTime)
+
+	ticker := time.NewTicker(time.Duration(s.jobServiceConfig.PurgeJobSetTime) * time.Second)
+	log := log.WithField("JobService", "ExpiredJobSetsPurge")
+
+	for range ticker.C {
```
owenthomas17, line 327: It would be useful to explicitly log that the purge is starting once the loop starts. This way, when parsing through the logs in production, you'll be able to see when each purge cycle runs.

internal/jobservice/repository/postgres.go
```diff
@@ -86,6 +68,10 @@ func (s *JSRepoPostgres) Setup(ctx context.Context) {
 			panic(err)
 		}
 	}
+
+	if s.jobServiceConfig.PurgeJobSetTime > 0 {
+		go s.purgeExpired(ctx)
```
owenthomas17, line 73: Is `Setup` the right place to start this routine? My initial read is that `Startup` would be a better place for this to live.

owenthomas17 (commented):
@Mo-Fatah - I've left a couple of comments, I hope they're helpful. I think overall this implementation is an improvement over the previous one, once a few bugs have been ironed out. Feel free to take or leave the review.

internal/jobservice/repository/sqlite.go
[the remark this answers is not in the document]
Mo-Fatah: > Is there a reason that `purgeExpired` has two different implementations, one for sqlite and one for postgres? Naively I'd think that a better implementation would be to have a single `purgeExpired` function that can compose the cleanup operations for each kind of db backend.

I find it hard to think how we could have a single `purgeExpired` implementation due to differences in each db handler interface. `JSRepoSQLite` uses `database/sql.DB` as a db handler while `JSRepoPostgres` uses `pgxpool.Pool`. 

However, I believe we can add `purgeExpired` to the `SQLJobService` interface and expose its implementations so it can be launched from `Startup` instead of `Setup` as you suggested below

Mo-Fatah:
waiting until #2505 is merged since it will change how we purge expired jobs/jobsets

richscott (approved):
This looks great - a nice improvement!

ClifHouck (approved):
LGTM

PR #2504, opened 2023-05-25
fix dotnet release

Kotwic4:
Fixes #


Kotwic4:
it is ugly but fixes the problem as an workaround. Strategic move is to go out of circle ci, so I don't want to spend more time to properly fix this problem

severinson (approved):

PR #2505, opened 2023-05-25
Improve jobservice db

### LINKED ISSUE #2496: Job Service: Improve db table names, and and foreign-keys to db
Several tables and columns in the Job Service database have confusing names, and also there are no foreign-keys to indicate strict relationships between jobsets and jobs. Several changes should be made:

* rename the `jobservice` table (and all its references) to simply `jobs`
* rename `jobs.jobid` column to just `jobs.id`
* rename`jobsets.jobsetid` to `jobsets.id`
* add a foreign-key constraint on `jobs.jobsetid` to `jobsets.id`

In addition to the table creation code, all SQL references to these tables/columns should be updated, for both the Postgresql and SQLite versions.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-302) by [Unito](https://www.unito.io)

Mo-Fatah:

Fixes #2496 

- The foreign-key constraint added some restrictions on jobs creation. Before that, most tests were creating jobs without corresponding jobsets. Therefore, some tweaks happened in the tests to match the new constraint

- sqlite doesn't have foreign keys enabled by default, therefore I added `PRAGMA foreign_keys = ON` in the sqlite setup to be consistent with `postgres`


internal/jobservice/repository/postgres.go
```diff
@@ -245,11 +249,15 @@ func (s *JSRepoPostgres) SubscribeJobSet(ctx context.Context, queue string, jobS
 
 // UnSubscribe to JobSet and delete all the jobs in the database
 func (s *JSRepoPostgres) CleanupJobSetAndJobs(ctx context.Context, queue string, jobSet string) (int64, error) {
```
Mo-Fatah, line 251: with the new foreign-key constraint, we should delete the jobs first before the jobset. Also, the function is expected to return the number of affected jobs

suprjinx: could we cascade delete from jobset?

Mo-Fatah: Can you have a look at it now ? I am thinking also of deleting `CleanupJobSetAndJobs` from the `SQLJobService` interface since it is just calls `UnsubscribeJobSet` in both of it's implemetation to perform the deletion due to cascade delete.

richscott (approved):

richscott (approved):

PR #2512, opened 2023-05-26
Use correct priority to compute allocated resources

zuqq:
The issue that I'm addressing here is that `totalAllocationByPoolAndQueue` uses the `Priority` field of `JobSchedulingInfo`:

https://github.com/armadaproject/armada/blob/f3fa267c875dd7908a7efc6fd62dd38a8a014a77/internal/scheduler/scheduling_algo.go#L544-L551

… even though that field represents the per-queue priority of the job, not its priority class:

https://github.com/armadaproject/armada/blob/f3fa267c875dd7908a7efc6fd62dd38a8a014a77/internal/scheduler/schedulerobjects/schedulerobjects.pb.go#L621-L622

It might be worth having a separate type for the priority number associated with each priority class, in order to distinguish it from the other integers that are floating around.

### DIFF
--- a/internal/scheduler/scheduling_algo.go
+++ b/internal/scheduler/scheduling_algo.go
@@ -2,6 +2,7 @@ package scheduler
 
 import (
 	"context"
+	"fmt"
 	"math/rand"
 	"strings"
 	"time"
@@ -56,12 +57,9 @@ func NewFairSchedulingAlgo(
 	maxSchedulingDuration time.Duration,
 	executorRepository database.ExecutorRepository,
 	queueRepository database.QueueRepository,
-) *FairSchedulingAlgo {
-	priorityClasses := config.Preemption.PriorityClasses
-	// The comment on `DefaultPriorityClass` suggests that we should never take
-	// this branch, but we are being extra cautious here.
+) (*FairSchedulingAlgo, error) {
 	if len(config.Preemption.PriorityClasses) <= 0 {
-		priorityClasses = map[string]configuration.PriorityClass{config.Preemption.DefaultPriorityClass: {Priority: 0}}
+		return nil, fmt.Errorf("expected the priority class mapping in the preemption configuration to contain at least one element (associated with the default priority class %s), but it was empty", config.Preemption.DefaultPriorityClass)
 	}
 
 	indexedResources := config.IndexedResources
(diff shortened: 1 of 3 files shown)


internal/scheduler/scheduling_algo.go
```diff
@@ -57,13 +57,11 @@ func NewFairSchedulingAlgo(
...
-		for _, p := range config.Preemption.PriorityClasses {
-			priorities = append(priorities, p.Priority)
-		}
-	} else {
-		priorities = append(priorities, 0)
+	priorityClasses := config.Preemption.PriorityClasses
+	// The comment on `DefaultPriorityClass` suggests that we should never take
+	// this branch, but we are being extra cautious here.
+	if len(config.Preemption.PriorityClasses) <= 0 {
+		priorityClasses = map[string]configuration.PriorityClass{config.Preemption.DefaultPriorityClass: {Priority: 0}}
```
severinson, line 64: We should error on no PCs imo.

internal/scheduler/scheduling_algo.go
```diff
@@ -543,10 +541,12 @@ func (l *FairSchedulingAlgo) totalAllocationByPoolAndQueue(executors []*schedule
 			}
 			jobSchedulingInfo := job.JobSchedulingInfo()
 			if jobSchedulingInfo != nil {
-				allocation.AddResourceList(
-					int32(jobSchedulingInfo.Priority),
-					jobSchedulingInfo.GetTotalResourceRequest(),
-				)
+				priorityClass, ok := l.priorityClasses[jobSchedulingInfo.PriorityClassName]
```
severinson, line 544: We shouldn't schedule this job. Either skip the job entirely or return an error.

zuqq: This function is called at the start of the scheduling round, when constructing a `fairShareSchedulingContext`:

https://github.com/armadaproject/armada/blob/f3fa267c875dd7908a7efc6fd62dd38a8a014a77/internal/scheduler/scheduling_algo.go#L289-L290

If I understand correctly, the jobs that we are looping over have already been scheduled; in particular, encountering an unknown priority class here means that the scheduler is in an inconsistent state.

This function could return an error instead, but I'm worried about the implications of making this change. If we abort the scheduling round here, wouldn't the next leader just run into the same issue?

zuqq: Actually, returning an error would be consistent with some of the other error handling in `scheduleOnExecutor`:

https://github.com/armadaproject/armada/blob/f3fa267c875dd7908a7efc6fd62dd38a8a014a77/internal/scheduler/scheduling_algo.go#L373-L377

On the other hand, `constructNodeDb` seems to prefer to just log and carry on:

https://github.com/armadaproject/armada/blob/f3fa267c875dd7908a7efc6fd62dd38a8a014a77/internal/scheduler/scheduling_algo.go#L435-L442

severinson (approved):

PR #2523, opened 2023-05-31
Job Service: rework job-set subscription processing to use worker pool  (BATCH-303)

richscott:
These changes rework the job-set subscription processing logic in the Job Service from using ad-hoc-spawned goroutines and an atomic map for locking to instead use a fixed pool of worker goroutines that read from a single channel of subscription requests. This should result in smoother run-time behavior of goroutine usage, GRPC and database connections.

Fixes #BATCH-303


internal/jobservice/application.go
```diff
@@ -100,46 +127,35 @@ func (a *App) StartUp(ctx context.Context, config *configuration.JobServiceConfi
...
+		evClient := events.NewPooledEventClient(evConnPool)
+		subRequests := make(chan SubRequest)
 		ticker := time.NewTicker(30 * time.Second)
-		eventClient := events.NewPooledEventClient(pool)
-		var subscribeMap sync.Map
-		for range ticker.C {
 
-			jobSets, err := sqlJobRepo.GetSubscribedJobSets(ctx)
-			log.Infof("job service has %d subscribed job sets", len(jobSets))
+		for w := 1; w <= config.SubscribeJobSetPoolSize; w++ {
```
suprjinx, line 145: It's probably just me, but I kept reading this name as a verb -- "Subscribe!" -- but in reality is a thing -- "Pool". Would a name like "SubscriberPool" or something make that more clear? "JobSetObserverPool"?

richscott: Good catch - I've renamed it to `SubscriberPoolSize` in commit ae717b13.

suprjinx (approved):

PR #2528, opened 2023-06-01
Update armada-scheduler scheduling config

JamesMurkin:
This should better match the armada api scheduling config

It removes config that no longer exists / swaps to using the new names for config that has changed

### DIFF
--- a/config/scheduler/config.yaml
+++ b/config/scheduler/config.yaml
@@ -50,6 +50,7 @@ scheduling:
   executorTimeout: 10m
   enableAssertions: true
   preemption:
+    alwaysAttemptScheduling: false
     enabled: true
     nodeEvictionProbability: 1.0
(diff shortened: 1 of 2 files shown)


severinson:
Think you're missing these:
https://github.com/armadaproject/armada/blob/master/internal/armada/configuration/types.go#L184
https://github.com/armadaproject/armada/blob/master/internal/armada/configuration/types.go#L205
https://github.com/armadaproject/armada/blob/master/internal/armada/configuration/types.go#L210
https://github.com/armadaproject/armada/blob/master/internal/armada/configuration/types.go#L219

severinson (approved):

severinson:
I've approved. Does the new scheduler actually set the priorityClassNameOverride?

JamesMurkin:
Yes it seems to, well if the config is set it does

PR #2538, opened 2023-06-05
Bind jobs to nodes in scheduling algo test harness

zuqq:
This means that we also set fields like `AllocatedByJobId` on the node.


internal/scheduler/scheduling_algo_test.go
```diff
@@ -361,16 +362,23 @@ func TestLegacySchedulingAlgo_TestSchedule(t *testing.T) {
 
 			for executorId, jobsByNodeName := range tc.existingRunningIndices {
 				for nodeName, jobIndices := range jobsByNodeName {
+					node := nodes[executorId][nodeName]
```
severinson, line 365: Maybe move to where we initialise nodes for tests.

severinson (approved):

PR #2545, opened 2023-06-06
Clif/jobservice rework stream common/batch 305

ClifHouck:
### Improve Jobservice jobset subscription handling/streaming.

Highlights:

- Job set subscriptions and subscription management moves to a completely `select` & channel oriented execution model. Think event-driven vs polling. No sleeping.
- No more `O(n)` loops over job set subscriptions.
- Job set subscriptions stop and signal the `JobSetSubscriptionExecutor` when they are done.
- The gRPC server notifies the `JobSetSubscriptionExecutor` of new subs immediately via a channel. No waiting to start a new subscription.

TODO:

- [ ] Need reviewer consensus on when to unsubscribe from a job set.
- [x] Unit tests.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-335) by [Unito](https://www.unito.io)

Local testing has shown this to be an extremely responsive and efficient way of handling job set subscriptions.


dave-gantenbein (commented):
Should we have some tests here?

ClifHouck:
> Should we have some tests here?

See TODO in original post.

internal/jobservice/application.go
```diff
@@ -95,6 +82,12 @@ func (a *App) StartUp(ctx context.Context, config *configuration.JobServiceConfi
 
 	RectifyConfig(config)
 
+	if os.Getenv("JOBSERVICE_DEBUG") != "" {
+		log.SetLevel(log.DebugLevel)
+		log.Debug("Set logging to debug level")
+		log.Debugf("Subscription time: %d", config.SubscribeJobSetTime)
```
richscott, line 88: This suggestion isn't prompted by your changes here, but reminded me of something that I've thought we should change for a while:  can we rename `JobServiceConfiguration.SubscribeJobSetTime`? It's an `int64` value that actually denotes the number of seconds before a job-set is considered expired, and having it named with a "Time" suffix implies that it's some sort of changing timestamp. What do you think about renaming it to `SubscriptionExpirySecs` or somesuch?  The current name has misled reviewers in other PRs into thinking it's a timestamp value that should get regular updates, when it actually does not.

internal/jobservice/eventstojobs/manage_subs.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+func (jse *JobSetSubscriptionExecutor) addSubscription(sub *repository.SubscribedTuple) {
+	jse.mutex.Lock()
+	defer jse.mutex.Unlock()
+
+	_, ok := jse.subscriptions[sub.JobSetKey]
+	if ok {
+		// TODO: What to do if subscription already exists locally?
+		// Restart it? Do nothing?
```
richscott, line 134: 💯 Good comments for future consideration.

internal/jobservice/eventstojobs/manage_subs.go
```diff
@@ -0,0 +1,331 @@
...
+	g.Go(func() error {
+		nextRecv := time.After(1 * time.Nanosecond)
+
+		// this loop will run until the context is canceled
+		for {
+			select {
+			case <-js.ctx.Done():
+				log.WithFields(requestFields).Debug("context is done")
+				return nil
+			case <-nextRecv:
```
richscott, line 275: Would changing `case <-nextRecv:` here to just a plain `default:` (and dropping the `nextRecv` var) work?

richscott (commented):
Looks great! I had just one suggestion for a renaming that I've wanted for a whle (before this PR) and a question about a select.

internal/jobservice/application.go
ClifHouck: Fine by me. Better names make everything clearer.

internal/jobservice/eventstojobs/manage_subs.go
ClifHouck: It would work, but  if there's a recoverable stream error, we're going to hammer the `stream.Recv()` call until it stops erroring.

internal/jobservice/application.go
ClifHouck: Addressed in https://github.com/armadaproject/armada/pull/2545/commits/4d6715a1443583284a1b2b36438544fbf65003b1

internal/jobservice/eventstojobs/manage_subs.go
richscott: I don't understand - with `nextRecv` being a 1 ns ticker, wouldn't that also do the same?

ClifHouck: No, its not a ticker, its a `time.After` which only fires once, and when it hits an error the loop sets the `nextRecv` to `time.After(5 * time.Seconds)`.

richscott (approved):
Looks good - great work!

internal/jobservice/eventstojobs/manage_subs.go
richscott: Aha - yes, of course - my error. Now, ☕ .

PR #2547, opened 2023-06-08
Fix example `.armadactl.yaml`

zuqq:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-337) by [Unito](https://www.unito.io)


ShivangShandilya:
@zuqq I was thinking of removing this `example structure` section from armadactl as if you take a look at other command-line tools such sections are not present there basically it will look good if we can remove this and if it doesn't have any use, I also opened a PR for this but I just saw yours and thought about telling you about this

PR #2549, opened 2023-06-08
Update armadactl.sh

ShivangShandilya:
Fixes #

We need to check if instead of `go run cmd/armadactl/main.go create queue e2e-test-queue` user can run the same command using only `armadactl create queue e2e-test-queue`

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-339) by [Unito](https://www.unito.io)


Sharpz7 (approved):
LGTM :))

PR #2551, opened 2023-06-10
Removing `example structure` from armadactl

ShivangShandilya:

This PR just removes the `example structure` from the armadactl

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-341) by [Unito](https://www.unito.io)


cmd/armada-load-tester/cmd/root.go
```diff
@@ -17,19 +17,6 @@ func init() {
 var rootCmd = &cobra.Command{
 	Use:   "armada-load-tester command",
 	Short: "Command line utility to submit many jobs to armada",
-	Long: `
```
kannon92, line 20: Why did we remove the Long description?

ShivangShandilya: Oops thanks fo noticing, I only wanted to remove the `example structure`, let me correct that

ShivangShandilya: @zuqq has provided spaces and all in the same `example structure` that I'm trying to remove, I also commented on his PR but did'nt hear back from him. This deletion is not vital to the project its just that of you take a look at any other command-line tool eg: `kubectl` you won't any kind of structure, you only see what kubectl does and its command so only for the purpose to make `armadactl` look good I have proposed this change

kannon92 (approved):

PR #2552, opened 2023-06-12
NodeDb performance improvements

severinson:
The current NodeDb implementation effectively iterates over groups of nodes, where all nodes in a group have an equal amount  allocatable of some resource, e.g., CPU. This causes issues in fragmented clusters, where all nodes may differ in amount of allocatable resources.  With this PR, we introduce two optimisations:

1. Reducing the overhead associated with each group.
2. Bucketing nodes with similar amounts of allocatable resources, e.g., all nodes with [2, 3) CPU allocatable, thus reducing the number of groups.

The overall speedup for a particular benchmark of interest is about a factor 10x. It also significantly reduces allocations, which will reduce GC pressure.

Current:
```
> go.exe test -benchmem -run=^$ -bench ^BenchmarkNodeTypeIterator$ github.com/armadaproject/armada/internal/scheduler/nodedb
goos: windows
goarch: amd64
pkg: github.com/armadaproject/armada/internal/scheduler/nodedb
cpu: 12th Gen Intel(R) Core(TM) i7-12700H
BenchmarkNodeTypeIterator-20               39607             29767 ns/op           21900 B/op        698 allocs/op
PASS
ok      github.com/armadaproject/armada/internal/scheduler/nodedb       1.652s
```

New:
```
> go.exe test -benchmem -run=^$ -bench ^BenchmarkNodeTypeIterator$ github.com/armadaproject/armada/internal/scheduler/nodedb
goos: windows
goarch: amd64
pkg: github.com/armadaproject/armada/internal/scheduler/nodedb
cpu: 12th Gen Intel(R) Core(TM) i7-12700H
BenchmarkNodeTypeIterator-20              401348              2810 ns/op            1376 B/op         34 allocs/op
PASS
ok      github.com/armadaproject/armada/internal/scheduler/nodedb       1.324s
```


internal/scheduler/nodedb/nodedb.go
```diff
@@ -878,3 +901,31 @@ func (nodeDb *NodeDb) stringFromPodRequirementsNotMetReason(reason schedulerobje
 		return s
 	}
 }
+
+func (nodeDb *NodeDb) nodeDbKeyFromNode(out []byte, node *schedulerobjects.Node, priority int32) []byte {
```
severinson, line 905: Previously, we relied on memdb methods for constructing a combined key from the node nodeType and resources. However, this caused lots of unnecessary allocs. Now we create the key ourselves using this method instead.

internal/scheduler/schedulerobjects/nodetype.go
```diff
@@ -67,40 +66,44 @@ func NewNodeType(taints []v1.Taint, labels map[string]string, indexedTaints map[
...
+// nodeTypeIdFromTaintsAndLabels generates a id unique for each combination of taints, labels, and unset labels.
+// The id is based on the fnv1a hash. Hash collisions do not affect correctness, only the efficiency of sorting out nodes.
 //
 // We separate taints/labels by $, labels and values by =, and and groups by &,
 // since these characters are not allowed in taints and labels; see
 // https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
 // https://man.archlinux.org/man/community/kubectl/kubectl-taint.1.en
-func nodeTypeIdFromTaintsAndLabels(taints []v1.Taint, labels, unsetIndexedLabels map[string]string) string {
-	// TODO: To reduce key size (and thus improve performance), we could hash the string.
+func nodeTypeIdFromTaintsAndLabels(taints []v1.Taint, labels, unsetIndexedLabels map[string]string) uint64 {
```
severinson, line 76: I've changed this from a string to a uint64. As a result, it's of fixed-length. While unlikely, I believe the previous varying-length id could cause issues with the compound key used for node indexing. As a bonus, using a uint64 also improves performance.

internal/scheduler/schedulerobjects/schedulerobjects.proto
```diff
@@ -31,6 +31,8 @@ message Node {
     string id = 1;
     // Kubernetes node name. Not guaranteed to be unique across clusters.
     string name = 12;
+    // Keys by which nodes are sorted in the nodeDb.
+    repeated bytes nodeDbKeys = 18;
```
severinson, line 35: We now compute the index keys for the node once on upsert and store them on the node. Instead of re-computing these keys every time we iterate over the nodes, which caused performance issues.

internal/armada/server/lease.go
```diff
@@ -383,9 +383,6 @@ func (q *AggregatedQueueServer) getJobs(ctx context.Context, req *api.StreamingL
 		nodes = append(nodes, node)
 	}
 	indexedResources := q.schedulingConfig.IndexedResources
```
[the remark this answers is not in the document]
severinson: Agreed. Fixed.

internal/scheduler/gang_scheduler_test.go
```diff
@@ -136,10 +137,53 @@ func TestGangScheduler(t *testing.T) {
...
+				testfixtures.TestSchedulingConfig(),
+			),
+			Nodes: testfixtures.N32CpuNodes(3, testfixtures.TestPriorities),
+			Gangs: [][]*jobdb.Job{
+				testfixtures.N16CpuJobs("A", testfixtures.PriorityClass0, 1),
+				testfixtures.N16CpuJobs("A", testfixtures.PriorityClass0, 1),
+				testfixtures.N16CpuJobs("A", testfixtures.PriorityClass0, 1),
+			},
+			ExpectedScheduledIndices: testfixtures.IntRange(0, 2),
+		},
```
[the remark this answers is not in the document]
severinson: I've changed the names to (hopefully) be more clear. Let me know if you have ideas.

zuqq (approved):

PR #2553, opened 2023-06-12
Add number of preempted jobs and gangs to `SchedulingContext`

zuqq:


internal/scheduler/scheduling_algo_test.go
```diff
@@ -410,6 +413,26 @@ func TestLegacySchedulingAlgo_TestSchedule(t *testing.T) {
 
 			assert.Equal(t, len(expectedScheduledJobs), len(schedulerResult.ScheduledJobs))
 
+			schedulingContextByExecutor := schedulingContextRepo.GetMostRecentSchedulingContextByExecutor()
```
zuqq, line 416: The test harness is getting quite long, but it's just so convenient to test this here.

PR #2555, opened 2023-06-12
Upgrade to Pulsar 3.0.1 (BATCH-318)

richscott:
Update Pulsar images to version 3.0.0. That version now has direct support for ARM64 images, so remove references to locally-built ones. Update pulsar-client Go library to latest version (v0.10.0).

Also, change a `rmdir` invocation to `rm -r` in e2e test target of makefile, in case the .kube directory is not already empty.

Some mocks had to be regenerated (via `make generate` target) as the new `pulsar-client-go` version now has `Consumer.Ack(Message)` and `Consumer.AckID(msgID)` returning an `error`.

All unit tests, plus the `tests-e2e` and `tests-e2e-airflow` integration tests pass successfully, with these changes.

**These changes won't be merged until approved and/or tested by GR devs.**

Fixes https://gr-oss.atlassian.net/browse/BATCH-318

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-345) by [Unito](https://www.unito.io)


Sharpz7:
I see no problems with this, outside of changing the `go.mod`. I found that doing that broke the linting completely as well when I tried it with goreleaser.

Sharpz7:
Can you add in the new image here as well? https://github.com/armadaproject/armada/blob/master/magefiles/developer.go#L32

internal/lookout/repository/schema/statik/statik.go
```diff
@@ -9,6 +9,6 @@ import (
 const LookoutSql = "lookout/sql" // static asset namespace
 
 func init() {
```
kannon92, line 11: From this file and below, do you know why these were included?  Looks like we committed code unrelated to pulsar upgrade as I hope lookout is unrelated to Pulsar.

richscott: I agree - Lookout is unrelated to Pulsar (as far as I know). When I did the Pulsar 3 upgrade, I had to upgrade the `pulsar-client-go` library (see the (very few) changes in `go.mod`), and they changed the signature on about 3 functions to start returning an `error` (when they returned nothing before). So I fixed our interface definitions that needed those changes, which required regenerating the mocks - that last step caused all the changes in the lookout/lookoutv2 dirs, including that giant statik definition. I thought about seeing if I could revert some of the files that were changed by `generate`, but compilation and testing quickly failed, so I just stayed with clean generated updates (I did no manual tweaks to those files after running the generate step).

kannon92 (dismissed):

PR #2561, opened 2023-06-14
Job Service: remove redundant call

Mo-Fatah:
Partially addresses #2494 
This is a temporary fix for the redundant call of `UpdateJobSetDb`. 
It is not a prefect to remove subscription check from `UpdateJobSetDb` because we should update only subscribed jobsets. However, the logic in `server.go` has to do this check first to subscribe any new incoming jobsets. So, I believe the current fix is not perfect but it is safe until a further broader changes to how we perform new jobset subscriptions.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-350) by [Unito](https://www.unito.io)


richscott (approved):
Looks good!

PR #2563, opened 2023-06-15
Filter list of queues in verbose scheduling report

zuqq:
Verbose scheduling reports currently include all the queues, which can be too noisy.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-352) by [Unito](https://www.unito.io)


internal/scheduler/context/context.go
```diff
@@ -154,34 +154,32 @@ func (sctx *SchedulingContext) ReportString(verbosity int32) string {
...
+	if verbosity <= 0 {
+		fmt.Fprintf(w, "Scheduled queues:\t%v\n", maps.Keys(scheduled))
+	} else {
+		fmt.Fprint(w, "Scheduled queues:\n")
+		for queueName, qctx := range scheduled {
+			fmt.Fprintf(w, "\t%s:\n", queueName)
+			fmt.Fprintf(w, indent.String("\t\t", qctx.ReportString(verbosity-1)))
+		}
+	}
+	preempted := armadamaps.Filter(
```
zuqq, line 172: In principle this can overlap with `scheduled`; should I avoid duplication in that case?

PR #2565, opened 2023-06-15
NodeDb: consider all nodes within the same bucket

severinson:
This PR fixes a bug where not all nodes rounded to the same amount of allocatable resources were considered.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-354) by [Unito](https://www.unito.io)

### DIFF
--- a/internal/scheduler/testfixtures/testfixtures.go
+++ b/internal/scheduler/testfixtures/testfixtures.go
@@ -477,8 +477,8 @@ func Test1GpuPodReqs(queue string, jobId ulid.ULID, priority int32) *schedulerob
 		jobId,
 		priority,
 		v1.ResourceList{
-			"cpu":    resource.MustParse("4"),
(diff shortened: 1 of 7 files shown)


internal/scheduler/testfixtures/testfixtures.go
```diff
@@ -477,8 +477,8 @@ func Test1GpuPodReqs(queue string, jobId ulid.ULID, priority int32) *schedulerob
 		jobId,
 		priority,
 		v1.ResourceList{
-			"cpu":    resource.MustParse("4"),
-			"memory": resource.MustParse("16Gi"),
+			"cpu":    resource.MustParse("8"),
```
severinson, line 480: To make the request 1/8 of the node.

zuqq (approved):

PR #2590, opened 2023-06-20
Avoid repeatedly copying node when evicting jobs

zuqq:
Comparison (using [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat)) of `go test ./internal/scheduler -bench='^BenchmarkPreemptingQueueScheduler$' -count=10 -run='^$'` before (`old.txt`) and after (`new.txt`) commit d30946186e6a8a75e2d26f51450c03746a424dd0:

```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i5-8210Y CPU @ 1.60GHz
                                                            │   old.txt    │               new.txt                │
                                                            │    sec/op    │    sec/op     vs base                │
PreemptingQueueScheduler/1000_nodes_1_queue_320000_jobs-4      6.369 ±  5%    5.269 ± 20%  -17.27% (p=0.005 n=10)
PreemptingQueueScheduler/1000_nodes_10_queues_320000_jobs-4    5.724 ±  3%    4.406 ±  2%  -23.02% (p=0.000 n=10)
PreemptingQueueScheduler/1_node_1_queue_320_jobs-4            4.779m ±  2%   3.649m ±  3%  -23.65% (p=0.002 n=10)
PreemptingQueueScheduler/1_node_10_queues_320_jobs-4          5.779m ±  3%   6.538m ± 73%        ~ (p=0.143 n=10)
PreemptingQueueScheduler/10_nodes_1_queue_3200_jobs-4         49.49m ±  4%   38.21m ±  1%  -22.78% (p=0.000 n=10)
PreemptingQueueScheduler/10_nodes_10_queues_3200_jobs-4       62.02m ±  4%   47.82m ±  5%  -22.89% (p=0.000 n=10)
PreemptingQueueScheduler/100_nodes_1_queue_32000_jobs-4       544.5m ±  4%   424.8m ±  2%  -21.99% (p=0.000 n=10)
PreemptingQueueScheduler/100_nodes_10_queues_32000_jobs-4     789.9m ± 11%   534.5m ± 11%  -32.33% (p=0.000 n=10)
geomean                                                       184.3m         148.0m        -19.70%
```

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-379) by [Unito](https://www.unito.io)


internal/scheduler/nodedb/nodedb.go
```diff
@@ -556,59 +557,52 @@ func BindPodToNode(req *schedulerobjects.PodRequirements, node *schedulerobjects
...
 	if _, ok := node.AllocatedByQueue[queue]; !ok {
-		return nil, errors.Errorf("queue %s has no resources allocated on node %s", queue, node.Id)
+		return errors.Errorf("queue %s has no resources allocated on node %s", queue, node.Id)
 	}
+
 	if node.EvictedJobRunIds == nil {
 		node.EvictedJobRunIds = make(map[string]bool)
 	}
 	if _, ok := node.EvictedJobRunIds[jobId]; ok {
-		// TODO: We're using run ids instead of job ids for now.
```
zuqq, line 599: As far as I can tell we are actually using job IDs here. On `master`, the value of `jobId` comes from `JobIdFromPodRequirements`:

https://github.com/armadaproject/armada/blob/a678bb59a73f7dfe030f5ce62059a006208f09f8/internal/scheduler/nodedb/nodedb.go#L564

… which gets it from the job ID annotation:

https://github.com/armadaproject/armada/blob/a678bb59a73f7dfe030f5ce62059a006208f09f8/internal/scheduler/nodedb/nodedb.go#L674

… and that's set to the job ID:

https://github.com/armadaproject/armada/blob/a678bb59a73f7dfe030f5ce62059a006208f09f8/internal/scheduler/common.go#L263

internal/scheduler/preempting_queue_scheduler.go
```diff
@@ -808,41 +808,41 @@ func (evi *Evictor) Evict(ctx context.Context, it nodedb.NodeIterator) (*Evictor
...
 				continue
 			}
-			req := PodRequirementFromLegacySchedulerJob(job, evi.priorityClasses)
-			if req == nil {
-				continue
+			if affectedNode == nil {
+				affectedNode = originalNode.DeepCopy()
 			}
-			node, err = nodedb.EvictPodFromNode(req, node)
+			err = nodedb.EvictJobFromNodeInPlace(evi.priorityClasses, job, affectedNode)
```
zuqq, line 833: It would be nicer to use the `UnbindPodsFromNode`/`unbindPodFromNodeInPlace` pattern, but this way of making the change is much easier; there are a couple of things going on in this function.

internal/scheduler/nodedb/nodedb.go
severinson: Yes, that's right. In the current system, the executors (more specifically, Kubernetes) generate the run ids. As a result, the scheduler has no clue what they are and we use jobs ids instead. In the Pulsar-backed scheduler, the scheduler creates the run ids and we'll be able to switch to using run ids here.

Issue #2483, opened 2023-05-18
Resiliency for ArmadaOperator Job Service calls


iJanki-gr:
ArmadaOperator depends on jobservice and Armada client instances.

If the jobservice dies the task immediately fails. I think the proper way would be for it to have some backed off retries before giving up.

Same approach is used in many other operators such as KubernetesPodOperator. If an API call to K8s fails the call will be retried a few times "reattaching" to the pod. In particular KPO uses "tenacity" library that could be used here too.

```
@tenacity.retry(stop=tenacity.stop_after_attempt(3), wait=tenacity.wait_exponential(), reraise=True)
     def read_pod(self, pod: V1Pod) -> V1Pod:
         ...
```

I think this is the only way to have resilience for jobservice. Even if we have multiple jobservices using a postgres db, if one replica goes down the connection will be lost and need to be reattempted to connect to the other replica.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-297) by [Unito](https://www.unito.io)

iJanki-gr:
At the moment, if jobservice is down the task will fail. If the jobservice goes down while the task is running it will fail too.

Having multiple replicas in k8s won't help unless there is a retry mechanism built in the operator.
The production scenario would look like multiple replicas and a k8s service. There is no API gateway.

pavan12395:
Hi @kannon92 , I want to take up this issue.

Going through the code of JobService & Armada Operator and above comments , Work involved in this issue could be..

1) Adding back-off retry algorithm in ArmadaClient & JobServiceClient 
2) Adding replicas for JobServiceServer.

It would be helpful for me if anyone can provide feedback on this...

PR #2591, opened 2023-06-20
Resiliency for ArmadaOperator services calls 

pavan12395:
### This Pull Request makes the following changes

Fixes issue https://github.com/armadaproject/armada/issues/2483

- Adds BackOff retry algorithm for ArmadaClient & JobServiceClient
- Contains Changes for Deploying JobService as a Deployment rather than StatefulSet

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-380) by [Unito](https://www.unito.io)


client/python/armada_client/client.py
```diff
@@ -21,6 +21,7 @@
 from armada_client.k8s.io.api.core.v1 import generated_pb2 as core_v1
 from armada_client.permissions import Permissions
 from armada_client.typings import JobState
+import tenacity
```
kannon92, line 24: I think you will need to add tenacity to our pyproject.toml file as this is a new dependency.  

You will also need to run the python code formatter (Black via `tox -e format-code`) to get your code to pass CI.

third_party/airflow/armada/operators/jobservice.py
```diff
@@ -30,11 +32,13 @@ def get_job_status(
         :param job_id: The id of the job
         :return: A Job Service Request (State, Error)
         """
+        print("Hello world")
```
kannon92, line 35: Please remove.

client/python/armada_client/client.py
pavan12395: Done!

kannon92: Can you remove tenacity from the client?  Not sure I like this change being enforced on all users of the client.

I’d leave the changed in jobservice.  Can you also experiment with deploying jobservice with multiple replicas

pavan12395: I didnt get you but in the comments of the issue , It was mentioned that we need to include backoff for all the clients(ArmadaClient & JobServiceClient).

kannon92: Thanks for removing the client side.  The main user of this is mostly for jobservice and I'm not sure we want to add retries to all users of the client.

kannon92 (dismissed):
The code looks good.  Maybe you can work on having replicas for jobservice in a separate PR?

pavan12395:
Hey @ClifHouck , I found out that tenacity works fine with async functions as well , So I added the functionality for JobServiceAsyncIOClient as well.

third_party/airflow/armada/operators/jobservice.py
```diff
@@ -18,6 +20,11 @@ class JobServiceClient:
     def __init__(self, channel):
         self.job_stub = jobservice_pb2_grpc.JobServiceStub(channel)
 
+    @tenacity.retry(
```
kannon92, line 23: I'd love some kind of test that can verify that this library works with GRPC.  

Can you look into adding a unit test?  You could use our mock python grpc class for jobservice.

Otherwise this is looking really good!

pavan12395: The test should check whether tenacity makes retries or not , right?

kannon92: Yep.

client/python/armada_client/client.py
pavan12395: Hey @kannon92  , I added the test for tenacity.

kannon92:
Hey @pavan12395 We got a request form our user for this feature to go a different way.  We are going to close this PR.  

Nice work on it so far!  We decided to go with GRPC retries rather than tenacity.

PR #2596, opened 2023-06-21
add: armadactl.md

ShivangShandilya:
What type of PR is this?

This PR adds about `armadactl` along with its subcommands  on the Armada Website


armadactl.md
```diff
@@ -0,0 +1,48 @@
...
+If you need help, run `armadactl --help` from the terminal window.
+
+## Operations
+
+The following table includes short descriptions and the general syntax for all of the armadactl operations:
+
+Operation       | Syntax    |       Description
+-------------------- | -------------------- | --------------------
+`analyze`    | `armadactl analyze [path/to/manifests] [flags]` | The analyze subcommand can be used to analyze a set of manifests and provide a report on their compatibility with a given Kubernetes cluster version.
+`cancel`    | `armadactl cancel [deployment_name] [flags]` | The cancel subcommand can be used to cancel a running Armada deployment.
```
kannon92, line 31: Not really sure what you mean by Armada deployment.  Cancel takes either `queue` and `jobSet` or `jobId`.  And it either cancels all the jobs in a queue + jobSet or it cancels an individual job.

ShivangShandilya: Hmm I copied the content from the main repository of armada if its incorrect, we might wanna correct this there too
https://github.com/armadaproject/armada/tree/master/cmd/armadactl

kannon92: I see.  I'll see what the team thinks.  If we feel that this is misleading than updating those comments would be useful.

kannon92 (approved):

armadactl.md
```diff
@@ -0,0 +1,48 @@
+# Command line tool (armadactl)
+
+Armadactl is a command-line tool used for managing jobs in the Armada workload orchestration system. It provides functionality for creating, updating, and deleting jobs, as well as monitoring job status and resource usage.
```
Sharpz7, line 3: Yeah I agree with Kevin, I think a lot of this is missleading. For example, armadactl does not manage jobs, it just helps you submit, watch and cancel them. I would argue Armada itself is what is managing jobs.

ShivangShandilya: So, how should we proceed with this? Because we'll be adding `armadactl` on the website and then we'll have to make changes again

Sharpz7: I would want to see an attempted re-write. Try and make it clear that it is not managing, but communicating.

https://kubernetes.io/docs/reference/kubectl/ I think this is a good refrence

PR #2601, opened 2023-06-22
Separate resource limits for each priority class

severinson:
Currently, per-queue resource limits are based on resources allocated to jobs of a particular priority. This causes problems when several priority classes have the same priority, since all jobs with priority classes of equal priority count towards the same limit. In this PR, we change the limits to be based on resources allocated to jobs of a particular priority class. This allows us to apply separate limits to different priority classes even when they have the same priority.

If at some point we want to better support using several priority classes with different priorities, I suggest we introduce explicit cumulative limits in a separate PR, similar to those we've had previously but not defined as part of the priority classes (since each priority class may not have a unique priority).


internal/armada/server/lease.go
```diff
@@ -282,10 +283,31 @@ func (q *AggregatedQueueServer) getJobs(ctx context.Context, req *api.StreamingL
 		})
 	}
 
+	// Map queue names to priority factor for all active queues, i.e.,
```
severinson, line 286: This is just moved to earlier in the same function.

internal/armada/server/lease.go
```diff
@@ -282,10 +283,31 @@ func (q *AggregatedQueueServer) getJobs(ctx context.Context, req *api.StreamingL
...
+	priorityFactorByActiveQueue := make(map[string]float64, len(activeQueues))
+	for _, queue := range activeQueues {
+		priorityFactorByActiveQueue[queue.Name] = priorityFactorByQueue[queue.Name]
+	}
+
 	// Nodes to be considered by the scheduler.
 	lastSeen := q.clock.Now()
 	nodes := make([]*schedulerobjects.Node, 0, len(req.Nodes))
-	allocatedByQueueForCluster := make(map[string]schedulerobjects.QuantityByPriorityAndResourceType)
+	allocatedByQueueAndPriorityClassForCluster := make(map[string]schedulerobjects.QuantityByTAndResourceType[string], len(queues))
```
severinson, line 310: We now track resources by priority class name instead of priority class priority.

internal/scheduler/schedulerobjects/resourcelist.go
```diff
@@ -39,83 +39,85 @@ func V1ResourceListFromResourceList(rl ResourceList) v1.ResourceList {
 	return rv
 }
 
-type QuantityByPriorityAndResourceType map[int32]ResourceList
+type QuantityByTAndResourceType[T comparable] map[T]ResourceList
```
severinson, line 42: I've replaces `QuantityByPriorityAndResourceType` with the more generic `QuantityByTAndResourceType[int32]`, which also allows us to use `QuantityByTAndResourceType[string]`, to track resources by priority class name.

internal/armada/server/lease.go
```diff
@@ -282,10 +283,31 @@ func (q *AggregatedQueueServer) getJobs(ctx context.Context, req *api.StreamingL
...
+	apiQueues := make([]*api.Queue, len(queues))
+	for i, queue := range queues {
+		priorityFactorByQueue[queue.Name] = float64(queue.PriorityFactor)
+		apiQueues[i] = &api.Queue{Name: queue.Name}
+	}
+	activeQueues, err := q.jobRepository.FilterActiveQueues(apiQueues)
+	if err != nil {
+		return nil, err
+	}
+	priorityFactorByActiveQueue := make(map[string]float64, len(activeQueues))
```
JamesMurkin, line 302: Is this actually used anywhere?

PR #2614, opened 2023-06-27
Add dominant resource fairness

severinson:
Do not merge yet. Should be merged after https://github.com/armadaproject/armada/pull/2611
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-403) by [Unito](https://www.unito.io)


internal/armada/configuration/types.go
```diff
@@ -113,7 +113,11 @@ type SchedulingConfig struct {
 	DefaultJobTolerationsByResourceRequest map[string][]v1.Toleration
 	// Maximum number of times a job is retried before considered failed.
 	MaxRetries uint
-	// Weights used when computing fair share.
+	// Controls how fairness is calculated. Can be either AssetFairness or DominantResourceFairness.
```
zuqq, line 116: This list of possible values is bound to become out of date.

internal/armada/configuration/types.go
```diff
@@ -187,6 +191,33 @@ type SchedulingConfig struct {
...
+// In particular, each queue has a cost associated with it and the next job to attempt to schedule
+// is taken from the queue with the smallest cost associated with it.
+//
+// Can be either AssetFairness or DominantResourceFairness.
+//
+// If AssetFairness, the cost associated with a queue is a linear combination of its total allocation.
+// E.g., w_CPU * "CPU allocation" + w_memory * "memory allocation".
+//
+// If DominantResourceFairness, the cost associated with a queue is
+// max("CPU allocation" / "CPU capacity", "memory allocation" / "mamory capacity", ...).
```
zuqq, line 204: Same comment as above; also, I'd attach the explanations to their respective values.

internal/scheduler/context/context.go
```diff
@@ -34,6 +35,11 @@ type SchedulingContext struct {
 	PriorityClasses map[string]configuration.PriorityClass
 	// Default priority class.
 	DefaultPriorityClass string
+	// Determines how fairness is computed.
+	FairnessType configuration.FairnessType
+	// Used to convert one resource into another when computing fair share.
+	// Only applies to DominantResourceFairness.
+	FairnessResourceMappingBySourceResource map[string]configuration.ResourceMapping
```
zuqq, line 42: This doesn't seem to be used yet.

Also, it's not clear to me why resource mapping would only apply to `DominantResourceFairness`; the fractional GPU example you gave above seems like it also makes sense in the case of asset fairness.

internal/scheduler/context/context.go
```diff
@@ -490,6 +512,44 @@ func (qctx *QueueSchedulingContext) ClearJobSpecs() {
 	}
 }
 
+// TotalCostForQueueWithAllocation returns the cost for which this queue should be penalised when computing fairness,
+// if the total allocation of this queue is given by allocated.
```
zuqq, line 516: This is a minor point, but I think that talking about "penalization" here (in addition to cost) is more confusing than anything.

```suggestion
// TotalCostForQueueWithAllocation returns the total cost of the resources allocated to this queue.
```

internal/scheduler/context/context.go
```diff
@@ -490,6 +512,44 @@ func (qctx *QueueSchedulingContext) ClearJobSpecs() {
...
+func (qctx *QueueSchedulingContext) TotalCostForQueue() float64 {
+	return qctx.TotalCostForQueueWithAllocation(qctx.Allocated)
+}
+
+// TotalCostForQueueWithAllocation returns the cost for which this queue should be penalised when computing fairness,
+// if the total allocation of this queue is given by allocated.
```
zuqq, line 522: Same comment as above.

internal/armada/configuration/types.go
severinson: Once DRF is used everywhere, I want to remove asset fairness and leave DRF as the only option.

internal/armada/configuration/types.go
severinson: Done

internal/scheduler/queue_scheduler.go
```diff
@@ -348,7 +348,7 @@ func (it *CandidateGangIterator) fractionOfFairShareWithGctx(gctx *schedulercont
 	it.buffer.Zero()
 	it.buffer.Add(qctx.Allocated)
 	it.buffer.Add(gctx.TotalResourceRequests)
-	return qctx.FractionOfFairShareWithAllocation(it.buffer)
+	return qctx.TotalCostForQueueWithAllocation(it.buffer)
```
zuqq, line 351: The name of this method is now out of date, but this is obviously not a blocker.

zuqq (approved):

PR #2620, opened 2023-06-28
Pin docs python

kannon92:
Let's pin all dependencies for our client so we don't run into issues during CI.  

Dependabot will upgrade these over time but we will know what upgrades break us and don't outside of CI.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-409) by [Unito](https://www.unito.io)

### DIFF
--- a/client/python/pyproject.toml
+++ b/client/python/pyproject.toml
@@ -10,7 +10,7 @@ authors = [{ name = "G-Research Open Source Software", email = "armada@armadapro
 
 [project.optional-dependencies]
 format = ["black==23.3.0", "flake8==6.0.0", "pylint==2.17.4"]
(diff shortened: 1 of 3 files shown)


jayofdoom:
Not the path I'd choose, and worth noting it would not have prevented the earlier failures, but this is a valid implementation of pinning the dependencies if that's what the armada community wants to do.

jayofdoom (approved):
It's valid, even if I don't love the pattern of pinning deps.

jayofdoom (changes requested):
Just saw something: Please leave my comment in so we know why we're pinning a transitive dependency. Maybe adjust it for correctness.

jayofdoom (approved):

PR #2621, opened 2023-06-28
Avoid copying nodes where possible

zuqq:
Concretely, there are two situations in which we can avoid intermediate copies of nodes:

1. When constructing a `NodeDb`, we were making a separate call to `BindJobToNode` for each already-running job (which meant that we had to copy a `Node` for each already-running job); instead, we now first group the slice of already-running jobs according to the nodes that they're running on, and then only copy each node at most once.

   I added a benchmark for this case, which is much faster now (after b3ac7c46c8e1e1e3e8d7a418cb7990bec96b8e51, it runs in a few seconds on my machine; I don't know how long it took before that because I haven't had the patience to let it run for more than 15 minutes).
2. When evicting jobs at the start of the scheduling round, we were making a separate call to `EvictJobFromNode` for each preemptible job (which, again, meant that we had to copy a `Node` for each preemptible job); instead, we now evict all preemptible jobs on a given node in one go.

   I already attempted to make this change in #2590, but this PR comes with a cleaner API (no in-place node operations are exported anymore). As in the previous PR, this change makes a noticeable difference when running `go test ./internal/scheduler -bench='^BenchmarkPreemptingQueueScheduler$' -count=10 -run='^$'`:

   ```
   goos: darwin
   goarch: amd64
   cpu: Intel(R) Core(TM) i5-8210Y CPU @ 1.60GHz
                                                               │   old.txt    │                new.txt                │
                                                               │    sec/op    │    sec/op      vs base                │
   PreemptingQueueScheduler/100_nodes_1_queue_32000_jobs-4       541.0m ±  8%   414.8m ± 137%  -23.33% (p=0.023 n=10)
   PreemptingQueueScheduler/100_nodes_10_queues_32000_jobs-4     802.0m ± 17%   529.3m ±   9%  -34.00% (p=0.000 n=10)
   PreemptingQueueScheduler/1000_nodes_1_queue_320000_jobs-4      7.651 ± 35%    5.123 ±  12%  -33.05% (p=0.001 n=10)
   PreemptingQueueScheduler/1000_nodes_10_queues_320000_jobs-4    5.572 ± 47%    4.132 ±  25%  -25.85% (p=0.001 n=10)
   PreemptingQueueScheduler/1_node_1_queue_320_jobs-4            4.762m ±  6%   3.525m ±   4%  -25.99% (p=0.000 n=10)
   PreemptingQueueScheduler/1_node_10_queues_320_jobs-4          5.792m ±  6%   4.256m ±   5%  -26.51% (p=0.000 n=10)
   PreemptingQueueScheduler/10_nodes_1_queue_3200_jobs-4         49.45m ±  2%   38.04m ±   3%  -23.08% (p=0.000 n=10)
   PreemptingQueueScheduler/10_nodes_10_queues_3200_jobs-4       62.23m ±  6%   46.07m ±   2%  -25.97% (p=0.000 n=10)
   geomean                                                       188.1m         136.7m         -27.32%
   ```

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-410) by [Unito](https://www.unito.io)


internal/armada/server/lease.go
```diff
@@ -378,24 +378,8 @@ func (q *AggregatedQueueServer) getJobs(ctx context.Context, req *api.StreamingL
...
-					nodeInfo.GetName(), req.GetClusterId(), job.Id,
-				)
-				skipNode = true
-				break
-			}
-		}
-		if skipNode {
-			continue
+		if node, err = nodedb.BindJobsToNode(q.schedulingConfig.Preemption.PriorityClasses, jobs, node); err != nil {
+			return nil, err
```
zuqq, line 382: This behavior is now the same as in the new scheduler.

internal/scheduler/nodedb/nodedb.go
```diff
@@ -523,60 +541,89 @@ func BindJobToNode(priorityClasses map[string]configuration.PriorityClass, job i
...
 	if _, ok := node.EvictedJobRunIds[jobId]; ok {
-		// TODO: We're using run ids instead of job ids for now.
-		return nil, errors.Errorf("job %s is already evicted from node %s", jobId, node.Id)
-	} else {
-		node.EvictedJobRunIds[jobId] = true
+		return errors.Errorf("job %s is already evicted from node %s", jobId, node.Id)
 	}
+	node.EvictedJobRunIds[jobId] = true
 
+	if _, ok := node.AllocatableByPriorityAndResource[evictedPriority]; !ok {
```
zuqq, line 599: We can get rid of this check once we stop storing `Node` values in the node database.

internal/scheduler/scheduling_algo_test.go
```diff
@@ -677,3 +680,35 @@ func TestLegacySchedulingAlgo_TestSchedule_ExecutorOrdering(t *testing.T) {
...
+		numNodes := int(math.Pow10(e))
+		b.Run(fmt.Sprintf("%d nodes", numNodes), func(b *testing.B) {
+			jobs := testfixtures.N1CpuJobs("queue-alice", testfixtures.PriorityClass0, 32*numNodes)
+			nodes := testfixtures.N32CpuNodes(numNodes, testfixtures.TestPriorities)
+			for i, node := range nodes {
+				for j := 32 * i; j < 32*(i+1); j++ {
+					jobs[j] = jobs[j].WithNewRun("executor-01", node.Name)
+				}
+			}
+			rand.Shuffle(len(jobs), func(i, j int) { jobs[i], jobs[j] = jobs[j], jobs[i] })
```
zuqq, line 695: I'm shuffling this so that grouping `jobs` by node (which we didn't have to do before) is slightly harder; I haven't actually checked if it makes a big difference.

severinson (approved):

internal/scheduler/nodedb/nodedb.go
```diff
@@ -523,60 +541,89 @@ func BindJobToNode(priorityClasses map[string]configuration.PriorityClass, job i
...
-	queue := job.GetQueue()
-	requests := job.GetResourceRequirements().Requests
-
+// Specifically:
+//
+//   - The jobs that jobFilter returns true for are marked as evicted on the node.
+//   - Within AllocatableByPriorityAndResource, the resources allocated to these jobs are moved from
+//     the jobs' priorities to evictedPriority; they are not subtracted from AllocatedByJobId and
+//     AllocatedByQueue.
+func EvictJobsFromNode(
```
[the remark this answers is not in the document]
zuqq: That's a good idea; will follow up.

internal/scheduler/nodedb/nodedb.go
```diff
@@ -523,60 +541,89 @@ func BindJobToNode(priorityClasses map[string]configuration.PriorityClass, job i
...
-	// Ensure we track allocated resources at evictedPriority.
-	if _, ok := node.AllocatableByPriorityAndResource[evictedPriority]; !ok {
-		pMin := int32(math.MaxInt32)
-		ok := false
-		for p := range node.AllocatableByPriorityAndResource {
-			if p < pMin {
-				pMin = p
-				ok = true
-			}
+	if jobFilter == nil {
```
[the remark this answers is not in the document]
zuqq: Thanks for catching this, I didn't actually mean to change the semantics. I think that I was looking at this from the perspective of "we never pass in nil here, so this is just a defensive check", but I agree that nil can actually make sense here.

severinson (approved):

PR #2623, opened 2023-06-29
Feature - Create more detailed error responses when submitting batch jobs

### LINKED ISSUE #1748: Better Error Message When Rejecting Jobs

Raajheer1:
#### What type of PR is this?
This PR adds functionality to JobSubmitResponse when an error has occured. Previously the only thing returned was the error. The changes made in this PR allow the code to do 2 things:
- Build a collection of errors, if mulitple jobs error for the same return, these changes will return them at the same time to the user. Rather than them having to reattempt submissions over and over again with 1 fix each time.
- Return more detailed errors to the user by utilizing the JobSubmitResponse Struct

Fixed #1748 

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-412) by [Unito](https://www.unito.io)


internal/armada/server/lease.go
```diff
@@ -913,11 +913,11 @@ func (q *AggregatedQueueServer) addAvoidNodeAffinity(
 		}
 
 		changed := addAvoidNodeAffinity(jobs[0], labels, func(jobsToValidate []*api.Job) error {
-			if ok, err := validateJobsCanBeScheduled(jobsToValidate, allClusterSchedulingInfo); !ok {
+			if ok, responseItems, err := validateJobsCanBeScheduled(jobsToValidate, allClusterSchedulingInfo); !ok {
 				if err != nil {
-					return errors.WithMessage(err, "can't schedule at least 1 job")
+					return errors.WithMessagef(err, "can't schedule %d job", len(responseItems))
```
richscott, line 918: I'd change `job` to `job(s)`.

richscott (commented):
I think it looks good so far! I had just one suggestion re an error message spelling. Also, I think you'll want to add some tests, of course.

richscott (dismissed):

Raajheer1:
Armada Before though armadactl
Armada-Before

Armada After through armadactl
Armada-After

Submitted a bad sleep yaml job containing two jobs, both have invalid podSpecs as the requested and limit resources both contain "memory: 0"

richscott (approved):

richscott:
This changes in this PR were merged into master a couple weeks ago, via a separate PR - closing this.

PR #2624, opened 2023-06-30
SQL Databases Interfaces

### LINKED ISSUE #2121: Build interfaces around our postgres access
We should make sure that users could bring their own SQL databases into Armada.  Currently we only allow postgres as our db but we should create an interface and potentially allow other databases.

This could help with test improvement as we can add mocks for SQL calls and allow us to run unit tests without an active database.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-110) by [Unito](https://www.unito.io)

em-r:
#### What type of PR is this?
Design Proposal.

This PR represents a proposal for defining database-agnostic interfaces that will assist in removing the tight coupling between Armada and the Postgres client `pgx`.

Fixes #2121

This PR can be considered an in-progress PR, the proposed interfaces were verified against the existing Postgres implementation on `Lookout`, `LookoutV2`, and `Scheduler`. New changes might be introduced after completing the investigation of the remaining components (e.g. JobService)

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-413) by [Unito](https://www.unito.io)


docs/design/database_interfaces.md
```diff
@@ -0,0 +1,186 @@
...
+
+1. Introduce base common database interfaces that can be shared reused by all components (Lookout, Scheduler, Scheduler Ingester).
+2. Add interfaces that abstracts the hardcoded Postgres configuration.
+3. Add interfaces around `pgx` structs.
+
+### Functional Specification (API Description)
+
+#### Database Connection
+
+Most of the components (Lookout, Scheduler, Scheduler Ingester) rely on [PostgresConfig](github.com/armadaproject/armada/internal/armada/configuration) to connect to external databases, we can avoid hardcoding the configuration of those components to use `PostgresConfig` but defining a generic `DatabaseConfig` interface that's when implemented will provide those components with the necessary details to connect to databases.
```
satyampsoni, line 17: The link produces an error 404. 
```suggestion
Most of the components (Lookout, Scheduler, Scheduler Ingester) rely on [PostgresConfig](https://github.com/armadaproject/armada/tree/master/internal/armada/configuration) to connect to external databases, we can avoid hardcoding the configuration of those components to use `PostgresConfig` but defining a generic `DatabaseConfig` interface that's when implemented will provide those components with the necessary details to connect to databases.
```

satyampsoni (changes requested):

docs/design/database_interfaces.md
em-r: fixed, thanks!

suprjinx (approved):

kannon92 (approved):
LGTM

PR #2642, opened 2023-07-05
Unified scheduling across clusters

severinson:
This PR adds a new scheduler flag, which if enabled causes the Pulsar-backed scheduler to schedule across all executors in the same pool in a unified manner. Specifically, all nodes managed by executors in the same pool are added to the same NodeDb before scheduling.

The tests have been reworked quite a bit to support this.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-429) by [Unito](https://www.unito.io)


internal/common/slices/slices.go
```diff
@@ -108,6 +118,14 @@ func MapAndGroupByFuncs[S ~[]E, E any, K comparable, V any](s S, keyFunc func(E)
 	return rv
 }
 
+// Pop removes the last item from s and returns it.
+// Calling pop on an empty slice causes a panic.
```
JamesMurkin, line 122: Is this function used?

Mildly concerned about the panic on pop when empty

JamesMurkin: It is used, sorry github was hiding a files content

internal/scheduler/scheduling_algo.go
```diff
@@ -62,66 +62,81 @@ func NewFairSchedulingAlgo(
...
 
-		log.Infof("scheduling on %s", executor.Id)
-		schedulerResult, sctx, err := l.scheduleOnExecutor(
-			timeout,
-			accounting,
-			txn,
-			executor,
-			jobDb,
+		// Schedule across the executors in this group.
+		// Assume pool and minimumJobSize are consistent within the group.
```
JamesMurkin, line 124: Is this a good assumption? Definitely it historically hasn't been true all clusters in a pool have the same minimumJobSize

I think its fine, we may just need to adjust our usage

severinson: I suppose the question is what the definition of a pool is. My intuition is that clusters in the same pool should be the same and that we should use other mechanisms (e.g., taints/tolerations) for more finer-grained control within clusters.

JamesMurkin (approved):

PR #2655, opened 2023-07-10
Implement tls into our grpc servers

JamesMurkin:
This PR makes it so our grpc endpoints can handle TLS connections
 - Previously we only handle plain connections and let the kubernetes ingress terminate the tls before sending it to our applications


internal/common/certs/cached_certificate.go
```diff
@@ -0,0 +1,120 @@
...
+
+	if c.certFileInfo == nil || updatedCertFileInfo.ModTime().After(c.certFileInfo.ModTime()) {
+		modified = true
+	}
+
+	if c.keyFileInfo == nil || updatedKeyFileInfo.ModTime().After(c.keyFileInfo.ModTime()) {
+		modified = true
+	}
+
+	if modified {
```
d80tb7, line 91: I'd probably log here that we are reloading.

d80tb7 (approved):

d80tb7 (approved):

PR #2659, opened 2023-07-11
SQL Database Interfaces Definition

em-r:
#### What type of PR is this?
This PR introduces the definition of interfaces proposed in this design https://github.com/armadaproject/armada/blob/master/docs/design/database_interfaces.md

Those interfaces are meant to establish an abstraction over the existing Postgres implementation which will eventually lead to removing the tight coupling between Armada and Postgres.

This is a follow PR to https://github.com/armadaproject/armada/pull/2624, and it's meant particularly to introduce the interfaces that were proposed in the design. Follow PRs will be raised regarding the implementation of the interfaces as well as integrating those implementations with Armada.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-445) by [Unito](https://www.unito.io)


internal/armada/configuration/types.go
```diff
@@ -88,6 +88,28 @@ type PulsarConfig struct {
 	BackoffTime time.Duration
 }
 
+// DatabaseConfig represents the configuration of the database connection.
+type DatabaseConfig interface {
```
kannon92, line 92: I don't think we allow functions in our yaml configurations so this may be a bit weird..  

Looking at our configuration we only allow things you can pass into raw yaml so not sure what this interface would be used for.

suprjinx: agree, can we just use the struct properties without the getter functions?

em-r: yes makes sense, a struct is more convenient, I'll update the PR accordingly. thanks!

kannon92 (approved):

kannon92:
@Mo-Fatah

internal/common/database/types/types.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Next moves the iterator to the next row in the result set, it returns false if the result set is exhausted, otherwise true.
+	Next() bool
+
+	// Err returns the error, if any, encountered during iteration over the result set.
+	Err() error
+
+	// Scan reads the values from the current row into dest values positionally. It returns an error if any occurred during the read operation.
+	Scan(dest ...any) error
+}
```
Mo-Fatah, line 118: I think we should add another interface that acts as a generic placeholder for both `DatabasePool` and `DatabaseConn` . In various parts of the armada, such as [here](https://github.com/armadaproject/armada/blob/f041857ec91427da7e133ec0c472c5dd9a4549bd/internal/common/database/migrations.go#L33C5-L33C5) , the `Querier` interface is a placeholder for both `pgxpool.Pool` `pgx.Conn`. It is a loose interface that requires only 3 functions https://pkg.go.dev/github.com/jackc/pgtype@v1.13.0/pgxtype#Querier

You can introduce a new similar one here for `DatabasePool` and `DatabaseConn` that requires maybe only `Exec` and `Query` .

Mo-Fatah (commented):
I think this looks great! I added a few comments that you can consider in another PR since this is already merged. Let me know what you thinks!

PR #2688, opened 2023-07-17
fixing goreleaserdockerNNNN directories filling up /tmp

### LINKED ISSUE #2685: Very large goreleaserdockerNNNN directories filling up /tmp
When doing Armada development, using the usual local developer build/run steps of `mage localdev full testsuite`, the Goreleaser tool generates several large temporary directories and does not clean them up. 

On my Linux system, a clean initial build and test (using the above commands), results in the following very large directories in `/tmp`:

```
$ du --summarize -BM /tmp/gore*
1762M   /tmp/goreleaserdocker3478908114
48M     /tmp/goreleaserdocker4155803608
253M    /tmp/goreleaserdocker4165426849
```

It’s not clear if there is a bug in `mage` or `goreleaser` that causes these remainders, but we have seen the root filesystem on two EC2 instances fill up when developers do repeated builds, as `/tmp` resides on the root fs.

┆Issue is synchronized with this [Jira Bug](https://gr-oss.atlassian.net/browse/BATCH-470) by [Unito](https://www.unito.io)

nitishchauhan0022:
Fixes #2685

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-473) by [Unito](https://www.unito.io)

### DIFF
--- a/tools.yaml
+++ b/tools.yaml
@@ -3,7 +3,7 @@
 tools:
 - github.com/go-swagger/go-swagger/cmd/swagger@v0.29.0
 - github.com/gordonklaus/ineffassign@v0.0.0-20210914165742-4cc7213b9bc8
-- github.com/goreleaser/goreleaser@v1.18.2
+- github.com/goreleaser/goreleaser@v1.19.3-0.20230717115619-dbcebd00ae61
 - github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway@v1.16.0
 - github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger@v1.16.0
 - github.com/jstemmer/go-junit-report@v1.0.0
--- a/tools.yaml
+++ b/tools.yaml
@@ -3,7 +3,7 @@
 tools:
 - github.com/go-swagger/go-swagger/cmd/swagger@v0.29.0
 - github.com/gordonklaus/ineffassign@v0.0.0-20210914165742-4cc7213b9bc8
-- github.com/goreleaser/goreleaser@v1.19.3-0.20230717115619-dbcebd00ae61
+- github.com/goreleaser/goreleaser@v1.20.0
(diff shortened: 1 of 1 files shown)


tools.yaml
```diff
@@ -3,7 +3,7 @@
 tools:
 - github.com/go-swagger/go-swagger/cmd/swagger@v0.29.0
 - github.com/gordonklaus/ineffassign@v0.0.0-20210914165742-4cc7213b9bc8
-- github.com/goreleaser/goreleaser@v1.18.2
+- github.com/goreleaser/goreleaser@v1.19.3-0.20230717115619-dbcebd00ae61
```
kannon92, line 6: I think we should probably wait for an official release for this.  Looks to be on 1.19.2 at the moment.

nitishchauhan0022: yes, for now it is the latest. 1.19.3 should work fine.
I will upgrade to the latest whenever it will available.

kannon92: I don't really understand where this package is coming from if its not official released yet.  So I think we should wait until 1.19.3 is official and then just use the official package.

nitishchauhan0022: To install from the main branch yesterday i did `go install github.com/goreleaser/goreleaser@main`
And than i checked the version of the goreleaser --version it was  to the last commit `v1.19.3-0.20230717115619-dbcebd00ae61`
so i added that version rather than pointing to the main branch because main can have bug, but upto that all was working fine for our use cases

richscott (commented):
Rather than using the tip of `main`, I'd very strongly prefer to wait and use the next official tagged release of Goreleaser, mostly to be sure they have tested all changes since 1.19.2.

nitishchauhan0022:
sure

kannon92 (approved):

richscott (approved):

PR #2690, opened 2023-07-18
feat: update release CI

dejanzele:
Fixes #


.github/workflows/release.yml
```diff
@@ -90,3 +90,10 @@ jobs:
           DOCKER_BUILDX_BUILDER: "${{ steps.buildx.outputs.name }}"
           DOCKER_BUILDX_CACHE_FROM: "type=gha"
           DOCKER_BUILDX_CACHE_TO: "type=gha,mode=max"
+  invoke-chart-push:
+    name: Invoke Chart push
+    needs: release
+    uses: armadaproject/charts/.github/workflows/invoke-push.yaml@master
```
jgiannuzzi, line 96: I think you mean G-Research/charts?

pavlovic-ivan, line 96: This does not exist currently. And i suspect you will use this [one](https://github.com/G-Research/charts/blob/master/.github/workflows/invoke-push.yaml). Am i right?

dejanzele: good catch

kannon92 (approved):

Sharpz7 (approved):

PR #2694, opened 2023-07-19
Upgrade pgx/v4 to pgx/v5

### LINKED ISSUE #2692: Upgrade pgx/v4 to pgx/v5 across the codebase
Most of the codebase uses `github.com/jackc/pgx/v4` as the Postgres driver, while some other parts (e.g. JobService) uses `"github.com/jackc/pgx/v5"` . For consistency and avoiding unnecessary dependencies, we should upgrade our database drivers to `v5` .

One thing I noticed that should be handled is the Interface change of the database handlers in ``v5`` . This led to incompatibility between the new handlers (e.g. `v5/pgxpool.Pool`, `v5/pgx.Conn` ) and the `pgxtype.Querier` interface used [here](https://github.com/armadaproject/armada/blob/f041857ec91427da7e133ec0c472c5dd9a4549bd/internal/common/database/migrations.go#L33C5-L33C5). Also, `pgx/v5` didn’t add a new interface for the database handlers, see: [https://github.com/jackc/pgx/issues/1188#issuecomment-1105867098](https://github.com/jackc/pgx/issues/1188#issuecomment-1105867098|smart-link)  

An easy fix is to introduce a new wrapper interface that can replace the deprecated interface.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-477) by [Unito](https://www.unito.io)

Mo-Fatah:
Fixes #2692
Notable changes:
- `"github.com/jackc/pgconn"` is replaced with `"github.com/jackc/pgx/v5/pgconn"`.
- `pgxpool.Connect` replaced with `pgxpool.New`. see: https://github.com/jackc/pgx/blob/master/CHANGELOG.md#pgxpool
-  instead of the method `db.BeginTxFunc` to perform a tx, `pgx/v5` provided a general function [pgx.BeginTxFunc](https://github.com/jackc/pgx/blob/master/tx.go#L401) 
- The old `pgxtype.Querier` interface is not supported in `pgx/v5`, so I introduced a new temp interface in `internal/common/database/querier.go` until the database interface work is complete.


richscott (dismissed):

kannon92 (dismissed):

go.mod
```diff
@@ -33,10 +33,9 @@ require (
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/hashicorp/golang-lru v0.5.4
 	github.com/instrumenta/kubeval v0.0.0-20190918223246-8d013ec9fc56
-	github.com/jackc/pgconn v1.13.0
 	github.com/jackc/pgerrcode v0.0.0-20201024163028-a0d42d470451
 	github.com/jackc/pgtype v1.13.0
-	github.com/jackc/pgx/v4 v4.17.2
+	github.com/jackc/pgx/v4 v4.17.2 // indirect
```
d80tb7, line 38: do we know what's bringing in a dependency on v4?

d80tb7 (changes requested):
Are you sure that your replacement code for `BeginTxFunc` is doing the right thing?

Previously `BeginTxFunc` took a function that returned an error.  If the returned error was null ``BeginTxFunc` ` would take care of committing the transaction for you, otherwise it would roll it back.  The replacement code doesn't look like it is doing this although maybe I am missing something?

go.mod
richscott: I took a quick look at `go mod graph`, and it looks like `github.com/jackc/pgtype` is pulling it in - even in the latest tagged version (1.14) - see  https://github.com/jackc/pgtype/blob/master/go.mod

Mo-Fatah:
> Are you sure that your replacement code for `BeginTxFunc` is doing the right thing?
> 
> Previously `BeginTxFunc` took a function that returned an error. If the returned error was null `BeginTxFunc` would take care of committing the transaction for you, otherwise it would roll it back. The replacement code doesn't look like it is doing this although maybe I am missing something?
> 
> I _think_ you want to convert these to use [pgx.BeginTxFunc](https://github.com/jackc/pgx/blob/master/tx.go#L401)

Yes, you are right. I thought that they dropped `BeginTxFunc` but they just converted into a general function https://github.com/jackc/pgx/blob/4091eedf03d661ba905f1a8711f2eaa99c508edc/CHANGELOG.md?plain=1#L277-L278 .
I will switch to `pgx.BeginTxFunc`

internal/lookoutv2/repository/getjobrunerror.go
```diff
@@ -29,24 +29,16 @@ func NewSqlGetJobRunErrorRepository(db *pgxpool.Pool, decompressor compress.Deco
...
-		for rows.Next() {
-			err := rows.Scan(&rawBytes)
-			if err != nil {
-				return err
-			}
-			return nil
-		}
-		return errors.Errorf("no error found for run with id %s", runId)
+		return err
 	})
```
Mo-Fatah, line 42: I noticed that the code here, and other parts, used `tx.Query` just to query a single row and found it confusing. switched to `tx.QueryRow` instead.

internal/lookoutv2/repository/getjobrunerror.go
```diff
@@ -29,24 +29,16 @@ func NewSqlGetJobRunErrorRepository(db *pgxpool.Pool, decompressor compress.Deco
 
 func (r *SqlGetJobRunErrorRepository) GetJobRunError(ctx context.Context, runId string) (string, error) {
 	var rawBytes []byte
-	err := r.db.BeginTxFunc(ctx, pgx.TxOptions{
+	err := pgx.BeginTxFunc(ctx, r.db, pgx.TxOptions{
 		IsoLevel:       pgx.RepeatableRead,
 		AccessMode:     pgx.ReadOnly,
 		DeferrableMode: pgx.Deferrable,
 	}, func(tx pgx.Tx) error {
-		rows, err := tx.Query(ctx, "SELECT error FROM job_run WHERE run_id = $1 AND error IS NOT NULL", runId)
```
Mo-Fatah, line 37: I noticed that the code here, and other parts, used tx.Query just to query a single row and found it confusing. switched to tx.QueryRow instead.

go.mod
d80tb7: I assume that pgx/v5 works fine with the old pgxtype?  If so I'd leave it for now and see if we can migrate to pgx/v5/pgtype in a separate pr.

internal/lookoutv2/repository/getjobrunerror.go
d80tb7: looks good- as an aside (and I know you didn't write the original query here) I don't think we need a transaction function here.  We only execute a single sql statement so we can probably just do `r.db.QueryRows()`.  Maybe one for a subsequent PR.

d80tb7 (approved):

richscott (approved):

PR #2696, opened 2023-07-19
[armada-scheduler] Fix node selector being nil after JobSchedulingInfo update

JamesMurkin:
Currently we rely on nodeSelector and annotations never being nil
 - If they are, we can't evict the job

As the object is a proto object it is hard to ensure its never created with nil values

For now just make sure any updates to job.JobSchedulingInfo result in a valid configuration

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-481) by [Unito](https://www.unito.io)

### DIFF
--- a/internal/scheduler/jobdb/job.go
+++ b/internal/scheduler/jobdb/job.go
@@ -71,21 +71,7 @@ func NewJob(
 	cancelled bool,
 	created int64,
 ) *Job {
-	// Initialise the annotation and nodeSelector maps if nil.
-	// Since those need to be mutated in-place.
-	if schedulingInfo != nil {
-		for _, req := range schedulingInfo.ObjectRequirements {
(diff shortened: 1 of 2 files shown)


internal/scheduler/jobdb/job.go
```diff
@@ -422,6 +427,7 @@ func (job *Job) WithCreated(created int64) *Job {
 func (job *Job) WithJobSchedulingInfo(jobSchedulingInfo *schedulerobjects.JobSchedulingInfo) *Job {
 	j := copyJob(*job)
 	j.jobSchedulingInfo = jobSchedulingInfo
+	// j.ensureJobSchedulingInfoFieldsInitialised()
```
kannon92, line 430: Could you delete the comment?

d80tb7 (approved):

PR #2699, opened 2023-07-20
Set default executorTimeout in armada config

JamesMurkin:
This is currently defaulting to having an infinite timeout

Which if a cluster dies, any jobs on it when it died forever contribute to fair share calculations, which isn't an ideal default

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-484) by [Unito](https://www.unito.io)


kannon92 (dismissed):
Not ideal at all!

d80tb7 (dismissed):

d80tb7 (approved):

PR #2700, opened 2023-07-20
Tidy logging in pod_issuer_handler

JamesMurkin:
Primarily the aim here is to reduce logging, but also make it clearer from logs when an issue starts/ends

Currently when there are a lot of issues, we log megabytes a second - due to logging the full issue of every pod - every 5 seconds

 - Skip issue detection again if pod already has issue registered
   - During issue detection we log the full issue
 - Reduce how often we log the full issue
 - Explicitly log issue registered/resolved
 - Perform issue handling slightly less frequently

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-485) by [Unito](https://www.unito.io)

### DIFF
--- a/internal/executor/service/pod_issue_handler.go
+++ b/internal/executor/service/pod_issue_handler.go
@@ -125,14 +125,15 @@ func (p *IssueHandler) registerIssue(issue *runIssue) {
 
 	runId := issue.RunId
 	if runId == "" {
-		log.Debugf("Not registering an issue for job %s as run id was empty", issue.JobId)
+		log.Warnf("Not registering an issue for job %s as run id was empty", issue.JobId)
 		return
 	}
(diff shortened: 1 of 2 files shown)


internal/executor/service/pod_issue_handler.go
```diff
@@ -107,6 +107,18 @@ func NewIssueHandler(
 	return issueHandler
 }
 
+func (p *IssueHandler) hasIssue(runId string) bool {
+	p.podIssueMutex.Lock()
+	defer p.podIssueMutex.Unlock()
+
+	if runId == "" {
```
d80tb7, line 114: probably should comment here as to why this would happen

d80tb7 (approved):

Issue #2332, opened 2023-04-04
[Magefiles] Add Scheduler and Scheduler Ingester to Localdev


Sharpz7:
Make sure the new components can be accessed from the docker-compose used in localdev.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-11) by [Unito](https://www.unito.io)

Sharpz7:
Closed since @theAntiYeti's change to add pulsar-backed localdev.

PR #2719, opened 2023-07-21
Support pulsar backed scheduler with LocalDev

theAntiYeti:
This PR modifies the configuration and magefile to make the pulsar backed scheduler the default when using LocalDev and also to support it in docker compose to begin with.

It adds a new command `mage LocalDev minimal-pulsar` which packages and builds the new scheduler components and postgres migration as a standard.

Closes https://github.com/armadaproject/armada/issues/2332


Sharpz7:
Can I check, how many parts of that config are new? Ideally, we don't want to have too much extra config.

I added to methods to localdev to add extra config without having to repeat yourself.

See https://github.com/armadaproject/armada/tree/master/developer/config and https://github.com/armadaproject/armada/tree/master/developer/env. I think by using these we can keep the codebase cleaner :))

theAntiYeti:
Does `--config` do an ad-hoc overwrite of configuration values?

Sharpz7:
I think so: except it only does it at the first level. I.e entire sections past the first level will be replaced if it is different.

In those situations, that's when the env vars are more useful (see the 2nd link).

PR #2729, opened 2023-07-23
fix: yarn.go to return yarn instead of yarn.exe

### LINKED ISSUE #2728: ```mage ui``` not working on Windows OS
**Describe the bug**
Currently, ```mage ui``` command is not working on Windows OS, and giving the following error

image

- git clone the armada project
- install necessary dependencies
- run ```mage ui```

**Expected behavior**
The command was expected to install yarn packages for the ```internal/lookout/``` directory

**Actual behavior**
It is giving an error that the executable is not found in environment variables, though there exists a yarn bin inside the system environment variables.

**Why is the error occurring?**
We are using the ```binaryWithExt``` function to get the ```.exe``` version of ```yarn```, i.e., ```yarn.exe```, which is not the case for Windows OS, since yarn is a binary which can be directly executed with ```yarn``` command, without the ```.exe``` extension.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-514) by [Unito](https://www.unito.io)

anshgoyalevil:
#### What type of PR is this?
This PR modifies the ```yarn.go``` file to return ```yarn``` instead of ```yarn.exe``` for Windows OS.

Fixes #2728 

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-515) by [Unito](https://www.unito.io)


magefiles/yarn.go
```diff
@@ -8,7 +8,7 @@ import (
 
 // Create golang code to build the UI
 func yarnBinary() string {
-	return binaryWithExt("yarn")
+	return "yarn"
```
kannon92, line 11: I think this change is incorrect.  Does this work on native windows?

kannon92 (approved):
Reading about `yarn` on windows this seems to be correct.

PR #2730, opened 2023-07-23
add `kubectl-armadactl` to goreleaser.yml

ShivangShandilya:
This PR builds another archive named `kubectl-armadactl` in our Armada project which will be later used to integrate `armadactl` with `kubectl`.

The contents of the new archive are exactly same as that of `armadactl`.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-516) by [Unito](https://www.unito.io)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -169,15 +169,12 @@ archives:
   - id: kubectl-armadactl
     builds:
       - armadactl
+    binary: kubectl-armadactl
     allow_different_binary_count: true
     name_template: 'kubectl-armadactl_{{ replace .Version "-" "_" }}_{{ .Os }}_{{ .Arch }}'
     format_overrides:
       - goos: windows
         format: zip
-    rename:
-      # Here, we specify the desired filename for the new archive.
-      - binary: armadactl
-        to: kubectl-armadactl
     files:
       - LICENSE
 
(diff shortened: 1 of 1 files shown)


kannon92 (commented):
Change looks good to me but I think there is a syntax error:

```
Error: yaml: unmarshal errors:
  line 177: field rename not found in type config.Archive
```

ShivangShandilya:
> Change looks good to me but I think there is a syntax error:
> 
> ```
> Error: yaml: unmarshal errors:
>   line 177: field rename not found in type config.Archive
> ```

I thought so this might give out error, is there any other way we can change binary name cause if we don't do this a new archive will be created `kubectl-armadactl` but will have binary with the name `armadactl` which we don't want.

.goreleaser.yml
```diff
@@ -166,6 +166,21 @@ archives:
...
 
+  - id: kubectl-armadactl
+    builds:
+      - armadactl
+    allow_different_binary_count: true
+    name_template: 'kubectl-armadactl_{{ replace .Version "-" "_" }}_{{ .Os }}_{{ .Arch }}'
+    format_overrides:
+      - goos: windows
+        format: zip
+    rename:
```
kannon92, line 177: Let's just try removing this.

ShivangShandilya: Done 😄

PR #2734, opened 2023-07-23
fix: add windows specific script for mage ui

### LINKED ISSUE #2733: ```mage-ui``` is not using Window's specific arguments for ```docker run```
**Describe the bug**
Currently, ```mage ui``` command is not working on Windows OS, and giving the following error

image

- git clone the armada project
- install necessary dependencies
- run ```mage ui```

**Expected behavior**
The command was expected to run the container for Lookout UI

**Actual behavior**
It is giving the error as described in the screenshot above.

**Why is the error occurring?**
We are using the command ```docker run --rm -u $(id -u ${USER}):$(id -g ${USER}) -v "${PWD}/../../../:/project" openapitools/openapi-generator-cli:v5.4.0 /project/internal/lookout/ui/openapi.sh``` for window os but window doesn't support variables like ```USER```, ```-g```, etc, so it is giving an error.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-519) by [Unito](https://www.unito.io)

anshgoyalevil:
#### What type of PR is this?
This PR fixes the yarn.go file that was failing due to the incompatibility between Windows and Linux-specific arguments which were passed to the docker run command in package.json of Lookout UI.

Fixes #2733 

A new go function to check the OS is added, though it can be used directly inside the yarn.go file if necessary.

#### Is the problem resolved?

Yes.

image


magefiles/utils.go
```diff
@@ -17,6 +17,13 @@ func binaryWithExt(name string) string {
 	return name
 }
 
+func checkOs() string {
+	if runtime.GOOS == "windows" {
```
kannon92, line 21: Why can't you just return `runtime.GOOS` for windows?

anshgoyalevil: Thanks for pointing that out. That ```if``` condition was un-necessary.

magefiles/utils.go
```diff
@@ -17,6 +17,10 @@ func binaryWithExt(name string) string {
 	return name
 }
 
+func checkOs() string {
```
Sharpz7, line 20: Hmmm. I would want to see this merged into one function with https://github.com/armadaproject/armada/blob/042d12ea31a20da8bdda9eba94372fb57d41d2bb/magefiles/utils.go#L50C1-L53C2.

Or, at the very least, use the same naming scheme for the function

anshgoyalevil: Would this be good to go?

```go
func onWindows() bool {
	return runtime.GOOS == "windows"
}
```

anshgoyalevil:
@Sharpz7 I have made the changes. Can you please re-review them? Thanks 😄

kannon92 (approved):

Sharpz7 (approved):

PR #2752, opened 2023-07-25
alpine to alpine 3

kannon92:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-538) by [Unito](https://www.unito.io)


kannon92:
@Sharpz7 I tried to resolve the Synk notifications but they keep coming up.  I think you could change this PR to be alpine:3 if the Synk stuff is to be trusted.

d80tb7 (changes requested):
Are you sure we want this.  This would mean we had non-repeatable builds, which sounds like it would be an issue.

kannon92:
> Are you sure we want this. This would mean we had non-repeatable builds, which sounds like it would be an issue.

Technically we would always have this problem as we don't have patch versions in our containers ie `alpine:3.1`.  

I mostly added this to silent the Synk security warnings that were suggesting that we use `alpine:3` so we keep up to date and not have to keep updating the alpine version.  

There is an issue with this PR so I will have to fix that before we merge this if you think it is a good idea.

PR #2754, opened 2023-07-25
Support pulsar backed scheduler with LocalDev

theAntiYeti:
Fixes #

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-540) by [Unito](https://www.unito.io)


magefiles/developer.go
```diff
@@ -10,7 +10,7 @@ import (
 
 var services = []string{"pulsar", "redis", "postgres"}
 
-var componentsStr string = "server,lookout,lookoutingester,lookoutv2,lookoutingesterv2,executor,binoculars,eventingester,jobservice"
```
JamesMurkin, line 13: Why isn't `postgresPulsarMigration` needed in this list?

magefiles/main.go
```diff
@@ -147,6 +150,9 @@ func LocalDev(arg string) error {
 	case "minimal":
 		os.Setenv("ARMADA_COMPONENTS", "executor,server")
 		mg.Deps(StartComponents)
+	case "minimal-pulsar":
+		os.Setenv("ARMADA_COMPONENTS", "executor-pulsar,server-pulsar,postgresPulsarMigration,scheduler,scheduleringester")
```
JamesMurkin, line 154: Why isn't `eventingester` in this list? Surely you'd need it started to get events

magefiles/developer.go
```diff
@@ -10,7 +10,7 @@ import (
...
+var componentsStr string = "server,lookout,lookoutingester,lookoutv2,lookoutingesterv2,executor,binoculars,eventingester,jobservice,scheduler,scheduleringester"
```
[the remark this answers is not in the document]
theAntiYeti: I'm going to remove that actually. This variable represents the state of the world when you run "full". As currently we want "full" to represent the legacy scheduler I think it's worth just leaving them out for now.

magefiles/main.go
theAntiYeti: Because the scheduler depends on the event ingester. Technically you don't need anything in that list if it's depended on in the docker compose file.

PR #2759, opened 2023-07-27
[armada-scheduler] Add separate config for maxJobsLeasedPerCall

JamesMurkin:
Currently we use the same config for how many job we scheduler per round and how many jobs we lease per call

This was fine historically, however now we schedule jobs per pool (not per executor) we'd like to increaes the number of jobs we scheduler per round
 - Which will have the unintended impact of increasing how many jobs we send per lease call

Adding separate config, so we can tune these 2 things separately

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-545) by [Unito](https://www.unito.io)

### DIFF
--- a/internal/scheduler/schedulerapp.go
+++ b/internal/scheduler/schedulerapp.go
@@ -138,7 +138,7 @@ func Run(config schedulerconfig.Configuration) error {
 		executorRepository,
 		legacyExecutorRepository,
 		allowedPcs,
-		config.Scheduling.MaximumJobsToSchedule,
+		config.MaxJobsLeasedPerCall,
 		config.Scheduling.Preemption.NodeIdLabel,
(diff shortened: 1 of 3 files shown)


internal/scheduler/schedulerapp.go
```diff
@@ -138,7 +138,7 @@ func Run(config schedulerconfig.Configuration) error {
 		executorRepository,
 		legacyExecutorRepository,
 		allowedPcs,
-		config.Scheduling.MaximumJobsToSchedule,
+		config.MaxJobsLeasedPerCall,
```
d80tb7, line 141: I not that we've got *Maximum*JobsToSchedule but *Max*JobsLeasedPerCall- we should probably be consistent.

d80tb7 (approved):

PR #2602, opened 2023-06-22
feat: add new test workflow

dejanzele:
Fixes #

#### Special notes for your reviewer:

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-391) by [Unito](https://www.unito.io)

PR #2771, opened 2023-07-28
mage test update and go code coverage

### LINKED ISSUE #2765: CodeCov no longer shows our go coverage
**Describe the bug**

CodeCov shows that our coverage is now only 16%.  Looking at the report, it is only showing our Typescript unit tests and some clients.  

I traced it to https://github.com/armadaproject/armada/pull/2602.  From then on, our coverage dropped from 58% to 16%.  

https://github.com/armadaproject/armada/pull/2602

**Expected behavior**

We should have go coverage in our repo.

**Actual behavior**

**Additional context**

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-551) by [Unito](https://www.unito.io)

nitishchauhan0022:
Fixes #2765

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-557) by [Unito](https://www.unito.io)


nitishchauhan0022:
hey @kannon92 I think the issue is in https://github.com/armadaproject/armada/blob/1538fefbdbded3b6ee72d6730324c4684eea80b1/makefile#L405
gotestsum `$(go list ./internal/... | grep -v 'jobservice/repository')` is not considering the output as files causing these tests to never run resulting in missing of internal package tests from the report, finally reducing the go code coverage to 16%.

kannon92:
> hey @kannon92 I think the issue is in
> 
> https://github.com/armadaproject/armada/blob/1538fefbdbded3b6ee72d6730324c4684eea80b1/makefile#L405
> 
> 
> gotestsum `$(go list ./internal/... | grep -v 'jobservice/repository')` is not considering the output as files causing these tests to never run resulting in missing of internal package tests from the report, finally reducing the go code coverage to 16%.

This actually worked before.  I think the issue is that we switched how we run the tests in the PR i linked above.  I don't think we even use this target anymore since we switched to mage and I believe the tests are run different in the CI.

nitishchauhan0022:
we haven't switched to mage for test. It is added now.

nitishchauhan0022:
> > hey @kannon92 I think the issue is in
> > https://github.com/armadaproject/armada/blob/1538fefbdbded3b6ee72d6730324c4684eea80b1/makefile#L405
> > 
> > gotestsum `$(go list ./internal/... | grep -v 'jobservice/repository')` is not considering the output as files causing these tests to never run resulting in missing of internal package tests from the report, finally reducing the go code coverage to 16%.
> 
> This actually worked before. I think the issue is that we switched how we run the tests in the PR i linked above. I don't think we even use this target anymore since we switched to mage and I believe the tests are run different in the CI.

Previously the targets were different https://github.com/armadaproject/armada/pull/2602/files#diff-beda42571c095172ab63437d050612a571d0d9ddd3ad4f2aecbce907a9b7e3d0L407

kannon92:
The comment is making me think you got this working!  Just need to fix the conflicts and I'll do a review!

Sharpz7:
Hey, I have a feeling my PR https://github.com/armadaproject/armada/pull/2770 will cause you problems here.

Let me know if you need any help.

nitishchauhan0022:
hey @Sharpz7, i have updated the merging to include the test target, and kept the e2e tests target removed as you shifted them from make code to use localdev setup #2770. I am not having any problems for now, will definitely ask if needed. Thankyou.

magefiles/tests.go
```diff
@@ -40,6 +42,70 @@ func gotestsum() error {
 	return nil
 }
 
+// Tests is a mage target that runs the tests and generates coverage reports.
```
Sharpz7, line 45: I still do not think this is needed. We should, be able to use:

`mage localdev minimal testsuite` here and get the result we want.

nitishchauhan0022: Hey, [mage localdev no-build](https://github.com/armadaproject/armada/blob/b498d64ac52a98facee494eaa5720e7f5af5b4ff/magefiles/main.go#L180) also does the kind setup thing, which is not required as per the instructions in [make](https://github.com/armadaproject/armada/blob/b498d64ac52a98facee494eaa5720e7f5af5b4ff/makefile#L399) to run these tests, that's why I have not used that.

nitishchauhan0022:
@Sharpz7 Also, one more thing which I am not clear is that how the test suite part works. Like it is equal to those tests which are previously done in make at various place like at target  [tests-e2e ](https://github.com/armadaproject/armada/blob/b498d64ac52a98facee494eaa5720e7f5af5b4ff/makefile#L527C1-L527C10). And now mage localdev minimal testsuite. The test suite looks like contains many things.

Sharpz7:
Testsuite = e2e tests

Sharpz7 (approved):

PR #2775, opened 2023-07-28
fixing cert tests

nitishchauhan0022:
#### What type of PR is this?
In the github actions, where we are doing the unit tests by the command [make tests](https://github.com/armadaproject/armada/blob/5c5081a907bd0a07e040a3301b2d9dc494cf4e6c/.github/workflows/test.yml#L77), the target is not executing as we are expecting it to.
This commad: https://github.com/armadaproject/armada/blob/5c5081a907bd0a07e040a3301b2d9dc494cf4e6c/makefile#L405
Output this:
`/home/nitish/cncf/Opensource/armada/bin/gotestsum --format short-verbose --junitfile test-reports/unit-tests.xml --jsonfile test-reports/unit-tests.json -- -coverprofile=test-reports/coverage.out -covermode=atomic ./cmd/... ./pkg/...`

Making the tests runs only for cmd and pkg.
Make is trying to replace it as a variable.

The command should be:
`$(GOTESTSUM) --format short-verbose --junitfile test-reports/unit-tests.xml --jsonfile test-reports/unit-tests.json -- -coverprofile=test-reports/coverage.out -covermode=atomic ./cmd/... ./pkg/... $$(go list ./internal/... | grep -v 'jobservice/repository')
` having $$ instead of $

If you replace it as this, the internal tests will run as well which were not previously and some of them failing, can be looked easily locally, while running the tests. 

Certs tests which is updated in this pr is one of them.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-561) by [Unito](https://www.unito.io)


nitishchauhan0022:
Cert tests failing in this pr #2771

kannon92:
Can you explain a bit more about this?  Why did these tests start failing in that PR?  Please write up a little description for the reader so they can understand why this is needed.

nitishchauhan0022:
> Can you explain a bit more about this? Why did these tests start failing in that PR? Please write up a little description for the reader so they can understand why this is needed.

It is because when I updated from the make target to mage target in the PR #2771.  And the mage tests target is running those tests which were not previously, that's why they are failing there.

internal/common/certs/cached_certificate_test.go
```diff
@@ -105,6 +105,8 @@ func TestCachedCertificateService_ReloadsCertPeriodically_WhenUsingRun(t *testin
 }
 
 func writeCerts(t *testing.T, certData *bytes.Buffer, keyData *bytes.Buffer) {
+	err := os.MkdirAll("testdata", 0o755)
```
kannon92, line 108: Nit: can you make testdata a constant rather than a hardcoded string?

kannon92 (dismissed):
Lgtm just one nit around using a constant variable

kannon92 (approved):

Sharpz7 (approved):

JamesMurkin (approved):

PR #2778, opened 2023-07-28
Jobservice performance testing tool

ClifHouck:
Needs a lot of user interface type work, but the functionality is generally there. Right now jobservice appears to be *very* CPU efficient. Trying to ID IO bottlenecks (likey DB or gRPC related).

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-564) by [Unito](https://www.unito.io)


richscott (approved):
Looks good to me. I too would like to see some tunable settings via command-line options, but that can be another PR.

PR #2811, opened 2023-07-31
Sql interfaces jobservice integration

em-r:
#### What type of PR is this?
Integration.

This PR introduces implementations of existing SQL interfaces for Postgres databases and integrates the new interfaces with JobService.

Fixes #

The PR is still missing tests which will be added later on as well, but includes enough changes for reviewers to review.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-597) by [Unito](https://www.unito.io)


config/jobservice/config.yaml
```diff
@@ -9,6 +9,7 @@ databaseType: "postgres"
 # storage file when using database type 'sqlite'
 databasePath: "/var/jobservice.db"
 # Connection details when using database type 'postgres'
+# to be replaced with databaseConfig
```
suprjinx, line 12: I think we should try to be backwards compatible with this config key -- there are a lot configuration repos out there that I wouldn't want to track down!

internal/common/database/types/types.go
```diff
@@ -81,12 +80,15 @@ type DatabasePool interface {
 	Close()
 
 	// Exec executes a query that doesn't return rows. It returns any error encountered.
-	Exec(context.Context, string, ...any) (any, error)
+	Exec(context.Context, string, ...any) (DatabaseQueryResult, error)
```
suprjinx, line 83: nice improvement over `any` :-)

config/jobservice/config.yaml
em-r: I agree, and instead of immediately switching to a new field and removing the existing `postgresConfig` field, we can have them both temporarily. if a user sets `postgresConfig` we can then display a warning message saying that this field is getting deprecated in future releases and that the new field that's meant to provide database configuration to job service is `databaseConfig`. 
this logic is now available in my latest commit ^^

Mo-Fatah:
@em-r Hello :wave: 
Just checking with you about any progress here. We can help if something blocking you.

richscott:
Closing due to inactivity.

PR #2825, opened 2023-08-01
Fix jobservice unit tests

ClifHouck:
┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-611) by [Unito](https://www.unito.io)

### DIFF
--- a/internal/jobservice/eventstojobs/manage_subs_test.go
+++ b/internal/jobservice/eventstojobs/manage_subs_test.go
@@ -36,6 +36,9 @@ func (m *MockEventClient) Recv() (*api.EventStreamMessage, error) {
 	m.messagesSent += 1
 
 	if m.messagesSent > 3 {
+		// Sleep a bit to mimick the stream being open but no events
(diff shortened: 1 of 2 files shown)


richscott (approved):

internal/jobservice/eventstojobs/manage_subs_test.go
```diff
@@ -36,6 +36,9 @@ func (m *MockEventClient) Recv() (*api.EventStreamMessage, error) {
 	m.messagesSent += 1
 
 	if m.messagesSent > 3 {
+		// Sleep a bit to mimick the stream being open but no events
+		// currently available.
+		time.Sleep(time.Second * 5)
```
kannon92, line 41: Maybe we should put this under a little load and verify that this won't be flaky?

ClifHouck: Can we cross that bridge if it happens? Right now they don't pass at all.

PR #2834, opened 2023-08-04
Add Scheduler Metrics

theAntiYeti:
Fixes #

### DIFF
--- a/internal/scheduler/scheduler_metrics.go
+++ b/internal/scheduler/scheduler_metrics.go
@@ -1,0 +1,48 @@
+package scheduler
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const NAMESPACE = "armada"
+const SUBSYSTEM = "scheduler"
+
+type SchedulerMetrics struct {
+	CycleTime          prometheus.Gauge
+	CycleTimeHistogram prometheus.Histogram
+}
+
+func NewSchedulerMetrics() *SchedulerMetrics {
+	cycleTime := prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace:   NAMESPACE,
+			Subsystem:   SUBSYSTEM,
+			Name:        "last_cycle_time",
+			Help:        "Time to complete most recent scheduling cycle.",
+			ConstLabels: map[string]string{},
+		})
+
+	cycleTimeHistogram := prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Namespace: NAMESPACE,
+			Subsystem: SUBSYSTEM,
+			Name:      "cycle_times",
+			Help:      "Histogram of cycle times observed.",
+			Buckets:   []float64{0, 1, 2, 3, 4, 5},
+		},
+	)
+
+	prometheus.MustRegister(cycleTime)
+	prometheus.MustRegister(cycleTimeHistogram)
+
+	return &SchedulerMetrics{
+		CycleTime:          cycleTime,
+		CycleTimeHistogram: cycleTimeHistogram,
+	}
+
+}
+
+func (metrics *SchedulerMetrics) ReportCycleTime(cycleTime float64) {
+	metrics.CycleTime.Add(cycleTime)
+	metrics.CycleTimeHistogram.Observe(cycleTime)
+}


internal/scheduler/scheduler.go
```diff
@@ -108,6 +110,7 @@ func NewScheduler(
 		nodeIdLabel:                nodeIdLabel,
 		jobsSerial:                 -1,
 		runsSerial:                 -1,
+		Metrics:                    NewSchedulerMetrics(),
```
severinson, line 113:
```suggestion
		metrics:                    NewSchedulerMetrics(),
```

internal/scheduler/scheduler.go
```diff
@@ -221,6 +234,11 @@ func (s *Scheduler) cycle(ctx context.Context, updateAll bool, leaderToken Leade
 			return err
 		}
 
+		// Report counts of scheduled jobs per queue.
+		// TODO: preemptible jobs, possibly other metrics here.
+		s.metrics.ReportScheduledJobs(overallSchedulerResult.ScheduledJobs)
```
d80tb7, line 239: we probably only want to set these if  `shouldSchedule && leaderToken.leader`

internal/scheduler/scheduler_metrics.go
```diff
@@ -0,0 +1,145 @@
...
+		if err != nil {
+			// A metric failure isn't reason to kill the programme.
+			log.Error(err)
+		}
+
+		observer.Observe(float64(count))
+	}
+}
+
+func (metrics *SchedulerMetrics) ReportPreemptedJobs(preemptedJobs []interfaces.LegacySchedulerJob) {
```
d80tb7, line 113: Feels like ` ReportScheduledJobs` and `ReportPreemptedJobs`  are 90% the same and thus could delegate to a common function

internal/scheduler/scheduler_metrics.go
```diff
@@ -0,0 +1,145 @@
...
+		priorityClassName := key.priorityClass
+
+		observer, err := metrics.scheduledJobs.GetMetricWithLabelValues(queue, priorityClassName)
+
+		if err != nil {
+			// A metric failure isn't reason to kill the programme.
+			log.Error(err)
+		}
+
+		observer.Observe(float64(count))
```
d80tb7, line 109: if you have an a error above you'll still end up calling this line.  I suspect observer will be undefined in this case and quite possibly `nil`, in which case you'll panic here.  

You probably just want to put this in an `else` clause.

internal/scheduler/scheduler.go
theAntiYeti: hmm I see what you mean. I thought it would exit out earlier if not leader but I guess that neglects the fact that it could lose leadership partway through scheduling.

internal/scheduler/scheduler_metrics.go
theAntiYeti: I agree but it also somewhat feels like a premature optimization. I'm worried we might end up locking ourselves into inflexibility by overusing DRY

internal/scheduler/scheduler.go
```diff
@@ -167,7 +170,18 @@ func (s *Scheduler) Run(ctx context.Context) error {
 				logging.WithStacktrace(log, err).Error("scheduling cycle failure")
 				leaderToken = InvalidLeaderToken()
 			}
-			log.Infof("scheduling cycle completed in %s", s.clock.Since(start))
+
+			cycleTime := s.clock.Since(start)
+
+			if shouldSchedule && leaderToken.leader {
+				// Only the leader token does real scheduling rounds.
+				s.metrics.ReportScheduleCycleTime(float64(cycleTime.Milliseconds()))
```
severinson, line 178: Better to give a `time.Duration` imo since that captures the unit.

internal/scheduler/scheduler_metrics.go
```diff
@@ -0,0 +1,161 @@
...
+	SUBSYSTEM = "scheduler"
+)
+
+type SchedulerMetrics struct {
+	// Cycle time when scheduling, as leader.
+	scheduleCycleTime prometheus.Histogram
+	// Cycle time when reconciling, as leader or follower.
+	reconcileCycleTime prometheus.Histogram
+	// Number of jobs scheduled per queue.
+	scheduledJobs prometheus.HistogramVec
```
severinson, line 23:
```suggestion
	scheduledJobsPerQueue prometheus.HistogramVec
```
I think this is more clear.

internal/scheduler/scheduler_metrics.go
```diff
@@ -0,0 +1,161 @@
...
+var (
+	schedulerMetricsPointer *SchedulerMetrics
+	singletonLock           sync.Mutex
+)
+
+func GetSchedulerMetrics() *SchedulerMetrics {
+	singletonLock.Lock()
+	defer singletonLock.Unlock()
+
+	if schedulerMetricsPointer == nil {
```
severinson, line 37: You can do this with init instead.

internal/scheduler/scheduler_metrics.go
```diff
@@ -0,0 +1,161 @@
...
+	// Cycle time when reconciling, as leader or follower.
+	reconcileCycleTime prometheus.Histogram
+	// Number of jobs scheduled per queue.
+	scheduledJobs prometheus.HistogramVec
+	// Number of jobs preempted per queue.
+	preemptedJobs prometheus.HistogramVec
+}
+
+var (
+	schedulerMetricsPointer *SchedulerMetrics
```
severinson, line 29:
```suggestion
	schedulerMetrics *SchedulerMetrics
```
No need to add the type to the variable name imo.

internal/scheduler/scheduler_metrics.go
```diff
@@ -0,0 +1,161 @@
...
+			Buckets:   prometheus.LinearBuckets(0, 5, 20),
+		},
+	)
+
+	reconcileCycleTime := prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Namespace: NAMESPACE,
+			Subsystem: SUBSYSTEM,
+			Name:      "reconcile_cycle_times",
+			Help:      "Cycle time when outside of a scheduling round.",
```
severinson, line 60: This is the cycle time when not leader, no?

internal/scheduler/scheduler_metrics.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+type collectionKey struct {
+	queue         string
+	priorityClass string
+}
+
+// aggregateJobs takes a list of jobs and counts how many there are of each queue, priorityClass pair.
+func aggregateJobs[S ~[]E, E interfaces.LegacySchedulerJob](scheduledJobs S) map[collectionKey]int {
+	groups := make(map[collectionKey]int)
```
severinson, line 136: This is a histogram? I'd call it histogram if so.

internal/scheduler/scheduler_metrics.go
theAntiYeti: When not leader or when not scheduling. That is when you're just reading the jobs to get up to date in the state.

severinson: I'd say that in the description.

internal/scheduler/scheduler.go
theAntiYeti: histogram doesn't handle time.Duration well.

severinson: You could do the conversion in (metrics *SchedulerMetrics) ReportScheduleCycleTime tho.

severinson (approved):

PR #2850, opened 2023-08-09
Self-hosting `armadactl` plugin in armada

ShivangShandilya:
This PR self-hosts our armadactl plugin via krew-index inside Armada root repository.
After successful merging of this PR, users will be able to use our plugin with the help of krew.

Reference: https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes/

### DIFF
--- a/plugins/armadactl.yml
+++ b/plugins/armadactl.yml
@@ -1,0 +1,36 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: armadactl
+spec:
+  version: v0.3.8655
+  homepage: https://github.com/armadaproject/armada
+  shortDescription: Command line utility to submit many jobs to armada
+  description: |
+    armadactl is a command-line tool used for managing jobs in the Armada workload orchestration system.
+    It provides functionality for creating, updating, and deleting jobs, as well as monitoring job status and resource usage.
+  caveats: |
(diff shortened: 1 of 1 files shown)


plugins/armadactl.yml
```diff
@@ -0,0 +1,36 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: armadactl
+spec:
+  version: v0.3.8655
```
kannon92, line 6:
```suggestion
  version: v0.3.88
```

Can you look into ways to get this plugin manifest updated with our latest release?  Maybe as a followup.

kannon92 (dismissed):

plugins/armadactl.yml
kannon92: After you apply this suggestion I think this is mergable actually.

kannon92 (approved):

Sharpz7:
I am happy to merge this PR as is, but I want to make sure we have docs on using it as well later down the line - or in this PR.

Whatever is prefered.

ShivangShandilya:
> I am happy to merge this PR as is, but I want to make sure we have docs on using it as well later down the line - or in this PR.
> 
> Whatever is prefered.

Yeah right I was planning to make another PR which will add README to this directory explaining how one can use this plugin, the reason I'm not including the same in this PR because I was thinking of attaching a video tutorial for this once I record it

PR #2857, opened 2023-08-09
Better Deduplicaton Logic

d80tb7:
Previously we permanently associated client ids with job ids before we published to pulsar.  This left us open to the following situation.

- User submits jobs with client ids
- Armada saves client ids
- Publish to pulsar fails
- User resubmits with same client id
- User is told the job is a duplicate and is redirected to a non-existent job id

Importunately closing this hole is impossible as we have two systems storing state  (pulsar and postgres) and no transactionality between the two. That said, we can make the situation a great deal better here.  The new code works as follows:
- Check jobs to see if client id says they are duplicates
- Publish messages to pulsar
- Store clients ids

This should be  better because although we still don't have transactionality, this does at least mean that the fail scenario is that jobs run twice rather than the user being given a non  existent duplicate.  Additionally we now *don't* return an error if the dedupe logic fails which means :
  a) if the final store of the client id fails we shouldn't see a resubmission regardless 
  b) users can  still submit if postgres is down.

In the very short term we should be able to close the duplicate jobs hole further by using Pulsar transactions.  Longer term, the only way to completely close this off would be to make the scheduler decide whether the job is a duplicate.   This is a much nicer solution  but unfortunately is a breaking  api change as we would no longer be able to return a response saying that the job is a duplicate- instead we'd have to notify the client async via the events.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-644) by [Unito](https://www.unito.io)


internal/armada/server/submit_test.go
```diff
@@ -469,38 +469,6 @@ func TestSubmitServer_SubmitJob_ReturnsJobItemsInTheSameOrderTheyWereSubmitted(t
 	})
 }
 
-func TestSubmitServer_SubmitJobs_HandlesDoubleSubmit(t *testing.T) {
```
d80tb7, line 472: This check has been removed because this is tersting deduplicationlogic on the old submitserver.  We've now removed this logic because we don't need dedupe logic in two places.

JamesMurkin (approved):

PR #2862, opened 2023-08-09
Feat/improved version checking output

Raajheer1:
This PR does two things:
- Refactors the version checking code to use the already written constraintCheck() function
- Expands the Constraint Check function to also output the actual dependency that is out dated.

The current functionality if a constraint is failed for example docker the output is:
"Error: found version xxx but it failed constraint >= xxx"

This PR changes the output to be:
"Error: found **docker** version xxx  but it failed constraint >= xxx"

Other than "docker" it could print out: "kind", "Go", "kubectl", "golangci-lint", "protoc", and "sqlc".

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-649) by [Unito](https://www.unito.io)

### DIFF
--- a/magefiles/docker.go
+++ b/magefiles/docker.go
@@ -74,13 +74,13 @@ func dockerVersion() (*semver.Version, error) {
 	return version, nil
 }
 
-func constraintCheck(version *semver.Version, versionRequirement string) error {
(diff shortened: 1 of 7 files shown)


Sharpz7 (approved):
Nice change!

richscott (approved):

PR #2880, opened 2023-08-14
Improved preemption logic

severinson:
Better account for fairness when choosing what to preempt.

Despite a more complex algorithm, performance doesn't change much.
```
goos: windows
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i7-12700H
                                                             │   old.txt   │            new.txt             │
                                                             │   sec/op    │   sec/op     vs base           │
PreemptingQueueScheduler/1000_nodes_10_queues_320000_jobs-20   1.267 ± ∞ ¹   1.343 ± ∞ ¹  ~ (p=1.000 n=1) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                                                             │    old.txt    │             new.txt              │
                                                             │     B/op      │     B/op       vs base           │
PreemptingQueueScheduler/1000_nodes_10_queues_320000_jobs-20   1.647Gi ± ∞ ¹   1.645Gi ± ∞ ¹  ~ (p=1.000 n=1) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                                                             │   old.txt    │             new.txt             │
                                                             │  allocs/op   │  allocs/op    vs base           │
PreemptingQueueScheduler/1000_nodes_10_queues_320000_jobs-20   14.47M ± ∞ ¹   14.40M ± ∞ ¹  ~ (p=1.000 n=1) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05
```


internal/scheduler/testfixtures/testfixtures.go
```diff
@@ -100,6 +100,7 @@ func TestSchedulingConfig() configuration.SchedulingConfig {
 		DominantResourceFairnessResourcesToConsider: TestResourceNames,
 		ExecutorTimeout:                  15 * time.Minute,
 		MaxUnacknowledgedJobsPerExecutor: math.MaxInt,
+		EnableNewPreemptionStrategy:      true,
```
severinson, line 103: I've run the tests both with and without this flag enabled. They succeed in both cases.

JamesMurkin (approved):

PR #2882, opened 2023-08-15
Armada auth examples

ClifHouck:
Rebase and slight revision of Adam's work to demonstrate auth with the `ArmadaOperator`.


third_party/airflow/examples/hello_armada_auth.py
```diff
@@ -0,0 +1,146 @@
...
+
+if __name__ == "__main__":
+    from sqlalchemy import create_engine
+    from sqlalchemy.orm.session import Session
+
+    engine = create_engine("postgresql://airflow:airflow@172.20.0.3:5432/airflow")
+    session = Session(engine)
+
+    # Will need to run `airflow db init` first
+    dag.test(
```
Sharpz7, line 141: Can I ask why you added this?

ClifHouck: I probably shouldn't leave this in, but it wouldn't connect to the airflow postgres instance without it

Sharpz7: I managed to get it to run by running `airflow db init` locally. I think this would be an easy win for more CI tests?

config/armadactl/config.yaml
```diff
@@ -0,0 +1,4 @@
+target: "server:50051"
```
Sharpz7, line 1: These configs won't be able to stay in the repo. We probably want to allow localdev to switch between them either using:

- Env Vars like in https://github.com/armadaproject/armada/tree/master/developer/env

- Extra configs like in https://github.com/armadaproject/armada/tree/master/developer/config

Happy to assist in this if you would like.

magefiles/airflow.go
```diff
@@ -39,7 +39,13 @@ func Airflow(arg string) error {
 func startAirflow() error {
 	fmt.Println("Starting airflow...")
 
-	err := sh.Run("mkdir", "-p", "developer/airflow/opt")
+	// Remove developer/airflow/opt
+	err := sh.Run("rm", "-rf", "developer/airflow/opt/")
```
Sharpz7, line 43: I found this to be helpful. Without it, I got problems with nested copies when repeatedly trying to run airflow?

config/armadactl/config.yaml
ClifHouck: That's fine, but beyond the scope of this PR I think.

magefiles/airflow.go
ClifHouck: Nested copies seems strange. I don't think I had that behavior. However, I don't think this is a problem.

third_party/airflow/examples/hello_armada_auth.py
ClifHouck: `airflow db init` shouldn't be necessary after `mage airflow start`

PR #2896, opened 2023-08-19
Upgrade Pulsar Client to v0.11

d80tb7:
There is one breaking change in the client which is that `consumer.Ack` now returns an error.  In almost all cases I think we simply want to continually retry acking the message, else we'll end up out of order.

┆Issue is synchronized with this [Jira Task](https://gr-oss.atlassian.net/browse/BATCH-681) by [Unito](https://www.unito.io)


internal/armada/server/submit_from_log.go
```diff
@@ -766,3 +765,20 @@ func (srv *SubmitFromLog) ReprioritizeJobSet(
...
+
+func (srv *SubmitFromLog) ack(ctx context.Context, msg pulsar.Message) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			err := srv.Consumer.Ack(msg)
+			if err == nil {
+				return
```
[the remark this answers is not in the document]
theAntiYeti: I guess the alternative fix here would involve labelled break?

internal/armada/server/eventsprinter.go
```diff
@@ -73,16 +74,21 @@ func (srv *EventsPrinter) Run(ctx context.Context) error {
...
 				break
 			} else if err != nil {
 				logging.WithStacktrace(log, err).Warnf("receiving from Pulsar failed")
 				break
 			}
-			consumer.Ack(msg)
+			util.RetryUntilSuccess(
+				context.Background(),
+				func() error { return consumer.Ack(msg) },
+				func(err error) { logging.WithStacktrace(log, err).Warnf("acking pulsar message failed") },
```
d80tb7, line 90: di we want a backoff here?

internal/common/pulsarutils/async.go
```diff
@@ -118,7 +118,18 @@ func Ack(ctx context.Context, consumers []pulsar.Consumer, msgs chan []*Consumer
 						"Asked to ack message belonging to consumer %d, however this is outside the bounds of the consumers array which is of length %d",
 						id.ConsumerId, len(consumers)))
 			}
-			consumers[id.ConsumerId].AckID(id.MessageId)
+			for {
```
d80tb7, line 121: I think we should chnage this to use your new `util.RetryUntilSuccess` function

internal/armada/server/eventsprinter.go
theAntiYeti: you didn't have one initially so I avoided it. But I'm sure a 1sec backoff won't hurt

d80tb7 (commented):
lgtm

theAntiYeti (approved):

internal/armada/server/submit_from_log.go
```diff
@@ -113,8 +113,7 @@ func (srv *SubmitFromLog) Run(ctx context.Context) error {
 			// If this message isn't for us we can simply ack it
 			// and go to the next message
 			if !schedulers.ForLegacyScheduler(msg) {
-				srv.Consumer.Ack(msg)
-				break
+				srv.ack(ctx, msg)
```
JamesMurkin, line 116: Is removing the break here intentional?

internal/armada/server/submit_from_log.go
```diff
@@ -146,7 +145,7 @@ func (srv *SubmitFromLog) Run(ctx context.Context) error {
 			messageLogger.WithField("numEvents", len(sequence.Events)).Info("processing sequence")
 			// TODO: Improve retry logic.
 			srv.ProcessSequence(ctxWithLogger, sequence)
-			srv.Consumer.Ack(msg)
+			srv.ack(ctx, msg)
```
JamesMurkin, line 148: If the context expires before getting here, we don't even ack here at all right?

JamesMurkin: Actually the context here never expires unless the serer is shutting down, so probably fine

internal/armada/server/submit_from_log.go
d80tb7: no- it wasn't.  @theAntiYeti can we put this back

theAntiYeti: done

JamesMurkin (approved):

PR #2925, opened 2023-08-29
Improved etcd protection

severinson:


internal/scheduler/jobiteration.go
```diff
@@ -51,6 +52,8 @@ type InMemoryJobRepository struct {
 	// If true, jobs are sorted first by priority class priority.
 	// If false, priority class is ignored when ordering jobs.
 	sortByPriorityClass bool
+	// Protects the above fields.
+	mu sync.Mutex
```
severinson, line 56: I had a test fail due to concurrent map modification in this class so I've added a lock.

JamesMurkin (approved):

config/executor/config.yaml
```diff
@@ -36,10 +36,6 @@ kubernetes:
   impersonateUsers: false
   trackedNodeLabels:
     - kubernetes.io/hostname
-  etcd:
```
severinson, line 39: Need to provide the metric urls anyway so we may as well leave this empty by default.

JamesMurkin (approved):

PR #2932, opened 2023-08-30
Stop executor requesting more jobs when it still has leased jobs

JamesMurkin:
Currently we "queue" jobs to be submitted on the executor - which sit the leased state until they are submitted to kubernetes

However this causes 2 issues with our current setup:
 - It prevents back-pressure from working well on the scheduler side. As the scheduler sees all these "Leased" jobs as active, so just keep scheduling more
 - In the case we are slowing submission due to etcd going over its limit. We "queue" lots of jobs, and as soon as etcd goes under its limit we hit it with potentially thousands of jobs

This flow needs further work and thought - however for now this is the minimal fix to prevent bad behaviour


internal/executor/configuration/types.go
```diff
@@ -19,6 +19,9 @@ type ApplicationConfiguration struct {
 	UseExecutorApi         bool
 	UseLegacyApi           bool
 	JobLeaseRequestTimeout time.Duration
+	// MaxLeasedJobs is the maximum jobs the executor should have in Leased state ay any one time (i.e jobs not submitted to kubernetes)
+	// It is largely used to calculate how many new jobs to request from the scheduler
+	MaxLeasedJobs int
```
severinson, line 24:
```suggestion
	MaxLeasedJobs int `validate:"gt=0"`
```

severinson (approved):

PR #2934, opened 2023-08-31
ArmadaContext that includes a logger

d80tb7:
This defines `armadacontext.Context` that works exactly like `context.Context` but also includes a `logrus.Entry`.  This allows us to pass around a contextual logger in a typesafe manner without having to add an extra log parameter to every function.

This touches a very large number of files but the vast majority of these are simply swapping out `context.Context` with `armadacontext.Context`.

Once this is merged we can raise follow-up prs to rationalise some of the way we do logging


internal/armada/server/submit_from_log.go
```diff
@@ -121,29 +119,18 @@ func (srv *SubmitFromLog) Run(ctx context.Context) error {
 			lastPublishTime = msg.PublishTime()
 			numReceived++
 
-			// Incoming gRPC requests are annotated with a unique id,
```
d80tb7, line 124: I've removed this because a) this code is going away soon and b) we've largely broken the link between pulsar message and grpc request so it's not that useful.

internal/common/ingest/ingestion_pipeline.go
```diff
@@ -206,7 +207,7 @@ func (ingester *IngestionPipeline[T]) Run(ctx context.Context) error {
 			} else {
 				for _, msgId := range msg.GetMessageIDs() {
 					util.RetryUntilSuccess(
-						context.Background(),
+						armadacontext.Background(),
```
d80tb7, line 210: The fact we create a new Background context here looks wrong but wasn't introduced in this PR.  I'll take a look at changing this in a follow up.

severinson (approved):

PR #2938, opened 2023-08-31
Add per-queue scheduling rate-limiting

severinson:
This PR adds global and per-queue job scheduling rate-limiting via token-bucket rate-limiters. This to prevent overloading downstream systems, e.g., etcd, with large numbers of jobs. This form of rate-limiting is more precise than the previous `maximumJobsToSchedule`, which did not account for the amount of time between scheduling rounds.


internal/scheduler/gang_scheduler.go
```diff
@@ -84,24 +100,13 @@ func (sch *GangScheduler) Schedule(ctx context.Context, gctx *schedulercontext.G
...
-	// Try scheduling the gang.
 	if _, err = sch.schedulingContext.AddGangSchedulingContext(gctx); err != nil {
 		return
 	}
 	gangAddedToSchedulingContext = true
 	if !gctx.AllJobsEvicted {
-		// Check that the job is large enough for this executor.
-		// This check needs to be here, since it relates to a specific job.
-		// Only perform limit checks for new jobs to avoid preempting jobs if, e.g., MinimumJobSize changes.
-		if ok, unschedulableReason = requestsAreLargeEnough(gctx.TotalResourceRequests, sch.constraints.MinimumJobSize); !ok {
```
severinson, line 97: These checks have been moved into the constraints package.

internal/armada/configuration/types.go
```diff
@@ -124,10 +124,25 @@ type SchedulingConfig struct {
...
-	MaximumGangsToSchedule uint
+	// Token bucket global job scheduling rate limiter settings; see
+	// https://pkg.go.dev/golang.org/x/time/rate#Limiter
+	//
+	// Rate-limiting is based on the number of tokens available at the start of each scheduling round,
+	// i.e., tokens accumulated while scheduling are only available at the start of the next scheduling round.
+	//
+	// MaximumSchedulingRate controls the number of jobs scheduled per second in steady-state,
+	// i.e., once the burst capacity has been exhausted.
+	MaximumSchedulingRate float64 `validate:"gt=0"`
```
d80tb7, line 135: I understand why you've made the rate limits act over a given scheduling cycle, but I think longer term we want to make these apply over some timer horizon.  That's because what users and administrators care about isn't the scheduling rate in given scheduling cycle but rather the scheduling rate over various time horizons.

d80tb7, line 135: It also might be worth exploring whether  `MaximumSchedulingRate` `MaximumPerQueueSchedulingRate`  could be  ints here- the benefit being that config would be slighly simpler.  `MaximumSchedulingRate` almost certainly could be an int, but `MaximumPerQueueSchedulingRate` maybe not?

internal/scheduler/constraints/constraints.go
```diff
@@ -12,32 +13,46 @@ import (
 )
 
 const (
-	UnschedulableReasonMaximumResourcesScheduled        = "maximum resources scheduled"
-	UnschedulableReasonMaximumNumberOfJobsScheduled     = "maximum number of jobs scheduled"
-	UnschedulableReasonMaximumNumberOfGangsScheduled    = "maximum number of gangs scheduled"
-	UnschedulableReasonMaximumResourcesPerQueueExceeded = "maximum total resources for this queue exceeded"
+	// Indicates that the limit on resources scheduled per round has been exceeded.
+	MaximumResourcesScheduledUnschedulableReason = "maximum resources scheduled"
```
d80tb7, line 17: it might be worth making these reasons into a type just so that functions like `IsTerminalUnschedulableReason` can be more explicit about what they operate on.

internal/scheduler/constraints/constraints.go
```diff
@@ -99,47 +112,79 @@ func absoluteFromRelativeLimits(totalResources schedulerobjects.ResourceList, re
...
-	}
-
-	// MaximumGangsToSchedule check.
-	if constraints.MaximumGangsToSchedule != 0 && sctx.NumScheduledGangs == int(constraints.MaximumGangsToSchedule) {
-		return false, UnschedulableReasonMaximumNumberOfGangsScheduled, nil
-	}
+// ScaleQuantity scales q in-place by a factor f.
+// This functions overflows for quantities the milli value of which can't be expressed as an int64.
+// E.g., 1Pi is ok, but not 10Pi.
+func ScaleQuantity(q resource.Quantity, f float64) resource.Quantity {
```
d80tb7, line 118: I know this wasn't added here but functions like this scare the hell out of me!

internal/scheduler/constraints/constraints.go
```diff
@@ -99,47 +112,79 @@ func absoluteFromRelativeLimits(totalResources schedulerobjects.ResourceList, re
...
+		return false, "", errors.Errorf("no QueueSchedulingContext for queue %s", gctx.Queue)
+	}
+
+	// Check that the job is large enough for this executor.
+	if ok, unschedulableReason := RequestsAreLargeEnough(gctx.TotalResourceRequests, constraints.MinimumJobSize); !ok {
+		return false, unschedulableReason, nil
+	}
+
+	// Global rate limiter check.
+	if sctx.Limiter != nil {
```
d80tb7, line 146: given that sctx/qctx are passed around al over the place, I think it's would be better if we could guarantee that the limiters would alwaysbe non-nil.  Benefits would be:
- less chance of someone forgetting a nil check
- simpler functions as callers could skip the nil checks.

Unless we think perf is a major consideration I'd just make sure we always instantiate a limiter, using infinite limits if limits are disabled

internal/scheduler/constraints/constraints.go
```diff
@@ -99,47 +112,79 @@ func absoluteFromRelativeLimits(totalResources schedulerobjects.ResourceList, re
...
+		tokens := sctx.Limiter.TokensAt(sctx.Started)
+		if tokens <= 0 {
+			return false, GlobalRateLimitExceededUnschedulableReason, nil
+		}
+		if sctx.Limiter.Burst() < len(gctx.JobSchedulingContexts) {
```
d80tb7, line 151: I'd consider abstracting  len(gctx.JobSchedulingContexts) into a `gctx.NumJobsScheduled()` function just to make it clearer what is going on here.

d80tb7: or even a `gctx.GangSize` field if a gang cannot change size over the course of a scheduler cycle.

internal/armada/configuration/types.go
```diff
@@ -124,10 +124,25 @@ type SchedulingConfig struct {
...
+	// MaximumSchedulingBurst controls the burst capacity of the rate-limiter.
+	//
+	// There are two important implications:
+	// - Armada will never schedule more than MaximumSchedulingBurst jobs per scheduling round.
+	// - Gang jobs with cardinality greater than MaximumSchedulingBurst can never be scheduled.
```
d80tb7, line 140: Do we check for this at submit time?  I.e. is is possible for a user to submit a gang job that can never be scheduled?

internal/armada/configuration/types.go
severinson: I've updated the docstring to make clear the rate-limiter persists between rounds.

severinson: I think we can leave them as floats. We can use either int or float values in config as-is.

internal/armada/configuration/types.go
severinson: Yes. As discussed, let's add a max gang size separately and then have some warning if that is greater than the burst.

internal/scheduler/constraints/constraints.go
severinson: I agree. But didn't want to do that in this PR.

internal/scheduler/constraints/constraints.go
severinson: We may actually want to update this to return an error (or panic) if it would overflow.

internal/scheduler/constraints/constraints.go
severinson: Agreed. I've removed the nil checks.

internal/scheduler/constraints/constraints.go
severinson: Agreed. Done.

d80tb7 (dismissed):

d80tb7 (approved):

PR #2940, opened 2023-08-31
Create more detailed error responses when submitting batch jobs

richscott:
_NOTE: this PR is a continuation of the work/PR that our summer 2023 MLH fellow/intern Raaj Patel had created. Unfortunately, his internship ended and he has returned to university before that PR (https://github.com/armadaproject/armada/pull/2623) was fully finished, approved, and merged. The only change to this code since he departed I have made is to fix a conflict in go.mod and go.sum._ 

_Following is his description in the original PR:_

**What type of PR is this?**

This PR adds functionality to JobSubmitResponse when an error has occured. Previously the only thing returned was the error. The changes made in this PR allow the code to do 2 things:
- Build a collection of errors, if mulitple jobs error for the same return, these changes will return them at the same time to the user. Rather than them having to reattempt submissions over and over again with 1 fix each time.
- Return more detailed errors to the user by utilizing the JobSubmitResponse Struct

**Which issue(s) this PR fixes:**

Fixed https://github.com/armadaproject/armada/issues/1748

Original Description from issue 1748 in JIRA, opened by @d80tb7 :
> When we reject a job (for example because it can't be scheduled on any node) the whole grpc call fails.
> There is an error message that is returned, but users have said that this doesn't give enough infomation
> about what was valid. Instead they ask:

> is it possible for these errors to be returned on the Error field on the JobResponseItem that we're given?


richscott:
Screenshots from Raaj Patel, showing the logging output before and after this change. 
> Submitted a bad sleep yaml job containing two jobs, both have invalid podSpecs as the requested and limit resources both contain "memory: 0"

before-log-changes
after-log-changes

internal/armada/server/job_validation.go
```diff
@@ -13,16 +15,30 @@ import (
...
+) (bool, []*api.JobSubmitResponseItem, error) {
 	activeClusterSchedulingInfo := scheduling.FilterActiveClusterSchedulingInfoReports(allClusterSchedulingInfo)
+	responseItems := make([]*api.JobSubmitResponseItem, 0, len(jobs))
 	for i, job := range jobs {
 		if ok, err := scheduling.MatchSchedulingRequirementsOnAnyCluster(job, activeClusterSchedulingInfo); !ok {
 			if err != nil {
-				return false, errors.WithMessagef(err, "%d-th job can't be scheduled", i)
+				response := &api.JobSubmitResponseItem{
+					JobId: job.Id,
+					Error: errors.WithMessagef(err, "%d-th job can't be scheduled", i).Error(),
```
d80tb7, line 26: I think here you're creating an error only to immediately throw it away.  It might be simpler to just use a `fmt.sprintf`

internal/armada/server/lease.go
```diff
@@ -954,11 +954,11 @@ func (q *AggregatedQueueServer) addAvoidNodeAffinity(
 		}
 
 		changed := addAvoidNodeAffinity(jobs[0], labels, func(jobsToValidate []*api.Job) error {
-			if ok, err := validateJobsCanBeScheduled(jobsToValidate, allClusterSchedulingInfo); !ok {
+			if ok, responseItems, err := validateJobsCanBeScheduled(jobsToValidate, allClusterSchedulingInfo); !ok {
 				if err != nil {
-					return errors.WithMessage(err, "can't schedule at least 1 job")
+					return errors.WithMessagef(err, "can't schedule %d job(s)", len(responseItems))
```
d80tb7, line 959: What does the error say here if you e.g. submit a batch of 100 jobs and 99 of them are fine but 1 can be scheduled.  Does it say "can't schedule 100 job(s)" or "can't schedule 1 jobs(s)"?

internal/armada/server/submit.go
```diff
@@ -765,29 +804,45 @@ func (server *SubmitServer) createJobsObjects(request *api.JobSubmitRequest, own
...
 	}
 
 	if request.Queue == "" {
-		return nil, errors.Errorf("[createJobs] queue not specified")
+		return nil, nil, errors.Errorf("[createJobs] queue not specified")
 	}
 
+	responseItems := make([]*api.JobSubmitResponseItem, 0, len(request.JobRequestItems))
 	for i, item := range request.JobRequestItems {
+		jobId := getUlid()
```
d80tb7, line 822: This looks like a change in behaviour in that previously we used to return `nil` for the id when we rejected jobs but now we return a jobId.  I think this is good in that we now have an id that the user can give us to search logs etc, but does have a couple of downsides:

- Users may get confused if they see jobIds that don't appear anywhere else in the system (e.g. lookout)
- It's a mildly breaking interface change- e.g. if clients previously had code that did `if jobResponse.JobId == nil...`  this will now behave differently.

I think I'm mildly in favour of this change, but it might be good to get a second opinion- @severinson ?

internal/armada/server/submit.go
```diff
@@ -253,13 +268,30 @@ func (server *SubmitServer) DeleteQueue(ctx context.Context, request *api.QueueD
 func (server *SubmitServer) SubmitJobs(ctx context.Context, req *api.JobSubmitRequest) (*api.JobSubmitResponse, error) {
 	principal := authorization.GetPrincipal(ctx)
 
-	jobs, e := server.createJobs(req, principal.GetName(), principal.GetGroupNames())
+	jobs, responseItems, e := server.createJobs(req, principal.GetName(), principal.GetGroupNames())
 	if e != nil {
```
d80tb7, line 272: Is there any limit to the size of the response we can generate here?  E.g. if the user submits a batch of 100 jobs and they all fail with some massive error are we sure we're not going to generate a response that exceeds  the max grpc response size? Likewise (although admittedly loess likely) could we end up generaling such a massive message that we would OOM.

If it were me I'd try and limit the size of the error message by saying:
- how may jobs failed validation (out of how may submitted)
- detailed error messages for the first n jobs that failed.  (maybe n=5 or something)

internal/armada/server/job_validation.go
richscott: Good catch - fixed in commit f1abe37a

internal/armada/server/lease.go
richscott: Good question - I've changed it so hopefully it's more clear - the code is now
```
return errors.WithMessagef(err, "can't schedule %d (out of %d submitted) job(s)", len(responseItems), len(jobsToValidate))
```

internal/armada/server/submit.go
richscott: @d80tb7  I'll replace that potentially-huge error with a  concise/distilled one, using that logic as you suggest.

richscott: @d80tb7 An attempt to implement the logic described above is now in commit 894c265a - let me know what you think.

d80tb7 (approved):

PR #2951, opened 2023-09-06
Updated correct address on which Armada-Airflow-UI is available & cred!

itsaviral2609:
#### What type of PR is this?

Updated correct address on which Armada-Airflow-UI is available http://localhost:8081/login/ and credentials airflow/airflow as user/password

image

Suggestion
Docs on armada-airflow setup can be made more detailed as folks will face issues on setting it up on local machine to test it due to python dependencies etc!


kannon92 (approved):
Thanks for this! I always had to check what those creds were.

Sharpz7 (approved):

richscott (approved):

PR #2952, opened 2023-09-06
Airflow: always set credentials from args in GRPC channel constructor

richscott:
In the `GrpcChannelArguments` constructor, always set the `credentials_callback_args` member from what is given. Add a unit test to verify serialization round-tripping is complete, and a `__eq__` implementation for GrpcChannelArguments. 

All this should allow the Airflow Deferrable Operator work with GRPC secure credentials in the same fashion as the non-Deferrable Operator.


ClifHouck (approved):
LGTM, thanks Rich!

dave-gantenbein (approved):

PR #2968, opened 2023-09-13
Run `on.push` only for master

### LINKED ISSUE #2959: Remove duplicated CI jobs
Some jobs get scheduled twice in certain situations.

[Python Airflow Operator](https://github.com/armadaproject/armada/blob/master/.github/workflows/airflow-operator.yml) runs twice on a PR.

```
Python Airflow Operator / airflow-tox (1.20) (pull_request) Queued — Waiting to run this check...
Python Airflow Operator / airflow-tox (1.20) (push) Queued — Waiting to run this check...
```

It gets scheduled for the `push` and the `pull_request` event.

We should look at examples in [FastTrackML](https://github.com/G-Research/fasttrackml) and [ParquetSharp](https://github.com/G-Research/ParquetSharp/).

Goal is to investigate which jobs get trigger multiple times and optimise them.

Mo-Fatah:
Fixes #2959 
- If a PR was originated from a branch from the upstream repo (not from a fork), this will trigger both the `pull_request` and the `push` workflows for this PR at the same time resulting in a duplicated CI checks. 
- The `not-**` workflows are no longer needed


dave-gantenbein (approved):

Sharpz7 (approved):

Sharpz7:
I am very keen to make sure we monitor PR's to make sure this doesn't do anything. Its been like this since the days of Jay being around - I doubt he did it without a reason.

This said, I think whats happened it github has changed their definition of whats triggers a "push" and "PR" action, and that is what has caused this.

PR #2977, opened 2023-09-18
LookoutV2 table should not overflow beyond the viewport

vthemelis:
* Make the table header and footer sticky;
* Only overflow the table content, not the whole of the V2 component;

Effectively change from this:
before

to this:
after


vthemelis:
I think that tests are now fixed. Could I get another run of the CI checks?

Sharpz7 (approved):

d80tb7 (approved):

PR #2994, opened 2023-09-21
Gang min cardinality

msumner91:
Context:
- This PR allows jobs in a gang to successfully schedule if a minimum number of jobs are deemed schedulable (previously we required that the entire gang was schedulable)
- In the case that the minimum threshold is schedulable, and the minimum is less than the maximum gang size, the remainder of jobs are failed back to the client for them to do with what they please
- This paves the way for clients such as DogeFarm to re-submit jobs to an existing gang in a different batch


severinson (dismissed):

severinson (approved):

msumner91:
Looks good to me

richscott (approved):

Issue #2987, opened 2023-09-20
Normalize workflow naming and mage usage [WIP]


Sharpz7:
This ticket will track potential things to cleanup in our CI.

A lot of this can be based on what is present in https://github.com/G-Research/fasttrackml/tree/main/.github/workflows

## New Structure

The only yamls that should be in the main `workflows` folder are the ones with the `on` tag. The goal of this is to have a structure that is easier to follow.

I think we should have a https://github.com/armadaproject/armada/blob/master/.github/workflows/ci.yml for each "type" of CI job:

- ci.yml: handles the main CI stuff
- ci-release.yml: handles releases
- ci-page.yml: handles the website
- ci-clients.yml handles the clients 
- ci-third-party.yml: handles third-party elements (i.e. airflow)

We can then have other folders to keep everything else. I.e `go`, `python` etc. This should help CI look cleaner and keep to the DRY principle, and make it more "natural" to keep our CI consistent across everything. (Based on code here: https://github.com/m4rs-mt/ILGPU/pull/1046)

## Mage / Tox Changes

All tox commands should go through mage to make CI consistent. It also removed the problem of needing to be in a different folder for it to work. Everything should be runnable from the root of the repo.

## Mage Update / Verify

These are targets that Kubernetes has. Verify is equivalent to running CI locally in full. Update is everything like docs, linting, builds, being done for you in one bundled command.

## Lookout CI Additions

We should:
- Lint and build Lookout in CI
- Cache lookout builds
- Copy the build instead of re-building inside the image for the "prepare" job in CI.

dave-gantenbein:
Won't do

PR #2996, opened 2023-09-25
Patch for CI

Sharpz7:
# Strange Temp Note
Sometimes (Like on commit ad47bf5), the cache will be hit, but not used. I have no idea what causes this.

Partially mentioned in #2987 

I tested it by merging it into my forks main and:
- Checking that the release still passes
- Checking that the cache was created
- Created another PR on my fork to verify that the cache works, and the required check system functions: https://github.com/Sharpz7/armada/pull/17
- Check the CI runs on my master fork okay: https://github.com/Sharpz7/armada/actions/runs/6591442058

The Integration Tests on master create the cache, and that is shared everywhere, including with new PR's (i.e no more long waits on your first push)

No Cache: https://photos.app.goo.gl/MGSWW51YY3BSwvPG9 (Integration Tests: 17m)
Old Cache: https://photos.app.goo.gl/CtuqvmgoKLfWNJpc6 (Integration Tests: 14m)
New Cache: https://photos.app.goo.gl/qyWcsKxUtN6RYdFk8 (Integration Tests: 7m) Half!!!

# IMPORTANT (Taken from https://github.com/G-Research/fasttrackml/pull/458#issue-1945354284)

This PR improves how we create the required check "All required checks succeeded" in the following ways:

- We now use a GitHub app to create the check instead of using the token provided by GitHub Actions. This allows us to create the check run with its own check suite instead of attaching it to the first GHA check suite, which changes when a pull request is closed and reopened, or when a new run attempt is made. See https://github.com/orgs/community/discussions/24616#discussioncomment-6088422 for more details on this very annoying GitHub limitation.
- A check is created and set as queued or in_progress when the corresponding workflow run's status changes.
This is needed so that the check is reset when a new run attempt is made or a PR is reopened.
- Logging is improved.
- Start/Completion times are added.

The following tasks need to be completed before merging this PR:

 - create and install a new GitHub app with the check:write permissions
-  add its credentials to a new create-check environment that can only be accessed by the main branch
The following tasks need to be completed after merging this PR:

 update branch protection (or ruleset, if https://github.com/G-Research/fasttrackml/pull/452 has been merged in the meantime) to enforce the required check to originate from the new GitHub app


Sharpz7:
When this is merged, the job `All required checks succeeded` is what needs to be required, not `All required checks done`

Sharpz7:
https://github.com/actions/runner-images/issues/668 - looks like this is the problem

Sharpz7:
Also implementing https://github.com/actions/cache/blob/main/caching-strategies.md#make-cache-read-only--reuse-cache-from-centralized-job

.github/workflows/build.yml
```diff
@@ -6,7 +6,7 @@ on:
 jobs:
   prepare:
     runs-on: ubuntu-latest-16-cores
-    if: github.repository_owner == 'armadaproject'
+    if: github.repository_owner == 'Sharpz7'
```
Sharpz7, line 9: Need to undo this before merging

Sharpz7:
The cache of `go/bin` is needed because of tool.yaml. It takes a long time to install the tools in here, therefore caching it can give a speedup

.github/workflows/build.yml
```diff
@@ -5,7 +5,7 @@ on:
 
 jobs:
   prepare:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-16-cores
```
dave-gantenbein, line 8: That's a lot of cores! Do we need this many?

.github/workflows/check-required.yml
```diff
@@ -0,0 +1,97 @@
...
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+
+      - uses: actions/github-script@v6
+        with:
+          github-token: ${{ steps.app-token.outputs.token }}
+          script: |
+            const ghaAppId = 15368;
```
pavlovic-ivan, line 39: Should this value be a secret? Also, const names are not concise. Would be good to rename to something more descriptive

pavlovic-ivan:
I don't quite understand the purpose of the checks-required workflow, together with the change made to the `all-required-checks-done` in `ci.yml`. Can you elaborate please? The echo in that job can be misleading since the checks are done in the other workflow, that can fail, but CI will always print out "All required checks done"

Sharpz7:
@pavlovic-ivan This is something jonathan created that is used in other GR-OSS repos. Basically, the echo you are refering to is not the one you use. You make your required job the one that prints "All required checks succeeded". To the best of my knowledge, this is a work around for Github Actions treating skipped jobs in a way we do not want.

.github/workflows/check-required.yml
Sharpz7: I don't think so. and since we use this check in other repos, I think we should keep it consistent between them in terms of naming

.github/workflows/test.yml
```diff
@@ -156,6 +177,22 @@ jobs:
...
 
+      # Only save on the intergration tests, as it is the most "complete" cache
+      # We want to save if the primary key and the matched key are different, or
+      # if the cache was not hit. The step we care about is steps.cache-restore.outputs
+      # Outputs
+      # cache-hit - A boolean value to indicate an exact match was found for the key.
+      # cache-primary-key - Cache primary key passed in the input to use in subsequent steps of the workflow.
+      # cache-matched-key - Key of the cache that was restored, it could either be the primary key on cache-hit or a partial/complete match of one of the restore keys.
+      - uses: actions/cache/save@v3
+        if: steps.cache-restore.outputs.cache-hit == 'false' || steps.cache-restore.outputs.cache-primary-key != steps.cache-restore.outputs.cache-matched-key
```
Sharpz7, line 188: This is to try and save the 20-90s we lose from the cache compressing, and then checking if the key exists. That should really happen the other way round.

Sharpz7:
Currently Broken:
- Cannot get release to work
- GitHub Pages Failure
- check-required.yml not working

No idea how to fix any of them currently, I will investigate.

.github/workflows/check-required.yml
Sharpz7: @jgiannuzzi do you know more about how to pick this ID?

jgiannuzzi: oh this is the public app ID of GitHub Actions 😄

jgiannuzzi:
@Sharpz7 please add information in your PR description about what needs to happen before and after your PR is merged for `check-required.yml` to work. See https://github.com/G-Research/fasttrackml/pull/458#issue-1945354284 for more info.

jgiannuzzi:
and for anyone looking for more context about the "new" required checks, see the following PRs:
- https://github.com/G-Research/fasttrackml/pull/241
- https://github.com/G-Research/fasttrackml/pull/458
- https://github.com/G-Research/fasttrackml/pull/470
- https://github.com/G-Research/fasttrackml/pull/473
- https://github.com/G-Research/fasttrackml/pull/481

.github/workflows/check-required.yml
pavlovic-ivan: > I don't think so. and since we use this check in other repos, I think we should keep it consistent between them in terms of naming

If this is the case, then the best thing would be for this workflow to exist in a dedicated "reusable" repo, and then called (if such a thing is acceptable for Armada). What happens when we have 10 repos all having the same check-required.yml workflow, and something changes in one of them? Are we going to track down the 9 other repos that don't have the latest change and create a PR? 

@jgiannuzzi @Sharpz7 feel free to tell me i'm nitpicking

jgiannuzzi: that's a great point @pavlovic-ivan

pavlovic-ivan: if you agree, i think we should create an action out of this

.github/workflows/airflow-operator.yml
```diff
@@ -90,15 +90,27 @@ jobs:
       - uses: actions/checkout@v3.3.0
       - run: docker buildx create --name ${DOCKER_BUILDX_BUILDER} --driver docker-container --use
       - run: docker buildx install
-      - uses: actions/setup-go@v4
+
+      - name: Set up Go
```
richscott, line 94: In the test, lint, and release workflows, we use the `magnetikonline/action-galng-cache` plugin, which functionally replaces the functionality of actions/setup-go and actions/cache/restore together - can that work here?

.github/workflows/release.yml
```diff
@@ -54,10 +54,23 @@ jobs:
       - name: Fetch Git tags
         run: git fetch --force --tags
 
-      - name: Setup Golang with Cache
```
richscott, line 57: And here we have the inverse - does `magnetikonline/action-golang-cache` not work, is buggy, or...?

.github/workflows/airflow-operator.yml
Sharpz7: Yes - Now that I deleted the extra folder we cache, this can potentially change. But I do think having separate save and cache steps (Which is not possible with `magnetikonline/action-galng-cache`) is better?

.github/workflows/release.yml
Sharpz7: See above comment

.github/workflows/airflow-operator.yml
richscott: I don't see the advantage of "separate save and cache steps" with using `actions/cache/restore`  - I believe the `magnetikonline` cache plugin automatically sets up the build and module cache paths as needed before, and persists those when the job is done.

.github/workflows/check-required.yml
jgiannuzzi: Sure! You may want to first prototype it in some repo and have your forks of armada and fml use it. If it works well, we can probably create a GR/actions repo to collect all of our shared actions/pipelines, starting with this one.

pavlovic-ivan: You are reading my mind (with GR/actions)!

jgiannuzzi:
Have a look at the work done in https://github.com/G-Research/fasttrackml/pull/552 (and its sibling PRs: https://github.com/G-Research/fasttrackml/pull/553 and https://github.com/G-Research/fasttrackml/pull/554).

This should show you a rationale for what to save/restore from the cache in an optimised way.

pavlovic-ivan:
FYI @dave-gantenbein @Sharpz7 @jgiannuzzi : i will be creating several PRs from now on, for this entire thing, split up into smaller pieces, and refer co-authors in those PRs (depending of course on the change itself). So i guess this PR can be closed, as others will contain changes from this one, + other necessary changes.

Sharpz7:
Closing. Good luck @pavlovic-ivan !

PR #3025, opened 2023-10-05
Add queue ttl

msumner91:
Context:
- Currently gang jobs are all submitted in a single batch to Armada and are rejected as a submit check if they aren't of a sufficient size
- This meant that you would either get all the jobs for a gang to yield a result or it is rejected
- In order to support elasticity between a min size and max size for gangs, we need to be able to submit gang jobs across batches
- With this it is possible that the gang size is not satisfied and jobs remain in the queue indefinitely
- Hence, we are adding the ability to set a TTL to indicate the max time a job can remain in the queued state in Armada forever


msumner91:
(Failing checks for airflow-integration-tests and python-client-integration-tests are cause by permissioning: "pull access denied for armada-airflow-operator-builder, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.")

d80tb7 (approved):

Issue #3032, opened 2023-10-09
Lookout: enhance logs tab of Job Details pop-up to render image-pull errors


richscott:
When there is an image-pull error in K8S, causing an Armada job to fail, the Logs tab of the Job Details pop-up just shows the (cryptic) error message `[x-request-id: "9gToCrFMA4FCKyNCcrt8pM"] pods "armada-01hc0gjpbpex5w944w85593yfg-0" not found` - this is due to there not being any Binoculars logs for the container, which never actually started. This can happen when a user specifies a non-existent image name in their job spec.

When an image-pull error occurs, Lookout should add additional content from the `Details` tab - directly from the `Error` section of the Details tab, which shows the image-pull error.

PR #3036, opened 2023-10-10
Show job run errors below container logs in details (#3032)

richscott:
In the job details pop-up, in the LOGS tab, if there are job run errors, render them in a new warning div, below the container logs div.

This is intended to be most useful when the container was not even able to start - quite often when an incorrect image name/version was specified, and the system was unable to pull the image. Without this, the container logs div just shows an error that it was unable to pull any logs from Binoculars.

Fixes https://github.com/armadaproject/armada/issues/3032

job-details-errimagepull


internal/lookout/ui/src/components/job-dialog/JobLogs.tsx
```diff
@@ -104,6 +105,13 @@ export default function JobLogs(props: JobLogsProps) {
           {props.error}
         </Alert>
       )}
+      <p>
```
dave-gantenbein, line 108: This will insert a <p> in all cases, probably you only want that block when there's an actual error

dave-gantenbein (changes requested):

internal/lookout/ui/src/components/job-dialog/JobLogs.tsx
richscott: Yes - good catch - fixed in commit 22f12d1c.

dave-gantenbein (commented):
lgtm

Mo-Fatah (approved):

dave-gantenbein (approved):
LGTM

PR #3038, opened 2023-10-11
Stop serving version 1 of the Lookout UI

zuqq:
This is a quick-and-dirty change to stop serving version 1 of the Lookout UI (keeping its API around for now); version 2 of the Lookout UI, which is currently served by `lookout` in tandem with version 1, is served by `lookoutv2` instead (together with the corresponding API).

Concretely, after running a sequence of commands like the following I end up with version 2 of the Lookout UI running on `localhost:10000`, and the API corresponding to version 1 of the Lookout UI running on `localhost:8089`:

```bash
mage ui
mage builddockers bundle,lookout-bundle
mage kind
docker compose up --detach postgres pulsar redis
# Before running this, wait for the containers from the preceding command to start.
docker compose up --detach executor lookout lookoutv2
```

I have only tested this change on my machine; there is likely a bit more work to do before this can be deployed.

### DIFF
--- a/cmd/lookout/main.go
+++ b/cmd/lookout/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"encoding/json"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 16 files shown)


zuqq:
I'm going to split up this PR into multiple pieces, so that we can roll them out separately.

PR #3040, opened 2023-10-12
Stop serving version 1 of the Lookout UI

zuqq:
This is the part of #3038 that only touches the Lookout UI, without changing either of the Lookout REST APIs.

On `master`, the Lookout UI has a navigation bar with four tabs, and `/` is routed to the "overview" tab:

lookout_v1

This PR removes the tabs and routes `/` to version 2 of the Lookout UI (which is accessible via the "v2" tab on `master`):

lookout_v2

This means that version 1 of the Lookout UI won't be accessible anymore, but I'm not deleting its code just yet.

In order to make the migration easier for our users, I'm redirecting `/v2` to `/` (being careful to preserve the query string), which means that existing links to version 2 of the Lookout UI will continue to work; I'm also adding a wildcard route (it just redirects to `/`), so that links to `/job-sets` or `/jobs` (corresponding to the "job sets" and "jobs" tabs) don't just show a mostly blank page.

### DIFF
--- a/internal/lookout/ui/src/App.tsx
+++ b/internal/lookout/ui/src/App.tsx
@@ -5,15 +5,13 @@ import { createGenerateClassName } from "@material-ui/core/styles"
 import { ThemeProvider as ThemeProviderV5, createTheme as createThemeV5 } from "@mui/material/styles"
 import { JobsTableContainer } from "containers/lookoutV2/JobsTableContainer"
 import { SnackbarProvider } from "notistack"
-import { Route, BrowserRouter, Routes } from "react-router-dom"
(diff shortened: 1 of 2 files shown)


Sharpz7:
I would want to see this correctly be completed in our "localdev" setup from mage. 

`mage localdev full` and `mage ui` allow for a developer to spin up armada using docker and have it all function. Happy to help and guide with this.

zuqq:
> I would want to see this correctly be completed in our "localdev" setup from mage.
> 
> `mage localdev full` and `mage ui` allow for a developer to spin up armada using docker and have it all function. Happy to help and guide with this.

This PR only touches the Lookout UI, so it is fully compatible with the existing setup.

Issue #3045, opened 2023-10-13
Scheduler: add timestamps to job runs in the scheduler db


Mo-Fatah:
Notes:
- Setting timestampes through a call that takes a list of `(run_id, timestamp)` pairs for performance

PR #3046, opened 2023-10-13
Update schedulerdb with runs timestamps

Mo-Fatah:
#3045


internal/scheduler/database/sql.yaml
```diff
@@ -11,4 +11,9 @@ sql:
...
         emit_prepared_queries: true
         emit_db_tags: true
-        emit_interface: false
\ No newline at end of file
+        emit_interface: false
+        overrides:
+          - db_type: "uuid"
+            go_type: "github.com/google/uuid.UUID"
+          - db_type: "timestamptz"
+            go_type: "time.Time"
```
Mo-Fatah, line 19: The new `sqlc` versions replaces some types with new types from the `pgx/v5/pgtype` package, some of them aren't interoperable with the codebase, specifically:
- `sqlc` now uses `pgtype.UUID` instead of `"github.com/google/uuid.UUID"`, and this breaks a lot of the codebase
- `sqlc` now uses `pgtype.Timestamptz` instead of "time.Time", this breaks the code in multiple places.

internal/scheduler/database/models.go
```diff
@@ -56,18 +57,22 @@ type Queue struct {
...
+	Succeeded           bool               `db:"succeeded"`
+	Failed              bool               `db:"failed"`
+	Returned            bool               `db:"returned"`
+	RunAttempted        bool               `db:"run_attempted"`
+	Serial              int64              `db:"serial"`
+	LastModified        time.Time          `db:"last_modified"`
+	LeasedTimestamp     pgtype.Timestamptz `db:"leased_timestamp"`
+	PendingTimestamp    pgtype.Timestamptz `db:"pending_timestamp"`
+	RunningTimestamp    pgtype.Timestamptz `db:"running_timestamp"`
+	TerminatedTimestamp pgtype.Timestamptz `db:"terminated_timestamp"`
```
Mo-Fatah, line 77: the new fields here uses `pgytype.Timestamptz`. I tried to force switch it to `time.Time` so we become consistent with previous fields, but it appears that `sqlc` uses `pgtype.Timestamptz` for nullable time types.

severinson: Is it not possible to override the default types? I remember seeing config for that n the past.

severinson: https://docs.sqlc.dev/en/latest/howto/overrides.html

Mo-Fatah: yeah, I actually did the override here https://github.com/armadaproject/armada/pull/3046/files/f72838670aaf0eaadea7f079e451f53fb10211a7#diff-5d7662fd33b2fd535f7090bf737e198d36400a8597c740e8380b88ed9f4747f8R18-R19
but it didn't seem to have effect. trying something hope it works

Mo-Fatah: yeah it worked
```yaml
          - db_type: "timestamptz"
            go_type: "time.Time"
            nullable: true
          - db_type: "timestamptz"
            go_type: "time.Time"
```
one for nullable timestamps and other for the non-nullable ones

internal/scheduler/database/query/query.sql
```diff
@@ -97,3 +97,16 @@ SELECT executor_id, last_updated FROM executors;
 INSERT INTO executors (executor_id, last_request, last_updated)
 VALUES(sqlc.arg(executor_id)::text, sqlc.arg(last_request)::bytea, sqlc.arg(update_time)::timestamptz)
 ON CONFLICT (executor_id) DO UPDATE SET (last_request, last_updated) = (excluded.last_request,excluded.last_updated);
+
+-- name: UpdateLeasedTime :exec
```
severinson, line 101: I'd call these `SetX` instead of `UpdateX`.

severinson: Actually, these queries don't make sense? We don't need a query that sets the same timestamp to multiple different runs. Instead, create a query that sets a specific timestamp for a specific run.

internal/scheduler/database/sql.yaml
severinson: This looks nice

internal/scheduler/database/query/query.sql
Mo-Fatah: > Actually, these queries don't make sense? We don't need a query that sets the same timestamp to multiple different runs. Instead, create a query that sets a specific timestamp for a specific run.

Yep, you are right. It doesn't make sense that way

severinson (approved):

Sharpz7 (approved):

PR #3047, opened 2023-10-16
Fix go-integration-tests in Github CI workflow.

richscott:
In `docker-compose.yaml`, remove Armada `server` as dependency for `executor` startup, and have the former wait 30 seconds before starting, so the Executor should be running and healthy before the server expects it.  Also save docker-compose logs and Pulsar container logs to GH artifacts for debugging purposes.


Sharpz7:
Nice Job! We probably don't need all the runners to be large, but lets just get this merged.

dejanzele (approved):

PR #3052, opened 2023-10-19
Bring back `/job-sets`

zuqq:
A user has pointed out that version 2 of the Lookout UI currently struggles to cancel large job sets (containing hundreds of thousands to millions of steps) through the "cancel selected" dialog, because it always calls `/v1/job/cancel` (instead of calling `/v1/jobset/cancel` if the user is cancelling entire job sets).

While we will ultimately want the "cancel selected" dialog to be smarter about this, I am proposing the following change as a way of unblocking the move to version 2 of the Lookout UI: keep the "job counts" variant of `/job-sets` alive,[^1] but populate it using version 2 of the Lookout API. The advantage of making this change is that it unblocks the rest of the migration while only requiring us to reroute a single API call.

[^1]: `/job-sets` previously also had "runtime" and "queued time" variants, but these require data that is not present in version 2 of the Lookout API.

### DIFF
--- a/internal/lookout/ui/src/components/NavBar.tsx
+++ b/internal/lookout/ui/src/components/NavBar.tsx
@@ -1,14 +1,54 @@
 import React from "react"
 
-import { AppBar, Toolbar, Typography } from "@material-ui/core"
+import { AppBar, Tab, Tabs, Toolbar, Typography } from "@material-ui/core"
+import { Link } from "react-router-dom"
+
+import { Router, withRouter } from "../utils"
 
 import "./NavBar.css"
 
-interface NavBarProps {
-  customTitle: string
+interface Page {
+  title: string
+  location: string
 }
 
-export default function NavBar({ customTitle }: NavBarProps) {
+const PAGES: Page[] = [
+  {
+    title: "Jobs",
+    location: "/",
+  },
+  {
+    title: "Job Sets",
+    location: "/job-sets",
+  },
+]
+
+// Creates mapping from location to index of element in ordered navbar
+function getLocationMap(pages: Page[]): Map<string, number> {
+  const locationMap = new Map<string, number>()
+  pages.forEach((page, index) => {
+    locationMap.set(page.location, index)
+  })
+  return locationMap
+}
+
+const locationMap = getLocationMap(PAGES)
+
+function locationFromIndex(pages: Page[], index: number): string {
+  if (pages[index]) {
+    return pages[index].location
+  }
+  return "/"
+}
+
+interface NavBarProps {
+  customTitle: string
+  router: Router
+}
+
+function NavBar({ customTitle, router }: NavBarProps) {
+  const currentLocation = router.location.pathname
+  const currentValue = locationMap.has(currentLocation) ? locationMap.get(currentLocation) : 0
   return (
     <AppBar position="static">
       <Toolbar className="toolbar">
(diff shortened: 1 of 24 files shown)


internal/lookout/ui/src/components/NavBar.tsx
```diff
@@ -1,14 +1,54 @@
 import React from "react"
```
zuqq, line 1: The content of this file comes from the parent of 3323ad7f74b7e7a6a07c29e2698736f0ec38c93c; the resulting navigation bar looks like this:

image

internal/lookout/ui/src/components/job-sets/JobSetTable.tsx
```diff
@@ -18,21 +17,16 @@ interface JobSetTableProps {
...
 }
 
-function cellRendererForState(
-  cellProps: TableCellProps,
-  onJobSetClick: (jobSet: string, state: string) => void,
-  state: string,
-) {
+function cellRendererForState(cellProps: TableCellProps) {
   if (cellProps.cellData) {
-    return <LinkCell onClick={() => onJobSetClick((cellProps.rowData as JobSet).jobSetId, state)} {...cellProps} />
```
zuqq, line 35: The numbers in the table used to link to `/jobs`; these links are now gone.

internal/lookout/ui/src/containers/JobSetsContainer.tsx
```diff
@@ -266,8 +245,44 @@ class JobSetsContainer extends React.Component<JobSetsContainerProps, JobSetsCon
...
+        queue: getJobSetsRequest.queue,
+        jobsQueued: state[JobState.Queued] || 0,
+        jobsPending: state[JobState.Pending] || 0,
+        jobsRunning: state[JobState.Running] || 0,
+        jobsSucceeded: state[JobState.Succeeded] || 0,
+        jobsFailed: state[JobState.Failed] || 0,
+        jobsCancelled: state[JobState.Cancelled] || 0,
+        latestSubmissionTime: group.aggregates.submitted as string,
+      }
+    })
```
zuqq, line 285: This is the real change: we now retrieve the status counts from version 2 of the Lookout API.

internal/lookout/ui/src/services/lookoutV2/UpdateJobSetsService.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export interface ReprioritizeJobSetsResponse {
+  reprioritizedJobSets: JobSet[]
+  failedJobSetReprioritizations: {
+    jobSet: JobSet
+    error: string
+  }[]
+}
+
+export class UpdateJobSetsService {
```
zuqq, line 21: The job set API calls now have a separate type, so that we no longer have to pass a `JobService` (which can also talk to version 1 of the Lookout API) to `JobSetsContainer`; this makes it clearer that this component no longer talks to version 1 of the Lookout API (and we will need this type anyway once the "cancel selected" dialog in version 2 of the Lookout UI learns to cancel job sets, in analogy with the existing `UpdateJobsServce` type that it already uses). The code below is from `JobService`, without any changes.

internal/lookoutv2/repository/querybuilder.go
```diff
@@ -989,6 +989,14 @@ func (qb *QueryBuilder) getQueryColumn(col string, queryTables map[string]bool)
...
 
 func limitOffsetSql(skip, take int) string {
+	// Asking for zero rows is not useful to us, so we take a value of zero to
+	// mean "no limit"; this is consistent with go-swagger, which uses zero as
+	// the default value for optional integers:
+	//
+	//     https://github.com/go-swagger/go-swagger/issues/1707
+	if take == 0 {
+		return fmt.Sprintf("OFFSET %d", skip)
+	}
```
zuqq, line 999: This change made it easier to swap out the implementation of `fetchJobSets` above, because `/job-sets` doesn't use pagination (and adding pagination to the API call it makes seems tricky).

severinson (approved):

PR #3054, opened 2023-10-19
Ingest jobruns timestamps

### LINKED ISSUE #3050: Ingest job runs phases timestampes to the schedulerDb through the scheduleringester
This should be after #3045 is merged

Mo-Fatah:
Fixes #3050


internal/scheduler/database/query/query.sql
```diff
@@ -53,7 +53,7 @@ UPDATE runs SET run_attempted = true WHERE run_id = ANY(sqlc.arg(run_ids)::UUID[
 UPDATE runs SET running = true WHERE run_id = ANY(sqlc.arg(run_ids)::UUID[]);
 
 -- name: MarkRunsCancelledByJobId :exec
-UPDATE runs SET cancelled = true WHERE job_id = ANY(sqlc.arg(job_ids)::text[]);
+UPDATE runs SET cancelled = true, terminated_timestamp = $1 WHERE job_id = ANY(sqlc.arg(job_ids)::text[]);
```
Mo-Fatah, line 56: I believe we can generalize this to all job runs marking functions ( MarkRunsFailed, MarkRunsSucceeded....). We don't a dedicated `SetX` function to update the timestamp for a jobrun, we can just embed the timestamp update into the status update like this one.

internal/scheduleringester/dbops.go
```diff
@@ -119,15 +122,15 @@ type (
...
 	MarkJobsSucceeded          map[string]bool
 	MarkJobsFailed             map[string]bool
 	UpdateJobPriorities        map[string]int64
 	UpdateJobSchedulingInfo    map[string]*JobSchedulingInfoUpdate
 	UpdateJobQueuedState       map[string]*JobQueuedStateUpdate
-	MarkRunsSucceeded          map[uuid.UUID]bool
+	MarkRunsSucceeded          map[uuid.UUID]time.Time
 	MarkRunsFailed             map[uuid.UUID]*JobRunFailed
-	MarkRunsRunning            map[uuid.UUID]bool
+	MarkRunsRunning            map[uuid.UUID]time.Time
```
Mo-Fatah, line 133: Not Sure what the `bool` was being used for? I didn't find it being used anywhere. This why I replaced the timestamp value so we can consume it in the schedulerdb

internal/scheduler/database/query/query.sql
severinson: Looks good to me.

internal/scheduleringester/dbops.go
severinson: I originally wanted to use a set, but there isn't a set datatype in go so I used map[string]bool. Using time instead makes sense for this change.

internal/scheduleringester/schedulerdb.go
```diff
@@ -218,6 +224,16 @@ func (s *SchedulerDb) WriteDbOp(ctx *armadacontext.Context, tx pgx.Tx, op DbOper
...
 		}
+		for runId, succeededTime := range o {
+			params := schedulerdb.SetTerminatedTimeParams{
+				RunID:               runId,
+				TerminatedTimestamp: &succeededTime,
+			}
+			err = queries.SetTerminatedTime(ctx, params)
+			if err != nil {
+				return errors.WithStack(err)
+			}
```
severinson, line 235:
```suggestion
			if err := queries.SetTerminatedTime(ctx, params); err != nil {
				return errors.WithStack(err)
			}
```

severinson: This is a bit neater imo. We use both versions in the code tho.

internal/scheduleringester/schedulerdb.go
```diff
@@ -242,12 +258,32 @@ func (s *SchedulerDb) WriteDbOp(ctx *armadacontext.Context, tx pgx.Tx, op DbOper
...
+				TerminatedTimestamp: &failInfo.FailureTime,
+			}
+			err = queries.SetTerminatedTime(ctx, params)
+			if err != nil {
+				return errors.WithStack(err)
+			}
+		}
 	case MarkRunsRunning:
 		runIds := maps.Keys(o)
 		err := queries.MarkJobRunsRunningById(ctx, runIds)
```
severinson, line 273: Upserting many rows in one call is much more efficient than making one call per job. There should be some way to create a function
```
queries.MarkJobRunsRunningById(ctx, runIds, timestamps)
```
Or
```
queries.MarkJobRunsRunningById(ctx, map[runIds]timestamps)
```
Of these, the first is probably easier to write in sql.

Mo-Fatah: I made it to be `queries.MarkJobRunsRunningById(ctx, runIds, timestamp)` , the function will receive a single timestamp object, couldn't completely figure out how to make sqlc update multiple rows with an array of values. However, the `JobRunsRunning` instruction will have a single `timestamp` value, not multiple times, so we can update the job status with this single value provided ( The way it is after my last commit)

internal/scheduleringester/instructions.go
```diff
@@ -177,7 +177,7 @@ func (c *InstructionConverter) handleSubmitJob(job *armadaevents.SubmitJob, subm
 	}}}, nil
 }
 
-func (c *InstructionConverter) handleJobRunLeased(jobRunLeased *armadaevents.JobRunLeased, meta eventSequenceCommon) ([]DbOperation, error) {
+func (c *InstructionConverter) handleJobRunLeased(jobRunLeased *armadaevents.JobRunLeased, leaseTime time.Time, meta eventSequenceCommon) ([]DbOperation, error) {
```
severinson, line 180: Consider adding the leaseTime to the JobRunLeased event instead.

internal/common/ingest/testfixtures/event.go
```diff
@@ -18,16 +16,21 @@ import (
 // Standard Set of events for common tests
 const (
 	JobIdString                  = "01f3j0g1md4qx7z5qb148qnh4r"
+	JobIdString2                 = "01f3j0g1md4qx7z5qb148qnh5r"
```
severinson, line 19: Duplicating all these is probably not the best approach. Have a look at what we do, e.g., for the GangScheduler tests. A similar constructor approach would probably be nicer here.

internal/scheduleringester/schedulerdb.go
```diff
@@ -230,22 +262,27 @@ func (s *SchedulerDb) WriteDbOp(ctx *armadacontext.Context, tx pgx.Tx, op DbOper
...
 			return errors.WithStack(err)
 		}
 	case MarkRunsRunning:
 		runIds := maps.Keys(o)
-		err := queries.MarkJobRunsRunningById(ctx, runIds)
-		if err != nil {
+		updateVals := make([]string, 0, len(runIds))
+		runningTimes := make([]interface{}, 0, len(runIds))
+		for i, runId := range runIds {
+			updateVals = append(updateVals, fmt.Sprintf("('%v'::uuid, true, $%d::timestamptz)", runId, i+1))
```
severinson, line 281: Break this out into separate functions. Where possible, strings should be constants defined at the top of the file instead of created dynamically.

internal/scheduleringester/schedulerdb.go
```diff
@@ -289,3 +317,17 @@ func execBatch(ctx *armadacontext.Context, tx pgx.Tx, batch *pgx.Batch) error {
...
+		idPgType = "text"
+	}
+	return fmt.Sprintf(`update runs set
+	%[2]v = runs_temp.%[2]v,
+	%[3]v = runs_temp.%[3]v
+	from (select * from unnest($1::%[4]v[], $2::boolean[] ,$3::timestamptz[])) 
+	as runs_temp(%[1]v, %[2]v, %[3]v) 
+	where runs.%[1]v = runs_temp.%[1]v;`,
+		id, phaseColumn, timeStampColumn, idPgType)
+}
```
Mo-Fatah, line 333: `unnest` seems more suitable here, much more performant and cleaner ( since we directly pass the array of vals into pg instead of destructing it on the `tx.Exec()` function)

severinson (approved):

richscott (approved):

PR #3061, opened 2023-10-23
Actually test `MaxUnacknowledgedJobsPerExecutor`

zuqq:
The existing test was dubious; it had no connection to `MaxUnacknowledgedJobsPerExecutor` at all.

### DIFF
--- a/internal/scheduler/scheduling_algo_test.go
+++ b/internal/scheduler/scheduling_algo_test.go
@@ -70,24 +70,41 @@ func TestSchedule(t *testing.T) {
 			queuedJobs:               testfixtures.N16Cpu128GiJobs(testfixtures.TestQueue, testfixtures.PriorityClass3, 10),
 			expectedScheduledIndices: []int{0, 1},
 		},
-		"do not schedule onto executors with too many unacknowledged jobs": {
-			// TODO: This test doesn't look right; we never set MaxUnacknowledgedJobsPerExecutor.
-			schedulingConfig: testfixtures.TestSchedulingConfig(),
+		"schedule onto executors with some unacknowledged jobs": {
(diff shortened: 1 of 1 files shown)


internal/scheduler/scheduling_algo_test.go
```diff
@@ -70,24 +70,41 @@ func TestSchedule(t *testing.T) {
 			queuedJobs:               testfixtures.N16Cpu128GiJobs(testfixtures.TestQueue, testfixtures.PriorityClass3, 10),
 			expectedScheduledIndices: []int{0, 1},
 		},
+		"schedule onto executors with some unacknowledged jobs": {
+			schedulingConfig: testfixtures.WithMaxUnacknowledgedJobsPerExecutorConfig(16, testfixtures.TestSchedulingConfig()),
```
zuqq, line 74: If the executor has 16 unacknowledged jobs and `MaxUnacknowledgedJobsPerExecutor` is set to 16, then we give it more jobs.

internal/scheduler/scheduling_algo_test.go
```diff
@@ -70,24 +70,41 @@ func TestSchedule(t *testing.T) {
...
+						acknowledged: false,
+					},
+				},
+			},
+			expectedScheduledIndices: testfixtures.IntRange(0, 47),
+		},
 		"do not schedule onto executors with too many unacknowledged jobs": {
-			// TODO: This test doesn't look right; we never set MaxUnacknowledgedJobsPerExecutor.
-			schedulingConfig: testfixtures.TestSchedulingConfig(),
+			schedulingConfig: testfixtures.WithMaxUnacknowledgedJobsPerExecutorConfig(15, testfixtures.TestSchedulingConfig()),
```
zuqq, line 92: If the executor has 16 unacknowledged jobs and `MaxUnacknowledgedJobsPerExecutor` is set to 15, then we do not give it any more jobs.

severinson (approved):

PR #3063, opened 2023-10-23
Pprof endpoints for all services

severinson:
This PR adds the option to expose pprof http profiling endpoints to all Armada services. Previously, this was only possible for the main Armada server.

Ideally we'd run the profiler http server in an errgroup everywhere. But that's a change we can make later.

### DIFF
--- a/internal/common/profiling/http.go
+++ b/internal/common/profiling/http.go
@@ -1,0 +1,24 @@
+package profiling
+
+import (
+	"fmt"
+	"net/http"
+	_ "net/http/pprof"
+)
+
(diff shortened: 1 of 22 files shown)


internal/common/profiling/http.go
```diff
@@ -0,0 +1,24 @@
+package profiling
+
+import (
+	"fmt"
+	"net/http"
+	_ "net/http/pprof"
```
severinson, line 6: Importing the internal/common/profiling package causes pprof endpoints to be registered.

d80tb7 (approved):

PR #3067, opened 2023-10-25
Delete version 1 of Lookout

zuqq:
Commit 061aa822d237331d09b3686f89bc1564aba8d2c4 is the part of #3038 that makes `lookoutv2` serve the Lookout UI; the subsequent commits delete all the code of version 1 of Lookout that is not shared with version 2 of Lookout.


Sharpz7:
Just to check, have you verified this works with `mage localdev full ui`, and that lookout becomes available on localhost?

@dejanzele this might also affect the operator

zuqq:
> Just to check, have you verified this works with `mage localdev full ui`, and that lookout becomes available on localhost?

It does work,[^1] because I am updating `docker-compose.yaml` and `magefiles/developer.go` as part of this PR. The Lookout UI is served on `localhost:1000`, together with version 2 of the Lookout UI.

[^1]: Strictly speaking I did encounter a problem with the `sed` version used by `mage buildpython`, but that is unrelated to this change; I'll create a separate PR.

Sharpz7 (approved):
This is great. Thank you :)

PR #3070, opened 2023-10-26
Explain that `/job-sets` requires a queue name

zuqq:
Consider this programmer art; it's just a sentence in the middle of the screen:

image


internal/lookout/ui/src/containers/JobSetsContainer.tsx
```diff
@@ -229,6 +229,9 @@ class JobSetsContainer extends React.Component<JobSetsContainerProps, JobSetsCon
   }
 
   private async loadJobSets() {
+    if (this.state.queue === "") {
+      return
+    }
```
zuqq, line 234: This is a drive-by change that prevents us from making requests if no queue name is given.

severinson (approved):

PR #3072, opened 2023-10-26
Fix loading of job table state from local storage

zuqq:
When the job table is mounted, it retrieves the user's job table settings from the query string and local storage, and then merges them by calling `mergeQueryParamsAndLocalStorage`:

https://github.com/armadaproject/armada/blob/0dc6f4e3ae323108fb71a6f6ed269a4c4b8c228b/internal/lookout/ui/src/services/lookoutV2/JobsTablePreferencesService.ts#L222-L224

If the user navigates to `/` (i.e., the query string is empty), then `mergeQueryParamsAndLocalStorage` is supposed to use the settings from local storage:

https://github.com/armadaproject/armada/blob/0dc6f4e3ae323108fb71a6f6ed269a4c4b8c228b/internal/lookout/ui/src/services/lookoutV2/JobsTablePreferencesService.ts#L170

This is currently not the case on `master`, because `fromQueryStringSafe` (which is called by `getPrefsFromQueryParams`) turns the empty object `{}` into `{ groupedColumns: [] }`:

https://github.com/armadaproject/armada/blob/0dc6f4e3ae323108fb71a6f6ed269a4c4b8c228b/internal/lookout/ui/src/services/lookoutV2/JobsTablePreferencesService.ts#L124-L129

This PR fixes this issue by not setting `g` to the empty array (which converts to `true` in boolean contexts) if it was originally `undefined`.

### DIFF
--- a/internal/lookout/ui/src/services/lookoutV2/JobsTablePreferencesService.ts
+++ b/internal/lookout/ui/src/services/lookoutV2/JobsTablePreferencesService.ts
@@ -62,7 +62,7 @@ type QueryStringJobFilter = {
 // Keys are shortened to keep URL size lower
 export interface QueryStringPrefs {
   // Grouped columns
-  g: string[] | [null]
+  g: string[]
   // Expanded rows
   e: string[]
   // Current page number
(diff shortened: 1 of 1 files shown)


internal/lookout/ui/src/services/lookoutV2/JobsTablePreferencesService.ts
```diff
@@ -62,7 +62,7 @@ type QueryStringJobFilter = {
 // Keys are shortened to keep URL size lower
 export interface QueryStringPrefs {
   // Grouped columns
-  g: string[] | [null]
+  g: string[]
```
zuqq, line 65: This is a drive-by change, because I noticed that we never produce `[null]` here anymore (see 85fdace5d616c83e42ef9853744b5c6bc92ba6c0).

d80tb7 (approved):

PR #3079, opened 2023-10-27
Update `build-python-client.sh` Docker image

zuqq:
I ran into the following error trying to run `mage buildpython` on my machine:

```console
sed: couldn't open temporary file proto/armada/sedgxb1ga: Permission denied
```

Indeed, none of the file's permission bits were set:

```console
$ stat --printf=%f proto/armada/sedgxb1ga 
8000
```

The internet tells me that this is an issue with versions of `sed` prior to 4.8 (see the "Bug fixes" section of https://savannah.gnu.org/forum/forum.php?forum_id=9647), and upgrading to a Debian version that ships a newer version of `sed` (see https://packages.debian.org/bookworm/sed) does indeed fix this issue on my machine.


zuqq:
The test failures look unrelated:

```console
E           grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
E           	status = StatusCode.UNKNOWN
E           	details = "[x-request-id: "SCAfkYvNRr5w5XPvuqv23N"] job 01hdsj4cntd4gxfe26x4v4h9nb unschedulable: failed to schedule onto legacy scheduler because 0-th job unschedulable:
E           no executor clusters available"
E           	debug_error_string = "UNKNOWN:Error received from peer  {created_time:"2023-10-27T21:55:02.971995066+00:00", grpc_status:2, grpc_message:"[x-request-id: \"SCAfkYvNRr5w5XPvuqv23N\"] job 01hdsj4cntd4gxfe26x4v4h9nb unschedulable: failed to schedule onto legacy scheduler because 0-th job unschedulable:\nno executor clusters available"}"
E           >
```

Sharpz7:
I think I know what could be causing that. I will investigate

zuqq:
The Python client workflows have been fixed on `master`, so this is ready to go in now.

d80tb7 (approved):

PR #3080, opened 2023-10-30
Fix PreemptingQueueScheduler benchmarks

severinson:


internal/scheduler/preempting_queue_scheduler_test.go
```diff
@@ -1714,6 +1715,9 @@ func BenchmarkPreemptingQueueScheduler(b *testing.B) {
 	}
 	for name, tc := range tests {
 		b.Run(name, func(b *testing.B) {
+			ctx := armadacontext.Background()
+			ctx.FieldLogger = logging.NullLogger
```
d80tb7, line 1719: Are we sure this is what we want?  If there's too much logging, could we just make a logger with loglevel set to warn?  That way we won't lose error messages.

d80tb7 (approved):

internal/scheduler/preempting_queue_scheduler_test.go
severinson: Not writing to stdout would be my preference for benchmarks.

PR #3093, opened 2023-11-09
Ensure consistent job order when re-scheduling

severinson:
Previously, evicted jobs were not guaranteed to be ordered consistently when re-scheduling. This may prevent jobs from preempted correctly.


internal/scheduler/jobdb/job.go
```diff
@@ -77,6 +80,82 @@ func (job *Job) ensureJobSchedulingInfoFieldsInitialised() {
 	}
 }
 
+// Equal returns true if job is equal to other and false otherwise.
+// Scheduling requirements are assumed to be equal if both jobs have equal schedulingKey.
+func (job *Job) Equal(other *Job) bool {
```
severinson, line 85: We need this for testing. Since testify.Assert can't compare unexported fields.

internal/scheduler/jobdb/comparison.go
```diff
@@ -0,0 +1,131 @@
...
+	return Compare(job, other.(*Job))
+}
+
+// Compare defines the order in which jobs in a particular queue should be scheduled,
+// both when scheduling new jobs and when re-scheduling evicted jobs.
+// Specifically, compare returns
+//   - 0 if the jobs have equal job id,
+//   - -1 if job should be scheduled before other,
+//   - +1 if other should be scheduled before other.
+func Compare(job, other *Job) int {
```
d80tb7, line 73: rename

d80tb7 (dismissed):

d80tb7 (approved):

PR #3107, opened 2023-11-14
Updated scheduler metrics

severinson:


internal/common/math/math.go
```diff
@@ -0,0 +1,12 @@
+package math
+
+import "golang.org/x/exp/constraints"
+
+// Max returns a if a > b and b otherwise.
```
severinson, line 5: I needed a max function. We may as well have a generic one.

internal/scheduler/jobdb/reconciliation.go
```diff
@@ -0,0 +1,248 @@
+package jobdb
```
severinson, line 1: This file contains the reconciliation logic previously in `syncState`. The long-term goal is to move all reconciliation logic into one place and then return armadaevents instead of `JobStateTransitions` as we do now.

JamesMurkin (approved):

PR #3117, opened 2023-11-21
[CI-IMPROVEMENT-1] Rename confusing step names, replace setup-go actions with official version

pavlovic-ivan:
#### What type of PR is this?
This is an improvement PR

First one in a series of PRs for improvements. This one just replaces unofficial setup GO actions with official ones. The unofficial action is written as a composite action and it does caching out-of-the-box. Replacing those actions with official ones, there is no caching, which will enable us to pin point where and how to use cache


pavlovic-ivan:
Something is off with my PR. Closing and investigating

PR #3121, opened 2023-11-22
Fix "best fit" computation for variable-sized gangs

zuqq:
This PR makes the "best fit" computation in `GangScheduler`'s `trySchedule` method compatible with variable-sized gangs (i.e., gangs that set the annotation `armadaproject.io/gangMinimumCardinality`).

### DIFF
--- a/internal/scheduler/context/context.go
+++ b/internal/scheduler/context/context.go
@@ -240,8 +240,7 @@ func (sctx *SchedulingContext) AddGangSchedulingContext(gctx *GangSchedulingCont
 			return false, err
 		}
 		allJobsEvictedInThisRound = allJobsEvictedInThisRound && evictedInThisRound
-		isSuccess := jctx.IsSuccessful()
-		if isSuccess {
+		if jctx.IsSuccessful() {
 			numberOfSuccessfulJobs++
 		}
 	}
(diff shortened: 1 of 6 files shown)


internal/scheduler/context/context.go
```diff
@@ -585,6 +584,36 @@ func (gctx *GangSchedulingContext) Cardinality() int {
...
 
+type GangSchedulingFit struct {
+	// The number of jobs in the gang that were successfully scheduled.
+	NumScheduled int
+	// The mean PreemptedAtPriority among successfully scheduled pods in the gang.
+	MeanPreemptedAtPriority float64
+}
+
+func (f GangSchedulingFit) Less(other GangSchedulingFit) bool {
+	return f.NumScheduled < other.NumScheduled || f.NumScheduled == other.NumScheduled && f.MeanPreemptedAtPriority > other.MeanPreemptedAtPriority
```
zuqq, line 595: I have not done a rigorous analysis of this way of measing "best fit"; we may well want to switch to something else later.

internal/scheduler/context/context.go
```diff
@@ -718,9 +754,8 @@ type PodSchedulingContext struct {
 	NodeId string
 	// Score indicates how well the pod fits on the selected node.
 	Score int
-	// Priority class priority at which this pod was scheduled.
-	// Only set if NodeId is.
-	ScheduledAtPriority int32
+	// Maximum priority that this pod preempted other pods at.
+	PreemptedAtPriority int32
```
zuqq, line 758: I'm not too fond of this name, because it's not immediately clear (at least without sufficient context) whether this means that this pod was preempted or that it preempted other pods, but I couldn't think of anything better.

internal/scheduler/gang_scheduler.go
```diff
@@ -216,20 +214,9 @@ func (sch *GangScheduler) tryScheduleGang(ctx *armadacontext.Context, gctx *sche
...
-		// Clear any node bindings on failure to schedule.
-		jctx.PodSchedulingContext.NodeId = ""
-	}
-}
-
 func (sch *GangScheduler) tryScheduleGangWithTxn(_ *armadacontext.Context, txn *memdb.Txn, gctx *schedulercontext.GangSchedulingContext) (ok bool, unschedulableReason string, err error) {
 	if ok, err = sch.nodeDb.ScheduleManyWithTxn(txn, gctx.JobSchedulingContexts); err == nil {
 		if !ok {
-			for _, jctx := range gctx.JobSchedulingContexts {
-				clearNodeBindings(jctx)
```
zuqq, line 230: This now happens further up in the call tree, when we set `UnschedulableReason` for these jobs; I made this change because I prefer mutating both fields in tandem.

internal/scheduler/nodedb/nodedb.go
```diff
@@ -516,7 +516,10 @@ func (nodeDb *NodeDb) ScheduleManyWithTxn(txn *memdb.Txn, jctxs []*schedulercont
 	// Attempt to schedule pods one by one in a transaction.
 	numScheduled := 0
 	for _, jctx := range jctxs {
-		// Defensively reset `ShouldFail` (this should always be false as the state is re-constructed per cycle but just in case)
+		// In general, we may attempt to schedule a gang multiple times (in
+		// order to find the best fit for this gang); clear out any remnants of
+		// previous attempts.
+		jctx.UnschedulableReason = ""
```
zuqq, line 522: I observed dirty `UnschedulableReason` fields in the tests I added (this issue is specific to variable-sized gangs, where the gang scheduling outcome may contain a mix of successful and unsuccessful job scheduling outcomes).

severinson (approved):

PR #3125, opened 2023-11-27
Use priority class name instead of priority in scheduling key

zuqq:
This is another change in preparation for home-away preemption: once we've added home-away preemption, different priority classes with the same "home" priority may nevertheless be different from the scheduler's point of view (because they defined different sets of "away" priorities). Note that this change will reduce the cache hit rate on `master`, because preemptible and non-preemptible priority classes of the same priority now produce different scheduling keys.


internal/scheduler/schedulerobjects/schedulerobjects.pb.go
```diff
@@ -842,10 +842,6 @@ type PodRequirements struct {
 	PreemptionPolicy string `protobuf:"bytes,5,opt,name=preemptionPolicy,proto3" json:"preemptionPolicy,omitempty"`
 	// Sum of the resource requirements for all containers that make up this pod.
 	ResourceRequirements v1.ResourceRequirements `protobuf:"bytes,6,opt,name=resourceRequirements,proto3" json:"resourceRequirements"`
-	// This key is a hash of the scheduling requirements.
-	// The hash is 20 bytes in size and is guaranteed to be the same for equivalent PodRequirements,
-	// unless Affinity is used, in which case they may differ as a result of unordered map keys.
-	CachedSchedulingKey []byte `protobuf:"bytes,8,opt,name=cachedSchedulingKey,proto3" json:"cachedSchedulingKey,omitempty"`
```
zuqq, line 848: The last references to this field were deleted in 7542cd0c197f1a8a75dca3eec5566a0f3c2b00f4.

severinson (approved):

PR #3140, opened 2023-12-01
Add additional annotations to JobRunLeased event

msumner91:
Goal(s):
- Add AdditionalAnnotations to JobRunLeased events
- Expose gang_runtime_cardinality as an AdditionalAnnotation on JobRunLeased

A follow-up PR will:
- Consume these values & persist them into PG
- Add them as annotations on the resulting PodSpec


pkg/api/api.swagger.go
```diff
@@ -9,7 +9,7 @@ package api
 // SwaggerJsonTemplate is a generated function returning the template as a string.
 // That string should be parsed by the functions of the golang's template package.
 func SwaggerJsonTemplate() string {
-	var tmpl = "{\n" +
+	tmpl := "{\n" +
```
zuqq, line 12: You'll have to revert this change (and the analogous one below), or you'll break the tests; see #3084.

pkg/armadaevents/events.proto
```diff
@@ -290,6 +290,8 @@ message JobRunLeased {
     string node_id = 4;
     // Used by the scheduler to maintain a consistent state
     int32 update_sequence_number = 5;
+    // Additional labels to be added to the PodSpec.
+    map<string, string> additional_labels = 6;
```
zuqq, line 294: Do we really want to use labels instead of annotations?

msumner91: I don't have a preference really - if an annotation makes more sense happy to use that

internal/scheduler/gang_scheduler.go
```diff
@@ -204,18 +204,18 @@ func (sch *GangScheduler) trySchedule(ctx *armadacontext.Context, gctx *schedule
 	return sch.tryScheduleGang(ctx, gctx)
 }
 
-func (sch *GangScheduler) tryScheduleGang(ctx *armadacontext.Context, gctx *schedulercontext.GangSchedulingContext) (ok bool, unschedulableReason string, err error) {
+func (sch *GangScheduler) tryScheduleGang(ctx *armadacontext.Context, gctx *schedulercontext.GangSchedulingContext) (ok bool, runtimeGangCardinality int, unschedulableReason string, err error) {
```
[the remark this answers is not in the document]
msumner91: Ah possibly - that didn't exist when I last looked at this stuff.

Will take a look and see

pkg/armadaevents/events.proto
msumner91: The only reason we went for a label is is because we currently use those for similar things like job id / jobset / queue id.

internal/scheduler/gang_scheduler.go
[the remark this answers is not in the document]
msumner91: Changed this to use GangFit instead as it's cleaner to compute from the context -- thanks for the heads up

pkg/armadaevents/events.proto
zuqq: Looking at https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#attaching-metadata-to-objects, using an annotation might be more appropriate. Do we currently set any gang scheduling metadata on gang jobs (e.g., gang ID would arguably fall into the label category)?

msumner91: Cool - I'll rename this to annotations then as it's not really something we want to select/identify with.

DogeFarm sets GangId etc. as metadata in DogeFarm and as an annotation to Armada

E.g. Yaml spec:
annotations:
      armadaproject.io/gangCardinality: '4'
      armadaproject.io/gangId: 28cf33a61192490a87974b0ad6eeb698-7

JamesMurkin (approved):

PR #3141, opened 2023-12-01
Use the Pulsar scheduler in end-to-end tests

zuqq:
I've added a "-legacy" suffix to the non-Pulsar objects, in order to prevent us from accidentally using them.

### DIFF
--- a/.github/workflows/python-client.yml
+++ b/.github/workflows/python-client.yml
@@ -83,7 +83,7 @@ jobs:
           # Manually create folders to ensure perms are correct.
           mkdir -p .kube/internal
           mkdir -p .kube/external
-          go run github.com/magefile/mage@v1.14.0 -v localdev minimal-legacy
+          go run github.com/magefile/mage@v1.14.0 -v localdev minimal-pulsar
       - name: Install Protoc
         uses: arduino/setup-protoc@v2
         with:


.github/workflows/airflow-operator.yml
```diff
@@ -100,7 +100,7 @@ jobs:
           # Manually create folders to ensure perms are correct.
           mkdir -p .kube/internal
           mkdir -p .kube/external
-          go run github.com/magefile/mage@v1.14.0 -v localdev minimal
+          go run github.com/magefile/mage@v1.14.0 -v localdev minimal-legacy
```
zuqq, line 103: I'm not familiar with this part of Armada, so I've refrained from moving it to the Pulsar scheduler.

docker-compose.yaml
```diff
@@ -77,15 +77,15 @@ services:
...
+      - "./developer/config/insecure-armada.yaml:/config/insecure-armada.yaml"
       - "go-cache:/root/.cache/go-build:rw"
       - "gomod-cache:/go/pkg/mod:rw"
     depends_on:
       - eventingester
     working_dir: /app
     env_file:
       - ./developer/env/server-pulsar.env
-    command: ./server --config /config/insecure-armada.yaml
+    command: sh -c "sleep 30 && ./server --config /config/insecure-armada.yaml"
```
zuqq, line 88: We can surely find a better way of doing this, but for now I've just made this command mirror the one in `server-legacy` (the `sleep 30` part was added in #3047).

docker-compose.yaml
```diff
@@ -97,18 +97,18 @@ services:
...
-      - postgresPulsarMigration
+      - scheduleringester
     working_dir: /app
     env_file:
       - ./developer/env/scheduler.env
     command: ./scheduler run --config /config/insecure-armada.yaml
 
-  postgresPulsarMigration:
-    container_name: postgresPulsarMigration
+  scheduler-migration:
```
zuqq, line 110: This is now consistent with the container name in the Helm package.

docker-compose.yaml
```diff
@@ -160,10 +162,8 @@ services:
     ports:
       - 9001:9001
       - 4001:4000
-    depends_on:
-      - server-pulsar
```
zuqq, line 164: As above, I've removed this because it was removed from `executor-legacy` in #3047.

docker-compose.yaml
```diff
@@ -254,7 +251,7 @@ services:
       - 60003:60003
       - 4008:4000
     depends_on:
-      - server
+      - server-legacy
```
zuqq, line 254: I'm not familiar with `jobservice`, so I refrained from changing this.

internal/scheduler/scheduler.go
```diff
@@ -936,7 +936,7 @@ func (s *Scheduler) ensureDbUpToDate(ctx *armadacontext.Context, pollInterval ti
 				ctx.Infof("Successfully ensured that database state is up to date")
 				return nil
 			}
-			ctx.Infof("Recevied %d partitions, still waiting on  %d", numReceived, numSent-numReceived)
+			ctx.Infof("Received %d partitions, still waiting on %d", numReceived, numSent-numReceived)
```
zuqq, line 939: Unrelated fix, because I happened to see this log message.

magefiles/main.go
```diff
@@ -183,21 +183,19 @@ func LocalDev(arg string) error {
...
 
 	switch arg {
-	case "minimal":
-		os.Setenv("ARMADA_COMPONENTS", "executor,server")
+	case "minimal-legacy":
+		os.Setenv("ARMADA_COMPONENTS", "executor-legacy,server-legacy")
 		mg.Deps(StartComponents)
 	case "minimal-pulsar":
-		// This 20s sleep is to remedy an issue caused by pods coming up too fast after pulsar
-		// TODO: Deal with this internally somehow?
```
zuqq, line 199: `CheckForPulsarRunning` already deals with this.

testsuite/testcases/basic/failure_oom_1x1.yaml
```diff
@@ -22,9 +22,9 @@ jobs:
               memory: 10Mi
               cpu: 100m
 ---
-timeout: "150s"
```
zuqq, line 25: This test now takes just over two and a half minutes in GitHub Actions (but it's much faster when I run it in isolation on my machine), so I've increased this timeout.

PR #3184, opened 2023-12-08
GoLand Run Configurations

MustafaI:
Adds a number of GoLand run configurations, allowing developers to run and debug all Armada components from one IDE. 

Fixes #


Sharpz7:
It would be great if you could add some docs for this to in the developer guide and verify that `mage localdev full testsuite` still works and you can access Lookout and Binoculars

PR #3187, opened 2023-12-11
Add support for home-away preemption

zuqq:
# Summary

This PR adds initial support for "home-away" preemption, which is a way of reserving nodes for certain classes of jobs (e.g., "whale" jobs, which request an entire node at a time) while still allowing other jobs to be scheduled on these nodes at a lower priority. This allows cluster operators to boost workloads that are not well served by the scheduling algorithm without unnecessarily reducing cluster utilization.

# Implementation

Nodes are reserved for certain classes of jobs by tainting them, which means that jobs need to have the corresponding tolerations in order to be scheduled on these nodes. On `master`, tolerations can be set for individual jobs by populating the `Tolerations` field of `PodRequirements`, and for entire classes of job by populating these fields of `SchedulingConfig`:

https://github.com/armadaproject/armada/blob/7a10e03e6f805c5a9bd782df4375e372685a0999/internal/armada/configuration/types.go#L161-L166

This PR adds a third possibility: the `PriorityClass` type gains the field `AwayNodeTypes`, which represents the sets of nodes that jobs of this priority class can be scheduled on as "away" jobs (i.e., at a lower priority than "home" jobs that the nodes have been reserved for); each set of nodes ("node type") is defined by a set of taints.

When it encounters a job with such a priority class, the scheduler first tries to schedule the job as a home job (i.e., it schedules it like it normally would); if that fails, it then tries to schedule on each away node type described by `AwayNodeTypes`, adding the tolerations corresponding to the taints defining each of these nodes types in turn. Because away jobs are scheduled at a lower priority than home jobs, the scheduler will prefer to schedule home jobs whenever it can; in particular, away jobs are preempted by home jobs through the existing mechanism of urgency-based preemption.


internal/scheduler/simulator/simulator_test.go
```diff
@@ -297,6 +299,149 @@ func TestSimulator(t *testing.T) {
...
+				armadaslices.Repeat(24, &armadaevents.EventSequence{Queue: "queue-1", JobSetName: "job-set-1", Events: []*armadaevents.EventSequence_Event{JobRunLeased()}}),
+				armadaslices.Repeat(1, &armadaevents.EventSequence{Queue: "queue-0", JobSetName: "job-set-0", Events: []*armadaevents.EventSequence_Event{JobSucceeded()}}),
+				[]*armadaevents.EventSequence{
+					{Queue: "queue-0", JobSetName: "job-set-0", Events: armadaslices.Repeat(4, SubmitJob())},
+				},
+				armadaslices.Repeat(8, &armadaevents.EventSequence{Queue: "queue-1", JobSetName: "job-set-1", Events: []*armadaevents.EventSequence_Event{JobRunLeased()}}),
+				armadaslices.Repeat(24, &armadaevents.EventSequence{Queue: "queue-1", JobSetName: "job-set-1", Events: []*armadaevents.EventSequence_Event{JobRunPreempted()}}),
+				armadaslices.Repeat(3, &armadaevents.EventSequence{Queue: "queue-0", JobSetName: "job-set-0", Events: []*armadaevents.EventSequence_Event{JobRunLeased()}}),
+				armadaslices.Repeat(24, &armadaevents.EventSequence{Queue: "queue-1", JobSetName: "job-set-1", Events: []*armadaevents.EventSequence_Event{SubmitJob()}}),
```
zuqq, line 433: This sequence of events is hard to read, but it shows an interesting scheduling phenomenon:

In this test there are four nodes, let's call them A<sub>0</sub>, A<sub>1</sub>, B<sub>0</sub>, and B<sub>1</sub>, where B<sub>0</sub> and B<sub>1</sub> have the taint `gpu-whale=true:NoSchedule`. At the start of this scheduling round the nodes are occupied by 32 jobs from `queue-0`. The rest of the scheduling round proceeds like this:

1. The scheduler evicts all of the jobs from `queue-0`, because they are preemptible.
2. The scheduler schedules seven evicted jobs from `queue-0` back on A<sub>0</sub>.
3. The scheduler schedules one job from `queue-1` on A<sub>1</sub>.
4. The scheduler schedules the remaining evicted job from `queue-0` that was already running on A<sub>0</sub>.
5. None of the jobs from `queue-0` that we already running on A<sub>1</sub> can be scheduled anymore, because a job from `queue-1` has already been scheduled there at the same priority. None of the away jobs from `queue-0` that are running on B<sub>0</sub> or B<sub>1</sub> can be scheduled anymore, because home jobs from `queue-1` are preferred instead.

The outcome of this scheduling round is that 24 out of 32 running jobs from `queue-0` are preempted, and three jobs from `queue-1` are scheduled in their place.

pkg/api/util.go
```diff
@@ -225,6 +225,10 @@ func (job *Job) GetPriorityClassName() string {
 	return podSpec.PriorityClassName
 }
 
+func (job *Job) GetScheduledAtPriority() (int32, bool) {
+	return -1, false
```
zuqq, line 229: I've avoided implementing home-away preemption for the old scheduler, because it's on its way out.

pkg/armadaevents/events.proto
```diff
@@ -290,6 +291,17 @@ message JobRunLeased {
     string node_id = 4;
     // Used by the scheduler to maintain a consistent state
     int32 update_sequence_number = 5;
+    // This bool is here because it is the most ergonomic way of distinguishing
+    // the case where `scheduled_at_priority` was explicitly set to zero from
+    // the case where it was not set at all.
```
zuqq, line 296: The alternatives are:

1. Use `Int32Value` instead of `int32` for `scheduled_at_priority`; this has the downside that switching between these types is a breaking change.
2. Wrap `scheduled_at_priority` in `oneof`; this has the downside that the generated code is ugly (and no safer than this approach; for example, `github.com/gogo/protobuf` still generates a `GetScheduledAtPriority` getter that returns an `int32`).
3. Use the `optional` keyword; this is not supported by `github.com/gogo/protobuf`.

All of these seem worse to me.

internal/scheduler/nodedb/nodedb.go
```diff
@@ -190,7 +197,15 @@ func (nodeDb *NodeDb) CreateAndInsertWithJobDbJobsWithTxn(txn *memdb.Txn, jobs [
 		return err
 	}
 	for _, job := range jobs {
-		if err := bindJobToNodeInPlace(nodeDb.priorityClasses, job, entry); err != nil {
+		priority, ok := job.GetScheduledAtPriority()
+		if !ok {
+			priorityClass, err := interfaces.PriorityClassFromLegacySchedulerJob(nodeDb.priorityClasses, job)
```
severinson, line 202: You're guaranteed to have a jobdb.Job here.

internal/scheduler/nodedb/nodedb.go
```diff
@@ -607,17 +631,76 @@ func (nodeDb *NodeDb) SelectNodeForJobWithTxn(txn *memdb.Txn, jctx *schedulercon
...
+	}
+
+	// Save the number of additional tolerations that the job originally had; we
+	// use this value to restore the slice of additional toleration at the end
+	// of each loop iteration.
+	numAdditionalTolerations := len(jctx.AdditionalTolerations)
+	for _, awayNodeType := range priorityClass.AwayNodeTypes {
+		wellKnownNodeType, ok := nodeDb.wellKnownNodeTypes[awayNodeType.WellKnownNodeTypeName]
+		if !ok {
+			return nil, fmt.Errorf("unknown well-known node type %s; must be in %v", awayNodeType.WellKnownNodeTypeName, nodeDb.wellKnownNodeTypes)
```
severinson, line 665: We could get jobs into the system with unknown well-known node types, e.g., if we remove a well-known node type after a job is submitted. Do we set the error as the failure reason for the job and then continue scheduling?

severinson: Do we validate the config to ensure that there are no unknown well-known node types referenced in a PC?

severinson: Skip this well-known node type if we can't find it and log a warning.

internal/scheduler/nodedb/nodedb.go
zuqq: I have a weak preference for using the `LegacySchedulerJob` interface here, since that's what's used in all other methods of this type.

dave-gantenbein:
Love the thorough description here! Can we consider adding some of this clarification to the scheduler section of the docs?

zuqq:
> Love the thorough description here! Can we consider adding some of this clarification to the scheduler section of the docs?

Noted; I have a couple of follow-up changes coming.

PR #3192, opened 2023-12-12
Add a metric for home-away preemption

zuqq:
This piggybacks on the updated scheduler metrics from #3107, by adding a `nodeType` label to the metric for scheduled jobs; initially, this label will only be set for away jobs.

In order to populate this label without going through the event stream, I'm adding the `WellKnownNodeTypeName` field to `PodSchedulingContext` and replacing `[]interfaces.LegacySchedulerJob` by `[]*armadacontext.JobSchedulingContext` everywhere in `SchedulerResult` (so that we can access the `PodSchedulingContext` of the scheduled jobs when updating the metrics).

### DIFF
--- a/internal/armada/server/lease.go
+++ b/internal/armada/server/lease.go
@@ -574,7 +574,8 @@ func (q *AggregatedQueueServer) getJobs(ctx *armadacontext.Context, req *api.Str
 
 	// Publish preempted + failed messages.
 	sequences := make([]*armadaevents.EventSequence, len(result.PreemptedJobs))
-	for i, job := range result.PreemptedJobs {
+	for i, jctx := range result.PreemptedJobs {
+		job := jctx.Job
 		jobId, err := armadaevents.ProtoUuidFromUlidString(job.GetId())
 		if err != nil {
 			return nil, err
@@ -619,7 +620,8 @@ func (q *AggregatedQueueServer) getJobs(ctx *armadacontext.Context, req *api.Str
 	}
 
 	preemptedApiJobsById := make(map[string]*api.Job)
-	for _, job := range result.PreemptedJobs {
+	for _, jctx := range result.PreemptedJobs {
+		job := jctx.Job
 		if apiJob, ok := job.(*api.Job); ok {
 			preemptedApiJobsById[job.GetId()] = apiJob
 		} else {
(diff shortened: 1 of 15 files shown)


internal/armada/server/lease.go
```diff
@@ -682,7 +685,7 @@ func (q *AggregatedQueueServer) getJobs(ctx *armadacontext.Context, req *api.Str
 	// Update resource cluster report to account for preempted/leased jobs and write it to Redis.
 	allocatedByQueueAndPriorityClassForCluster = updateAllocatedByQueueAndPriorityClass(
 		allocatedByQueueAndPriorityClassForCluster,
-		subtract, result.PreemptedJobs,
+		subtract, maps.Values(preemptedApiJobsById),
```
zuqq, line 688: This change incurs extra allocation.

internal/scheduler/scheduler.go
```diff
@@ -286,35 +286,21 @@ func (s *Scheduler) cycle(ctx *armadacontext.Context, updateAll bool, leaderToke
...
-			},
-			nil,
-		); err != nil {
+	for _, jctx := range overallSchedulerResult.PreemptedJobs {
+		if err := s.schedulerMetrics.UpdatePreempted(jctx); err != nil {
 			return err
 		}
 	}
-	for _, job := range overallSchedulerResult.FailedJobs {
+	for _, jctx := range overallSchedulerResult.FailedJobs {
```
zuqq, line 299: The "failed" case is also needed by `UpdateMany` (where we only have jobs, not job scheduling contexts), so it still goes through `Update` (unlike "scheduled" and "preempted").

internal/scheduler/scheduler_test.go
```diff
@@ -1226,23 +1227,11 @@ func NewSchedulerResultForTest[S ~[]T, T interfaces.LegacySchedulerJob](
...
-		castScheduledJobs[i] = job
-	}
-	castFailedJobs := make([]interfaces.LegacySchedulerJob, len(failedJobs))
-	for i, job := range failedJobs {
-		castFailedJobs[i] = job
-	}
 	return &SchedulerResult{
-		PreemptedJobs: castPreemptedJobs,
-		ScheduledJobs: castScheduledJobs,
+		PreemptedJobs: schedulercontext.JobSchedulingContextsFromJobs(testfixtures.TestPriorityClasses, preemptedJobs, GangIdAndCardinalityFromAnnotations),
```
zuqq, line 1231: I think that the last argument to `JobSchedulingContextsFromJobs` can be removed (it's always `GangIdAndCardinalityFromAnnotations`), but that's a separate topic.

severinson (approved):

PR #3197, opened 2023-12-14
Scheduler: handle run_preempted and run_pending events

### LINKED ISSUE #3174: Scheduler: Add `JobRunPreempted` and `JobRunAssigned` event handlers
Currently both `EventSequence_Event_JobRunPreempted` and `EventSequence_Event_JobRunAssigned` are being ignored in the scheduleringester: https://github.com/armadaproject/armada/blob/master/internal/scheduleringester/instructions.go#L110-L111

Since both events would be helpful in the metrics collection, we should handle both events and update the scheduler db to include them.

the changes should include:
- update the scheduler databse to add jobrun-preempted timestamp column, add assigned (or similar) boolean column and add preempted boolean column
- add the corresponding events handlers to the scheduler ingester
- update the unit tests

Mo-Fatah:
Fixes #3174 
This should be merge after #3054

### DIFF
--- a/internal/scheduler/database/migrations/004_add_preempted_pending.up.sql
+++ b/internal/scheduler/database/migrations/004_add_preempted_pending.up.sql
@@ -1,0 +1,3 @@
+ALTER TABLE runs ADD COLUMN preempted boolean NOT NULL DEFAULT false;
+-- Set to true once the run is assigned to an executor and waiting to start running.
+ALTER TABLE runs ADD COLUMN pending boolean NOT NULL DEFAULT false;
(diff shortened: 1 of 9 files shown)


internal/scheduler/database/migrations/004_add_preempted_pending.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE runs ADD COLUMN preempted boolean NOT NULL DEFAULT false;
+-- Set to true once the run is assigned to an executor and waiting to start running.
```
severinson, line 2:
```suggestion
-- Set to true once the run is sent to an executor, is assigned to a node, and is waiting to start running.
```

severinson (approved):

richscott (approved):

PR #3200, opened 2023-12-18
changes made to test.yml ci

### LINKED ISSUE #3102: Fail-fast in Test workflow
In `.github/workflows/test.yml`, since the `go-mod-up-to-date` job will run separately/concurrently with the `go-unit-tests` and `go-integration-tests` jobs, if there is an error in the go-mod-up-to-date job, it's probably not worthwhile to run the other two. We could speed up the failure case scenario by adding a `needs: go-mod-up-to-date` clause to the latter two jobs, this would speed up overall workflow processing when there's a module issue - see https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow for more info.

jaydee029:
Fixes #3102 

changes made in test.yml ci as specified in the issue description 
needs clause was added to `go-unit-tests` and `go-integration-tests` jobs.


jaydee029:
please review @Sharpz7

Sharpz7 (approved):

richscott (approved):

PR #3206, opened 2023-12-20
Bring back `dirWithIndexFallback`

zuqq:
#3202 broke the routing in Lookout, by making the HTTP server handle paths like `/job-sets` and `/oidc` that were supposed to be handled by the frontend.

### DIFF
--- a/internal/common/serve/static.go
+++ b/internal/common/serve/static.go
@@ -1,6 +1,7 @@
 package serve
 
 import (
+	"io/fs"
 	"net/http"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


internal/common/serve/static.go
```diff
@@ -9,6 +10,27 @@ import (
...
+	dir http.Dir
+}
+
+func CreateDirWithIndexFallback(path string) http.FileSystem {
+	return dirWithIndexFallback{http.Dir(path)}
+}
+
+func (d dirWithIndexFallback) Open(name string) (http.File, error) {
+	file, err := d.dir.Open(name)
+	if errors.Is(err, fs.ErrNotExist) {
```
zuqq, line 28: This is essentially the code that I overzealously removed in #3202, expect that I have replaced `err != nil` by `errors.Is(err, fs.ErrNotExist)` here.

JamesMurkin (approved):

PR #3229, opened 2024-01-05
[CI-IMPROVEMENT-6] Introduce targeted caching of tools

pavlovic-ivan:
#### What type of PR is this?
This is an improvement PR

Sixth in a series of PRs for improvements. This adds caching of tools from tools.yaml to the CI. This change cuts additional 2 minutes after cache is created. Iintegration tests will take less time, which will be the most visible improvement

#### Notes for maintainers
After PR is merged, i will add the same change with caching to workflows:

- codeql-analysis.yml
- airfow-operator.yml
- python-client.yml

### DIFF
--- a/magefiles/main.go
+++ b/magefiles/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


magefiles/main.go
```diff
@@ -14,27 +15,87 @@ import (
 
 // BootstrapTools installs all tools needed tobuild and release Armada.
 // For the list of tools this will install, see tools.yaml in the root directory
+type ToolsList struct {
+	Tools []string
+}
+
```
jgiannuzzi, line 21: this type definition could be moved back inside `BootstrapTools` as it is not used anywhere else

magefiles/main.go
```diff
@@ -20,21 +21,80 @@ func BootstrapTools() error {
 		Tools []string
 	}
 
-	tools := &ToolsList{}
-	err := readYaml("tools.yaml", tools)
+	requiredTools := &ToolsList{}
+	if err := readYaml("tools.yaml", requiredTools); err != nil {
+		return err
```
dave-gantenbein, line 26: Should this and the following return err fmt.Printf the message? The standard in this file seems inconsistent, but I'd imagine I'd like to see a specific message if in fact something went wrong. Not a big deal, but you could be saving your future self some heartache. :)

magefiles/main.go
```diff
@@ -211,6 +271,19 @@ func readYaml(filename string, out interface{}) error {
 	return err
 }
 
+// Helper function to write YAML data to a file
+func writeYAML(filename string, data interface{}) error {
```
dave-gantenbein, line 275: Is there a standard yaml package in this repo already we could use instead of hand rolling this?

magefiles/main.go
pavlovic-ivan: returning an error here is actually raising an error to the main process that executed the function. that means that the main process will actually handle and log the error. does that answer your question?

magefiles/main.go
pavlovic-ivan: i would assume there is, but i didn't look at that because of the approach used in the file, which is not using such library if exists. maybe it can be a future improvement, would you agree?

jgiannuzzi (approved):
LGTM 👍

magefiles/main.go
dave-gantenbein: I'll defer to the rest of the team. :) I would not block merging for this.

richscott: Theoretically, a `yaml.Encoder` object could be constructed (by opening the file, getting a `*File` handle (which implements the `io.Writer` interface), then call its `Encode(x interface{}) error` method, but it wouldn't be any more concise, because of all the boilerplate logic, e.g. (not tested):
```
func newWriteYAML(filename string, data interface{}) error {
	outfile, err := os.Open(filename)
	if err != nil {
		return err
	}
	encoder := yaml.NewEncoder(outfile)
	err = encoder.Encode(data)
	if err != nil {
		return err
	}
	err = encoder.Close()
	if err != nil {
		return err
	}
	return outfile.Close()
}
```
So I think the current code is about as brief & correct as any.

richscott (approved):
Looks great!

PR #3236, opened 2024-01-15
Add `GangInfo`

zuqq:
This PR reduces the number of places in which we extract gang metadata from a job's annotations by adding the `GangInfo` type and adding a field of this type to `JobSchedulingContext`; it also removes the need to pass `GangIdAndCardinalityFromAnnotations` to `JobSchedulingContextFromJob`, which simplifies the code.

### DIFF
--- a/internal/scheduler/context/context.go
+++ b/internal/scheduler/context/context.go
@@ -2,6 +2,7 @@ package context
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 	"text/tabwriter"
 	"time"
@@ -230,7 +231,7 @@ func (sctx *SchedulingContext) AddGangSchedulingContext(gctx *GangSchedulingCont
 			numberOfSuccessfulJobs++
 		}
 	}
-	if numberOfSuccessfulJobs >= gctx.GangMinCardinality && !allJobsEvictedInThisRound {
+	if numberOfSuccessfulJobs >= gctx.GangInfo.MinimumCardinality && !allJobsEvictedInThisRound {
 		sctx.NumScheduledGangs++
 	}
 	return allJobsEvictedInThisRound, nil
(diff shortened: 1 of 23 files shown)


internal/scheduler/context/context.go
```diff
@@ -699,33 +681,95 @@ func (jctx *JobSchedulingContext) GetNodeSelector(key string) (string, bool) {
...
 	for i, job := range jobs {
-		jctxs[i] = JobSchedulingContextFromJob(priorityClasses, job, extractGangInfo)
+		jctxs[i] = JobSchedulingContextFromJob(priorityClasses, job)
 	}
 	return jctxs
 }
 
-func JobSchedulingContextFromJob(priorityClasses map[string]types.PriorityClass, job interfaces.LegacySchedulerJob, extractGangInfo func(map[string]string) (string, int, int, bool, error)) *JobSchedulingContext {
-	// TODO: Move cardinality to gang context only and remove from here.
-	// Requires re-phrasing nodedb in terms of gang context, as well as feeding the value extracted from the annotations downstream.
```
zuqq, line 712: I've deleted this comment, because I think that it's misleading: the scope of the required changes is much larger (for example, this would also require fundamental changes to the in-memory job database).

internal/scheduler/nodedb/nodedb.go
```diff
@@ -512,34 +512,10 @@ func NodeJobDiff(txnA, txnB *memdb.Txn) (map[string]*Node, map[string]*Node, err
 	return preempted, scheduled, nil
 }
 
-// ScheduleMany assigns a set of jobs to nodes.
-// If N jobs can be scheduled, where N >= `GangMinCardinality`, it will return true, nil and set ShouldFail on any excess jobs.
-// Otherwise, it will return false, nil.
-// TODO: Pass through contexts to support timeouts.
-func (nodeDb *NodeDb) ScheduleMany(jctxs []*schedulercontext.JobSchedulingContext) (bool, error) {
```
zuqq, line 519: This was only used in tests.

internal/scheduler/nodedb/nodedb_test.go
```diff
@@ -74,7 +71,7 @@ func TestSelectNodeForPod_NodeIdLabel_Success(t *testing.T) {
 		map[string]string{schedulerconfig.NodeIdLabel: nodeId},
 		testfixtures.N1Cpu4GiJobs("A", testfixtures.PriorityClass0, 1),
 	)
-	jctxs := schedulercontext.JobSchedulingContextsFromJobs(testfixtures.TestPriorityClasses, jobs, func(_ map[string]string) (string, int, int, bool, error) { return "", 1, 1, true, nil })
```
zuqq, line 77: I'm glad to be getting rid of this argument to `JobSchedulingContextsFromJobs`: for example, in this case we're returning an empty gang ID and a value of `true` for `isGangJob`, which is illegal.

severinson (approved):

PR #3239, opened 2024-01-19
PR 3180 follow-up

severinson:


internal/scheduler/scheduler.go
```diff
@@ -774,9 +776,8 @@ func (s *Scheduler) generateUpdateMessagesFromJob(job *jobdb.Job, jobRunErrors m
 		events = append(events, jobReprioritised)
 	}
 
-	if origJob != job {
-		err := txn.Upsert([]*jobdb.Job{job})
-		if err != nil {
+	if !origJob.Equal(job) {
```
severinson, line 779: The previous behaviour would be to always upsert jobs we've made any changes to even if those changes were no-ops.

internal/scheduler/jobdb/reconciliation.go
```diff
@@ -174,19 +174,19 @@ func (jobDb *JobDb) reconcileRunDifferences(jobRun *JobRun, jobRepoRun *database
 			rst.Running = true
 		}
 		if jobRepoRun.Succeeded && !jobRun.Succeeded() {
-			jobRun = jobRun.WithSucceeded(true)
+			jobRun = jobRun.WithSucceeded(true).WithRunning(false)
```
severinson, line 177: Caught by assertions in dev. Runs should only be in one state.

PR #3347, opened 2024-01-30
Scheduler: resource waste metric

Mo-Fatah:


internal/scheduler/metrics/metrics.go
```diff
@@ -189,48 +210,77 @@ func (m *Metrics) Update(
...
+	}
+	if err := m.updateCounterVecFromJob(m.transitions, labels[1:], job); err != nil {
 		return err
 	}
 	return nil
 }
 
 func (m *Metrics) UpdateCancelled(job *jobdb.Job) error {
 	labels := m.buffer[0:0]
+	labels = append(labels, running)
```
severinson, line 242: Not necessarily true. Jobs can be cancelled while queued.

internal/scheduler/metrics/metrics.go
```diff
@@ -189,48 +210,77 @@ func (m *Metrics) Update(
...
 		return err
 	}
 	return nil
 }
 
 func (m *Metrics) UpdateFailed(ctx *armadacontext.Context, job *jobdb.Job, jobRunErrorsByRunId map[uuid.UUID]*armadaevents.Error) error {
-	labels := m.buffer[0:0]
 	category, subCategory := m.failedCategoryAndSubCategoryFromJob(ctx, job, jobRunErrorsByRunId)
+	labels := m.buffer[0:0]
+	labels = append(labels, running)
```
severinson, line 259: Job can go from queued to failed.

internal/scheduler/metrics/metrics.go
```diff
@@ -252,11 +306,15 @@ func (m *Metrics) UpdateLeased(jctx *schedulercontext.JobSchedulingContext) erro
 func (m *Metrics) UpdatePreempted(jctx *schedulercontext.JobSchedulingContext) error {
 	labels := m.buffer[0:0]
 	job := jctx.Job.(*jobdb.Job)
+	labels = append(labels, running)
```
severinson, line 309: Jobs can be preempted in leased/pending.

internal/scheduler/metrics/metrics.go
```diff
@@ -383,7 +441,7 @@ func (m *Metrics) updateCounterVecFromJob(vec *prometheus.CounterVec, labels []s
 		labels[i] = string(resourceName)
 		if c, err := vec.GetMetricWithLabelValues(labels...); err != nil {
 			return err
-		} else {
+		} else if len(labels) == cap(labels) { // We consume all the buffer, therefore it
```
severinson, line 444: Is this needed? Comparing against the cap seems fragile.

internal/scheduler/metrics/metrics.go
```diff
@@ -392,3 +450,57 @@ func (m *Metrics) updateCounterVecFromJob(vec *prometheus.CounterVec, labels []s
...
+// updateResourceSecondsCounterVec is a helper method to increment vector counters by the number of seconds per resource a jobs has consumed in a given state.
+func (m *Metrics) updateResourceSecondsCounterVec(vec *prometheus.CounterVec, labels []string, job *jobdb.Job) error {
+	i := len(labels)
+
+	// Number of jobs.
+	labels = append(labels, jobsResourceLabel)
+	if c, err := vec.GetMetricWithLabelValues(labels...); err != nil {
+		return err
+	} else {
+		c.Add(1)
```
severinson, line 463: Need the multiple by the duration in state here too.

internal/scheduler/metrics/metrics.go
```diff
@@ -392,3 +450,57 @@ func (m *Metrics) updateCounterVecFromJob(vec *prometheus.CounterVec, labels []s
...
+			stateDuration := stateDuration(job.LatestRun(), labels[1]).Seconds()
+			q := requests[resourceName]
+			v := float64(q.MilliValue()) / 1000
+			c.Add(v * stateDuration)
+		}
+	}
+	return nil
+}
+
+func stateDuration(run *jobdb.JobRun, state string) time.Duration {
```
severinson, line 480: This code assumes a particular set of state transitions, e.g., that only running jobs are preempted, which is not true in general. Would it not be easier to return the difference between the largest and second largest timestamp? Since the largest timestamp will correspond to the most recent state transition and the second largest to the one before that.

internal/scheduler/jobdb/job_run.go
```diff
@@ -188,35 +203,27 @@ func MinimalRun(id uuid.UUID, creationTime int64) *JobRun {
...
-	nodeName string,
-	scheduledAtPriority *int32,
-	running bool,
-	succeeded bool,
-	failed bool,
-	cancelled bool,
-	returned bool,
-	runAttempted bool,
-) *JobRun {
+func (jobDb *JobDb) CreateRun(nodeId string, dbRun *database.Run) *JobRun {
```
severinson, line 208: I guess you did this because you had to add a lot more arguments? My concern with this is that the database.Run now leaks into the JobDb.

internal/scheduler/metrics/metrics.go
Mo-Fatah: I was thinking of something, and forgot to delete it

internal/scheduler/metrics/metrics.go
Mo-Fatah: Sounds good to me. I thought of this logic that may doesn't require sorting/comparing timestamps:
```
func getPriorState(run *jobdb.JobRun, state string) string {
	switch state {
	case queued:
		return ""
	case leased:
		return queued
	case pending:
		return leased
	case succeeded:
		return running
	case running:
		return pending
	case failed, cancelled, preempted:
		if run.Running() {
			return running
		} else if run.Pending() {
			return pending
		} else if run.Leased() {
			return leased
		} else {
			return queued
		}
	}
	return ""
}
```
what do you think of this approach ? also does it sounds logically correct ?

severinson: I think the run will have already been marked as failed at this point.

internal/scheduler/jobdb/job_run.go
Mo-Fatah: yes, the arguments number was unmanageable after the additional fields I added to `JobRun`. Can elaborate what are the leaks that concerns you exactly ?

internal/scheduler/metrics/metrics.go
```diff
@@ -392,3 +461,63 @@ func (m *Metrics) updateCounterVecFromJob(vec *prometheus.CounterVec, labels []s
...
+		cancelled: run.TerminatedTime(),
+		succeeded: run.TerminatedTime(),
+		failed:    run.TerminatedTime(),
+	}
+	diff := stateTime.Sub(time.Time{}).Seconds()
+	for s, t := range stateToTime {
+		if t == nil {
+			continue
+		}
+		if priorTime == nil || stateTime.Sub(*t).Seconds() < diff {
```
Mo-Fatah, line 516: diff not updated, will fix it now

internal/scheduler/jobdb/job_run.go
severinson: The leak is having the jobdb package depend on the database package. Ideally the jobdb package does not depend on the mechanism by which we load data from postgres.

internal/scheduler/metrics/metrics.go
```diff
@@ -189,74 +210,122 @@ func (m *Metrics) Update(
...
 		return err
 	}
 	return nil
 }
 
 func (m *Metrics) UpdatePreempted(jctx *schedulercontext.JobSchedulingContext) error {
-	labels := m.buffer[0:0]
 	job := jctx.Job.(*jobdb.Job)
+	latestRun := job.LatestRun()
+	priorState, priorStateTime := getPriorState(latestRun, latestRun.PreemptedTime())
```
severinson, line 318: Think this should be terminated time.

internal/scheduler/metrics/metrics.go
```diff
@@ -392,3 +461,64 @@ func (m *Metrics) updateCounterVecFromJob(vec *prometheus.CounterVec, labels []s
...
+		return end.Sub(*start)
+	}
+	return time.Duration(0)
+}
+
+func getPriorState(run *jobdb.JobRun, stateTime *time.Time) (prior string, priorTime *time.Time) {
+	if stateTime == nil {
+		return "", nil
+	}
+	stateToTime := map[string]*time.Time{
```
severinson, line 502: Shouldn't need to create a map here. I'm concerned it will cause unnecessary GC pressure, which we've had issues with in the past. Just check all of these one by one and return the state/time of the largest timestamp smaller than stateTime?

severinson (approved):

severinson (approved):

severinson (approved):

richscott (approved):

PR #3351, opened 2024-01-31
Cache error regex matching for metrics

severinson:


internal/scheduler/metrics/metrics.go
```diff
@@ -250,14 +267,44 @@ func (m *Metrics) failedCategoryAndSubCategoryFromJob(ctx *armadacontext.Context
...
+		return 0, false
+	}
+	i, ok := m.matchedRegexIndexByErrorMessage.Get(message)
+	if !ok {
+		return 0, false
+	}
+	return i.(int), true
+}
+
+func (m *Metrics) regexIndexFromErrorMessage(message string) (int, bool) {
```
zuqq, line 301: It occurs to me that it would be slightly cleaner to move all the interactions with the LRU cache into this function.

zuqq (approved):

internal/scheduler/metrics/metrics.go
severinson: That's prob. Let me try it.

zuqq (approved):

PR #3369, opened 2024-02-05
[CI-IMPROVEMENT-9] Implement check-required workflow that checks if all required checks are successful and not skipped

pavlovic-ivan:
#### What type of PR is this?
This is an improvement PR

9th in a series of PRs for improvements. This implements `check-required` workflow that verifies if all required checks are successful, which is the case with the job in the CI workflow.

#### Notes for maintainers
Prerequisites:
- [ ] Create a Github app that has permissions: "Read access to metadata" and "Read and write access to checks"
- [ ] Create an env `create-check` with a protection rule for `master` branch with two secrets `APP_ID` and `APP_PRIVATE_KEY` and copy the apps creds created in the previous step

After merging the PR
- [ ] update branch protection to enforce the required check to originate from the new GitHub app

Work based on these two PRs as a reference:
- https://github.com/G-Research/fasttrackml/pull/241
- https://github.com/G-Research/fasttrackml/pull/458

### DIFF
--- a/.github/workflows/check-required.yml
+++ b/.github/workflows/check-required.yml
@@ -1,0 +1,97 @@
+name: Check required jobs
+
+# This workflow is triggered when a workflow run for the CI is completed.
+# It checks if the "All required checks done" job was actually successful
+# (and not just skipped) and creates a check run if that is the case. The
+# check run can be used to protect the main branch from being merged if the
+# CI is not passing. We need to use a GitHub app token to create the check
+# run because otherwise the check suite will be assigned to the first workflow
+# run for the CI, which might not be the latest one. See
+# https://github.com/orgs/community/discussions/24616#discussioncomment-6088422
+# for more details.
+
+on:
+  workflow_run:
+    workflows: [CI]
+
+permissions:
+  actions: read
+  checks: write
+
+jobs:
+  required-jobs:
+    name: Check required jobs
+    if: ${{ !github.event.repository.fork }}
+    environment: create-check
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


severinson (approved):

.github/workflows/check-required.yml
```diff
@@ -0,0 +1,97 @@
...
+              },
+            };
+
+            core.summary.addHeading('The following required checks have been considered:', 3);
+            ghaChecks.forEach(check => {
+              core.summary
+                .addLink(check.name, check.html_url)
+                .addCodeBlock(JSON.stringify(check, ['status', 'conclusion', 'started_at', 'completed_at'], 2), 'json');
+
+              if (check.status === 'completed' && check.conclusion === 'success') {
```
dave-gantenbein, line 74: Do we want to do any handling of cases where we're not completed successfully here?

pavlovic-ivan: This `else` block here is only informational. It's not necessary, but since the code is taken from FML, maybe they benefit from this info, which then makes this contribution identical to what they use. @jgiannuzzi what do you think of this? does FML benefit from it or not? should i remove it, so both are identical (making it easier to extract to separate repo later)

richscott (approved):

.github/workflows/check-required.yml
jgiannuzzi: @dave-gantenbein any other case is handled by `newCheck`'s default status that is set to `in_progress`. Basically we _only_ create our successful check when we have found a successful check, otherwise we mark it as pending.

PR #3373, opened 2024-02-05
Use separate db for jobservice integration tests (#521)

richscott:
Create and use a separate database for the Job-Service integration tests (known as the `teste2eAirflow` Mage target). The localdev Armada setup uses the default/built-in database 'postgres', and since there are table-name collisions between the Armada and the Job Service databases (e.g. both have a 'jobs' table), create and use a db named 'jobservice' for the Job Service integration tests.

Fixes https://github.com/G-Research/gr-oss/issues/521


pavlovic-ivan (approved):
LGTM. Thanks @richscott

Mo-Fatah (approved):

suprjinx (approved):
LGTM

PR #3392, opened 2024-02-11
Remove Some Unneeded Config values

d80tb7:
Remove some config values that are no longer needed.

There are two changes here where I've moved things around in a way that has resulte din non-backwards compatible config:

- The ingesters no longer have a separate `metrics` subsection.  Therefore `metrics.port` is now just `metricsPort`
- The  eventIngester used to have an option for never expiring anything.  I've removed this option.  If people really want to never expire anything they can just set the expiration tine to 999 years


config/armada/config.yaml
```diff
@@ -125,9 +125,6 @@ pulsar:
   maxAllowedMessageSize: 4194304 # 4MB
   receiverQueueSize: 100
 postgres:
-  maxOpenConns: 100
```
d80tb7, line 128: these aren't used by pgx and are thus misleading.

JamesMurkin (approved):

Issue #3390, opened 2024-02-09
Improve repo test coverage in various Go packages


richscott:
This is simply a list of suggested Golang files and functions that currently have low or no test coverage, and would be helpful to write new or enhanced tests for these, to verify functionality. Not every entry in here needs to be addressed in a single pull request - we will close out this ticket and create new ones with a trimmed down list of the remaining entries as necessary.  Note that some of the functions are private - it may be required to change their visibility to public, depending on if the test code has Go-package access to the tested function.

- [ ] internal/armada/server/submit_to_log.go:857:                	createJobs                              	0.0%
- [ ] internal/armada/server/submit_to_log.go:861:                	createJobsObjects                       	0.0%
- [ ] internal/common/database/migrations.go:91:                  	ReadMigrations                          	0.0%
- [ ] internal/common/database/migrations.go:125:                 	ReadMigrationsFromStatik                	0.0%
- [ ] internal/executor/job/submit.go:54:                         	SubmitJobs                              	0.0%
- [ ] internal/executor/job/submit.go:58:                         	submitJobs                              	0.0%
- [ ] internal/executor/job/submit.go:84:                         	submitWorker                            	0.0%
- [ ] internal/executor/job/submit.go:113:                        	submitPod                               	0.0%
- [ ] internal/executor/job/submit.go:153:                        	applyExecutorSpecificIngressDetails     	0.0%
- [ ] internal/executor/job/util.go:14:                           	CreateSubmitJobFromExecutorApiJobRunLease	0.0%
- [ ] internal/lookoutv2/repository/getjobs.go:205:               	sortRuns                                	60.0%
- [ ] internal/lookoutv2/repository/getjobs.go:221:               	getJobRunTime                           	40.0%
- [ ] internal/lookoutv2/repository/getjobs.go:231:               	makeJobRows                             	84.6%
- [ ] internal/lookoutv2/repository/getjobs.go:292:               	makeRunRows                             	84.6%
- [ ] internal/lookoutv2/repository/getjobs.go:337:               	makeAnnotationRows                      	84.6%
- [ ] internal/scheduler/database/executor_repository.go:77:      	StoreExecutor                           	72.7%
- [ ] internal/scheduler/scheduler_metrics.go:205:                	observeJobAggregates                    	0.0%
- [ ] internal/scheduler/scheduler_metrics.go:221:                	reportNumberOfJobsConsidered            	0.0%
[...]

PR #3397, opened 2024-02-13
Writing Test coverage for ReadMigrations function

Bharadwajshivam28:
#### What type of PR is this?
 I am creating Test Coverage for the function [ReadMigrations]( https://github.com/armadaproject/armada/blob/82398c859ae2bc42d1c1cd32fe11e162d109bbaf/internal/common/database/migrations.go#L91C1-L122C2)
 
refs: #3390


Bharadwajshivam28:
@richscott I am working on creating Test coverage for the function `ReadMigrations` . While seeing the function and what it does I have created few tests:-

1) Checks file format which will be read (currently the file name is just for test we need to change it as per our req)
2) Checks slice format
3) Checks the ID length (adjust the maximum ID length as needed)

There are many changes required like we can add/remove conditions which will make things more better ....

internal/common/database/migrations_test.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"testing"
+	"strconv"
+	"strings"
+	"io/fs"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestReadMigrations(t *testing.T) {
+	mockFiles := []string{
```
richscott, line 12: This is a good start - using a mocked filesystem with `io/fs` is good. Try adding some more files in here, with varying differences to match (or not) the expected patterns) - see the existing migrations in `internal/scheduler/database/migrations/*.sql` and also in `internal/lookoutv2/schema/migrations/*.sql`.

internal/common/database/migrations_test.go
```diff
@@ -0,0 +1,46 @@
...
+		assert.True(t, strings.HasSuffix(migration.name, "migration.sql"), "Invalid file format for %v", migration.name)
+	}
+	
+	// Checks slice format
+	expectedMigrations := []Migration{
+		{id: 0, name: "", sql: ""},
+	}
+	assert.Len(t, migrations, 1, "Unexpected slice length")
+
+	// Slice first value should be int
```
richscott, line 29: The comments for these assertions are nice, but probably not necessary - the code is very clear.

internal/common/database/migrations_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	// Checks ID length
+	for _, migration := range migrations {
+		assert.True(t, len(strconv.Itoa(migration.id)) <= 5, "Invalid ID length for %v", migration.name)
+	}
+	assert.NoError(t, err, "Unexpected error in ReadMigrations")
+}
+
+/*
+func createMockFileSystem(files []string) fs.FS {
```
richscott, line 44: Here you just need to declare a new struct type, and then have it implement the same method(s) that are defined for `fs.FS` - see https://pkg.go.dev/io/fs@go1.20.12

richscott (commented):
Just a few suggestions for now - this is a good start.

internal/common/database/migrations_test.go
Bharadwajshivam28: Sure, will look into it

Bharadwajshivam28:
Hello @richscott I have modified tests. I aligned on to match the `expected pattern` which we want from the `ReadMigrations function.` I compared with .`sql files` in the location `internal/scheduler/database/migrations/*.sql` `and internal/lookoutv2/schema/migrations/*.sql`

Currently the Tests includes-
1) Checks file format
2) Checks file is sorted or not
3) Checks for extracted ID for each file
4) Checks for ID length
5) Checks for slice formatting and types
6) checks if the sql (content not empty)

btw.... it needs modifcation as currently `mockfiles` part is unfinished and also the `current tests needs changes`

internal/common/database/migrations_test.go
Bharadwajshivam28: Can you please elaborate this one?

richscott: This would be just an instance of creating a classic Go interface implementation. You'd create a struct (for example, `type struct testFS { ... }`) and then create method(s) that implement the `fs.FS` interface - in this case, there's just one method, so the new function definition would be 
`func (myFS *testFS) Open(name string) (File, error) { /* return a fake File object */ }`.

If you search for `golang interfaces`, there are several good articles on the basics of implementing interfaces in Go, like https://golangbot.com/interfaces-part-1/ .

Bharadwajshivam28: got it...

Bharadwajshivam28:
can you please @richscott review the current tests which i have pushed

internal/common/database/migrations_test.go
Bharadwajshivam28: I have created struct `testFS` for `Mockfiles` handling

internal/common/database/migrations.go
```diff
@@ -98,7 +98,7 @@ func ReadMigrations(fsys fs.FS, basePath string) ([]Migration, error) {
 
 	var migrations []Migration
 	for _, f := range files {
-
+
```
richscott, line 101: This change should be removed.

internal/common/database/migrations_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	for _, migration := range migrations {
+		assert.IsType(t, 0, migration.id, "Should be of type int")
+		assert.IsType(t, "", migration.name, "Should be of type string")
+		assert.IsType(t, "", migration.sql, "Should be of type string")
+	}
+    
+	assert.Equal(t, migrations[0].id, 1, "Incorrect ID for 001_initial_schema.sql")
+	assert.Equal(t, migrations[1].id, 2, "Incorrect ID for 002_cancel_reason.sql")
+	assert.Equal(t, migrations[2].id, 3, "Incorrect ID for 003_run_leased_column.sql")
```
richscott, line 60: I don't think it's necessary to exhaustively test for ID matching for every existing SQL file - just two or three sample tests should be enough.

Bharadwajshivam28: Thanks...

internal/common/database/migrations.go
Bharadwajshivam28: Okay...

Bharadwajshivam28:
Hey @richscott  I have made the changes..

internal/common/database/migrations_test.go
```diff
@@ -0,0 +1,77 @@
+package database
+
+import (
+	"testing"
```
richscott, line 4: Can you run the `goimports` tool on this file? It will sort the imports, and separate out third-party/non-standard-library modules, automatically.

Bharadwajshivam28: Okay let me do it

PR #3402, opened 2024-02-15
Remove UnifiedSchedulingByPool Config Flag

d80tb7:
UnifiedSchedulingByPool is onw standard.  Let's remove the config flag.

### DIFF
--- a/internal/scheduler/scheduling_algo_test.go
+++ b/internal/scheduler/scheduling_algo_test.go
@@ -338,26 +338,6 @@ func TestSchedule(t *testing.T) {
 			},
 			expectedScheduledIndices: []int{0},
 		},
-		"UnifiedSchedulingByPool": {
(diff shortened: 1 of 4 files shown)


internal/scheduler/scheduling_algo_test.go
```diff
@@ -338,26 +338,6 @@ func TestSchedule(t *testing.T) {
 			},
 			expectedScheduledIndices: []int{0},
 		},
-		"UnifiedSchedulingByPool": {
```
severinson, line 341: I think we want to keep these tests.

d80tb7: I've put back `Schedule gang job over multiple executors` but I think `UnifiedSchedulingByPool` is now the same as `scheduling`

severinson (approved):

PR #3403, opened 2024-02-15
Add Unacknowledged jobs metric

Mo-Fatah:


internal/scheduler/scheduling_algo.go
```diff
@@ -35,6 +35,8 @@ type SchedulingAlgo interface {
 	// Schedule should assign jobs to nodes.
 	// Any jobs that are scheduled should be marked as such in the JobDb using the transaction provided.
 	Schedule(*armadacontext.Context, *jobdb.Txn) (*SchedulerResult, error)
+	// Config returns the configuration used by the scheduling algorithm.
+	Config() *configuration.SchedulingConfig
```
severinson, line 39: I don't think we should pass config back up. If the parent wants the config it should retain it when creating the algo instead of passing it back up again.

internal/scheduler/schedulerobjects/schedulerobjects.proto
```diff
@@ -21,6 +21,8 @@ message Executor {
     google.protobuf.Timestamp lastUpdateTime = 5 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
     // Jobs that are owned by the cluster but are not assigned to any node.
     repeated string unassigned_job_runs = 9;
+    
+    int64 num_unacknowledged_jobs = 10;
```
severinson, line 25: Why do we need this? The scheduler should already know which jobs haven't been acked.

internal/scheduler/metrics/metrics.go
```diff
@@ -378,6 +389,23 @@ func (m *Metrics) UpdateRunning(job *jobdb.Job) error {
...
+		return err
+	}
+	for _, executor := range executors {
+		labels := m.buffer[0:0]
+		labels = append(labels, executor.Id)
+		g, err := m.unacknowledgedJobsPerExecutor.GetMetricWithLabelValues(labels...)
+		if err != nil {
+			continue
+		}
+		g.Set(float64(executor.NumUnacknowledgedJobs) / float64(maxUnacknowledgedJobsPerExecutor))
```
severinson, line 404: I'd prefer to export an integer with the number of jobs instead of a ratio.

internal/scheduler/metrics/metrics.go
```diff
@@ -59,8 +60,9 @@ type Metrics struct {
 	matchedRegexIndexByErrorMessage *lru.Cache
 
 	// Job metrics.
-	transitions     *prometheus.CounterVec
-	resourceSeconds *prometheus.CounterVec
+	transitions                   *prometheus.CounterVec
+	resourceSeconds               *prometheus.CounterVec
+	unacknowledgedJobsPerExecutor *prometheus.GaugeVec
```
severinson, line 65: I think we should do something more general. I'd call it
```
numUnacknoledgedJobsPerExecutor: prometheus.NewGaugeVec(
			prometheus.GaugeOpts{
				Namespace: namespace,
				Subsystem: subsystem,
				Name:      "unacknowledged_jobs_per_executor",
				Help:      "Number of unacknowledged jobs per executor.",
			},
			[]string{"executor"},
		),
```

severinson: Separately, we may want to add another metric called something like `executorHealth`. But not now.

PR #3408, opened 2024-02-16
Add `jsonb` backend to Lookout

zuqq:
This PR is the second step toward storing job annotations in a column of type `jsonb` (after adding such a column in #3407): `QueryBuilder` can now construct queries that use this column, and the endpoints `/jobs` and `/jobGroups` allow users to opt in to this new backend by adding `backend=jsonb` to the query string. I have also changed the `/jobs` and `/jobGroups` test suites to run with both variants (i.e., pulling annotations from `user_annotation_lookup` or `annotations`), to make sure that these return the same results.


internal/lookoutv2/repository/getjobs.go
```diff
@@ -70,14 +72,80 @@ type annotationRow struct {
...
-		db:            db,
-		lookoutTables: NewTables(),
+		db:              db,
+		lookoutTables:   NewTables(),
+		useJsonbBackend: useJsonbBackend,
 	}
 }
 
 func (r *SqlGetJobsRepository) GetJobs(ctx *armadacontext.Context, filters []*model.Filter, activeJobSets bool, order *model.Order, skip int, take int) (*GetJobsResult, error) {
+	if r.useJsonbBackend {
```
d80tb7, line 84: Personally I'd create two private functions here, `getJobsLegacy` and `getJobsJsonB`,  and just have this function delegate based on `r.useJsonbBackend`.  Same for everywhere else we do this.

internal/lookoutv2/model/model.go
```diff
@@ -42,16 +42,52 @@ type Job struct {
...
+}
+
+func NewPostgreSQLTime(t *time.Time) *PostgreSQLTime {
+	if t == nil {
+		return nil
+	}
+	return &PostgreSQLTime{Time: *t}
+}
+
+func (t PostgreSQLTime) MarshalJSON() ([]byte, error) {
```
d80tb7, line 69: I  presume this is  is needed when you have timestamps int he jsonb column?  If so is it ok that this isn't nil safe? (c.f. `NewPostgreSQLTime` which will happily return you  a nil).

d80tb7 (approved):

internal/lookoutv2/repository/getjobs.go
zuqq: I agree that it's a little silly here; this is the only place where the code for the two branches significantly, though (e.g., in `GroupJobs` we just make two different calls to `QueryBuilder` methods, and everything else is shared).

internal/lookoutv2/model/model.go
zuqq: In a previous version of this code `MarshalJSON` operated on `*PostgreSQLTime` (and blindly dereferenced `t`); in the version you're commenting on it operates on `PostgreSQLTime`, because I had similar concerns. However, both versions are fine (`json.Marshal` doesn't call `MarshalJSON` on nil pointers): https://go.dev/play/p/JpvUKtlWILQ

zuqq: We discussed the analogous question for `UnmarshalJSON` offline; the upshot is that `json.Unmarshal` also guarantees that the pointer it calls `UnmarshalJSON` on is not nil (I modeled this after the [implementation of `UnmarshalJSON` for `time.Time`](https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/time/time.go;l=1375-1389), which also just dereferences the pointer).

d80tb7 (approved):

PR #3416, opened 2024-02-20
Fix metrics double counting

Mo-Fatah:


internal/scheduler/scheduler.go
```diff
@@ -378,7 +378,7 @@ func (s *Scheduler) updateMetricsFromSchedulerResult(ctx *armadacontext.Context,
 		}
```
severinson, line 378: We should remove the call to UpdatePreempted here. Since preemptions are reported when the run is actually reported as preempted.

internal/scheduler/metrics/metrics.go
```diff
@@ -273,21 +273,33 @@ func (m *Metrics) UpdateCancelled(job *jobdb.Job) error {
...
-func (m *Metrics) UpdateFailed(ctx *armadacontext.Context, job *jobdb.Job, jobRunErrorsByRunId map[uuid.UUID]*armadaevents.Error) error {
+// For post-scheduling cycle, jctx is expected to be NOT nil and only the transitions metric is updated.
+// For pre-scheduling cycle, job is expected to be updated through JobDb reconciliation and only the resourceSeconds metric is updated.
+func (m *Metrics) UpdateFailed(ctx *armadacontext.Context, job *jobdb.Job, jctx *schedulercontext.JobSchedulingContext, jobRunErrorsByRunId map[uuid.UUID]*armadaevents.Error) error {
+	if job == nil {
+		job = jctx.Job.(*jobdb.Job)
+	}
 	category, subCategory := m.failedCategoryAndSubCategoryFromJob(ctx, job, jobRunErrorsByRunId)
+	if category == jobRunPreempted {
+		return nil // Preemption metric is handled separately.
```
severinson, line 284: We should call UpdatePreempted here.

internal/scheduler/metrics/metrics.go
```diff
@@ -186,11 +186,6 @@ func (m *Metrics) Update(
 			return err
 		}
 	}
-	if jst.Leased {
```
severinson, line 189: I'd add a comment to this function explaining why we're not calling leased here since it may look surprising.

internal/scheduler/metrics/metrics.go
```diff
@@ -273,8 +268,14 @@ func (m *Metrics) UpdateCancelled(job *jobdb.Job) error {
 	return nil
 }
 
-func (m *Metrics) UpdateFailed(ctx *armadacontext.Context, job *jobdb.Job, jobRunErrorsByRunId map[uuid.UUID]*armadaevents.Error) error {
+func (m *Metrics) UpdateFailed(ctx *armadacontext.Context, job *jobdb.Job, jctx *schedulercontext.JobSchedulingContext, jobRunErrorsByRunId map[uuid.UUID]*armadaevents.Error) error {
```
severinson, line 271: Why do we need to pass in both a job and jctx, which has an embedded job?

internal/scheduler/metrics/metrics.go
```diff
@@ -273,8 +268,14 @@ func (m *Metrics) UpdateCancelled(job *jobdb.Job) error {
...
+	if job == nil {
+		job = jctx.Job.(*jobdb.Job)
+	}
 	category, subCategory := m.failedCategoryAndSubCategoryFromJob(ctx, job, jobRunErrorsByRunId)
+	if category == jobRunPreempted {
+		return m.UpdatePreempted(job, jctx)
```
severinson, line 277: Add a comment explaining why we're doing this.

internal/scheduler/metrics/metrics.go
```diff
@@ -315,7 +316,11 @@ func (m *Metrics) UpdateLeased(job *jobdb.Job, jctx *schedulercontext.JobSchedul
 		job = jctx.Job.(*jobdb.Job)
 	}
 	latestRun := job.LatestRun()
-	priorState, priorStateTime := getPriorState(job, latestRun, latestRun.LeaseTime())
+	leaseTime := time.Now()
```
severinson, line 319: Can we instead set the lease time on the job from within the scheduler when creating the lease?

internal/scheduler/scheduler.go
```diff
@@ -372,13 +372,8 @@ func (s *Scheduler) updateMetricsFromSchedulerResult(ctx *armadacontext.Context,
 			return err
 		}
 	}
-	for _, jctx := range overallSchedulerResult.PreemptedJobs {
```
severinson, line 375: Add a comment explaining why we're not updating preempted here.

internal/scheduler/metrics/metrics.go
Mo-Fatah: because UpdatedPreempted needed jctx. I removed jctx now it as it no longer needed for both UpdateFailed and UpdatePreempted

severinson (approved):

richscott (approved):

PR #3419, opened 2024-02-22
Better Resolved Query API

d80tb7:
Refactor of the query API to be a bit more useful.  Specifically:

* Update `GetJobStatus`  rpc to fetch statuses of multiple jobs
* Add `GetJobDetails` rpc to fetch information about multiple jobs by job id
* Add `GetJobRunDetails` rpc to fetch information about multiple job runs by run id
* Added a great deal more testing
* Moved the proto from `queryAPI` to `job` (there is an argument that at some point we should comobine with `submit`
* Excluded the queryapi from the C# client as the api is not yet stable

Note that this API is still experimental and not for production use


internal/armada/queryapi/database/query.sql
```diff
@@ -1,2 +1,12 @@
...
+SELECT job_id, state FROM job WHERE job_id = ANY(sqlc.arg(job_ids)::text[]);
+
+-- name: GetJobDetails :many
+SELECT job_id, queue, jobset, namespace, state, submitted, cancelled, cancel_reason, last_transition_time, latest_run_id, job_spec FROM job WHERE job_id = ANY(sqlc.arg(job_ids)::text[]);
+
+-- name: GetJobRunsByRunIds :many
+SELECT * FROM job_run WHERE run_id = ANY(sqlc.arg(run_ids)::text[]);
+
+-- name: GetJobRunsByJobIds :many
+SELECT * FROM job_run WHERE job_id = ANY(sqlc.arg(job_ids)::text[]) order by leased  desc;
```
JamesMurkin, line 11: Does this need the order by?

internal/armada/queryapi/query_api.go
```diff
@@ -2,51 +2,208 @@ package queryapi
...
+	for _, jobId := range req.JobIds {
+		dbStatus, ok := dbStatusById[jobId]
+		if ok {
+			apiStatus, ok := JobStateMap[dbStatus]
+			if !ok {
+				apiStatus = api.JobState_UNKNOWN // We know about this job but we can't map its state
+			}
+			apiStatusById[jobId] = apiStatus
+		} else {
+			apiStatusById[jobId] = api.JobState_UNKNOWN // We don't know about this job
```
JamesMurkin, line 184: Will we ever need a way to differentiate between these 2 cases? I guess in theory the first case should never happen

JamesMurkin (approved):

internal/armada/queryapi/database/query.sql
d80tb7: Yes- two reasons:
1.  It means that the job runs appear in chronological order (i.e. the active run will be first)
2. It makes unit testing much easier becuase you have a deterministic result.

The performance overhead of this *should* be be negligable unless we are retrieving a very large number of runs.

internal/armada/queryapi/query_api.go
d80tb7: The first case should really be an error and I might move it to be one once I'm more confident.  For now I'll just log a warning.

PR #3425, opened 2024-02-27
Split scheduler metrics by resource

Mo-Fatah:


internal/scheduler/configuration/configuration.go
```diff
@@ -73,6 +73,10 @@ type MetricsConfig struct {
 	TrackedErrorRegexes []string
 	// Metrics are exported for these resources.
 	TrackedResourceNames []v1.ResourceName
+	// Some resource names aren't generic and can be attached to a specific vendor.
+	// e.g. "nvidia.com/gpu" is a vendor specific resource name.
+	// ResourceRenaming maps these vendor specific resource names to a generic name to be used in the exported metrics.
```
severinson, line 78:
```suggestion
	// Optionally rename resources in exported metrics.
	// E.g., if ResourceRenaming["nvidia.com/gpu"] = "gpu", then metrics for resource "nvidia.com/gpu" use resource name "gpu" instead.
	// This can be used to avoid illegal Prometheus metric names (e.g., for "nvidia.com/gpu" as "/" is not allowed).
	// It can also be used to track multiple resources within the same metric, e.g., "nvidia.com/gpu" and "amd.com/gpu".
```

internal/scheduler/metrics/metrics.go
```diff
@@ -59,8 +59,19 @@ type Metrics struct {
 	matchedRegexIndexByErrorMessage *lru.Cache
 
 	// Job metrics.
-	transitions     *prometheus.CounterVec
-	resourceSeconds *prometheus.CounterVec
+	jobs             *prometheus.CounterVec
```
severinson, line 62: Make this a map instead. We can't know in advance which resources exist within the system as nodes may export arbitrary resources.

internal/scheduler/metrics/metrics.go
```diff
@@ -59,8 +59,19 @@ type Metrics struct {
...
+	cpu              *prometheus.CounterVec
+	memory           *prometheus.CounterVec
+	gpu              *prometheus.CounterVec
+	storage          *prometheus.CounterVec
+	ephemeralStorage *prometheus.CounterVec
+
+	jobsSeconds             *prometheus.CounterVec
```
severinson, line 69: Should also be a map.

internal/scheduler/metrics/metrics.go
```diff
@@ -477,72 +577,72 @@ func errorTypeAndMessageFromError(ctx *armadacontext.Context, err *armadaevents.
...
 	return nil
 }
 
-func stateDuration(start *time.Time, end *time.Time) time.Duration {
-	if start != nil && end != nil {
-		return end.Sub(*start)
+// maps resource names to their respective counter vectors
+// e.g. cpu -> m.cpu, m.cpuSeconds
+func (m *Metrics) resourceToMetrics(resourceName v1.ResourceName) (*prometheus.CounterVec, *prometheus.CounterVec) {
+	switch resourceName {
```
severinson, line 623: Once you're storing the counters in a map you won't need this switch anymore.

config/scheduler/config.yaml
```diff
@@ -24,6 +24,10 @@ schedulerMetrics:
     - "memory"
     - "ephemeral-storage"
     - "nvidia.com/gpu"
+  resourceRenaming:
+    nvidia.com/gpu: "gpu"
+    amd.com/gpu: "gpu"
+    ephemeral-storage: "ephemeral_storage"
```
severinson, line 30: Is `ephemeralStorage` more idiomatic? I don't know what people normally do in Prometheus.

internal/scheduler/configuration/configuration.go
```diff
@@ -73,6 +73,13 @@ type MetricsConfig struct {
 	TrackedErrorRegexes []string
 	// Metrics are exported for these resources.
 	TrackedResourceNames []v1.ResourceName
+	// Optionally rename resources in exported metrics.
+	// E.g., if ResourceRenaming["nvidia.com/gpu"] = "gpu", then metrics for resource "nvidia.com/gpu" use resource name "gpu" instead.
+	// This can be used to avoid illegal Prometheus metric names (e.g., for "nvidia.com/gpu" as "/" is not allowed).
+	// Allowed characters in resource names are [a-zA-Z_:][a-zA-Z0-9_:]*
+	// If a tracked resource contains illegal characters, it is should be renamed here.
```
severinson, line 80: I'd remove this line. Seems redundant.

internal/scheduler/metrics/metrics.go
```diff
@@ -477,72 +459,91 @@ func errorTypeAndMessageFromError(ctx *armadacontext.Context, err *armadaevents.
...
 	}
 	return nil
 }
 
-func stateDuration(start *time.Time, end *time.Time) time.Duration {
-	if start != nil && end != nil {
-		return end.Sub(*start)
+// getOrCreateMetrics returns the counter and counterSeconds Vec for the given resource name.
+// If the counter and counterSeconds Vecs do not exist, they are created and stored in the resourceCounters map.
+func (m *Metrics) getOrCreateMetrics(resource v1.ResourceName) (*prometheus.CounterVec, *prometheus.CounterVec) {
```
severinson, line 507:
```suggestion
func (m *Metrics) counterVectorsFromResource(resource v1.ResourceName) (*prometheus.CounterVec, *prometheus.CounterVec) {
```

severinson: You can do this also for jobs to avoid creating it eagerly in New(). A bit cleaner.

internal/scheduler/configuration/configuration.go
```diff
@@ -73,6 +73,13 @@ type MetricsConfig struct {
...
+	// It can also be used to track multiple resources within the same metric, e.g., "nvidia.com/gpu" and "amd.com/gpu".
+	ResourceRenaming map[v1.ResourceName]string
```
severinson, line 82: We're using the validator package for config validation elsewhere. https://pkg.go.dev/github.com/go-playground/validator/v10#section-readme
Could you use this pre-defined regex to validate that these metric names are valid? https://github.com/go-playground/validator/blob/v10.19.0/regexes.go#L7

Mo-Fatah: this doesn't include underscores or colons https://github.com/go-playground/validator/blob/a0f74b0fb2a7ae1750c0f0b0a49550d8b6e2e708/regexes.go#L7

Also, it allows a metric that starts with a digit, which prometheus will reject later.  

most tracked resources doesn't include underscores or colons themselves, however this might reject valid resource names in the future. 

The regex i added acts the same as prometheus naming check (in the current version we are using ) https://github.com/prometheus/common/blob/49b36038aead362ce78f09337dc5872000bd048a/model/metric.go#L97

severinson (approved):

richscott (approved):

PR #3427, opened 2024-02-28
Updated failure estimator

severinson:
This PR introduces a new learning algorithm for node and queue failure probability estimation. Simulations indicate this new method is much more accurate than the one we use now.

Adding the gonum package as a dependency resulted in a version increase of the slices package we use. As a result, I had to update all calls to `slices.SortFunc` as the API had changed.

The snyk error reported seemingly only affects image decoding. It shouldn't matter to us.


internal/common/interfaces/interfaces.go
```diff
@@ -1,13 +1,20 @@
...
 
-// Equaler expresses that objects can be compared for equality via the Equals method.
+// Equaler represents objects can be compared for equality via the Equals method.
 type Equaler[T any] interface {
 	// Returns true if both objects are equal.
 	Equal(T) bool
 }
+
+// Number represents any integer or floating-point number.
+type Number interface {
```
severinson, line 18: Curiously this interface isn't in the default collection.

internal/scheduler/failureestimator/simulator.jl
```diff
@@ -0,0 +1,352 @@
+using Random
```
severinson, line 1: I've included my simulator here for anyone working on this in the future.

d80tb7 (approved):

internal/scheduler/database/executor_repository_test.go
```diff
@@ -61,8 +61,14 @@ func TestExecutorRepository_LoadAndSave(t *testing.T) {
 				}
 				retrievedExecutors, err := repo.GetExecutors(ctx)
 				require.NoError(t, err)
-				executorSort := func(a *schedulerobjects.Executor, b *schedulerobjects.Executor) bool {
-					return a.Id > b.Id
+				executorSort := func(a *schedulerobjects.Executor, b *schedulerobjects.Executor) int {
```
severinson, line 64: We have several uses of `SortFunc` where the sorting is opposite that of what I would think the intuitive one.

d80tb7 (dismissed):

internal/common/optimisation/optimisation.go
```diff
@@ -0,0 +1,8 @@
+package optimisation
+
+import "gonum.org/v1/gonum/mat"
+
+type Optimiser interface {
+	Update(out, p *mat.VecDense, g mat.Vector) *mat.VecDense
```
robertdavidsmith, line 8: Better names for parameters? It's not clear what these do from the signature.

severinson: Fixed. I left p and g as-is in the implementations as I think the long names make the code a bit less readable there.

internal/scheduler/failureestimator/failureestimator.go
```diff
@@ -84,88 +75,41 @@ type FailureEstimator struct {
...
-	// Prevents concurrent map modification issues when scraping metrics.
+	// Prevents concurrent map modification when scraping metrics.
 	mu sync.Mutex
 }
 
-// New returns a new FailureEstimator. Parameters have the following meaning:
-// - {node, queue}SuccessProbabilityCordonThreshold: Success probability below which nodes (queues) are considered unhealthy.
-// - {node, queue}CordonTimeout: Amount of time for which nodes (queues) remain unhealthy in the absence of any job successes or failures for that node (queue).
-// - {node, queue}EquilibriumFailureRate: Job failures per second necessary for a node (queue) to remain unhealthy in the absence of any successes for that node (queue).
+type Update struct {
```
robertdavidsmith: better names/comment?

robertdavidsmith (approved):

PR #3434, opened 2024-03-01
Strip annotations prefix in `CreateUserAnnotationInstruction`

zuqq:
My mistake was not passing in `annotations` as the last argument here:

https://github.com/armadaproject/armada/blob/8e523538f38596d3ff798771559764cdf0d9edd0/internal/lookoutingesterv2/instructions/instructions.go#L215

Unfortunately the tests didn't catch this, because they weren't looking at all the fields of `InstructionSet`.

### DIFF
--- a/internal/lookoutingesterv2/instructions/instructions_test.go
+++ b/internal/lookoutingesterv2/instructions/instructions_test.go
@@ -208,6 +208,23 @@ func TestConvert(t *testing.T) {
 		},
 	}
 
+	expectedCreateUserAnnotations := []*model.CreateUserAnnotationInstruction{
+		{
+			JobId:  testfixtures.JobIdString,
+			Key:    "a",
+			Value:  "0",
+			Queue:  testfixtures.Queue,
+			Jobset: testfixtures.JobSetName,
+		},
+		{
+			JobId:  testfixtures.JobIdString,
+			Key:    "b",
+			Value:  "1",
+			Queue:  testfixtures.Queue,
+			Jobset: testfixtures.JobSetName,
+		},
+	}
+
 	otherJobIdUlid := util.ULID()
 	otherJobId := util.StringFromUlid(otherJobIdUlid)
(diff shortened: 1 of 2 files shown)


internal/lookoutingesterv2/instructions/instructions_test.go
```diff
@@ -598,6 +622,8 @@ func TestConvert(t *testing.T) {
 			assert.Equal(t, tc.expected.JobsToUpdate, instructionSet.JobsToUpdate)
 			assert.Equal(t, tc.expected.JobRunsToCreate, instructionSet.JobRunsToCreate)
 			assert.Equal(t, tc.expected.JobRunsToUpdate, instructionSet.JobRunsToUpdate)
+			assert.Equal(t, tc.expected.UserAnnotationsToCreate, instructionSet.UserAnnotationsToCreate)
+			assert.Equal(t, tc.expected.MessageIds, instructionSet.MessageIds)
```
zuqq, line 626: This is another field of `InstructionSet` that this test suite didn't check.

internal/lookoutingesterv2/instructions/instructions_test.go
```diff
@@ -675,33 +701,6 @@ func TestTruncatesStringsThatAreTooLong(t *testing.T) {
...
-			Queue:  testfixtures.Queue,
-			Jobset: testfixtures.JobSetName,
-		},
-	}
-	instructions := createUserAnnotationInstructions(
-		testfixtures.JobIdString,
-		testfixtures.Queue,
-		testfixtures.JobSetName,
-		extractUserAnnotations(userAnnotationPrefix, annotations),
-	)
```
zuqq, line 701: I think that we should just get rid of this test, because it reimplements part of `handleSubmitJob` (which means that it's bound to diverge from the actual code in production, just like it did here).

d80tb7 (approved):

PR #3435, opened 2024-03-01
Improved NodeTypeIterator loop detection

severinson:


internal/scheduler/nodedb/nodeiteration.go
```diff
@@ -377,10 +389,18 @@ func (it *NodeTypeIterator) NextNode() (*Node, error) {
 			return nil, nil
 		}
 		node := v.(*Node)
-		if node.Id == it.previousNodeId {
-			panic(fmt.Sprintf("iterator received the same node twice consecutively: %s", node.Id))
+		if it.keyIndex >= len(node.Keys) {
+			return nil, errors.Errorf("keyIndex is %d, but node %s has only %d keys", it.keyIndex, node.Id, len(node.Keys))
+		}
+		nodeKey := node.Keys[it.keyIndex]
+		if it.previousKey != nil && bytes.Compare(it.previousKey, nodeKey) != -1 {
```
severinson, line 396: This is the new loop detection check.

internal/scheduler/nodedb/nodedb.go
```diff
@@ -1204,38 +1223,40 @@ func (nodeDb *NodeDb) AddEvictedJobSchedulingContextWithTxn(txn *memdb.Txn, inde
...
 	indexNameByPriority := make(map[int32]string, len(priorities))
+	keyIndexByPriority := make(map[int32]int, len(priorities))
 	for i, priority := range priorities {
 		name := nodeIndexName(i)
 		indexNameByPriority[priority] = name
+		keyIndexByPriority[priority] = i
 		indexes[name] = &memdb.IndexSchema{
 			Name:    name,
-			Unique:  false,
+			Unique:  true,
```
severinson, line 1252: Enforce that the nodeDb keys are unique. We can do this now since we've associated a unique index with each node that is appended to the database keys associated with the node.

d80tb7 (approved):

PR #3438, opened 2024-03-02
Creating Tests for the function submitJobs

Bharadwajshivam28:
refs #3390 

Hey @richscott  I am creating Test for the function [submitJobs](https://github.com/armadaproject/armada/blob/ed0fd4634b5ae3553f673ef346bfc39b680ebec4/internal/executor/job/submit.go#L58)

### DIFF
--- a/internal/executor/job/submit.go
+++ b/internal/executor/job/submit.go
@@ -53,7 +53,7 @@ type FailedSubmissionDetails struct {
 
 func (submitService *SubmitService) SubmitJobs(jobsToSubmit []*SubmitJob) []*FailedSubmissionDetails {
 	return submitService.submitJobs(jobsToSubmit)
-}
+} 
 
 func (submitService *SubmitService) submitJobs(jobsToSubmit []*SubmitJob) []*FailedSubmissionDetails {
 	wg := &sync.WaitGroup{}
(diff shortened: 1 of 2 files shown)


internal/executor/job/submit.go
```diff
@@ -53,7 +53,7 @@ type FailedSubmissionDetails struct {
 
 func (submitService *SubmitService) SubmitJobs(jobsToSubmit []*SubmitJob) []*FailedSubmissionDetails {
 	return submitService.submitJobs(jobsToSubmit)
-}
+}
```
richscott, line 56: Please remove this change - there shouldn't be extra whitespace at the end of lines. (I think running `golangci-lint` will warn about things like this, and using `go fmt file_xxx.go` will fix it for you.

internal/executor/job/submit.go
```diff
@@ -110,7 +110,7 @@ func (submitService *SubmitService) submitWorker(wg *sync.WaitGroup, jobsToSubmi
 // submitPod submits a pod to k8s together with any services and ingresses bundled with the Armada job.
 // This function may fail partly, i.e., it may successfully create a subset of the requested objects before failing.
 // In case of failure, any already created objects are not cleaned up.
-func (submitService *SubmitService) submitPod(job *SubmitJob) (*v1.Pod, error) {
+func (submitService *SubmitService) submitPod(job *SubmitJob) (*v1.Pod, error) { //iska krna h
```
richscott, line 113: This looks like a leftover note or typo - please remove it.

internal/executor/job/submit_test.go
```diff
@@ -94,3 +94,30 @@ func newArbitraryError(message string) error {
 func newArmadaErrCreateResource() error {
 	return &armadaerrors.ErrCreateResource{}
 }
+
+func TestSubmitJobs(t *testing.T){
+	submitService := &SubmitService{
```
richscott, line 99: I think running `go fmt` on this file would collapse the two lines of the SubmitService reference into one, since there are no members declared in it.

internal/executor/job/submit.go
Bharadwajshivam28: Thanks @richscott I will make these changes

internal/executor/job/submit.go
Bharadwajshivam28: Oh I missed it to remove

richscott:
@Bharadwajshivam28 Can you rebase this branch from `master`, and then we can verify all issues are addressed and checks succeed? Thanks.

PR #3454, opened 2024-03-11
Fix Scheduling Weights

d80tb7:
In  #3429 we introduced a bug whereby we mixed up a weight and priority factor.  As weights are 1/priority_factor this has led to all queue weights being inverted.

This PR addresses this issue and also adds a test that will allow us to spot this if it ever happens again.

### DIFF
--- a/internal/scheduler/testfixtures/testfixtures.go
+++ b/internal/scheduler/testfixtures/testfixtures.go
@@ -790,7 +790,7 @@ func Test1Node32CoreExecutor(executorId string) *schedulerobjects.Executor {
 func MakeTestQueue() queue.Queue {
 	return queue.Queue{
 		Name:           TestQueue,
-		PriorityFactor: 0.01,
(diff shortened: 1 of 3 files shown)


internal/scheduler/testfixtures/testfixtures.go
```diff
@@ -790,7 +790,7 @@ func Test1Node32CoreExecutor(executorId string) *schedulerobjects.Executor {
 func MakeTestQueue() queue.Queue {
 	return queue.Queue{
 		Name:           TestQueue,
-		PriorityFactor: 0.01,
+		PriorityFactor: 100,
```
d80tb7, line 793: This had also changed in the previous PR.  PriorityFactors are generally > 1 and weights (i.e. 1/PF) are generally fractional.  For all tests the exact value ere doesn't matter, but semantically it makes more sense to change this back to how is was previously.

samclark (approved):

PR #3458, opened 2024-03-13
Log warning on unused config keys

severinson:
This PR introduces a warning printed at startup if there are config keys in the yaml that don't match a config item in the struct the yaml is decoded into.

For example, adding the following unused keys to the server scheduling config:
```
  myUnusedField: "abc"
  myOtherUnusedField: 1.0
  myThirdUnusedField: true
  myMapUnusedField:
    - foo: "bar"
```

Results in a warning being printed at startup:
```
> go run .\cmd\armada\main.go
INFO[2024-03-13T11:12:41.828Z]startup.go:53 Read base config from C:\Users\albin\git\armada\config\armada\config.yaml
WARN[2024-03-13T11:12:41.83Z]startup.go:83 Unused keys: [Scheduling.mythirdunusedfield Scheduling.mymapunusedfield Scheduling.myunusedfield Scheduling.myotherunusedfield]
INFO[2024-03-13T11:12:41.83Z]main.go:47 Starting...
```

It also logs the set and unset config keys at a debug level.

### DIFF
--- a/internal/common/startup.go
+++ b/internal/common/startup.go
@@ -11,6 +11,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/mitchellh/mapstructure"
+	"golang.org/x/exp/slices"
+
 	"github.com/armadaproject/armada/internal/common/certs"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -39,7 +42,7 @@ func BindCommandlineArguments() {
 }
 
 // TODO Move code relating to config out of common into a new package internal/serverconfig
-func LoadConfig(config interface{}, defaultPath string, overrideConfigs []string) *viper.Viper {
+func LoadConfig(config any, defaultPath string, overrideConfigs []string) *viper.Viper {
 	v := viper.NewWithOptions(viper.KeyDelimiter("::"))
 	v.SetConfigName(baseConfigFileName)
 	v.AddConfigPath(defaultPath)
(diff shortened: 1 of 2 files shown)


internal/common/startup.go
```diff
@@ -63,12 +69,26 @@ func LoadConfig(config interface{}, defaultPath string, overrideConfigs []string
...
-	err := v.Unmarshal(config, commonconfig.CustomHooks...)
-	if err != nil {
+	var metadata mapstructure.Metadata
+	customHooks := append(slices.Clone(commonconfig.CustomHooks), func(c *mapstructure.DecoderConfig) { c.Metadata = &metadata })
+	if err := v.Unmarshal(config, customHooks...); err != nil {
 		log.Error(err)
 		os.Exit(-1)
 	}
 
+	// Log a warning if there are config keys that don't match a config item in the struct the yaml is decoded into.
```
severinson, line 79: This is the new bit that logs warnings.

config/armada/config.yaml
```diff
@@ -98,17 +98,10 @@ submission:
   defaultActiveDeadline: "72h" # 3 days.
   defaultActiveDeadlineByResourceRequest:
     nvidia.com/gpu: "336h" # 14 days.
-eventRetention:
```
severinson, line 101: All config changes in this PR are to remove yaml that doesn't match any config defined in the corresponding struct.

masipauskas (approved):

internal/common/startup.go
severinson: There's an argument for just failing to start instead.

PR #3463, opened 2024-03-18
Log where we mutate containers with inconsistent requests/limits

d80tb7:
We have some code that looks to see if a user has specified a resource in requests but not limits or vice versa.  If they have then the missing resource value is automatically filled in.  A practical example here would be if a user has specified a 1 core cpu limit but does not specify a cpu request.  In  this case we would autopopulate a 1 core cpu request.

I'm largely against this logic because:
- It has a somewhat complex interaction with the other defaulting/checking we do
- It will hide mistakes from the user.

Ideally I think we'd either a) remove this completely, or b) replace with something simpler where we will populate an entire missing requests/limits section but not individual resources.

As this would be a breaking api change, I've added a log line so we can determine if this feature is actually being used.

### DIFF
--- a/internal/armada/server/applydefaults.go
+++ b/internal/armada/server/applydefaults.go
@@ -131,7 +131,6 @@ func applyDefaultActiveDeadlineSecondsToPodSpec(spec *v1.PodSpec, config configu
 // memory limit, but does not specify a memory request, assign a memory request that matches the limit.
 // Similarly, if a Container specifies its own CPU limit, but does not specify a CPU request, automatically
 // assigns a CPU request that matches the limit.
-//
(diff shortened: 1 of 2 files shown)


kannon92:
Another reason to be against this is that not specifying limits is a valid case in kubernetes. Pods get assigned a QoS based on if limits=requests (guaranteed QoS) or request != limits (burstableable) or best effort in all other cases.

This influences the oom_score for OOMKiller so setting limits to equal requests means you are forcing the guaranteed Qos. 

https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#node-out-of-memory-behavior

JamesMurkin (approved):

PR #3468, opened 2024-03-21
Fix jobrun conflicting terminal states

Mo-Fatah:


internal/scheduler/jobdb/job.go
```diff
@@ -604,7 +604,11 @@ func (job *Job) ActiveRunTimestamp() int64 {
 
 // InTerminalState returns true if the job  is in a terminal state
 func (job *Job) InTerminalState() bool {
-	return job.succeeded || job.cancelled || job.failed
+	isRunInTerminal := false
+	if latestRun := job.LatestRun(); latestRun != nil {
+		isRunInTerminal = latestRun.Succeeded() || latestRun.Cancelled() || latestRun.Failed() || latestRun.Preempted()
```
Mo-Fatah, line 609: I thought that if we have the latest run reached a terminal state, we should also report the job as in terminal state and return `true`.

severinson (approved):

richscott (approved):

severinson:
This PR ensures jobs only have one terminal state associated with them. The scheduler should decide whether a job is succeeded, preempted, etc. But should decide on only one of these. In the past, we often had multiple terminal states set on the same job.

PR #3474, opened 2024-03-26
Improve executor event batching

JamesMurkin:
Currently the executor attempts to batch events it sends to the Executor API

The issue is it doesn't wait for a batch of events to accumulate, so it typically will only send events that occurred at almost the exact same time in a batch

This is bad as it means the executor makes many requests to the ExecutorAPI. This causes optimisations such as compressing event sequences to add more overhead than they reduce

We will now send all events in the buffer every 2 seconds OR instantly if we reach 200 events in the buffer
 - Under high load we'll just send lots of batches of 200, to prevent a big backlog growing
 - Under lighter load, we'll send all events that occurred in a 2 second interval, to allow us to group these into bigger messages


internal/executor/reporter/job_event_reporter.go
```diff
@@ -34,9 +33,17 @@ type JobEventReporter struct {
...
+	clock            clock.Clock
+	maxBatchSize     int
 }
 
-func NewJobEventReporter(clusterContext clusterContext.ClusterContext, jobRunState *job.JobRunStateStore, eventSender EventSender) (*JobEventReporter, chan bool) {
+func NewJobEventReporter(
+	clusterContext clusterContext.ClusterContext,
+	jobRunState *job.JobRunStateStore,
+	eventSender EventSender,
+	clock clock.Clock,
```
d80tb7, line 44: I'd remove the clock parameter here.  The only reason you'll need to set it is for unit tests and for that you are probably in the same package so you can just override the field directly.

internal/executor/reporter/job_event_reporter_test.go
```diff
@@ -79,7 +80,8 @@ func TestJobEventReporter_SendsPreemptionEvent(t *testing.T) {
 			_, executorContext, _, eventSender := setupTest(t, tc.podsInJobRunState)
 			preemptedClusterEvent := createPreemptedClusterEvent(preemptedPod, preemptivePod)
 			executorContext.SimulateClusterAddEvent(preemptedClusterEvent)
-			// Event processing is async, sleep shortly to give it time to process
+
+			// Event processing is async, give time for event to be processed
 			time.Sleep(time.Millisecond * 100)
```
d80tb7, line 85: you didn't add this, but the sleep here is mildly concerning!

internal/executor/reporter/job_event_reporter_test.go
```diff
@@ -104,7 +106,51 @@ func TestJobEventReporter_SendsPreemptionEvent(t *testing.T) {
...
 }
 
-func setupTest(t *testing.T, existingPods []*v1.Pod) (EventReporter, *fakecontext.SyncFakeClusterContext, *job.JobRunStateStore, *FakeEventSender) {
+func TestJobEventReporter_SendsEventImmediately_OnceNumberOfWaitingEventsMatchesBatchSize(t *testing.T) {
+	jobEventReporter, eventSender, _ := setupBatchEventsTest(2)
+	pod1 := createPod(1)
+	pod2 := createPod(2)
+
+	jobEventReporter.QueueEvent(EventMessage{CreateSimpleJobFailedEvent(pod1, "failed", "cluster1", api.Cause_Error), util.ExtractJobRunId(pod1)}, func(err error) {})
+	time.Sleep(time.Millisecond * 100) // Allow time for async event processing
```
d80tb7, line 115: I think the better way to do this would be to block until you've seen the expected event or some timeout has occurred.  That way you could set a long timeout (like 10 seconds) and in the common case (i.e. the test passes) you'll only be blocking for single digit millis.

That would probably take a bit of reworking because you'd need to do something like make the FakeEventSender put the events on a channel.  Happy to approve this as is and then we can rework it later.

d80tb7 (approved):

internal/executor/reporter/job_event_reporter.go
JamesMurkin: Sounds sensible, but I'd need to swap it to have a Run func as the clock gets used in the constructor.

internal/executor/reporter/job_event_reporter_test.go
JamesMurkin: It is a bit awkward as you are waiting on something being added to an internal buffer - there isn't always going to be an external thing you can wait for

We could add a onEventProcessed() or something, you could then use in testing to wait on

PR #3480, opened 2024-03-27
Update docs

severinson:


dave-gantenbein (changes requested):
All of the comparions here are against single cluster schedulers, so you're comparing apples to radishes. Kueue has a very beta multi cluster feature now, MCAD exists, etc. I'm not sure this adds any value to the documentation as currently written.

PR #3481, opened 2024-03-28
Migrate executor to produce pulsar events

JamesMurkin:


internal/executor/job/processors/preempt_runs.go
```diff
@@ -74,14 +74,20 @@ func (j *RunPreemptedProcessor) Run() {
 }
 
 func (j *RunPreemptedProcessor) reportPodPreempted(run *job.RunState, pod *v1.Pod) error {
-	preemptedEvent := reporter.CreateSimpleJobPreemptedEvent(pod, j.clusterContext.GetClusterId())
-	failedEvent := reporter.CreateSimpleJobFailedEvent(pod, "Run preempted", j.clusterContext.GetClusterId(), api.Cause_Error)
+	preemptedEvent, err := reporter.CreateSimpleJobPreemptedEvent(pod)
+	if err != nil {
+		return fmt.Errorf("failed creating preempted event because - %s", err)
+	}
+	failedEvent, err := reporter.CreateSimpleJobFailedEvent(pod, "Run preempted", j.clusterContext.GetClusterId(), armadaevents.KubernetesReason_AppError)
```
d80tb7, line 81: we should make it so that we don't need to create both preempted and failed events.  That's not for today though!

d80tb7 (approved):

PR #3503, opened 2024-04-06
Refactor Of Submit Endpoint

d80tb7:
This PR  begins the refactor of the `Submit` endpoint to (hopefully) make it simpler and more consistent.  Note that this PR just concerns itself with the `SubmitJobs` method.  The methods that deal with  cancelling and reprioritising jobs  will be changed in a subsequent PR.  The PR itslelf is rather large but here is a summary of the main changes.

- The endpoint has been moved into its own package (`submit`) this is in anticipation of having one package per endpoint.
- Deduplication and Publishing functionality have been moved into separate structs  (`deduplicator` and `publisher` respectively).  This is to both to aid testing and to reduce the amount of logic in the main `submit` struct.
- We no longer create an intermediate `api.Job` but rather convert straight from an `api.JobSubmitRequest` to an `armadaevents.EventSequence`.
- Validation is now organised as a series of validation functions which are applied in turn.  Validation is performed on the raw  `api.JobSubmitRequest` before modification.
- Conversion from `api.JobSubmitRequest` to `armadaevents.EventSequence` is now handled internally by a subpackage of `submit`.   
- ApplyDefaults has been renamed `postprocess` and has been given a structure similar to validation whereby a se  of post processors are applied in sequeence.  PostProcessing is done on a `armadaevents.SubmitJob` message.


internal/scheduler/submitcheck.go
```diff
@@ -134,11 +135,35 @@ func (srv *SubmitChecker) updateExecutors(ctx *armadacontext.Context) {
 	srv.jobSchedulingResultsCache.Purge()
 }
 
-func (srv *SubmitChecker) CheckApiJobs(jobs []*api.Job) (bool, string) {
+func (srv *SubmitChecker) CheckJobDbJobs(jobs []*jobdb.Job) (bool, string) {
 	return srv.check(schedulercontext.JobSchedulingContextsFromJobs(srv.priorityClasses, jobs))
 }
 
-func (srv *SubmitChecker) CheckJobDbJobs(jobs []*jobdb.Job) (bool, string) {
+func (srv *SubmitChecker) CheckApiJobs(es *armadaevents.EventSequence) (bool, string) {
```
d80tb7, line 142: This is a;l a bit messy, but I think we should refactor the way that the submitCheck works ina subsequent PR

internal/scheduler/submitcheck.go
```diff
@@ -87,6 +87,8 @@ func NewSubmitChecker(
 	}
 }
 
+var stringInterner = stringinterner.New(1000)
```
JamesMurkin, line 90: Should this only be 1000?

internal/scheduler/jobdb/jobdb.go
```diff
@@ -59,30 +56,22 @@ func (_ RealUUIDProvider) New() uuid.UUID {
...
 	)
 }
 
 func NewJobDbWithSchedulingKeyGenerator(
 	priorityClasses map[string]types.PriorityClass,
-	defaultPriorityClassName string,
 	skg *schedulerobjects.SchedulingKeyGenerator,
 	stringInterner *stringinterner.StringInterner,
 ) *JobDb {
-	defaultPriorityClass, ok := priorityClasses[defaultPriorityClassName]
```
JamesMurkin, line 74: Does something prevent invalid config to protect us here?

internal/armada/submit/submit.go
```diff
@@ -0,0 +1,822 @@
...
+			return armadaevents.MustProtoUuidFromUlidString(util.NewULID())
+		},
+	}
+}
+
+// SubmitJobs allows users to submit jobs to Armada.  On receipt of a request, the following actions are performed:
+//   - The request is validated to make sure it is well formed.  If this fails ten an error is returned
+//   - Each JobRequestItem inside the request is checked to see if it is a duplicate
+//   - Each non-duplicate is converted into an armadaevents.SubmitMessage
+//   - All SubmitMessages are checked to see if the job they define can be scheduled (an example of a job tht cannot
```
JamesMurkin, line 80:
```suggestion
//   - All SubmitMessages are checked to see if the job they define can be scheduled (an example of a job that cannot
```

internal/scheduler/submitcheck.go
d80tb7: it shouldn't matter too much as it's an LRU so it'll cache the last 10k strings.  We're not really getting anything from the string interner here as these strings are already in memory and only live for the lifetime of the request (which is hopefully measure in millis)- I needed it purely because the interface desires it.

Happy to make it 10K or 100K though if you prefer.

internal/scheduler/jobdb/jobdb.go
d80tb7: it's a fair point.  Let me see if I can put something like this back.

JamesMurkin (approved):

PR #3548, opened 2024-04-27
Fix the performance of LimitSequenceByteSize

JamesMurkin:
I changed this function recently (https://github.com/armadaproject/armada/pull/3483) as it was previously incorrectly not limiting the byte sequence in all cases

Unfortunately this change caused the complexity of the function to go from O(n) -> O(n^2), where n is the number of sequence.Events

This PR changes the complexity back to O(n) while still keeping the correct limit check

We now estimate the byte overhead added by the Event slice using the constant `sequenceEventListOverheadSizeBytes` - which is a very conservative estimate
 - This does mean the function is now less useful for very small limits, but that isn't a problem for us in practice. Typically we limit to 4Mb, so 100 byte overhead is insignificant
 - I've added a test to ensure the estimate of `sequenceEventListOverheadSizeBytes` remains valid

This change means the time to process an EventSequence with 170k events (approximately our worst case when using a 4Mb limit) goes from 3m10s -> 22.5ms (on my AWS machine)

The motivation for this change was that large events going through our system were causing significant pauses
 - This impacted the scheduler and ingesters, causing widespread dispruption on large events (cancel/repioritize)

### DIFF
--- a/internal/common/eventutil/eventutil_test.go
+++ b/internal/common/eventutil/eventutil_test.go
@@ -646,10 +646,10 @@ func TestSequenceEventListSizeBytes(t *testing.T) {
 	}
 
 	sequenceSizeBytes := uint(proto.Size(sequence))
-	// If this fails, it means that the sequenceEventListSizeBytes constant is possibly too small
+	// If this fails, it means that the sequenceEventListOverheadSizeBytes constant is possibly too small
 	// We are showing our safe estimate of the byte overhead added by the event list in proto is definitely large enough
 	//  by showing it is larger than a sequence with a single event (as that sequence contains the overhead added by the event list)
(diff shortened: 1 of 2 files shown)


internal/common/eventutil/eventutil_test.go
```diff
@@ -655,7 +685,6 @@ func TestLimitSequenceByteSize(t *testing.T) {
 
 	_, err = LimitSequenceByteSize(sequence, 1, false)
 	assert.NoError(t, err)
-	assert.Equal(t, []*armadaevents.EventSequence{sequence}, actual)
```
JamesMurkin, line 658: Removing this as it is the same as the assert on line 651, and the two variables being compared haven't changed between this line and 651

d80tb7 (approved):

robertdavidsmith (approved):

Issue #3553, opened 2024-04-29
Job Service subscription manager logic does not handle errors on job event message stream correctly


Sovietaced:
If the [job service subscription manager goroutine](https://github.com/armadaproject/armada/blob/master/internal/jobservice/eventstojobs/manage_subs.go#L271-L338) encounters a server error (besides an EOF) from the Armada API it will be stuck in a terminal state reading the same error from the stream and writing it to the database, endlessly.  That is because a stream error is terminal. The server ends the stream with an error and any calls to read from the stream will return the same error over and over again. This is a common mistake with gRPC stream client logic unfortunately.

See: https://groups.google.com/g/grpc-io/c/XcN4hA9HonI

> Client-side, io.EOF is returned directly from grpc-go steram.Send() calls when the server closes the stream.  At that point, Recv() should be called until it returns a non-nil error to receive the server's messages and eventually the status.  If io.EOF is returned by Recv(), that indicates a success; otherwise that is the final status of the RPC.  All errors are permanent from the standpoint of that stream -- subsequent streams are possible using the same client, even if it is a connection error, as the client automatically attempts to reconnect.  Make sure you always call Recv() on the stream until you get a non-nil error or cancel the context used when creating the stream, or else a goroutine and other resources will be leaked.

**Files Or Markdown that can reproduce the issue**

This diff will update the mock server to behave like a real gRPC server, which returns the terminal error over and over again.
https://github.com/armadaproject/armada/commit/a385e07e6a8c1471a3dfc77e1bbe04a7d4f9b3c5

Making this change will fail the existing unit tests.

**Expected behavior**

The expected behavior is that the job service will be robust to errors returned by the Armada API server and will eventually converge instead of getting stuck in a terminal state. 

**Actual behavior**

The Job Service will cache the last error received on the stream into the DB and continue to write the error into the DB endlessly, until the service is rebooted. 

**Additional context**

I have a prospective fix which simply returns from the goroutine when a stream error occurs, however the subscription timeout is currently set to 300 seconds by default which feels like a poor user experience. I am happy to provide a PR

dave-gantenbein:
Hi @Sovietaced! We'd be happy to look at the PR, but FYI we're in the process of phasing out the job service in favor of an API endpoint that returns the same info. This will simplify things overall from a deployment perspective, but if it unblocks you today send it along and we'll take a look. Thanks!

Sovietaced:
[...]

PR #3555, opened 2024-04-30
Attempt to reconnect to streams that are terminated with unexpected errors

Sovietaced:
#### What type of PR is this?
This is a bug fix

If the [job service subscription manager goroutine](https://github.com/armadaproject/armada/blob/master/internal/jobservice/eventstojobs/manage_subs.go#L271-L338) receives an error (besides an EOF) from the Armada API it will be stuck in a terminal state reading the same error from the stream and writing it to the database, endlessly.  That is because a stream error is terminal. The server ends the stream with an error and any calls to read from the stream will return the same error over and over again. This is a common mistake with gRPC stream client logic unfortunately.

We specifically saw this become an issue when the Armada API server was propagating redis connection pool errors back over the stream to the jobservice subscriptions under heavy concurrent load. 

This pull request updates the behavior to treat the stream receive errors as terminal and introduces a reconnection mechanism to overcome transient armada API issues that is fairly simple for now. It reconnects with a linear backoff. 

Stream creation errors will still result in halting subscription goroutines. 

Fixes https://github.com/armadaproject/armada/issues/3553

We considered updating the goroutine logic to return when an error is received on the stream but we were worried that this would remove any materialized state about the jobs from the Postgres database. More specifically we were concerned that if the subscription was restarted, clients of the job service API might see responses that jump backwards in time, back towards initial message IDs. As such, we felt it was best to keep the goroutine and the state in Postgres alive.


internal/jobservice/eventstojobs/manage_subs_test.go
```diff
@@ -42,11 +42,7 @@ func (m *MockEventClient) Recv() (*api.EventStreamMessage, error) {
 		return nil, io.EOF
 	}
 
-	// We only want to return the error once.
```
Sovietaced, line 45: Updates the mock server to behave like a real server would.

internal/jobservice/eventstojobs/manage_subs_test.go
```diff
@@ -79,33 +75,43 @@ func TestJobSetSubscriptionSubscribe(t *testing.T) {
 		name                 string
 		isJobSetSubscribedFn func(context.Context, string, string) (bool, string, error)
 		ttlSecs              time.Duration
-		err                  error
+		eventClients         []MockEventClient
```
Sovietaced, line 78: Refactored so that we can mock reconnecting and acquiring new streams.

richscott (dismissed):
Looks good - great fix!

ClifHouck (approved):
Seems OK to me as well. Should be fine to merge as long as tests pass.

Sovietaced:
Lint issue should be fixed now

Sovietaced:
@ClifHouck Can you please re-approve so the CI runs

ClifHouck (approved):

Sovietaced:
ping @richscott or any other maintainer. Just hit this issue today in production, however this time caused by what seems like some network error. 

```
could not obtain job set event message, retrying  \u001b[31merror\u001b[0m=\"rpc error: code = Unavailable desc = error reading from server: EOF
```

richscott (approved):

PR #3557, opened 2024-05-01
Remove Legacy Scheduler Job

d80tb7:
We used to need to schedule based on one of two job structs: jobdb.Job (for the new in-memory scheduler) and api.Job (for the old Redis Scheduler). To manage this we defined a LegacySchedulerJob interface that defined the basic functionality we needed. Now that the Redis Scheduler is no more we can remove this interface and simplify the code a bit.


internal/scheduler/jobiteration_test.go
```diff
@@ -7,76 +7,43 @@ import (
...
 	}
 	jctxs := make([]*schedulercontext.JobSchedulingContext, len(jobs))
 	for i, job := range jobs {
 		jctxs[i] = &schedulercontext.JobSchedulingContext{Job: job}
 	}
 	repo := NewInMemoryJobRepository()
 	repo.EnqueueMany(jctxs)
-	expected := []string{"0", "1", "2", "3", "4", "5"}
-	actual := make([]string, 0)
+	// expected := []string{"0", "1", "2", "3", "4", "5"}
```
JamesMurkin, line 34: Can we remove this line?

PR #3573, opened 2024-05-09
Remove UserAnnotationLookup Insertion

d80tb7:
Now that we are using jsonb to store annotations in lookout  we can remove the old user_annotation_lookup code.  This PR removes all the insertion code form the lookout ingester which should improve insertion performance.

I've also added some log lines so that we can easily see how long insertion is taking.

### DIFF
--- a/internal/lookoutingesterv2/benchmark/benchmark.go
+++ b/internal/lookoutingesterv2/benchmark/benchmark.go
@@ -44,10 +44,9 @@ func withDbBenchmark(b *testing.B, config configuration.LookoutIngesterV2Configu
 func benchmarkSubmissions1000(b *testing.B, config configuration.LookoutIngesterV2Configuration) {
 	const n = 1000
 	jobIds := makeUlids(n)
-	jobsToCreate, userAnnotationsToCreate := createJobInstructions(jobIds, n, 10*n)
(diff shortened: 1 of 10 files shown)


MustafaI (approved):

internal/lookoutingesterv2/benchmark/benchmark.go
```diff
@@ -194,21 +190,7 @@ func createJobInstructions(jobIds []string, numJobs int, numUserAnnotations int)
 			Annotations:               make(map[string]string),
 		}
 	}
-	createUserAnnotationInstructions := make([]*model.CreateUserAnnotationInstruction, numUserAnnotations)
-	for i := 0; i < numUserAnnotations; i++ {
-		job := createJobInstructions[i%len(createJobInstructions)]
-		k := uuid.NewString()
-		v := uuid.NewString()
-		job.Annotations[k] = v
```
zuqq, line 202: I think that we want to keep these lines (otherwise we're just inserting jobs without annotations), but this is not a blocker.

zuqq (approved):

internal/lookoutingesterv2/benchmark/benchmark.go
d80tb7: ok- yes- let me fix that in a follow up.

PR #3576, opened 2024-05-10
Fix job service gRPC stream cancellation handling

Sovietaced:
This is a follow up to https://github.com/armadaproject/armada/pull/3555 where I tried to get cute about handling stream cancellation. I updated the code from performing string matching to checking whether the error was of type `context.Canceled` and this introduced a cosmetic bug. The issue with this is that the errors returned in the stream are gRPC specific, so this pull request updates the unite test and fixes the logic to `continue` and not log the cancellation as some other error.


Sovietaced:
Unit test failures seem unrelated
```
     🧪 - com/armadaproject/armada/internal/lookoutv2/repository | Failed
     🧪 - com/armadaproject/armada/internal/lookoutv2/repository | Failed
     🧪 - com/armadaproject/armada/internal/lookoutv2/repository | Failed
```

richscott:
Closing this one, as the query API will supercede this effort.

PR #3583, opened 2024-05-13
Use A Publisher Everwhere

d80tb7:
Update everywhere that needs to publish Pulsar messages to use a `Publisher` interface.  This means:

* We're a bit more decoupled from publishes
* We publish all event messages in a consistent manner.  This means we now have a place where we can e.g. log a warning  if event messages don't have an optional field filled in.


JamesMurkin:
Replaced by https://github.com/armadaproject/armada/pull/3803

PR #3589, opened 2024-05-14
Python client: queue and job_set_id are now required parameters to Armada API fixing python client to enforce that

masipauskas:
As part of ongoing work simplify Armada API surface, now queue and job_set_id - are both required parameters for cancel / reprioritize requests.

This enables over longer period, for us to reduce redis dependency (for non-event stream use-cases)


masipauskas:
Closed in favour of - https://github.com/armadaproject/armada/pull/3588

PR #3600, opened 2024-05-18
Fix Integration Tests

d80tb7:
`ARMADA_SCHEDULING_EXECUTORUPDATEFREQUENCY` was being set in the magefile  where it was having no effect.  As a result executor state was only being updated every 60 seconds which meant that jobs were being rejected as state had not propagated.  This change moves `ARMADA_SCHEDULING_EXECUTORUPDATEFREQUENCY` to be set as part of the docker compose where it is correctly applied.

I've also tweaked some other values here which are more sensible defaults.


magefiles/main.go
```diff
@@ -201,9 +201,6 @@ func LocalDev(arg string) error {
 	mg.Deps(BootstrapTools)
 	fmt.Println("Time to bootstrap tools:", time.Since(timeTaken))
 
-	// Set the Executor Update Frequency to 1 second for local development
```
d80tb7, line 204: this was having no effect

robertdavidsmith (approved):

Issue #3618, opened 2024-05-22
Armada scheduler crashes due to zero'd job run UUID


Sovietaced:
Armada scheduler is in a crash loop. There appears to be a zero'd out job run UUID getting created or deserialized somewhere. I haven't looked into the source code yet but I'm under the impression that the job run UUID is generated by Armada. 

```
DEBU[2024-05-22T18:45:37.423Z]wrapper_logrus.go:63 Write data: 13                                local_addr="10.14.6.68:59620" remote_addr="pulsar://pulsar-broker-0.pulsar-broker.armada.svc.cluster.local:6650"
panic: uuid: Parse(): invalid UUID length: 0

goroutine 258 [running]:
github.com/google/uuid.MustParse({0x0, 0x0})
	/home/runner/go/pkg/mod/github.com/google/uuid@v1.4.0/uuid.go:169 +0xa5
github.com/armadaproject/armada/internal/scheduler.(*MetricsCollector).updateClusterMetrics(0xc000f96f60, 0xc00051a300?)
	/home/runner/work/armada/armada/internal/scheduler/metrics.go:273 +0x1926
github.com/armadaproject/armada/internal/scheduler.(*MetricsCollector).refresh(0xc000f96f60, 0xc00051a300)
	/home/runner/work/armada/armada/internal/scheduler/metrics.go:121 +0xb3
github.com/armadaproject/armada/internal/scheduler.(*MetricsCollector).Run(0xc000f96f60, 0xc00051a300)
	/home/runner/work/armada/armada/internal/scheduler/metrics.go:89 +0x14a
github.com/armadaproject/armada/internal/scheduler.Run.func9()
	/home/runner/work/armada/armada/internal/scheduler/schedulerapp.go:353 +0x25
golang.org/x/sync/errgroup.(*Group).Go.func1()
	/home/runner/go/pkg/mod/golang.org/x/sync@v0.6.0/errgroup/errgroup.go:78 +0x64
created by golang.org/x/sync/errgroup.(*Group).Go
	/home/runner/go/pkg/mod/golang.org/x/sync@v0.6.0/errgroup/errgroup.go:75 +0xa5
```	

Running on Armada 0.4.48

The postgres DB is showing zero runs with a zero'd UUID
```
postgres=# SELECT * from runs where run_id = '00000000-0000-0000-0000-000000000000';
[...]

PR #3620, opened 2024-05-23
Update IsManagedPod to check for job run id and job id

Sovietaced:
We had an issue where a v3 and v4 executor were running on the same cluster. The v4 executor was pulling empty job run ids from pods launched by the v3 executor. We realize that we shouldn't run two executors in the same cluster however it also feels like the filtering logic in the cluster utilization logic should just check for the job run id label if that is what it is ultimately going to pick out of the pod and assume is present. 

Fixes: https://github.com/armadaproject/armada/issues/3618

### DIFF
--- a/internal/executor/util/pod_util.go
+++ b/internal/executor/util/pod_util.go
@@ -61,8 +61,9 @@ func IsInTerminalState(pod *v1.Pod) bool {
 }
 
 func IsManagedPod(pod *v1.Pod) bool {
-	_, ok := pod.Labels[domain.JobId]
-	return ok
+	_, jobRunIdOk := pod.Labels[domain.JobRunId]
+	_, jobIdOk := pod.Labels[domain.JobId]
+	return jobRunIdOk && jobIdOk
 }
 
 func GetManagedPodSelector() labels.Selector {
--- a/internal/executor/util/pod_util_test.go
+++ b/internal/executor/util/pod_util_test.go
@@ -71,7 +71,7 @@ func TestHasIngress_WhenAnnotationNotSetToTrue(t *testing.T) {
 	assert.False(t, HasIngress(pod))
 }
 
-func TestIsManagedPod_ReturnsTrueIfJobRunIdLabelPresent(t *testing.T) {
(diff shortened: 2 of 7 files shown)


dave-gantenbein:
We're going to double check with one of the core devs tomorrow, but most likely we need both labels on the pod.  I think what we want to do is validate that the executor has given us valid run ids and reject their update if they haven't.

Sovietaced:
> We're going to double check with one of the core devs tomorrow, but most likely we need both labels on the pod. I think what we want to do is validate that the executor has given us valid run ids and reject their update if they haven't.

I think adding validation on the server side is definitely ideal. I'd be happy to add that as well just let me know!

dave-gantenbein:
@Sovietaced - by all means!

Sovietaced:
@dave-gantenbein should be good for review. I think the go mod CI job flaked so should be unrelated.

richscott:
@Sovietaced Can you rebase this branch from `master`, then if the unit tests pass again at least once, this should be good to go.

Sovietaced:
> @Sovietaced Can you rebase this branch from `master`, then if the unit tests pass again at least once, this should be good to go.

done

richscott (approved):

dave-gantenbein:
@Sovietaced - we discussed this and while it won't cause any negative impact, I'm wondering if you've updated things so the executors and server are on the same version, which would be the recommended pattern. If there's something blocking you from doing that let me know here or in slack and we can merge this, but again this isn't how we'd recommend you go about setting things up long term, as other issues could potentially creep in and we won't be testing this scenario...

Sovietaced:
> @Sovietaced - we discussed this and while it won't cause any negative impact, I'm wondering if you've updated things so the executors and server are on the same version, which would be the recommended pattern. If there's something blocking you from doing that let me know here or in slack and we can merge this, but again this isn't how we'd recommend you go about setting things up long term, as other issues could potentially creep in and we won't be testing this scenario...

We haven't updated things yet and have been running a fork but I have already mentioned I realize this is not a supported deployment. I don't think the PR makes the codebase any worse considering there is no validation being done in the scheduler API but you do you.

PR #3623, opened 2024-05-24
Add job APIs to armada client

Sovietaced:
This pull request adds a `get_job_status` method to the armada python client. This is something that we'd like to use in order to migrate to the new query API. I figure the armada airflow operator will probably need to use this sooner or later as well as it migrates to the new query API.

### DIFF
--- a/client/python/armada_client/client.py
+++ b/client/python/armada_client/client.py
@@ -7,7 +7,7 @@ https://armadaproject.io/api
 
 from datetime import timedelta
 import logging
-from typing import Dict, Iterator, List, Optional
+from typing import Dict, Iterable, Iterator, List, Optional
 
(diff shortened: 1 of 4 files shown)


Sovietaced:
> We should probably expose the other two endpoints as well: `GetJobDetails`, and `GetJobRunDetails`. Either here or in a follow-up PR.

I can do that here. Will ping when I'm done.

Sovietaced:
@ClifHouck should be good for review

MustafaI:
I'd hold off on this for a while as we'll be syncing out our improved Armada Airflow Operator, which as you say makes use of the query API as opposed to JobService. We'll get these exact changes as part of that work, which should be done in about a week.

MustafaI:
Closing, see #3650

PR #3624, opened 2024-05-24
WIP: Fix armada-airflow build

Sovietaced:
… builds

Fixes #

### DIFF
--- a/third_party/airflow/pyproject.toml
+++ b/third_party/airflow/pyproject.toml
@@ -9,9 +9,9 @@ requires-python = ">=3.7"
 # extremely difficult.
 dependencies = [
 	"armada-client",
-	"apache-airflow>=2.6.3",
-	"grpcio==1.64.0",
-	"grpcio-tools==1.64.0",
+	"apache-airflow@git+https://github.com/Sovietaced/airflow@protobuf5-test",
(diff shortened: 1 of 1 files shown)


richscott:
@Sovietaced We expect that the new query API in Armada will be imminent, and the Airflow projects will be switching to using that, and will retire the old Job Service service. Likely this PR may be obsolete in the near future.

Sovietaced:
> @Sovietaced We expect that the new query API in Armada will be imminent, and the Airflow projects will be switching to using that, and will retire the old Job Service service. Likely this PR may be obsolete in the near future.

This PR is about fixing the armada-airflow build. I don't think its related to the old jobs service at all. There has been a fix for the armada-airflow build [here](https://github.com/armadaproject/armada/commit/3a49d8d88192307337fdc3d58ed4f63e69e78b75), but the approach is not ideal considering it locks users of armada-airflow to a specific version of grpcio. Libraries pinning dependencies is an anti pattern imo. Applications should pin dependencies, libraries should not.

richscott:
@Sovietaced our armada-airflow users  have made a PR that indicates they are experiencing an incompatibility with the later versions of grpcio in using opentelemetry with airflow - see https://github.com/armadaproject/armada/pull/3631/files

PR #3625, opened 2024-05-27
Fix Fair Share

d80tb7:
Previously we were counting a queue as "active" if it had any non-terminal jobs regardless of pool.  This change means that we now have separate "active" queues per pool and only consider a queue as active for that pool accordingly.


internal/scheduler/scheduling_algo.go
```diff
@@ -290,13 +297,35 @@ func (l *FairSchedulingAlgo) newFairSchedulingAlgoContext(ctx *armadacontext.Con
...
+	isActiveByPoolByQueue := make(map[string]map[string]bool, len(queues))
 	jobsByExecutorId := make(map[string][]*jobdb.Job)
 	nodeIdByJobId := make(map[string]string)
 	jobIdsByGangId := make(map[string]map[string]bool)
 	gangIdByJobId := make(map[string]string)
 	for _, job := range txn.GetAll() {
-		isActiveByQueueName[job.Queue()] = true
+
+		// Mark a queue being active for a given pool.  A queue is defined as being active if it has a job running
+		// on a pool or oif a queued job is eligible for that pool
```
JamesMurkin, line 308:
```suggestion
		// on a pool or if a queued job is eligible for that pool
```

JamesMurkin (approved):

JamesMurkin (approved):

PR #3629, opened 2024-05-29
Upgrade proto dependencies to align with operator

masipauskas:
Fixes #

### DIFF
--- a/client/python/pyproject.toml
+++ b/client/python/pyproject.toml
@@ -4,7 +4,7 @@ version = "0.3.1"
 description = "Armada gRPC API python client"
 readme = "README.md"
 requires-python = ">=3.7"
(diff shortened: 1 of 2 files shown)


MustafaI (approved):

Sovietaced:
Updating airflow to the latest version of gRPC now guarantees that it cannot build due to conflicting dependencies with `apache-airflow`. At least  before armada and airflow could get one the same version that was less than gRPC 1.62.0. What is the point of this fix?

masipauskas:
You're right - that this PR end up being useless (our primary consumer of python client, is our airflow operator, will try to resolve proto version conflicts today on this). This was an attempt to fix airflow / python client incompatibility issue in a quick and simple way, but I didn't consider the opentelemetry-proto package and their pinned dependencies.

PR #3630, opened 2024-05-29
Scheduler: use efficient/immutable resource list implementation

robertdavidsmith:


internal/scheduler/internaltypes/node.go
```diff
@@ -149,3 +147,19 @@ func deepCopyLabels(labels map[string]string) map[string]string {
 	}
 	return result
 }
+
+func deepCopyStringToResourceList(m map[string]ResourceList) map[string]ResourceList {
```
d80tb7, line 151: I *think* this (and the below function) are equivalent to `maps.clone`

internal/scheduler/internaltypes/resource_list.go
```diff
@@ -1,16 +1,196 @@
...
+		}
+	}
+	return false
+}
+
+func (rl ResourceList) IsEmpty() bool {
+	return rl.factory == nil
+}
+
+func (rl ResourceList) ExceedsAvailable(available ResourceList) (string, k8sResource.Quantity, k8sResource.Quantity, bool) {
```
d80tb7, line 111: I'd be tempted to document this function as it's non trivial (e.g. it returns three values)

internal/scheduler/internaltypes/resource_list_factory.go
```diff
@@ -64,15 +69,30 @@ func (factory *ResourceListFactory) FromNodeProto(resources schedulerobjects.Res
...
 
-// Fail on unknown resources, round up.
-func (factory *ResourceListFactory) FromJobProto(resources schedulerobjects.ResourceList) (ResourceList, error) {
+// Ignore unknown resources, round up.
+func (factory *ResourceListFactory) FromJobResourceListIgnoreUnknown(resources map[string]k8sResource.Quantity) ResourceList {
 	result := make([]int64, len(factory.indexToName))
-	for k, v := range resources.Resources {
+	for k, v := range resources {
 		index, ok := factory.nameToIndex[k]
+		if ok {
```
d80tb7, line 77: do we want to log (even if only at debug) what we are ignoring here?

internal/scheduler/jobdb/job.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/armadaproject/armada/internal/scheduler/internaltypes"
```
d80tb7, line 7: This should be with the other armada imports (it's bit sad that goimports doesn't understand this!)

internal/scheduler/jobdb/job.go
```diff
@@ -49,6 +51,8 @@ type Job struct {
 	queuedVersion int32
 	// Scheduling requirements of this job.
 	jobSchedulingInfo *schedulerobjects.JobSchedulingInfo
+	// Scheduling requirements of this job stored in efficient form.
```
d80tb7, line 54:
```suggestion
	// Resource requirements of this job stored in efficient form.
```

d80tb7: (technically it's not the whole set of scheduling requirements, just the resources)

internal/scheduler/internaltypes/node.go
robertdavidsmith: Fixed

internal/scheduler/internaltypes/resource_list.go
robertdavidsmith: Fixed

internal/scheduler/jobdb/job.go
robertdavidsmith: done

internal/scheduler/jobdb/job.go
robertdavidsmith: done

internal/scheduler/scheduler_test.go
```diff
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/armadaproject/armada/internal/scheduler/internaltypes"
```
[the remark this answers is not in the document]
robertdavidsmith: done

internal/scheduler/internaltypes/resource_list_factory.go
robertdavidsmith: We discussed this, decided against it as it would require passing ctx down, and we check at validation time anyway.

d80tb7 (approved):

PR #3663, opened 2024-06-09
Show Rejections in  Lookout 

d80tb7:
Backend changes necessary to show rejection messages in lookout.  Changes here are:

- Adds a new `REJECTED` state to lookout
- Adds a new `job_error` table to the lookout db
- Jobs are marked as `REJECTED` if the event failure message is of type `Error_JobRejected`.  We also store the error in the `job_error` table
- New Repository as been added which allows a job error to be looked up by jobId.

We shouldn't merge this until the frontend code is ready.


internal/lookoutingesterv2/lookoutdb/insertion_test.go
```diff
@@ -593,6 +595,54 @@ func TestUpdateJobRunsScalar(t *testing.T) {
...
+		assert.Equal(t, expectedJobError, jobError)
+
+		// Insert again and test that it's idempotent
+		err = ldb.CreateJobErrorsBatch(armadacontext.Background(), defaultInstructionSet().JobErrorsToCreate)
+		assert.Nil(t, err)
+		jobError = getJobError(t, db, jobIdString)
+		assert.Equal(t, expectedJobError, jobError)
+
+		// If a row is bad then we should return an error and no updates should happen
+		_, err = ldb.db.Exec(armadacontext.Background(), "DELETE FROM job_error")
```
JamesMurkin, line 615: Would it be worth check that if you try to update it with a different value, it keeps the original value?

internal/lookoutv2/application.go
```diff
@@ -121,6 +122,19 @@ func Serve(configuration configuration.LookoutV2Config) error {
 		},
 	)
 
+	api.GetJobErrorHandler = operations.GetJobErrorHandlerFunc(
+		func(params operations.GetJobErrorParams) middleware.Responder {
+			ctx := armadacontext.New(params.HTTPRequest.Context(), logger)
+			result, err := getJobErrorRepo.GetJobErrorMessage(ctx, params.GetJobErrorRequest.JobID)
+			if err != nil {
+				return operations.NewGetJobRunDebugMessageBadRequest().WithPayload(conversions.ToSwaggerError(err.Error()))
```
JamesMurkin, line 130: Is this right? Looks like a copy paste error

JamesMurkin (approved):

PR #3679, opened 2024-06-17
Improve performance of selectNodeForJobWithFairPreemption

JamesMurkin:
Testing with the worst case, this makes the function 2-4 times faster
 - Combined with https://github.com/armadaproject/armada/pull/3678 it is about 3-5 times faster

Even the average case should be improved, as we now do less work

The main "trick" here is to not perform node.UnsafeCopy() and nodeDb.unbindJobFromNodeInPlace until it is needed (once you know which node is the selected one)
 - These functions are expensive, particularly UnsafeCopy

Instead just keep track of the evicted jobs + available resource per node locally until you've found one that matches.

There are more enhancements that could be made here:
 - Work out when it is better to check static requirements first (possibly even save these and use them for all jobs with the same scheduling key)
 - Improve nodeDb speed to look up a node (use a map?)
 - Improve nodeDb speed to iterate evicted jobs (use a slice? / effectively we just need a sorted hashmap, not a fully fledged radix tree)


internal/scheduler/nodedb/nodedb.go
```diff
@@ -802,6 +802,13 @@ func (nodeDb *NodeDb) selectNodeForPodWithItAtPriority(
 	return selectedNode, nil
 }
 
+type consideredNode struct {
```
d80tb7, line 805: define struct in function

PR #3707, opened 2024-06-19
Make ingesters limit how many events they process at once

JamesMurkin:
Currently we limit how many messages are processed at once

However a single message may contain thousands of events, so limit on messages is a bit flawed
 - A limit of 1000, could mean you're processing 1000 or 1000000+ events at once

This should make the ingesters less prone to long pauses if some large messages come along

Also adjusted event ingester config to be more standard with other config naming

### DIFF
--- a/config/eventingester/config.yaml
+++ b/config/eventingester/config.yaml
@@ -12,9 +12,9 @@ pulsar:
   receiverQueueSize: 100
 subscriptionName: "events-ingester"
 minMessageCompressionSize: 1024
-batchSize: 1048576  #1MB
+maxMessageBatchSizeBytes: 1048576  #1MB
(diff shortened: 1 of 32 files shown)


config/eventingester/config.yaml
```diff
@@ -12,9 +12,9 @@ pulsar:
   receiverQueueSize: 100
 subscriptionName: "events-ingester"
 minMessageCompressionSize: 1024
-batchSize: 1048576  #1MB
+maxMessageBatchSizeBytes: 1048576  #1MB
```
d80tb7, line 15: maybe call this maxOutputMessageSize

d80tb7 (approved):

PR #3714, opened 2024-06-20
Pre-emption Based On Adjusted Fair Share

d80tb7:
Introduces the concept of `adjustedFairShare` which is calculated as follows:

* Define `demand` as total resources of all active jobs for a queue
* Define `cappedFairShare` for a queue as `demand`\\`totalClusterResources`
* Define `excessShare` as `fairShare - cappedFairShare`
* For queues where `excessShare  > 0`,  `adjustedFairShare = cappedFairShare`
* For all other queues we repeated the calculation, to reshare the total `excessShare`

When choosing what queues are eligable for preemption,the default is now to use `adjustedFairShare` rather than `fairShare`.  This behaviour  can be controlled by a new `useAdjustedFairShareProtection` flag


cmd/fish/main.go
```diff
@@ -0,0 +1,54 @@
+package main
```
d80tb7, line 1: this was me playing with algos!  I'll remove it before I submit properly!

JamesMurkin (approved):

PR #3748, opened 2024-06-23
Allow Limits To Be Greater Than Requests

d80tb7:
Initial implementation of allowing limits to be less than requests.  This is controlled by a new `MaxOversubscriptionByResourceRequest` config flag that defines the maximum multiplier by which limits can be greater than requests on a per resource level.  Default value is 1.0, so if no config is given the limits must be equal to requests.

Note that scheduling is still done using requests only, although I've removed the bit of code that sets limits equal to requests to save memory.


internal/armada/submit/validation/submit_request_test.go
```diff
@@ -817,6 +818,30 @@ func TestValidateResources(t *testing.T) {
 			}),
 			expectSuccess: false,
 		},
+		"Requests and limits different, passes MaxResourceOversubscriptionByResourceRequest": {
```
JamesMurkin, line 821: Add tests for the other cases
 - limit less than request 
 - request and limit have different resources specified

PR #3756, opened 2024-06-26
Remove Gogo Nullable Properties From API Protos

d80tb7:
Removes all gogo nullable properties from all proto files except for those under `schedulerobjects`.  Main change here is the need to convert bateween `map[K]V` to `map[K]*V`.

This is the step 2 of many for removing gogo proto!


internal/executor/domain/resources.go
```diff
@@ -2,11 +2,6 @@ package domain
 
 import armadaresource "github.com/armadaproject/armada/internal/common/resource"
 
-const (
```
d80tb7, line 5: these were unused.

robertdavidsmith (approved):

PR #3758, opened 2024-06-27
Fix Issues Arising from Confusing Weighted Costs

d80tb7:
The cost provider returned weighted costs, this caused issues iin a couple of places where an unweighted cost was expected.

- protected fair share
- actual_share metric

I've modified the cost provider to also return unweighted costs and have renamed the existing methods soo that it's clear they are weighted.


internal/scheduler/preempting_queue_scheduler_test.go
```diff
@@ -1312,6 +1312,37 @@ func TestPreemptingQueueScheduler(t *testing.T) {
 				"D": 1,
 			},
 		},
+		"ProtectedFractionOfFairShare non equal weights": {
```
d80tb7, line 1315: This unit test demonstrates the issue

JamesMurkin (approved):

PR #3770, opened 2024-07-05
Add support for mixed pool clusters

JamesMurkin:
Currently an executor (k8s cluster) must belong to a pool.

This can be quite restricting if you have several nodes, all of which you want to be in their own pools. As you have to build a cluster for each group of nodes

With this change we are making it so a Node belongs to a pool (marked via label) - and defaulting to the cluster pool
 - At some point after this migration, we may remove the cluster having a pool and instead have a static "default" pool

As part of this change I've also removed the minimumJobSize functionality as it is unused
 - If we need it back, we'll need to add it as config at the scheduling level (minimum per pool)


internal/scheduler/poolutils_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/armadaproject/armada/internal/scheduler/testfixtures"
+)
+
+var (
+	nodeWithPool        = &schedulerobjects.Node{Pool: "node-pool"}
+	nodeWithoutPool     = &schedulerobjects.Node{}
+	executorWithPool    = &schedulerobjects.Executor{Pool: "executor-pool"}
+	executorWithoutPool = &schedulerobjects.Executor{}
+
+	runWithPool = testfixtures.JobDb.CreateRun(
```
d80tb7, line 21: this is mental!

internal/scheduler/schedulerobjects/schedulerobjects.proto
```diff
@@ -15,6 +15,7 @@ message Executor {
     string pool = 2;
     // The nodes available for scheduling via this executor.
     repeated Node nodes = 3;
+    // Deprecated - not currently in use
```
d80tb7, line 18: why would we not just remove this?

d80tb7 (approved):

PR #3771, opened 2024-07-05
Remove Gogo Proto From Everywhere except schedulerobjects

d80tb7:
This PR removes gogo proto from all files except schedulerobjects.proto.  This means that it is gone from all public facing apis and the armadaevents messages.

Final PR after this one will be to change the `mage proto` target to generate without using gogo


internal/common/ingest/testfixtures/event.go
```diff
@@ -3,6 +3,8 @@ package testfixtures
 import (
 	"time"
 
+	protoutil "github.com/armadaproject/armada/internal/common/proto"
```
robertdavidsmith, line 6: import ordering

robertdavidsmith (approved):

PR #3773, opened 2024-07-07
Improvements To Fairness Metrics

d80tb7:
* Adds a metric for `armada_scheduler_demand`  which is demand per queue per pool (in units of cost).  This helps us tell if `adjusted_fair_share` has been capped at demand.
* Adds a metric for `armada_scheduler_fairness_error` which is the sum of all deltas between `adjusted_fair_share`  and `actual_share` when `adjusted_fair_share` is greater than `actual_share`
* Chnages `actual_share` so that it is calculated as a percentage of available cluster capacity rather than as a percentage of running jobs.

### DIFF
--- a/internal/scheduler/scheduler_metrics.go
+++ b/internal/scheduler/scheduler_metrics.go
@@ -78,6 +78,23 @@ var actualSharePerQueueDesc = prometheus.NewDesc(
 	}, nil,
 )
 
+var demandPerQueueDesc = prometheus.NewDesc(
+	fmt.Sprintf("%s_%s_%s", NAMESPACE, SUBSYSTEM, "demand"),
+	"Demand of each queue and pool.",
+	[]string{
+		"queue",
+		"pool",
+	}, nil,
+)
+
+var fairnessErrorDesc = prometheus.NewDesc(
+	fmt.Sprintf("%s_%s_%s", NAMESPACE, SUBSYSTEM, "fairness_error"),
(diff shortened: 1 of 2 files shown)


JamesMurkin (dismissed):

internal/scheduler/scheduler_metrics.go
```diff
@@ -78,6 +78,23 @@ var actualSharePerQueueDesc = prometheus.NewDesc(
...
+	"Demand of each queue and pool.",
+	[]string{
+		"queue",
+		"pool",
+	}, nil,
+)
+
+var fairnessErrorDesc = prometheus.NewDesc(
+	fmt.Sprintf("%s_%s_%s", NAMESPACE, SUBSYSTEM, "fairness_error"),
+	"Cumulative delta between adjusted fair share and actual share for all users who\n// are below their fair share",
```
JamesMurkin, line 92:
```suggestion
	"Cumulative delta between adjusted fair share and actual share for all users who are below their fair share",
```

internal/scheduler/scheduler_metrics.go
```diff
@@ -78,6 +78,23 @@ var actualSharePerQueueDesc = prometheus.NewDesc(
...
+	[]string{
+		"pool",
```
JamesMurkin, line 94: I was just curious why we don't have this metric per queue also, we could also sum them to get the total number

JamesMurkin (approved):

Issue #3566, opened 2024-05-07
Ability to release from a branch


dave-gantenbein:
We need this in case we need to quickly patch a release that's currently in production, but we don't want to release changes that have already made it to master. This actually happened a couple of weeks ago and we ended up having to revert a bunch of changes from master and then re-apply them once the patched release had been made.  I think ideally we'd define a branch pattern for hotfixes that would mark the branch as releasable (maybe something like  .*-patch-[0-9]* ) thus triggering all the same protections as the master branch and causing a release to be made on PR merge

PR #3810, opened 2024-07-23
Implement patch releases

pavlovic-ivan:
Implement patch releases changing the validate ref jobs in release, release-rc workflows and conditions in build workflow

#### What type of PR is this?
Improvement PR

This PR implements releases of patch versions so that Armada team doesn't have to revert changes from a main branch to fix an issue in a latest release, and so it's obvious that a release has a patch if a bug is introduced.

Fixes: https://github.com/armadaproject/armada/issues/3566

Before merging this PR, an admin needs to setup similar configs for branches that start with `patch/v` because these branches will be used for the patched releases, and need access to environments, secrets and have the same protections rules as `master` branch.


d80tb7 (approved):

dave-gantenbein (approved):

richscott (approved):
Nice work!

PR #3841, opened 2024-07-25
Queue Scheduling Controls

MustafaI:
Fixes #

This is a pretty big PR, so here are a summary of changes:
 - Ability to pause/resume scheduling per-queue
   - Queued demand for paused queues not considered in fair share calculations
   - Modified `Queue` and `api.Queue` to contain queue scheduling status (see: `pkg/api`, `internal/scheduler/database/migrations`)
 - Ability to label queues
   - Modified `Queue` and `api.Queue` to contain queue labels (see: `pkg/api`, `internal/scheduler/database/migrations`)
 - Armadactl scheduling controls based on the above features:
   - Added `cordon` and `uncordon` commands on `queue` and `queues`
   - `match-labels`, `match-queues`, `dry-run` and `inverse` flags on `queues` allow better transparency and control on actions being taken. 
   - see: `cmd/armdactl/cmd`, `internal/armadactl`
 - Fix bug where `unfeasibleSchedulingKeys` was being populated incorrectly
   - This bug led to all jobs of a particular shape being banned for relevant scheduling rounds. 
   - We fixed this bug by doing the following:
     - Only add a scheduling key to `unfeasibleSchedulingKeys` if it's globally unschedulable (see: `internal/scheduler/gang_scheduler.go`) - for now a gang exceeding the global burst size would be the only constraint meeting this criteria.
     - Ensuring that jobs not considered due to constraint are marked as unsuccessful (see: `internal/scheduler/gang_scheduler.go`, `internal/scheduler/queue_scheduler.go` )
   - By fixing this bug we also broke some tests which relied on `UnsuccessfulJobSchedulingContexts` being populated in a certain way. However, this data structure was only populated if a job was marked as unsuccessful because it's scheduling key was in `unfeasibleSchedulingKeys`. 
   - Fixing this bug should also fix inaccuracies in metrics, as well as provide better job scheduling reporting.


cmd/armadactl/cmd/scheduling_control.go
```diff
@@ -0,0 +1,165 @@
...
+
+			dryRun, err := cmd.Flags().GetBool("dry-run")
+			if err != nil {
+				return fmt.Errorf("error reading dry-run flag: %s", err)
+			}
+
+			return a.CordonQueue(queue, dryRun)
+		},
+	}
+	cmd.Flags().Bool("dry-run", false, "Show selection of queues that will be modified in this operation")
```
d80tb7, line 140: presumably this is not a selection of queues, but just a single queue?

internal/armadactl/queue.go
```diff
@@ -2,10 +2,14 @@ package armadactl
...
 import (
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	"sigs.k8s.io/yaml"
 
+	"github.com/armadaproject/armada/internal/common/slices"
+	"github.com/armadaproject/armada/pkg/api"
+
```
d80tb7, line 12: remove the newline

internal/armadactl/queue.go
```diff
@@ -68,6 +81,44 @@ func (a *App) GetQueue(name string) error {
 	return nil
 }
 
+func (a *App) getAllQueuesAsAPIQueue(queueNames []string, labels []string, inverse bool) ([]*api.Queue, error) {
+	queueContainsAllLabelsAndInProvidedNames := func(q *api.Queue) bool {
```
d80tb7, line 85: Do we want/need to support a user giving both queue names and queue labels.  The logic here would be simpler if we restricted them to either queue names or queue labels.  I'd argue unless we have a good use case in mind we should stick to the simpler implementation and add something more complex if we find we need it.

internal/scheduler/database/migrations/013_add_queue_scheduling_paused.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE queues ADD COLUMN scheduling_paused bool DEFAULT FALSE;
```
JamesMurkin, line 1: I don't think this is doing anything. I'm almost certain this table isn't even used

internal/armadactl/queue.go
```diff
@@ -68,6 +81,44 @@ func (a *App) GetQueue(name string) error {
...
+	queuesToReturn, err := a.Params.QueueAPI.GetAll()
+	if err != nil {
+		return nil, errors.Errorf("[armadactl.getAllQueuesAsAPIQueue] error getting all queues: %s", err)
+	}
+
+	// Filter function seems expensive enough to justify parallel filter
+	return slices.ParallelFilter(queuesToReturn, queueContainsAllLabelsAndInProvidedNames), nil
+}
+
+// GetQueue calls app.QueueAPI.GetAll with the provided parameters.
```
d80tb7, line 107:
```suggestion
// GetQueues calls app.QueueAPI.GetAll with the provided parameters.
```

internal/armadactl/queue.go
```diff
@@ -68,6 +81,44 @@ func (a *App) GetQueue(name string) error {
...
+		})
+		inQueues := len(queueNames) == 0 || slices.Contains(queueNames, q.Name)
+		return inverse != (containsAllLabels && inQueues)
+	}
+	queuesToReturn, err := a.Params.QueueAPI.GetAll()
+	if err != nil {
+		return nil, errors.Errorf("[armadactl.getAllQueuesAsAPIQueue] error getting all queues: %s", err)
+	}
+
+	// Filter function seems expensive enough to justify parallel filter
```
JamesMurkin, line 103: Is this correct?

I'd be surprised if running a small function on a few thousand objects really justified the complexity of ParallelFilter - but possibly you've tested it and know otherwise

internal/armadactl/scheduling_control.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/armadaproject/armada/internal/common/slices"
+	"github.com/armadaproject/armada/pkg/api"
+	"github.com/armadaproject/armada/pkg/client/queue"
+)
+
+func (a *App) cordonQueue(q *api.Queue) error {
+	q.SchedulingPaused = true
+	newQueue, err := queue.NewQueue(q)
+	if err := a.Params.QueueAPI.Update(newQueue); err != nil {
```
d80tb7, line 14: I'd argue it would be better here to have a cordon rpc call that just takes the queue name. That way you know that the only thing that can be updated by this function is the cordon flag.  As it stands a user could inadvertantly update e.g. queuePriority by calling this function, which I'd argue would be somewhat uinexpected!

internal/armadactl/scheduling_control.go
```diff
@@ -0,0 +1,119 @@
...
+		if err != nil {
+			return fmt.Errorf("could not pause scheduling on queue %s: %s", queue, err)
+		} else {
+			fmt.Printf("%s paused\n", queue)
+		}
+	}
+	return nil
+}
+
+func (a *App) CordonQueues(matchQueues []string, matchLabels []string, dryRun bool, inverse bool) error {
```
d80tb7, line 41: how slow is this to cordon hundreds of queues?  Is it worth having an endpoint where you can cordon a large number of queues in a single call?

internal/armadactl/scheduling_control.go
```diff
@@ -0,0 +1,119 @@
...
+	return nil
+}
+
+func (a *App) uncordonQueue(q *api.Queue) error {
+	q.SchedulingPaused = false
+	newQueue, err := queue.NewQueue(q)
+	if err != nil {
+		return fmt.Errorf("error creating new queue type for %s: %s", q.Name, err)
+	}
+	if err = a.Params.QueueAPI.Update(newQueue); err != nil {
```
d80tb7, line 70: as before I'd have a dedicated uncordon api endpoint and have thus function use that.  That way there's now way of inadvertantly modifying other queue properties.

internal/common/slices/slices.go
```diff
@@ -109,6 +110,16 @@ func GroupByFuncUnique[S ~[]E, E any, K comparable](s S, keyFunc func(E) K) map[
 	return rv
 }
 
+// Apply applies the given function to all elements of the input slice
```
d80tb7, line 113: Should add a test for this

internal/common/slices/slices.go
```diff
@@ -159,6 +170,32 @@ func Subtract[T comparable](list []T, toRemove []T) []T {
 	return out
 }
 
+// ParallelFilter is semantically the same as Filter except that results can be returned out of order
+func ParallelFilter[S ~[]E, E any](s S, predicate func(e E) bool) S {
```
d80tb7, line 174: Do we have evidence that this is needed (e.g. it seems fairly marginal for your use case of filtering queues).  If not I'd just remove this as there's a new streams api coming to GO so imagine we'll want to deprecate most of the code here.

cmd/armadactl/cmd/scheduling_control.go
MustafaI: Yep

internal/common/slices/slices.go
```diff
@@ -172,6 +209,19 @@ func Filter[S ~[]E, E any](s S, predicate func(e E) bool) S {
 	return out
 }
 
+// Contains returns true if the slice s contains the element e.
+func Contains[S ~[]E, E comparable](s S, e E) bool {
```
d80tb7, line 213: I *think* this already exists in `golang.org/x/exp/slices`

internal/common/slices/slices.go
```diff
@@ -193,6 +243,18 @@ func AnyFunc[S ~[]T, T any](s S, predicate func(val T) bool) bool {
 	return false
 }
 
+// AllFunc returns true if predicate(v) returns true for all values v in s.
+func AllFunc[S ~[]T, T any](s S, predicate func(val T) bool) bool {
```
d80tb7, line 247: test

internal/scheduler/scheduling_algo.go
```diff
@@ -483,9 +489,21 @@ func (l *FairSchedulingAlgo) schedulePool(
...
 
-		queueLimiter.SetLimitAt(now, rate.Limit(l.schedulingConfig.MaximumPerQueueSchedulingRate))
+		queueContextDemand := demand.AggregateByResource()
 
-		if err := sctx.AddQueueSchedulingContext(queue, weight, allocatedByPriorityClass, demand.AggregateByResource(), cappedDemand.AggregateByResource(), queueLimiter); err != nil {
+		if !fsctx.schedulingStatusByQueue[queue] {
+			queueLimiter.SetLimitAt(now, rate.Limit(float64(0)))
+			queueLimiter.SetBurstAt(now, 0)
+
+			// Queued jobs should not be considered for paused queues, so demand := running
```
d80tb7, line 498: can't we just take this into account when we calculate demand? Rather than calculating demand and then changing that later on

pkg/client/queue/get_all.go
```diff
@@ -0,0 +1,87 @@
...
+						return nil, e
+					}
+				}
+				if e == io.EOF {
+					return nil, e
+				}
+				if !isTransportClosingError(e) {
+					log.Error(e)
+				}
+				time.Sleep(5 * time.Second)
```
d80tb7, line 62: why do we have this sleep?

d80tb7, line 62: I'm not sure we should do this.  When we watch events this is so that we can reconnect if there is a transient error, which is probably reasonable as watching events can take ages.  Getting queues should be quick, however so I'd say we should just fail immediately if there is an issue, rather than having the client hang for 5 seconds.

pkg/api/submit.proto
```diff
@@ -192,6 +192,8 @@ message Queue {
     // If provided for a priority class, global limits for that priority class do not apply to this queue.
     map<string, PriorityClassResourceLimits> resource_limits_by_priority_class_name = 7;
     repeated Permissions permissions = 6;
+    bool scheduling_paused = 8;
```
d80tb7, line 195: I think I'd add comments for both of these.

pkg/client/queue/get_all.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to obtain api connection details: %s", err)
+		}
+		conn, err := client.CreateApiConnection(connectionDetails)
+		if err != nil {
+			return nil, fmt.Errorf("failed to connect to api because %s", err)
+		}
+		defer conn.Close()
+
+		client := api.NewSubmitClient(conn)
```
d80tb7, line 31: Make a queueClient here

pkg/client/queue/queue.go
```diff
@@ -46,10 +49,19 @@ func NewQueue(in *api.Queue) (Queue, error) {
 		}
 	}
 
+	// Queue labels must be Kubernetes-like key-value labels
+	for _, label := range in.Labels {
+		if len(strings.Split(label, "=")) != 2 {
+			return Queue{}, fmt.Errorf("queue label must be key-value, not %s", label)
```
d80tb7, line 55:
```suggestion
			return Queue{}, fmt.Errorf("queue label must be key=value, not %s", label)
```

internal/scheduler/database/migrations/013_add_queue_scheduling_paused.sql
d80tb7, line 1: I don't think this is needed

PR #3846, opened 2024-07-30
fix: Fix armada executor job task interval variable

Sovietaced:
I believe the incorrect interval was used for the job lease background task


dejanzele:
@d80tb7 @JamesMurkin what are your thoughts?

JamesMurkin:
I think this is fine, however could you also update the default value so this becomes effectively a noop in terms of default behaviour.

Update this value to 5s, to match the value being passed currently:
https://github.com/armadaproject/armada/blob/master/config/executor/config.yaml#L13

Sovietaced:
ping @JamesMurkin @d80tb7

d80tb7 (approved):

JamesMurkin (approved):

PR #3864, opened 2024-08-03
Stop producing pod unschedulable errors

d80tb7:
We don't do anything with this error other than:

a) use it to set the node in the lookout ingester (which is unnecessary as this will now get set on lease)
b) turn it into a`JobUnableToScheduleEvent` on the public api, which causes confusion.

We can remove the handing code, once we're sure nothing is producing it any more.


robertdavidsmith (dismissed):

JamesMurkin:
Can we ensure this logs we have detected an error before we merge this?

PR #3866, opened 2024-08-04
Remove some unused pulsar properties

d80tb7:
Remove various Pulsar properties that are no longer used


config/armada/config.yaml
```diff
@@ -69,13 +69,9 @@ submission:
 pulsar:
   URL: "pulsar://pulsar:6650"
   jobsetEventsTopic: "events"
-  redisFromPulsarSubscription: "RedisFromPulsar"
```
robertdavidsmith, line 72: Remove `RedisFromPulsarSubscription` from `types.go` below too?

d80tb7: ah yes- good point!

robertdavidsmith (approved):

PR #3871, opened 2024-08-08
[Lookout] Fix jobspec yaml to include job labels

JamesMurkin:
Labels were unintentionally omitted from the yaml we display for a job.
 - This causes confusion and also prevents copy/pasting the job yaml to resubmit the same job


samclark (approved):
LGTM

PR #3876, opened 2024-08-12
Make the executor close LeaseJobRuns stream

JamesMurkin:
This makes it so that we properly wait for the EOF on the stream, which indicates the stream has closed successfully 

In addition I've made the stream handling simpler:
 - We no longer try to handle partial success, such as us timing out in the middle of receiving leased jobs
 - No longer special case handle timeout and instead just call recv which will return a timeout error if the context expires


internal/executor/service/lease_requester.go
```diff
@@ -69,45 +69,51 @@ func (requester *JobLeaseRequester) LeaseJobRuns(ctx *armadacontext.Context, req
...
+		case *executorapi.LeaseStreamMessage_End:
+			shouldEndStream = true
+		default:
+			log.Errorf("unexpected lease stream message type %T", typed)
 		}
 	}
 
+	err = closeStream(stream)
+	if err != nil {
+		return nil, err
```
d80tb7, line 95: couldn't we just log a warning here?  We've got all the information we need so do we need to propagate the error?

d80tb7 (approved):

PR #3886, opened 2024-08-22
Remove Armada UUIDs from Event Ingester and Event Api

d80tb7:
Phase two of removing the proto uuids and replacing them with the string versions


internal/server/event/conversion/conversions.go
```diff
@@ -494,39 +438,12 @@ func FromInternalJobRunPreempted(queueName string, jobSetName string, time time.
...
-	jobId, err := armadaevents.UlidStringFromProtoUuid(e.PreemptedJobId)
-	if err != nil {
-		return nil, err
-	}
-	runId, err := armadaevents.UuidStringFromProtoUuid(e.PreemptedRunId)
-	if err != nil {
-		return nil, err
-	}
-
-	preemptiveJobId := ""
```
d80tb7, line 506: we don't set these  anywhere.  We should remove these from the event but we can do that in a separate PR

JamesMurkin (approved):

PR #3890, opened 2024-08-23
Change RunId to be a string inside the scheduler

d80tb7:
Part 3 of removing the proto ids. This change makes runId a string inside the scheduler.  Note that we haven't yet changed the database type to string (this will happen in the next PR) so there is still a bit of awkwardness as the generated type is uuid.


internal/scheduler/jobdb/jobdb.go
```diff
@@ -41,23 +41,23 @@ type JobDb struct {
...
-// UUIDProvider is an interface used to mock UUID generation for tests.
-type UUIDProvider interface {
-	New() uuid.UUID
+// IDProvider is an interface used to mock run id  generation for tests.
+type IDProvider interface {
+	New() string
 }
 
-// RealUUIDProvider calls uuid.New.
+// RealUUIDProvider generates idsusing a UUID
```
JamesMurkin, line 56:
```suggestion
// RealUUIDProvider generates an id using a UUID
```

pkg/armadaevents/uuid.go
```diff
@@ -102,6 +102,16 @@ func ProtoUuidFromUuidString(uuidString string) (*Uuid, error) {
 	return ProtoUuidFromUuid(id), nil
 }
 
+// MustProtoUuidFromUuidString parses a ULID string into a proto UUID and returns it.
```
JamesMurkin, line 105:
```suggestion
// MustProtoUuidFromUuidString parses a UUID string into a proto UUID and returns it.
```

JamesMurkin (approved):

PR #3892, opened 2024-08-23
Add a generic log retrieval client to Armada Python Client

ClifHouck:
Adds a job log retrieval client to Armada's Python client library. Users of the library are meant to use `JobLogClient` while underneath the hood it relies on a new `BinocularsClient`.

### DIFF
--- a/client/python/armada_client/internal/binoculars_client.py
+++ b/client/python/armada_client/internal/binoculars_client.py
@@ -1,5 +1,6 @@
 from typing import Optional
 
+import grpc
 
 from armada_client.armada import (
     binoculars_pb2,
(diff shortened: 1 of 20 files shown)


d80tb7 (dismissed):

client/python/armada_client/internal/binoculars_client.py
```diff
@@ -0,0 +1,87 @@
...
+
+from armada_client.armada import (
+    binoculars_pb2,
+    binoculars_pb2_grpc,
+)
+
+from armada_client.k8s.io.api.core.v1 import generated_pb2 as core_v1
+
+
+def new_binoculars_client(url: str, disable_ssl: bool = False):
```
dejanzele, line 13: Should it by default expect a secure connection?

ClifHouck: The way it's written, by default it will try to use SSL.

ClifHouck:
I think this is good to go, if I can get approvals again.

dejanzele (approved):

d80tb7 (approved):

PR #3959, opened 2024-09-20
Remove assertions from preempting_queue_scheduler.go

d80tb7:


JamesMurkin:
I think you could also remove NodeJobDiff from NodeDb now - as you are removing the only place it is called

JamesMurkin (approved):

eleanorpratt (approved):

PR #3983, opened 2024-10-04
Scheduler: refactor scheduling_algo

robertdavidsmith:
Gradually pushing `schedulerobjects.Executor` and `schedulerobjects.Node` out of the lower level functions, so these are only used at the top level. Lower-level code should use `internaltypes.Node`.

The step after this will be to make `getCapacityForPool` work on `internaltypes`, but that will imply moving fairness and constraints to `internaltypes` too, so let's commit this small step first.


robertdavidsmith:
Very bad conflicts, replaced with https://github.com/armadaproject/armada/pull/3998

PR #3986, opened 2024-10-07
Marking flags under scheduling-report as deprecated.  (#240)

MustafaI:
* Marking flags under scheduling-report as deprecated. Opening docs in text viewer.

* Adding tty check before opening text viewer

* Removing comment


cmd/armadactl/cmd/docs.go
```diff
@@ -43,10 +48,45 @@ func docsCmd() *cobra.Command {
 				return fmt.Errorf("could not render content from documentation file: %s", err)
 			}
 
-			fmt.Println(out)
+			if term.IsTerminal(int(os.Stdout.Fd())) {
+				if err = openTextViewer(out); err != nil {
+					return fmt.Errorf("could open documentation in text viewer: %s", err)
```
robertdavidsmith, line 53: - Change `could` to `could not`
- If it can't open the text viewer suggest
   - Write that message as a warning to stderr
   - Dump the output to stdout

PR #3998, opened 2024-10-10
Scheduler: refactor - more schedulerobjects to internaltypes

robertdavidsmith:
- Replace `schedulerobjects` with `internaltypes` in more places
- Add `AllResourceRequirements` to `job.go`
- Improve unit tests for `job.go`


internal/scheduler/testfixtures/testfixtures.go
```diff
@@ -343,15 +344,22 @@ func WithNodeAffinityJobs(nodeSelectorTerms []v1.NodeSelectorTerm, jobs []*jobdb
 }
 
 func WithRequestsJobs(rl schedulerobjects.ResourceList, jobs []*jobdb.Job) []*jobdb.Job {
```
robertdavidsmith, line 346: Proper deep copy, rather than mutating things that would be immutable if only go had the `readonly` keyword.

d80tb7 (approved):

PR #4003, opened 2024-10-12
Move lookback limit into the QueuedJobsIterator 

d80tb7:
Move the lookback limit up the stack from the `QueuedGangIterator` to the `QueuedJobsIterator`.  This has a couple of benefits:

- Simplifies the `QueuedGangIterator`, which is already fairly complex
- Means we can get rid of all the logic about when "Evicted" jobs need to contribute to the lookback.  `QueuedJobsIterator` only ever contains queued jobs (as opposed to evicted jobs) so the lookback always applies.

### DIFF
--- a/internal/scheduler/scheduling/queue_scheduler_test.go
+++ b/internal/scheduler/scheduling/queue_scheduler_test.go
@@ -358,18 +358,6 @@ func TestQueueScheduler(t *testing.T) {
 			Queues:                   testfixtures.SingleQueuePriorityOne("A"),
 			ExpectedScheduledIndices: []int{0},
 		},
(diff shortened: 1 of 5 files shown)


internal/scheduler/scheduling/queue_scheduler_test.go
```diff
@@ -358,18 +358,6 @@ func TestQueueScheduler(t *testing.T) {
 			Queues:                   testfixtures.SingleQueuePriorityOne("A"),
 			ExpectedScheduledIndices: []int{0},
 		},
-		"MaxQueueLookback": {
```
d80tb7, line 361: This no longer applies

PR #4007, opened 2024-10-16
Some more simulator improvements

d80tb7:
- Remove some unused code
- Add back in fast forward 
- Hard code all scheduler rate limiting to be unlimited as any other value is objectively wrong until we make the rate limiter respect the simulator clock.
- Don't clone the config- this was only previously needed as the simulator accessed this config from multiple threads and this is no longer the case.
- Scheduler now logs every 15 seconds of elapsed time rather than every 60 seconds of simulated time.

### DIFF
--- a/internal/scheduler/simulator/simulator_test.go
+++ b/internal/scheduler/simulator/simulator_test.go
@@ -484,39 +484,6 @@ func TestSimulator(t *testing.T) {
 	}
 }
 
-func TestSchedulingConfigsFromPattern(t *testing.T) {
-	actual, err := SchedulingConfigsFromPattern("./testdata/configs/basicSchedulingConfig.yaml")
-	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


internal/scheduler/simulator/simulator_test.go
```diff
@@ -484,39 +484,6 @@ func TestSimulator(t *testing.T) {
...
 }
 
-func TestSchedulingConfigsFromPattern(t *testing.T) {
-	actual, err := SchedulingConfigsFromPattern("./testdata/configs/basicSchedulingConfig.yaml")
-	require.NoError(t, err)
-	expected := []configuration.SchedulingConfig{GetBasicSchedulingConfig()}
-	assert.Equal(t, expected, actual)
-}
-
-//func TestClusterSpecsFromPattern(t *testing.T) {
```
robertdavidsmith, line 494: Either delete or uncomment?

robertdavidsmith (approved):

PR #4012, opened 2024-10-18
Cleanup unused ownership authorization code

Sovietaced:
I have been digging into the authorization code just to validate what Armada is and isn't doing. We're running an old version of Armada from June and most of the unused code was cleaned up but I still found this lingering around.


dejanzele:
@d80tb7 this looks like a safe cleanup, is it good to be merged?

d80tb7 (approved):

dejanzele (approved):

PR #4014, opened 2024-10-21
Add restricted tolerations validator

d80tb7:
Add restricted tolerations validator to the api.  This rejects any job submission where the user has set a validation key that is on a denylist.  Aim here is to prevent users from targeting nodes that they shouldn't (e.g. unscheduled nodes)


internal/server/submit/validation/submit_request.go
```diff
@@ -384,3 +385,27 @@ func validateInitContainerCpu(j *api.JobSubmitRequestItem, config configuration.
...
+	}
+
+	jobTolerationKeys := map[string]bool{}
+	for _, s := range spec.Tolerations {
+		jobTolerationKeys[s.Key] = true
+	}
+
+	for _, restricted := range config.RestrictedTolerationKeys {
+		if jobTolerationKeys[restricted] {
+			return fmt.Errorf("toleration %s is not uer settable", restricted)
```
JamesMurkin, line 407:
```suggestion
			return fmt.Errorf("toleration %s is not user settable", restricted)
```

JamesMurkin (approved):

Issue #4018, opened 2024-10-21
Armada Scheduler Initialization Behavior Can Lead to OOMs


Sovietaced:
During initialization, the Armada Scheduler pulls the entire lifetime of jobs and job runs from the database in order to build its in-memory state. This leads to a memory consumption behavior that is directly related to the number of jobs and job runs in the database. This is a particularly sneaky issue when doing a rolling deployment of Armada after some stability because the Armada Scheduler ends up crash looping if memory requests/limits are not high enough. 

I understand there is a pruning functionality to garbage collect jobs and job runs from the database but the initialization behavior of the Armada Scheduler can be made much more efficient considering that for initialization it only needs to materialize state for non-terminal jobs. Afterwards, the normal sync logic that checks for diffs based on the serial number is adequate. Ideally with improvements to the initialization sync behavior the memory consumption should be much more predictable in most use cases. 

Below is an example of repeated OOM Killing until the scheduler's resources have been increased. At that point you can see the memory hump during initialization.

<img width="2676" alt="Screenshot 2024-10-21 at 2 25 09 PM" src="https://github.com/user-attachments/assets/0eeafe9f-d76f-46f7-aa9c-371a9697e9f2">

We have made a fix for this internally that I'm happy to upstream.

d80tb7:
Yes that would be really useful- it's an issue we've been meaning to fix for a while now but haven't got round to.  I'd be happy to look at a PR for this.

dejanzele:
Hey @Sovietaced,

Thanks for raising this issue!

We are looking forward to your PR.

Feel free to reach out if you need any support.

Sovietaced:
I'll have something up in a few days, gotta polish it up :)

PR #4026, opened 2024-10-27
Update scheduler initialization behavior to avoid OOM

Sovietaced:
Fixes https://github.com/armadaproject/armada/issues/4018

This pull request updates the scheduler logic such that it only pulls non-terminal jobs and their associated job runs to load into the in-memory job db. Additionally, it fetches that latest job and job run serials and sets its not he in-memory job db such that later syncs only fetch the true delta of jobs and job runs after initialization (since there may been jobs that completed after the most recent non-terminal job started running).

This has reduced our memory consumption considerably on startup however it will still grow related to the number of non-terminal jobs. Additionally, initialization still takes an awful long time for us (~2 mins) so I assume that there is a linear scan going without indexes on the various terminal-state columns. That is something I may try and optimize in another pull request.

### DIFF
--- a/internal/scheduler/database/job_repository.go
+++ b/internal/scheduler/database/job_repository.go
@@ -80,8 +80,8 @@ func NewPostgresJobRepository(db *pgxpool.Pool, batchSize int32) *PostgresJobRep
 }
 
 func (r *PostgresJobRepository) FetchInitialJobs(ctx *armadacontext.Context) ([]Job, []Run, error) {
-	var updatedJobs []Job = nil
-	var initialRuns []Run = nil
+	var updatedJobs []Job
+	var initialRuns []Run
 
 	start := time.Now()
 	defer func() {
@@ -104,6 +104,10 @@ func (r *PostgresJobRepository) FetchInitialJobs(ctx *armadacontext.Context) ([]
 		initialJobRows, err := fetch(0, r.batchSize, func(from int64) ([]SelectInitialJobsRow, error) {
 			return queries.SelectInitialJobs(ctx, SelectInitialJobsParams{Serial: from, Limit: r.batchSize})
 		})
+		if err != nil {
+			return err
+		}
+
 		updatedJobs = make([]Job, len(initialJobRows))
 		updatedJobIds := make([]string, len(initialJobRows))
 		for i, row := range initialJobRows {
@@ -127,10 +131,6 @@ func (r *PostgresJobRepository) FetchInitialJobs(ctx *armadacontext.Context) ([]
 				Serial:                  row.Serial,
 				Pools:                   row.Pools,
 			}
-		}
-
-		if err != nil {
-			return err
 		}
 
 		// Fetch dbRuns
@@ -366,8 +366,8 @@ func (r *PostgresJobRepository) CountReceivedPartitions(ctx *armadacontext.Conte
 }
 
 func (r *PostgresJobRepository) FetchLatestSerials(ctx *armadacontext.Context) (int64, int64, error) {
-	var jobSerial int64 = 0
-	var jobRunSerial int64 = 0
+	var jobSerial int64
+	var jobRunSerial int64
 
 	err := pgx.BeginTxFunc(ctx, r.db, pgx.TxOptions{
 		IsoLevel:       pgx.RepeatableRead,
--- a/internal/scheduler/database/job_repository.go
+++ b/internal/scheduler/database/job_repository.go
@@ -36,6 +36,13 @@ type JobRunLease struct {
 
 // JobRepository is an interface to be implemented by structs which provide job and run information
 type JobRepository interface {
+
+	// FetchInitialJobs returns all non-terminal jobs and their associated job runs.
+	FetchInitialJobs(ctx *armadacontext.Context) ([]Job, []Run, error)
+
+	// FetchLatestSerials returns the latest job and job runs serials in the DB for use during initialization.
+	FetchLatestSerials(ctx *armadacontext.Context) (int64, int64, error)
+
 	// FetchJobUpdates returns all jobs and job dbRuns that have been updated after jobSerial and jobRunSerial respectively
 	// These updates are guaranteed to be consistent with each other
 	FetchJobUpdates(ctx *armadacontext.Context, jobSerial int64, jobRunSerial int64) ([]Job, []Run, error)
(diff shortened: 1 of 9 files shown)


magefiles/tests.go
```diff
@@ -52,12 +52,20 @@ func Tests() error {
 		return err
 	}
 
-	err = dockerRun("run", "-d", "--name=redis", docker_Net, "-p=6379:6379", "redis:6.2.6")
+	if len(docker_Net) > 0 {
```
Sovietaced, line 55: This is a bit of an ugly fix for running tests on Mac OS. For Mac OS docker net is empty string which breaks the `dockerRun` call.

internal/scheduler/database/job_repository.go
```diff
@@ -73,6 +79,71 @@ func NewPostgresJobRepository(db *pgxpool.Pool, batchSize int32) *PostgresJobRep
 	}
 }
 
+func (r *PostgresJobRepository) FetchInitialJobs(ctx *armadacontext.Context) ([]Job, []Run, error) {
+	var updatedJobs []Job = nil
```
dejanzele, line 83:
```suggestion
	var updatedJobs []Job
```

Default value is `nil`

internal/scheduler/database/job_repository.go
```diff
@@ -73,6 +79,71 @@ func NewPostgresJobRepository(db *pgxpool.Pool, batchSize int32) *PostgresJobRep
...
+	var initialRuns []Run = nil
```
dejanzele, line 84: same

internal/scheduler/database/job_repository.go
```diff
@@ -73,6 +79,71 @@ func NewPostgresJobRepository(db *pgxpool.Pool, batchSize int32) *PostgresJobRep
...
+				Failed:                  row.Failed,
+				SchedulingInfo:          row.SchedulingInfo,
+				SchedulingInfoVersion:   row.SchedulingInfoVersion,
+				Serial:                  row.Serial,
+				Pools:                   row.Pools,
+			}
+		}
+
+		if err != nil {
+			return err
```
dejanzele, line 133: If there is useful additional info, should we wrap the error? If not, maybe use `errors.WithStack`?

internal/scheduler/database/job_repository.go
```diff
@@ -294,6 +365,33 @@ func (r *PostgresJobRepository) CountReceivedPartitions(ctx *armadacontext.Conte
 	return uint32(count), nil
 }
 
+func (r *PostgresJobRepository) FetchLatestSerials(ctx *armadacontext.Context) (int64, int64, error) {
+	var jobSerial int64 = 0
```
dejanzele, line 369: The `= 0` part can be omitted as that is the default value

internal/scheduler/database/job_repository.go
```diff
@@ -73,6 +79,71 @@ func NewPostgresJobRepository(db *pgxpool.Pool, batchSize int32) *PostgresJobRep
...
+				Succeeded:               row.Succeeded,
+				Failed:                  row.Failed,
+				SchedulingInfo:          row.SchedulingInfo,
+				SchedulingInfoVersion:   row.SchedulingInfoVersion,
+				Serial:                  row.Serial,
+				Pools:                   row.Pools,
+			}
+		}
+
+		if err != nil {
```
dejanzele, line 132: After looking at the `fetch` function, if it returns an error, the returned jobs will be `nil`.
Can we move the error check right after the `fetch` call?

internal/scheduler/database/job_repository.go
Sovietaced: Most of this is copy pasted from the existing `FetchJobUpdates` code. Do you want me to fix that code too?

internal/scheduler/database/job_repository.go
Sovietaced: Again, I am following the convention set in the code that is already here

internal/scheduler/database/job_repository.go
Sovietaced: Yeah that is fair, again just copy pasted behavior

internal/scheduler/database/job_repository_test.go
```diff
@@ -545,6 +719,66 @@ func TestFetchJobRunLeases(t *testing.T) {
...
+				require.NoError(t, err)
+
+				// Fetch updates
+				jobSerial, jobRunSerial, err := repo.FetchLatestSerials(ctx)
+				require.NoError(t, err)
+
+				// Assert results
+				assert.Equal(t, tc.latestJobSerial, jobSerial)
+				assert.Equal(t, tc.latestJobRunSerial, jobRunSerial)
+				cancel()
```
dejanzele, line 774: Maybe it is better to do the `cancel` in a `t.Cleanup` to prevent context leaks, even though this are tests, no reason to have lingering goroutines as the `require.NoError` will terminate in case of an error and the context won't get cancelled :)

Sovietaced: I was following convention here but yeah I dunno why the code doesn't already do that

dejanzele:
@d80tb7 can you please take a look?

internal/scheduler/scheduler.go
```diff
@@ -356,6 +356,54 @@ func (s *Scheduler) cycle(ctx *armadacontext.Context, updateAll bool, leaderToke
...
+			// We receive nil jobs from jobDb.ReconcileDifferences if a run is updated after the associated job is deleted.
+			// These nil job must be sorted out.
+			jobDbJobs = append(jobDbJobs, jst.Job)
+		}
+	}
+	if err := txn.Upsert(jobDbJobs); err != nil {
+		return nil, nil, err
+	}
+
+	latestJobSerial, latestJobRunSerial, err := s.jobRepository.FetchLatestSerials(ctx)
```
d80tb7, line 389: I don't think this is quite right.  We fetch this in a separate database transaction from the initial load so we'll miss any updates that have occurred between `jobRepository.FetchInitialJobs` and here.

I think we should be able to derive `latestJobSerial` and `latestJobRunSerial` as the max serials of `initialJobs` and `initialRuns` respectively.

internal/scheduler/scheduler.go
```diff
@@ -356,6 +356,54 @@ func (s *Scheduler) cycle(ctx *armadacontext.Context, updateAll bool, leaderToke
 	return overallSchedulerResult, nil
 }
 
+// syncInitialState updates jobs in jobDb to match state in postgres and returns all in-progress jobs.
+func (s *Scheduler) syncInitialState(ctx *armadacontext.Context) ([]*jobdb.Job, []jobdb.JobStateTransitions, error) {
```
d80tb7, line 360: I *think*  the only difference between this function and `syncState` needs to be the queries we use.  E.g. `s.jobRepository.FetchInitialJobs`  vs `s.jobRepository.FetchJobs`.  We could just modify syncState to take an `initialise` parameter of type bool which selects the appropriate query.

d80tb7: Looks good, but I think there's in issue with potentially missed updates in `syncInitialState`.  I think we can fix this by simply modifying `syncState` to use the new queries rather than creating a separate  ` syncInitialState` function but happy to entertain other solutions.

internal/scheduler/database/job_repository.go
dejanzele: I guess we can fix it

internal/scheduler/scheduler.go
Sovietaced: Ah I see that now, you're correct I'll fix this up.

internal/scheduler/scheduler.go
Sovietaced: I tI think an edge case is that the serials could still be quite old but in most cases they should be fairly up to date so I'm ok with simplifying things.

d80tb7 (approved):

d80tb7:
Looks great- thank you!

dejanzele (approved):

PR #4070, opened 2024-11-29
POC of Market Based Pricing

d80tb7:
Proof Of Concept Of Market Based Pricing.  Main changes are as follows:

- Job submissions can now have an `ExperimentalPricingInfo` section added.  This allows a `BidPrice` to be set.
- In the Scheduler configuration pools may be configured to be `MarketDriven`.  If this flag is set then the pool will be use price-based scheduling rahter than fair-share-based scheduling.
- When using price-based scheduling, jobs are shceduled in price order wioth no consideration to fair share.  Jobs are preempted as necessary including self-preemption if you outbid yourself.
- For `MarketDriven` pools a `SpotPrice` metric is produced.  This will be zero if the pool is less than 95% utilized, otherswise it will be eqaul to the lowest priced job (running or scheduled).
- Fair share scheduling is unaffected.


.run/Executor.run.xml
```diff
@@ -8,7 +8,7 @@
       <env name="ARMADA_EXECUTORAPICONNECTION_FORCENOTLS" value="true" />
       <env name="ARMADA_HTTPPORT" value="8085" />
       <env name="HOME" value="$USER_HOME$/" />
-      <env name="KUBECONFIG" value="$PROJECT_DIR$/.kube/internal/config" />
+      <env name="KUBECONFIG" value="$PROJECT_DIR$/.kube/external/config" />
```
d80tb7, line 11: This is intentional.  The run config for the executor is incorrect as `internal` will attempt to connect via the kind network, which won't work from Goland.  `external` will attempt to connect on loclahost which is correct.

internal/scheduler/jobdb/job.go
```diff
@@ -413,6 +423,13 @@ func (job *Job) WithPriority(priority uint32) *Job {
 	return j
 }
 
+// WithBidPrice returns a copy of the job with the bidPrice updated.
+func (job *Job) WithBidPrice(price float64) *Job {
+	j := copyJob(*job)
```
robertdavidsmith, line 428: Arguably you want DeepCopyJob, but then every other With* method has copyJob :(

internal/scheduler/jobdb/jobdb.go
```diff
@@ -134,6 +175,7 @@ func (jobDb *JobDb) NewJob(
 	jobSet string,
 	queue string,
 	priority uint32,
+	price float64,
```
robertdavidsmith, line 178: Call this bidPrice for consistency?

internal/scheduler/metrics/cycle_metrics.go
```diff
@@ -193,6 +194,14 @@ func newCycleMetrics() *cycleMetrics {
 		poolQueueAndResourceLabels,
 	)
 
+	spotPrice := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: prefix + "spot_price",
+			Help: "spop price for given pool",
```
robertdavidsmith, line 200: s/spop/spot/

internal/scheduler/jobdb/job.go
d80tb7: I *think* this is fine!

robertdavidsmith (approved):

PR #4124, opened 2025-01-06
Align aggregate counts for job statuses

mauriceyap:
On the Lookout UI, when grouping is enabled for the jobs table, change how counts for different statuses are displayed, so that they are aligned and are less visually noisy.


mauriceyap:
image
image
image

PR #4131, opened 2025-01-10
Improve handling when node is running jobs from multiple pools

JamesMurkin:
The scenario where a job is running jobs from multiple pools should only happen when a node has its pool changed
 - The jobs already running will have a pool that doesn't match the nodes current pool

Currently what happens is that the scheduler sees the node as empty and double schedules the node

This PR changes that, so we set any resource used by jobs in other pools as unallocatable, to prevent this double scheduling
 - This is a slightly more generic approach than the current approach where we just mark jobs from Away pools as unallocatable

I had to change when we calculate jobsByPool to calculate for running jobs of all pools


internal/scheduler/scheduling/scheduling_algo.go
```diff
@@ -238,35 +238,37 @@ func (l *FairSchedulingAlgo) newFairSchedulingAlgoContext(ctx *armadacontext.Con
...
 		}
-		if slices.Contains(otherPool.AwayPools, pool.Name) {
-			homeJobs = append(homeJobs, jobSchedulingInfo.jobsByPool[otherPool.Name]...)
+		if slices.Contains(pool.AwayPools, currentPool.Name) {
+			// Jobs from away pools need to be considered in the current scheduling round, so should be added here
+			// This is so the jobs are available for eviction, if a home job needs to take their place
+			currentPoolJobs = append(currentPoolJobs, jobSchedulingInfo.jobsByPool[pool.Name]...)
+		} else {
+			// Jobs not used by the current pool belong to other pools we aren't currently considering
+			// Add them here, so their resource can made unallocatable in the nodeDb, preventing us scheduling over them
```
robertdavidsmith, line 254: Add "this can happen if a node's pool is changed" ?

robertdavidsmith (approved):

PR #4147, opened 2025-01-16
[executor] Allow retrying fail pods that haven't had any containers start

JamesMurkin:
This PR makes it so you can configurable retry failed pods, much like the pending pod checks

The current caveat is that we only allow this for pods that have not started any containers, maybe we'll adjust this restriction at some point but we want to be conservative adding this feature for now

This feature for now is largely going to used to retry pods that have hit various Pod Allocation issues you can have (gpu drivers failing to allocate gpus, k8s failing to allocate containers etc)


internal/executor/service/job_state_reporter.go
```diff
@@ -0,0 +1,233 @@
...
+		return
+	}
+
+	event, err := reporter.CreateEventForCurrentState(pod, stateReporter.clusterContext.GetClusterId())
+	if err != nil {
+		log.Errorf("Failed to report event: %v", err)
+		return
+	}
+
+	if pod.Status.Phase == v1.PodFailed {
```
JamesMurkin, line 95: The code in this if statement is largely the only new code, the rest is just split out from job_event_reporter

internal/executor/podchecks/failedpodchecks/event_checks.go
```diff
@@ -0,0 +1,68 @@
...
+
+type PodEventsChecker struct {
+	checks []podEventCheck
+}
+
+func NewPodEventsChecker(checks []podchecks.PodEventCheck) (*PodEventsChecker, error) {
+	podEventChecks := make([]podEventCheck, 0, len(checks))
+
+	for _, check := range checks {
+		re, err := regexp.Compile(check.Regexp)
```
d80tb7, line 30: At some point it would be nice to see if we can get viper to marshall regexes and push stuff like this to config validation.

d80tb7 (dismissed):

d80tb7 (approved):

PR #4181, opened 2025-02-03
Add support for custom job service names

### LINKED ISSUE #4171: Make service name configurable during job submission
We have a use case where we are submitting a gang of jobs for scheduling and it would be beneficial to have a deterministic and stable service name during submission time. More specifically we are submitting a Ray cluster with a head node and many worker nodes. The worker nodes use an init-container to block until the head node is reachable via a service. Given that we are using third-party containers it would be nice if we could define the jobs and init containers such that they referenced a predefined service name instead of implementing service discovery at runtime to learn of the service name based on the non-deterministic job ID.

Sovietaced:
Fixes #4171

This pull request adds a fix for #4171 and includes a feature flag to enable the new features (defaults to false). It includes an update to the `ServiceConfig` message that allows for an optional service name to be configured.

As the PR currently sits, service name conflicts result in a recoverable error from the executor. I'm happy to change this but the error is plumbed to the user so I feel the user experience is ok and theoretically this could be a recoverable situation. 

Screenshot 2025-02-03 at 4 50 16 PM


internal/server/submit/conversion/conversions_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"math"
 	"testing"
 
+	"github.com/armadaproject/armada/internal/server/configuration"
```
d80tb7, line 7: this needs to be in the block with the armada imports at the bottom (annoyingly linting isn't intellligent enough to fix/catch this)

pkg/api/submit.proto
```diff
@@ -41,6 +41,7 @@ message IngressConfig {
 message ServiceConfig {
     ServiceType type = 1;
     repeated uint32 ports = 2;
+    string name = 3;
```
d80tb7, line 44: lets put a comment saying what this does and what the behaviour is if you don't set it.

internal/server/submit/conversion/conversions_test.go
dejanzele: this is weird, `goimports` should catch this, I will check

Sovietaced:
@d80tb7 good for another review

internal/server/submit/conversion/conversions.go
```diff
@@ -87,7 +89,8 @@ func convertIngressesAndServices(
 	for _, ingressConfig := range jobReq.Ingress {
 		ports := filterServicePorts(availableServicePorts, ingressConfig.Ports)
 		if len(ports) > 0 {
-			serviceObject := createService(jobId, serviceIdx, ports, v1.ServiceTypeClusterIP, ingressConfig.UseClusterIP)
+			serviceName := fmt.Sprintf("%s-service-%d", common.PodName(jobId), serviceIdx)
```
dejanzele, line 92: This can maybe be extracted into a function, as it is also used on line 118.

dejanzele (approved):

PR #4182, opened 2025-02-04
Scheduler To Use internaltypes.JobSchedulingInfo

d80tb7:
In this PR we introduce a new `internaltypes.JobSchedulingInfo` and us that inside the scheduler as a replacement for `schedulerobjects.JobSchedulingInfo`.  Motivation here is to ensure that the internals of the scheduler can use objects where we control the code, which is not the case for `schedulerobjects.JobSchedulingInfo` which use code generated by protoc.  The long term goal here is to make `internaltypes.JobSchedulingInfo` fully immutable, although that is not done in this PR so as to minimilse the number of code changes.


internal/scheduler/internaltypes/job_scheduling_info.go
```diff
@@ -0,0 +1,114 @@
...
+	if podRequirements == nil {
+		return nil, errors.Errorf("job must have pod requirements")
+	}
+	return &JobSchedulingInfo{
+		Lifetime:          j.Lifetime,
+		PriorityClassName: j.PriorityClassName,
+		SubmitTime:        j.SubmitTime,
+		Priority:          j.Priority,
+		PodRequirements: &PodRequirements{
+			NodeSelector:         podRequirements.NodeSelector,
```
robertdavidsmith, line 82: Should arguably deep copy here

internal/scheduler/internaltypes/job_scheduling_info.go
```diff
@@ -0,0 +1,114 @@
...
+		PriorityClassName: j.PriorityClassName,
+		SubmitTime:        j.SubmitTime,
+		Priority:          j.Priority,
+		ObjectRequirements: []*schedulerobjects.ObjectRequirements{
+			{
+				Requirements: &schedulerobjects.ObjectRequirements_PodRequirements{
+					PodRequirements: &schedulerobjects.PodRequirements{
+						NodeSelector:         podRequirements.NodeSelector,
+						Affinity:             podRequirements.Affinity,
+						Tolerations:          podRequirements.Tolerations,
```
robertdavidsmith, line 105: Should arguably deep copy here

robertdavidsmith (approved):

PR #4188, opened 2025-02-04
Improve configurability of ingress config

### LINKED ISSUE #4187: Improve configurability of Ingress during Job Submission
We are trying to use the ingress feature during job submission but it doesn't seem compatible with our environment for a few reasons. 
1. We are not able to configure the ingress class (in our case `nginx`)
2. We are not able to control the TLS hosts. We need this to [terminate TLS](https://kubernetes.github.io/ingress-nginx/examples/tls-termination/). 

Our use case is to setup a Ray cluster head node and expose the Ray dashboard outside the cluster. We're setting up an ephemeral DNS record for this ingress and we must be able to configure the DNS name as the TLS host in order for the ingress to terminate TLS connections targeting the DNS name. The existing code generates TLS hosts based on the service port name, job ID, and job namespace which is not suitable for us. 

I think the `IngressConfig` should be updated to support a configurable ingress class as well as additional TLS hosts to configure in addition to the autogenerated ones.

Sovietaced:
Fixes #4187


Sovietaced:
This pull requests needs more work. In order to get our setup to work we have to change even more of the code in `conversions.go` which seems like it would be pretty disruptive.

PR #4197, opened 2025-02-07
Improve allocatable resource calculation - nodeDb

JamesMurkin:
- We no longer add nodes to the nodeDb if they are unschedulable with no jobs assigned to them
   - These nodes should never get anything assigned to them, so no need to include them in the nodeDb
   - This means these nodes resource is not counted towards total allocatable resource, which is used for fairshare purposes
 - Don't allow Armada to schedule over unallocatableResources
   - node now has an allocatableResources resources field, which is calculated as  allocatableResources = totalResources - sum(unallocatableResources)
   - This is how much resource is actually alloctable, and we no longer use totalResources for fairshare calculation

The net result here is:
 - Resource used for fairshare calculate better represents the amount of resource available to be scheduled on - Unallocatable resource is properly respected - Nodes that cannot be scheduled on do not skew the fairshare calculation

The fairshare calculation still is "wrong" in the case an unschedulable node has jobs running on it, but those jobs don't fill the node
 - The remaning resource is used in the fairshare calculation, but can never actually be allocated


internal/scheduler/internaltypes/node.go
```diff
@@ -63,21 +68,20 @@ func FromSchedulerObjectsNode(node *schedulerobjects.Node,
...
 
 	allocatableByPriority := map[int32]ResourceList{}
 	for _, p := range allowedPriorities {
-		allocatableByPriority[p] = totalResources
-	}
-	for p, rl := range node.UnallocatableResources {
-		MarkAllocated(allocatableByPriority, p, resourceListFactory.FromJobResourceListIgnoreUnknown(rl.Resources))
+		allocatableByPriority[p] = allocatableResources
 	}
 	allocatableByPriority[EvictedPriority] = allocatableByPriority[minInt32(allowedPriorities)]
```
robertdavidsmith, line 83: Simplify to 
```
allocatableByPriority[EvictedPriority] = allocatableResources
```
?

d80tb7 (approved):

PR #4210, opened 2025-02-17
Remove internalTypes.Node.unallocatableResources

d80tb7:
The only place this is used is some logging.  Getting rid of it will be helpful because it paves the way for removing the need to pass per-priority class unavailable resources from the executor to the scheduler.


robertdavidsmith:
While this is a nice tidy, note
- We still need to pass it unavailable resources back to the scheduler because it's used at https://github.com/armadaproject/armada/blob/762900515f4b90cb1cdd80e33f1479f7bc9c3f0b/internal/scheduler/internaltypes/node.go#L75
- In this PR you can also delete https://github.com/armadaproject/armada/blob/762900515f4b90cb1cdd80e33f1479f7bc9c3f0b/internal/scheduler/internaltypes/node.go#L73

JamesMurkin (approved):

PR #4214, opened 2025-02-17
Scala client: enhance mock Armada server for unit tests

richscott:
These changes add better checking of pre-requisite resources (existence of specified queue when submitting a job, etc) in the Armada Server mock for the Scala client. Simple maps are used to track creation/deletion of queues and jobs.

Several Armada client convenience functions now more accurately return GRPC runtime exceptions (instead of just an empty List or object) in the same way that the Scala binding would with a real Armada cluster.  JobIDs are now generated using ULIDs.

A `client/scala/scala-armada-client/.sbtopts` has been created, with a Java define directive to squelch the sbt warnings about converting relative (e.g. `../../../`) paths to absolute, so a bare `sbt build` or `sbt test` should be quieter.


client/scala/scala-armada-client/src/test/scala/ArmadaClientSuite.scala
```diff
@@ -76,7 +93,23 @@ private class SubmitMockServer extends SubmitGrpc.Submit {
   }
 
   def submitJobs(request: JobSubmitRequest): scala.concurrent.Future[JobSubmitResponse] = {
-    Future.successful((new JobSubmitResponse(List(JobSubmitResponseItem("fakeJobId")))))
+    val q = queueMap.get(request.queue)
+    if (q == null) {
+      var resItems = Seq[JobSubmitResponseItem]()
+      request.jobRequestItems.foreach { reqItem =>
+        val jobId: String = ulidGen.base32().toLowerCase()
+        resItems = resItems :+ new JobSubmitResponseItem(jobId, "requested queue does not exist")
```
GeorgeJahad, line 101: I'm confused by this code.  It appears like it is creating a seq of response items, but that seq doesn't appear to get used anywhere.

what am i missing?

client/scala/scala-armada-client/src/test/scala/ArmadaClientSuite.scala
```diff
@@ -103,8 +135,12 @@ private class JobsMockServer extends JobsGrpc.Jobs {
   }
 
   def getJobStatus(request: JobStatusRequest): scala.concurrent.Future[JobStatusResponse] = {
-    val response = new JobStatusResponse(Map("fakeJobId" -> JobState.RUNNING))
-    Future.successful(response)
+    val statusMap = collection.mutable.Map[String,JobState]() // jobID -> state
+    val it = jobMap.keys.asIterator()
+
+    while (it.hasNext()) statusMap.put(it.next(), JobState.RUNNING)
+
```
GeorgeJahad, line 142: I'm still learning, but I think something like this is more scala like:
```
    jobMap.keySet.forEach(k => statusMap.put(k, JobState.RUNNING))
```

client/scala/scala-armada-client/src/test/scala/ArmadaClientSuite.scala
```diff
@@ -136,52 +176,71 @@ class ArmadaClientSuite extends munit.FunSuite {
...
+    val jobId = newJob.jobResponseItems(0).jobId
+    val jobStatus = ac.getJobStatus(jobId)
+    assert(jobStatus.jobStates(jobId).isRunning)
+
+    ac.deleteQueue(qName)
   }
 
-  test("ArmadaClient.GetJobStatus()") {
+  test("ArmadaClient.{get,create,delete}Queue()") {
```
GeorgeJahad, line 224: I'm not familiar with this test harness.  Do the '{}' characters have some special significance, or are they just meant to remind us of bash shell syntax?

GeorgeJahad (approved):
I had a few questions, but it looks good to me.

client/scala/scala-armada-client/src/test/scala/ArmadaClientSuite.scala
richscott: You're not missing a thing!  This was literally dead code that I neglected to remove when I learned that real Armada returns just the GRPC runtime exception, with no enumerated job request items. Removed in  commit 5d569f7d .

client/scala/scala-armada-client/src/test/scala/ArmadaClientSuite.scala
richscott: Good suggestion - I wasn't clear if the iterator pattern or the enumerator pattern was the now-preferred version.  Fixed in  commit 5d569f7d .

client/scala/scala-armada-client/src/test/scala/ArmadaClientSuite.scala
richscott: I don't think these strings have any particular semantic significance in the munit framework - they're just string labels that are printed when each test func is executed. I downcased the function names in the labels to be consistent with the actual function names (I previously missed that in my first client PR).  The `{...}` was just my shortcut to describe the 3 individual funcs that are being exercised in the test. I could break this out into 3 separate test funcs, but that seems like it would be unnecessarily verbose. Right now, the output of `sbt test` looks like:
```
[info] welcome to sbt 1.10.7 (Ubuntu Java 21.0.6)
[info] loading settings for project scala-armada-client-build from plugins.sbt...
[info] loading project definition from /home/rich/work/armada/client/scala/scala-armada-client/project
[info] loading settings for project root from build.sbt...
[info] set current project to Scala-Armada-Client (in build file:/home/rich/work/armada/client/scala/scala-armada-client/)
io.armadaproject.armada.ArmadaClientSuite:
  + ArmadaClient.eventHealth() 0.138s
  + ArmadaClient.submitHealth() 0.008s
  + ArmadaClient.submitJobs() 0.035s
  + ArmadaClient.getJobStatus() 0.017s
  + ArmadaClient.{get,create,delete}Queue() 0.01s
[info] Passed: Total 5, Failed 0, Errors 0, Passed 5
[success] Total time: 1 s, completed Feb 18, 2025, 1:21:46 PM
```
Let me know if you'd prefer it be decomposed into separate test funcs. Thanks.

GeorgeJahad (approved):
lgtm

PR #4224, opened 2025-02-19
Add node allocated/allocatable metrics

JamesMurkin:
This PR adds a metrics for each node showing:
 - allocatable by resource type
 - allocated by resource type

This is useful as it allows us to:
 - See in detail what nodes/resources the scheduler thinks it is scheduling on.
   - Currently we only have aggregated values that are for the clusters view of the world, which doesn't always line up with that nodes the scheduler is scheduling on
 - Allows us to see fragmentation of the nodes by resource type


internal/scheduler/metrics/cycle_metrics.go
```diff
@@ -364,6 +388,24 @@ func (m *cycleMetrics) ReportSchedulerResult(result scheduling.SchedulerResult)
...
 		}
 
+		nodes, err := schedulingStats.NodeDb.GetNodes()
+		if err != nil {
+			log.Errorf("unable to generate node stats as failed to get nodes from nodeDb %s", err)
+		} else {
+			for _, node := range nodes {
+				isSchedulable := strconv.FormatBool(!node.IsUnschedulable())
+				for _, resource := range node.GetAllocatableResources().GetResources() {
+					currentCycle.nodeAllocatableResource.WithLabelValues(node.GetPool(), node.GetName(), node.GetExecutor(), node.GetReportingNodeType(), resource.Name, isSchedulable).Set(float64(resource.RawValue))
```
robertdavidsmith, line 398: `RawValue` needs to be scaled.

You can call `QuantityAsFloat64` defined [here](https://github.com/armadaproject/armada/blob/27cfc6ed1639ef6cdca266e2f5f6353176b7fb00/internal/common/resource/resource.go#L48) on resource.Value

d80tb7 (approved):

PR #4237, opened 2025-02-24
Fix filter button in queue table cells

mauriceyap:
Clicking the filter button next to a value for a queue name in the jobs table currently causes a runtime error. It sets the value of the filter to a string, whereas the queue filter expects an array of strings.


mauriceyap:
Duplicates #4223

PR #4240, opened 2025-02-25
Add typecheck command for Lookout UI

mauriceyap:
Currently, only the source code for the Lookout UI is typechecked with TypeScript. This happens as part of the build process (run using `yarn build`). Test code written in TypeScript is not currently typechecked as part of any CI or developer workflow, so type errors in test code are able to be committed without causing visible build failures.

This change introduces a new `yarn typecheck` command which simply runs `tsc`. The README has been updated to encourage developers to use this command during development. The command is also run in the existing *Tests* GitHub workflow, which will therefore now fail if there are typechecker errors.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,9 +23,9 @@ jobs:
           cache: yarn
           cache-dependency-path: ./internal/lookoutui/yarn.lock
 
-      - name: Install Dependencies And Run Unit Tests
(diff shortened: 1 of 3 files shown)


mauriceyap:
Verified that the change to the GitHub actions workflow works as intended in the check for this PR:
image

d80tb7 (approved):

PR #4248, opened 2025-02-27
Fix bug when calculating fairshare in CostBasedCandidateGangIterator 

JamesMurkin:
The fairshare is supposed to be how much "cost" the queue can have fairly, which is a weighted value

However we were instead passing the unweighted value, which is typically far smaller than the weighted cost
 - As the weight is a fraction

This meant the logic for ordering when queues were below their fairshare was never getting entered

This PR fixes how we calculate the fairshare for a queue to use the weighed value


internal/scheduler/scheduling/scheduling_algo_test.go
```diff
@@ -399,11 +399,12 @@ func TestSchedule(t *testing.T) {
 			schedulingConfig: testfixtures.TestSchedulingConfig(),
 			executors:        []*schedulerobjects.Executor{test1Node32CoreExecutor("executor1")},
 			queues:           []*api.Queue{{Name: "queue1", PriorityFactor: 0.01}, {Name: "queue2", PriorityFactor: 0.01}},
-			queuedJobs:       testfixtures.N32Cpu256GiJobs("queue2", testfixtures.PriorityClass0, 1),
+			queuedJobs:       testfixtures.N16Cpu128GiJobs("queue2", testfixtures.PriorityClass0, 1),
```
JamesMurkin, line 402: I believe this unit test was asserting the wrong thing

It was showing a queue above fairshare would kick a queue below fairshare out of the way - which isn't how it should work

I've inverted it, so the preempting queue is at fairshare and the preempted queue is above fairshare, so it now asserts the expected behaviour

robertdavidsmith (approved):

robertdavidsmith (approved):

PR #4267, opened 2025-03-07
Add armada_scheduler_job_seconds_lost_to_preemption metrics

JamesMurkin:
This metric will:
 - Show time lost to preemption
 - Have configurable job checkpoint times, and calculate time lost considering each different checkpoint time

This metric is useful to determine how much wastage preemption is causing and how much better/worse it'd be for different checkpoint times

At some point we'll make jobs tell us when they last checkpointed so we don't need a configurable list for estimations


internal/scheduler/metrics/state_metrics.go
```diff
@@ -11,26 +13,35 @@ import (
 	"github.com/armadaproject/armada/pkg/armadaevents"
 )
 
+const noCheckpointLabelValue = "none"
```
mauriceyap, line 16: worth moving this into consts.go?

mauriceyap (approved):

internal/scheduler/metrics/state_metrics.go
JamesMurkin: Possibly, somewhat consts is shared constants but also I see your point.

I need to make a follow up PR so will fix it in that one.

PR #4275, opened 2025-03-10
feat: scala-client: update scala version and grpc-netty library

dejanzele:
Fixes #

### DIFF
--- a/client/scala/armada-scala-client/build.sbt
+++ b/client/scala/armada-scala-client/build.sbt
@@ -1,4 +1,4 @@
-val scala2Version = "2.13.15"
+val scala2Version = "2.13.16"
 
 lazy val root = project
   .in(file("."))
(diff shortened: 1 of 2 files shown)


EnricoMi (approved):
LGTM!

client/scala/armada-scala-client/build.sbt
```diff
@@ -20,7 +20,7 @@ Compile / PB.targets := Seq(
 Compile / PB.protoSources ++= Seq(file("./proto"))
 
 libraryDependencies ++= Seq(
-    "io.grpc" % "grpc-netty" % scalapb.compiler.Version.grpcJavaVersion,
+    "io.grpc" % "grpc-netty" % "1.71.0",
```
EnricoMi, line 23: As I mentioned in Slack:
The version picked here should work best with Armada and for any user, it should not be specifically picked for Spark. If Spark does not work with it, we need to deal with it where Spark comes into the game (Spark cluster manager), not here. This project should be Spark-agnostic.

EnricoMi:
@dejanzele what is the motivation for upgrading these dependecies?

PR #4279, opened 2025-03-11
Integrate external java-client into Armada repository

### LINKED ISSUE #4264: Integration external java-client into Armada repository
There is an Armada client library for the Java language at https://github.com/armadaproject/java-client/ . In order to more closely track any GRPC API changes in Armada, and to more easily add test logic for that library code, we should move that code directly into the Armada repo (https://github.com/armadaproject/armada/), just as we currently have the Python, C#/.Net, and Scala clients.

The code should be placed underneath the existing `client/` directory, e.g. `client/java/`. Probably we should preserve the existing layout of the current external `java-client` repo underneath this directory, and if possible, we should see if we can somehow use the original commit history, to preserve the original authorship.

Also, the current `java-client` repo does not have unit tests. After relocating the code, we should add tests for this code. The Scala client has basic unit tests, plus a mock Armada server setup - since Scala has interoperability with Java, we should investigate enhancing/adapting the Scala client unit tests to also link with the java-client library (instead of the Scala client library), and attempt using the Scala client tests to exercise the java client code.

sudiptob2:
Fixes #4264  

### Tasks  

- [x] Move the [Java client](https://github.com/armadaproject/java-client/) to the `client/java` directory while preserving the **original author's commit history.**  
- [x] Add build scripts (Dockerfile and Magefiles).  
- [x] Implement unit tests.
- [x] Add documentation
- [x] Update ci
- [x] Add unit test in the ci


client/java/.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
+name: CI
```
EnricoMi, line 1: This has to be integrated into the repo root directory, a good name would be `.github/workflow/java-client.yml`.

client/java/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
EnricoMi, line 1: Probably not needed as there is a LICENSE file in the root directory.

scripts/build-java-client.sh
```diff
@@ -0,0 +1,20 @@
...
+
+ROOT=$(pwd)
+JDIR=client/java
+cd $ROOT/$JDIR
+
+# Build the Java client using Maven
+mvn clean install
+
+if [ $? -eq 0 ]; then
+  jarfile=$(find target -type f -name 'java-client*.jar')
```
EnricoMi, line 12: This finds multiple jars if there are sources and javadoc jars as well (the case for the Scala client):
https://github.com/armadaproject/armada/blob/daac225a8549d98bb588213d9314c47df54a6fe5/scripts/build-scala-client.sh#L16-L17

client/java/.github/workflows/ci.yml
sudiptob2: updated

scripts/build-java-client.sh
EnricoMi: Yes, they are all in the `target` dir. Thanks!

.github/workflows/java-client-release-to-maven.yml
```diff
@@ -0,0 +1,42 @@
+name: Release Java Client to Maven
```
EnricoMi, line 1: Right, this is the release job. Do you have a test workflow? Or will this be separate as it did not exist before (and this is an integration effort)?

sudiptob2: I looked into other existing clients and found that tests for external clients are not run as part of the CI. I can add a step to run unit tests in the release pipeline if you suggest.

sudiptob2: @EnricoMi added a test workflow.

EnricoMi (approved):
LGTM!

sudiptob2:
> There are prerequisites to merging this PR. Environment needs to be created and secrets set. Admin needs to do this first

Updated with re-using the `maven-release` environment already created for the Scala client.

PR #4280, opened 2025-03-12
Remove GogoProto From Scheduler Objects

d80tb7:
Remove Gogo Proto from schedulerobjects.  These were the last use of gogo proto annotations in the Armada proto files and thus should allow us to completely remove gogo proto as a dependency in a subsequent PR.

The main effect here is that `schedulerobjects.ResourceList` will now unmarshall to `*schedulerobjects.ResourceList` and the quantity map it wraps will now be of type `map[string]*resource.Quantity` rather than `map[string]resource.Quantity`.


robertdavidsmith (approved):
LGTM 
Removing allocatableByPriorityAndResource may cause a nasty blip of you upgrade from before https://github.com/armadaproject/armada/pull/4127/ to after this PR in one go. Maybe warn on releases page?

Sovietaced:
This is great news. gogo proto has been annoying to use

PR #4286, opened 2025-03-13
Migrate Scala client from SBT to MVN

EnricoMi:
#### What type of PR is this?
This changes the build system used to build the Armada Scala client.

This migrates the Armada Scala client from sbt build system to mvn.

SBT build is Scala specific, MVN is generally known in Java land and can easily compile scala. So it is more widely known among JVM developers.

Further, Sonatype Central Portal added support to upload snapshots only recently. Our `io.armadaproject` namespace is registered with the Central Portal. The Maven publish plugin supports snapshots for the Central Portal, while the SBT publish plugin is still developing this feature.

This is a one-to-one replacement for the sbt build setup.

This now copies the protobuf files from the root of the repository into the client project when running `mvn`, before this had to be done via `./scripts/build-scala-client.sh`.

This migrates from MUnit to scalatest, which is more common in Scala land.


EnricoMi:
Looks like the mvn project is still not self-contained. You first need to bootstrap the proto files using Go via Mage files.

It is unusual that you need to install Go to compile a mvn project.

What about adding the proto files to the Scala client sources? They are 624k?

client/scala/armada-scala-client/src/main/scala/io/armadaproject/armada/ArmadaClient.scala
```diff
@@ -2,18 +2,30 @@ package io.armadaproject.armada
```
EnricoMi, line 2: This is only applying scala code style, no functional changes.

client/scala/armada-scala-client/src/test/scala/io/armadaproject/armada/ArmadaClientSuite.scala
```diff
@@ -0,0 +1,436 @@
+package io.armadaproject.armada
```
EnricoMi, line 1: - moved here from client/scala/armada-scala-client/src/test/scala/ArmadaClientSuite.scala
- applying scala code style
- only functional change is noted below

scripts/build-scala-client.sh
```diff
@@ -5,53 +5,21 @@
...
-done
-
-cd ..
-for pfile in \
-    pkg/api/event.proto pkg/api/submit.proto pkg/api/health.proto pkg/api/job.proto pkg/api/binoculars/binoculars.proto
-do
-  dir=$(dirname $pfile)
-  mkdir -p $ROOT/$SDIR/proto/$dir
-  cp $pfile $ROOT/$SDIR/proto/$dir/
-done
```
EnricoMi, line 37: All moved into pom.xml:
https://github.com/armadaproject/armada/blob/2bfdda7daa4e889a9485682774f566dcdad2c507/client/scala/armada-scala-client/pom.xml#L155-L184

scripts/build-scala-client.sh
```diff
@@ -5,53 +5,21 @@
...
 cd $ROOT/$SDIR
-sbt clean && \
-sbt -Dsbt.io.implicit.relative.glob.conversion=allow compile && \
-sbt -Dsbt.io.implicit.relative.glob.conversion=allow package
+mvn clean package
 
 if [ $? -eq 0 ]; then
-  jarfile=$(find . -type f -name 'armada-scala-client*.jar')
-  if [[ $jarfile ]]; then
+  jarfile=$(find . -type f -name 'armada-scala-client*.jar' | sort | head -n1)
```
EnricoMi, line 16: There are multiple jars (package, sources, javadocs), the first is the actual package.

dave-gantenbein (approved):

richscott:
> Looks like the mvn project is still not self-contained. You first need to bootstrap the proto files using Go via Mage files.
> 
> It is unusual that you need to install Go to compile a mvn project.
> 
> What about adding the proto files to the Scala client sources? They are 624k?

We have the static Armada proto declarations in `pkg/api/...`, but also many of the files in `/proto` are regenerated from fetching & massaging of the third-party packages enumerated in `magefiles/proto.go` - my understanding is they want to be sure to always get the current versions of those when doing a full build.

PR #4288, opened 2025-03-14
Add Scala client tests to CI

EnricoMi:
#### What type of PR is this?
Improves CI.

This adds Scala client tests to the CI, similar to Python client test.


.github/workflows/python-client.yml
```diff
@@ -28,22 +28,25 @@ on:
       - 'scripts/build-python-client.sh'
       - 'makefile'
       - '.github/workflows/python-tests/*'
+      - './magefiles/tests.go'
```
EnricoMi, line 31: Unrelated to the Scala client tests but I saw these discrepancies between `push.paths` and `pull_request.paths`.

.github/workflows/python-client.yml
```diff
@@ -28,22 +28,25 @@ on:
...
+      - './magefiles/python.go'
 
 jobs:
   python-client-tox:
     runs-on: ubuntu-22.04
     strategy:
+      fail-fast: false
```
EnricoMi, line 38: Having all matrix combinations run to completion is always useful to avoid incrementally uncovering issues in the CI.

.github/workflows/python-client.yml
```diff
@@ -52,10 +55,10 @@ jobs:
           cache-prefix: python-client-tox
       - uses: ./.github/workflows/python-tests
         with:
-          python-version: ${{ matrix.python }}
+          python-version: ${{ matrix.python-version }}
```
EnricoMi, line 58: Having different variable names for the same thing `with.python-version` vs. `matrix.python` is usually a good source of bugs.

richscott (approved):

Issue #4318, opened 2025-03-28
Scheduler Slow Memory Leak


Sovietaced:
I've observed that the Armada Scheduler has a slow memory leak. I just enabled pprof so I should have some results in the next few days but wanted to open this for tracking. 

![Image](https://github.com/user-attachments/assets/605e4d59-0073-418f-9eea-cba38c2875fb)

Sovietaced:
Turns out this memory leak is caused by a leak in the regular calls to fetch the latest queues.

dejanzele:
Hey,

Thanks for raising this PR!

Somebody from the team will check this out and review.

PR #4319, opened 2025-03-28
Fix queue cache memory leak

Sovietaced:
<!--  Thanks for sending a pull request!  Here are some tips for you:

Fixes https://github.com/armadaproject/armada/issues/4318


internal/scheduler/queue/queue_cache.go
```diff
@@ -72,20 +73,44 @@ func (c *ApiQueueCache) fetchQueues(ctx *armadacontext.Context) error {
...
+// This should be called after our end of stream message has been seen (StreamingQueueMessage_End)
+// We call recv one more time and expect an EOF back, indicating the stream is properly closed
+func closeStream(stream api.Submit_GetQueuesClient) error {
+	res, err := stream.Recv()
+	if err == nil {
+		switch typed := res.GetEvent().(type) {
+		default:
+			return fmt.Errorf("failed closing stream - unexpectedly received event of type %T", typed)
+		}
+	} else if err == io.EOF {
```
dejanzele, line 111: I'd always suggest `errors.Is`, even in straigtforward cases.

internal/scheduler/queue/queue_cache.go
```diff
@@ -72,20 +73,44 @@ func (c *ApiQueueCache) fetchQueues(ctx *armadacontext.Context) error {
...
+}
+
+// This should be called after our end of stream message has been seen (StreamingQueueMessage_End)
+// We call recv one more time and expect an EOF back, indicating the stream is properly closed
+func closeStream(stream api.Submit_GetQueuesClient) error {
+	res, err := stream.Recv()
+	if err == nil {
+		switch typed := res.GetEvent().(type) {
+		default:
+			return fmt.Errorf("failed closing stream - unexpectedly received event of type %T", typed)
```
dejanzele, line 109: Isn't it the same if you change `typed` to `res.GetEvent()` and remove the whole `switch` statement?

internal/scheduler/queue/queue_cache.go
```diff
@@ -72,20 +73,44 @@ func (c *ApiQueueCache) fetchQueues(ctx *armadacontext.Context) error {
...
+	if closeErr := closeStream(stream); closeErr != nil {
+		return fmt.Errorf("closing get queues stream: %w", closeErr)
+	}
+
+	return nil
+}
+
+// This should be called after our end of stream message has been seen (StreamingQueueMessage_End)
+// We call recv one more time and expect an EOF back, indicating the stream is properly closed
+func closeStream(stream api.Submit_GetQueuesClient) error {
```
dejanzele, line 104: where is the stream actually being closed?

internal/scheduler/queue/queue_cache.go
Sovietaced: This is the canonical way to do it for gRPC.

internal/scheduler/queue/queue_cache.go
Sovietaced: This code was actually copied from another part of Armada that has solved this issue already: https://github.com/armadaproject/armada/blob/08f6f8a690d10a7370ee63944afb65529656e043/internal/executor/service/lease_requester.go#L104-L118

internal/scheduler/queue/queue_cache.go
Sovietaced: The stream is closed by the server side but the client side doesn't consider it closed until it reads the `io.EOF`. Handling `io.EOF` is well documented and the gRPC folks discuss it here: https://groups.google.com/g/grpc-io/c/XcN4hA9HonI?pli=1

The code is a bit confusing because Armada added in an additional marker to close the stream at the application layer it looks like the stream is closed but it is not until you handle `io.EOF`.

internal/scheduler/queue/queue_cache.go
Sovietaced: I always prefer to keep convention but I'm happy to change both places.

robertdavidsmith (approved):
Happy to approve, this is clearly a step forward from when it was leaking memory.

robertdavidsmith (approved):

robertdavidsmith (approved):

masipauskas (approved):

PR #4321, opened 2025-03-30
Add index for initial job loading

Sovietaced:
Previously I made a change such that the scheduler would load non-terminal jobs into memory instead of all jobs. This fixed memory usage issues but now we've discovered that it can take up to 15 mins for the scheduler to startup because it has to do a linear scan through all the rows to find those that are non-terminal. 

This pull request adds an index that optimizes queries for non-terminal jobs. This query path is primarily used during initialization and blocks the scheduler from running until all non-terminal jobs are loaded into the database.

Before

```
armada=# EXPLAIN ANALYZE SELECT job_id, serial FROM jobs WHERE serial > 0 AND cancelled = 'false' AND succeeded = 'false' and failed = 'false' ORDER BY serial LIMIT 5;
                                                                QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.42..14.07 rows=5 width=35) (actual time=578.395..618.550 rows=5 loops=1)
   ->  Index Scan using idx_jobs_serial on jobs  (cost=0.42..278677.81 rows=102124 width=35) (actual time=578.393..618.547 rows=5 loops=1)
         Index Cond: (serial > 0)
         Filter: ((NOT cancelled) AND (NOT succeeded) AND (NOT failed))
         Rows Removed by Filter: 392592
 Planning Time: 0.246 ms
 Execution Time: 618.574 ms
(7 rows)
```
After
```
armada=# EXPLAIN ANALYZE SELECT job_id, serial FROM jobs WHERE serial > 0 AND cancelled = 'false' AND succeeded = 'false' and failed = 'false' ORDER BY serial LIMIT 5;
                                                                QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.42..12.39 rows=5 width=35) (actual time=30.423..32.503 rows=5 loops=1)
   ->  Index Scan using idx_jobs_terminal on jobs  (cost=0.42..244516.59 rows=102124 width=35) (actual time=30.419..32.496 rows=5 loops=1)
         Index Cond: ((serial > 0) AND (cancelled = false) AND (succeeded = false) AND (failed = false))
 Planning Time: 0.711 ms
 Execution Time: 32.595 ms
(5 rows)
```


robertdavidsmith (approved):
I'm happy to add the index, it's a reasonable change

However I'm surprised this is a problem. Is the scheduler pruner running correctly? If all is well, you should see two working CronJobs like this, and pruner pods regularly being created and going to `Completed`.



```
kubectl -n armada get cronjob 
NAME                         SCHEDULE       SUSPEND   ACTIVE   LAST SCHEDULE   AGE
armada-lookout-db-pruner     0 * * * *      False     0        15m             113d
armada-scheduler-db-pruner   */15 * * * *   False     1        25s             440d
```

masipauskas (approved):

PR #4328, opened 2025-04-02
Fixing relese java client workflow

pavlovic-ivan:
This is a bug fixing PR because publishing java client is not working for some reason. **This also might be one-in-several PRs because i can only test this by merging to master here**. Changes:

- Update setup-java action version
- Delete settings.xml from project, and remove it from workflow. Action setup-java will create a settings.xml file and add it to the HOME/.m2 dir.
- Update pom.xml. Align versions of plugins with scala client


sudiptob2 (approved):

.github/workflows/java-client-release-to-maven.yml
```diff
@@ -27,7 +27,6 @@ jobs:
 
     - name: Build
       run: |
-        cp client/java/settings.xml /home/runner/.m2
         cd client/java
         mvn clean install
       env:
```
EnricoMi, line 32: This env is not needed for `install` target, the artifacts are signed during `deploy`.

EnricoMi (approved):
LGTM! The downgrade of the plug-in is probably not needed.

EnricoMi:
Comparing th Java client `pom.xml` with the Scala client one, there is no `<distributionManagement>` in the latter.

.github/workflows/java-client-release-to-maven.yml
pavlovic-ivan: Sorry i don't get what you mean?

pavlovic-ivan:
> LGTM! The downgrade of the plug-in is probably not needed.

i only made it equal to the version used in Scala client, as suggested in Slack. @EnricoMi shall i revert it to the version previously used?

Issue #4330, opened 2025-04-03
Scala client: submitHealth() in ArmadaClient wrapper should return a Future


richscott:
The `SubmitGrpc.scala` generated from the original `submit.proto` defines the `health()` method as returning a `Future[HealthCheckResponse]` - make the ArmadaClient convenience wrapper's `submitHealth()` also return the
same. Without this change, an `UnsupportedOperationException` is raised in some tests.

PR #4331, opened 2025-04-03
Scala client: wrap in submitHealth() return value in a Future

richscott:
The `SubmitGrpc.scala` generated from the original `submit.proto` defines the `health()` method as returning a `Future[HealthCheckResponse]` - make the ArmadaClient convenience wrapper's `submitHealth()` also return the same. Without this change, an UnsupportedOperationException is raised in some tests.

Fixes https://github.com/armadaproject/armada/issues/4330


GeorgeJahad (approved):
lgtm

PR #4337, opened 2025-04-15
Publish Metrics To Pulsar

d80tb7:
Publish Cycle Metrics To Pulsar.  This is fairly experimental and therefore has been somewhat shoehorned into the existing metrics collection.

This functionality is feature-flagged and disabled by default.


internal/scheduler/configuration/configuration.go
```diff
@@ -58,6 +58,8 @@ type Configuration struct {
 	QueueRefreshPeriod time.Duration `validate:"required"`
 	// Allows queue priority overrides to be fetched from an external source.
 	PriorityOverride PriorityOverrideConfig
+	// Whether To Metrics To Pulsar.  This is currently experimental
```
robertdavidsmith, line 61: typos

Issue #4340, opened 2025-04-18
Native support for preemption retries


jparraga-stackav:
**Is your feature request related to a problem? Please describe.**

Armada has support for urgency and fair share based preemption which is a useful feature for prioritizing important workloads and enforcing fair share of resources. One of the pain points with enabling preemption is that when workloads are preempted they are not requeued or resubmitted. Currently it is the responsibility of the user to recognize that their workloads are preempted and then resubmit them. 

**Describe the solution you'd like**

Ideally Armada would have native support for retrying preempted workloads. Retries could be configured as global defaults or opted in at an individual job/gang level with annotations. When a job is preempted it returns to a queued state until it is able to be scheduled again. Each time a job is preempted it is captured with a preempted job run. 

**Describe alternatives you've considered**

We have prototyped an external entity which watches for preempted workloads and automatically resubmits them. This workflow feels a bit clunky and computationally wasteful as this entity must constantly either poll for preempted workloads or watch job sets. The APIs to do all of these operations are mixed between Armada Lookout and Armada Server. This can be particularly complicated when a single jobset has one or many different jobs/gangs that should be retried. 

Additionally, any system that submits to Armada has to be aware that their workloads may be resubmitted which creates additional work for integrations built on top of Armada. 

**Additional context**

We have prototyped support for this and I will include a pull request shortly.

PR #4342, opened 2025-04-19
Add support for native preemption retries

jparraga-stackav:
#### What type of PR is this?
New feature

This pull request add support for native Armada Preemption Retry Handling. Retry handling can be configured at the platform level as a default in the scheduling config as well as with two annotations:

armadaproject.io/preemptionRetryCountMax (defaults to 0 for now, ie. disabled if not configured)
armadaproject.io/preemptionRetryEnabled

The scheduler has been modified to only submit terminal job events when a job is preempted and is not retryable. If the job is preempted and retryable the job is requeued. 

The executor has two different code paths that send events when pods enter a failed state:

1. The job state reporter handles pod events and sends job failed events when jobs fail.
2. The pod issue handler handles pod deletion events and sends job failed events when a pod is unexpectedly deleted.

Both of these code paths have been updated not to send job failed events when the pod has been preempted as this would fail the overall job. 

Note: This introduces a subtle behavior regression where pods that are preempted out from underneath Armada (ie. system pods preempting armada jobs) will no longer fail the job. I have some ideas for how to fix that but I think it is a pretty exceptional case. 

Fixes: https://github.com/armadaproject/armada/issues/4340

### DIFF
--- a/internal/common/preemption/utils.go
+++ b/internal/common/preemption/utils.go
@@ -1,0 +1,39 @@
+package preemption
+
+import (
+	"github.com/armadaproject/armada/internal/server/configuration"
+	"strconv"
+)
+
+// AreRetriesEnabled determines whether preemption retries are enabled at the job level. Also returns whether the
+// annotation was set.
+func AreRetriesEnabled(annotations map[string]string) (bool, bool) {
+	preemptionRetryEnabledStr, exists := annotations[configuration.PreemptionRetryEnabledAnnotation]
+	if !exists {
+		return false, false
+	}
+
+	preemptionRetryEnabled, err := strconv.ParseBool(preemptionRetryEnabledStr)
+	if err != nil {
+		return false, true
+	}
+	return preemptionRetryEnabled, true
+}
+
+// GetMaxRetryCount gets the max preemption retry count at a job level. Also returns whether the annotation was set.
+func GetMaxRetryCount(annotations map[string]string) (uint, bool) {
+	var preemptionRetryCountMax uint = 0
+	preemptionRetryCountMaxStr, exists := annotations[configuration.PreemptionRetryCountMaxAnnotation]
+
+	if !exists {
+		return 0, false
+	}
+	maybePreemptionRetryCountMax, err := strconv.Atoi(preemptionRetryCountMaxStr)
+	if err != nil {
+		return 0, true
+	} else {
+		preemptionRetryCountMax = uint(maybePreemptionRetryCountMax)
+	}
+
+	return preemptionRetryCountMax, true
+}
--- a/internal/executor/service/job_state_reporter.go
+++ b/internal/executor/service/job_state_reporter.go
@@ -93,6 +93,11 @@ func (stateReporter *JobStateReporter) reportCurrentStatus(pod *v1.Pod) {
 	}
 
 	if pod.Status.Phase == v1.PodFailed {
+
+		if util.IsPodPreempted(pod) {
+			return
+		}
+
 		hasIssue := stateReporter.podIssueHandler.HasIssue(util.ExtractJobRunId(pod))
 		if hasIssue {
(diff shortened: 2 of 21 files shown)


internal/common/preemption/utils.go
```diff
@@ -0,0 +1,39 @@
+package preemption
+
+import (
+	"github.com/armadaproject/armada/internal/server/configuration"
+	"strconv"
+)
+
+// AreRetriesEnabled determines whether preemption retries are enabled at the job level. Also returns whether the
+// annotation was set.
+func AreRetriesEnabled(annotations map[string]string) (bool, bool) {
```
dejanzele, line 10: I'd name the return params, so it is easier to see what is what

d80tb7:
Hi,

thanks for this- it looks great.  One thing I think we should consider/discuss is whether we could add the preemption retryfields as first-class proto fields rather than relying on annotations.  Reasoning here is that anottions are very easy to add at first as they require no interface changes, but can soone get quite hard to work with as evwerything is just a map[string]string.

Personally I'd be in favour of adding these fields to `schedulerobjects.JobSchedulingInfo` right now with the view of adding them to `api.Job` once the feature is stable.




One comment I

robertdavidsmith (approved):

jparraga-stackav:
> Personally I'd be in favour of adding these fields to `schedulerobjects.JobSchedulingInfo` right now with the view of adding them to `api.Job` once the feature is stable.

I can do that for sure.

jparraga-stackav:
> Personally I'd be in favour of adding these fields to `schedulerobjects.JobSchedulingInfo` right now with the view of adding them to `api.Job` once the feature is stable.

I looked into this but it seems a bit difficult to keep it is an annotation and not a first class citizen. 

I'm not able to create the scheduling info from an `armadaevents.SubmitJob` since the annotations aren't there. I think there would also need to be some more thinking about how the global preemption retry config integrates into this. Might need to move that into the scheduler ingester to more elegantly handle that unless we want to reference the global preemption retry config all the time.

Cinojose:
@jparraga-stackav Thanks for the work on this. 
I was wondering if we could also handle imminent node shutdown scenarios. In addition to the current retry logic for pod evictions, it might be useful to check for pods [terminated](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#graceful-node-shutdown) with the reason:

`Pod was terminated in response to imminent node shutdown.`

We could incorporate a simple check on pod.Status.Message to capture these cases. This would help cover both graceful shutdowns and node preemptions for various reasons.

jparraga-stackav:
> @jparraga-stackav Thanks for the work on this. I was wondering if we could also handle imminent node shutdown scenarios. In addition to the current retry logic for pod evictions, it might be useful to check for pods [terminated](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#graceful-node-shutdown) with the reason:
> 
> `Pod was terminated in response to imminent node shutdown.`
> 
> We could incorporate a simple check on `pod.Status.Message` to capture these cases. This would help cover both graceful shutdowns and node preemptions for various reasons.

I will likely look into this as a follow up improvement.

internal/scheduler/database/migrations/022_add_job_run_index.sql
```diff
@@ -1 +1 @@
-ALTER TABLE runs ADD COLUMN IF NOT EXISTS run_index bigint NOT NULL DEFAULT 0;
+ALTER TABLE runs ADD COLUMN IF NOT EXISTS run_index bigint;
```
jparraga-stackav, line 1: During testing we found that the old scheduler ingester was trying to write rows with a null value which blocked ingestion from occurring. In order to make this smoother we've made this null-able and then handle null-able values in the app layer.

dave-gantenbein:
@dejanzele to review this...

PR #4351, opened 2025-04-30
Support silent OIDC signin in Lookout UI

mauriceyap:
Adds support and fixes the behaviour for silently refreshing the OIDC token in the Lookout UI. The `oidc-client-ts` library we use uses an iframe to refetch the auth token from the configured OIDC provider silently in the background one minute before the stored token expires.

Also remove unused code.

### DIFF
--- a/internal/lookoutui/src/App.tsx
+++ b/internal/lookoutui/src/App.tsx
@@ -3,7 +3,6 @@ import { useEffect } from "react"
 import { CssBaseline, styled, ThemeProvider } from "@mui/material"
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query"
 import { SnackbarProvider } from "notistack"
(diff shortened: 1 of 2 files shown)


d80tb7 (approved):

internal/lookoutui/src/App.tsx
```diff
@@ -48,19 +46,6 @@ type AppProps = {
...
-    client_id: config.clientId,
-    redirect_uri: `${window.location.origin}${OIDC_REDIRECT_PATHNAME}`,
-    scope: config.scope,
-    userStore: new WebStorageStateStore({ store: window.localStorage }),
-    loadUserInfo: true,
-  }
-
-  return new UserManager(userManagerSettings)
-}
-
```
mauriceyap, line 63: this is dead code

PR #4354, opened 2025-05-21
Remove priority class name override

jparraga-stackav:
This pull request removes the default configuration which adds a priority class name override. 

This configuration is puzzling to me since there is already a default priority class name set. I don't see why the intended priority class should be overridden when sent down to the executor. This can also conflict with the preemptible label that is unconditionally set prior to the priority class name override. 

I mainly find this problematic because we are unable to unset this when managing Armada using the Armada Operator.

### DIFF
--- a/config/scheduler/config.yaml
+++ b/config/scheduler/config.yaml
@@ -93,7 +93,6 @@ scheduling:
       priority: 1000
       preemptible: true
   defaultPriorityClassName: "armada-default"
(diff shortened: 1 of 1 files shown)


dejanzele:
@jparraga-stackav as Armada handles preemption, I described the flow here https://github.com/armadaproject/armada/pull/4600

Is there a reason why you would attach different priority classes to pods?
This does not impact Armada scheduling and preemption anyhow.

PR #4362, opened 2025-05-29
Fix potential row resource leak

jparraga-stackav:
This pull request fixes a potential row resource leak that can occur since `rows` can be reassigned during the loop but only closed once by `defer rows.Close()`

### DIFF
--- a/internal/scheduler/database/job_repository.go
+++ b/internal/scheduler/database/job_repository.go
@@ -189,6 +189,15 @@ func (r *PostgresJobRepository) FetchJobRunErrors(ctx *armadacontext.Context, ru
 		AccessMode:     pgx.ReadWrite,
 		DeferrableMode: pgx.Deferrable,
 	}, func(tx pgx.Tx) error {
+
+		var rows pgx.Rows
+		var queryErr error
+		defer func() {
+			if rows != nil {
+				rows.Close()
+			}
+		}()
+
 		for _, chunk := range chunks {
 			tmpTable, err := insertRunIdsToTmpTable(ctx, tx, chunk)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


d80tb7 (dismissed):

JamesMurkin (approved):

dejanzele:
@jparraga-stackav can you please run `golangci-lint run --fix` so it fixes the lint issue?

internal/scheduler/database/job_repository.go
```diff
@@ -203,11 +211,10 @@ func (r *PostgresJobRepository) FetchJobRunErrors(ctx *armadacontext.Context, ru
...
 		JOIN job_run_errors ON job_run_errors.run_id = tmp.run_id`
 
-			rows, err := tx.Query(ctx, fmt.Sprintf(query, tmpTable))
-			if err != nil {
-				return err
+			rows, queryErr = tx.Query(ctx, fmt.Sprintf(query, tmpTable))
+			if queryErr != nil {
+				return queryErr
 			}
-			defer rows.Close()
```
nikola-jokic, line 210: @dejanzele I don't think this should be an issue, and I don't think it leaks the memory.

The defer once evaluated pushes the arguments to the defer stack at the time defer statement is evaluated. In this case, the address of the object returned from the `tx.Query` will be pushed. Even if `rows` is re-assigned, the copy of the memory address is pushed to the stack, so once the defer gets executed, it would call close on the correct object.

Since this is already correct from the behavioral point of view, the only question is should we close the row right after we are done with it. If so, I'd argue that the logic should be extracted to a function that would do everything inside the function body. Otherwise, I don't think the fix is necessary.

PR #4363, opened 2025-06-06
added auth to scala client

GeorgeJahad:
Port of code from java client:
https://github.com/armadaproject/java-client/blob/main/src/main/java/io/armadaproject/ArmadaClient.java#L36-L51

Fixes https://github.com/G-Research/spark/issues/86

### DIFF
--- a/client/scala/armada-scala-client/src/main/scala/io/armadaproject/armada/ArmadaClient.scala
+++ b/client/scala/armada-scala-client/src/main/scala/io/armadaproject/armada/ArmadaClient.scala
@@ -16,7 +16,9 @@ import api.submit.{
 }
 import api.health.HealthCheckResponse
 import com.google.protobuf.empty.Empty
-import io.grpc.{ManagedChannel, ManagedChannelBuilder}
+import io.grpc.netty.NettyChannelBuilder
+import io.grpc.stub.MetadataUtils
+import io.grpc.{ManagedChannel, ManagedChannelBuilder, Metadata}
 
 import scala.concurrent.Future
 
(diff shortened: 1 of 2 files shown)


richscott (dismissed):
LGTM!

EnricoMi:
Looks like this does two things, use SSL and authentication, you cannot use SSL without authentication.

Maye a signature like this makes more sense (where `useSsl=true` is implied if `bearerToken.nonEmpty`:
```scala
def apply(host: String, port: Int, useSsl: Boolean, bearerToken: Option[String]): ArmadaClient
```

GeorgeJahad:
@EnricoMi I think I've done what you wanted.  Please take a look.

richscott (approved):

dave-gantenbein (approved):

client/scala/armada-scala-client/src/main/scala/io/armadaproject/armada/ArmadaClient.scala
```diff
@@ -75,14 +77,45 @@ class ArmadaClient(channel: ManagedChannel) {
...
+          "Bearer " + bearerToken.get
+        )
+        val channel =
+          NettyChannelBuilder
+            .forAddress(host, port)
+            .useTransportSecurity()
+            .intercept(MetadataUtils.newAttachHeadersInterceptor(metadata))
+            .build
+        ArmadaClient(channel)
+    }
```
EnricoMi, line 119: I'd factor out the `ArmadaClient(channel)`, this removes a bit of redundant code:
```scala
    val channel = (useSsl, bearerToken) match {
      case (false, None) =>
        NettyChannelBuilder.forAddress(host, port).usePlaintext().build()
      case (true, None) =>
        NettyChannelBuilder
          .forAddress(host, port)
          .useTransportSecurity()
          .build()
      case (_, Some(token)) =>
        val metadata = new Metadata()
        metadata.put(
          Metadata.Key.of("Authorization", Metadata.ASCII_STRING_MARSHALLER),
          "Bearer " + bearerToken.get
        )
        NettyChannelBuilder
          .forAddress(host, port)
          .useTransportSecurity()
          .intercept(MetadataUtils.newAttachHeadersInterceptor(metadata))
          .build
    }
    ArmadaClient(channel)
```

GeorgeJahad:
@EnricoMi I addressed your comment here: https://github.com/armadaproject/armada/pull/4368

PR #4368, opened 2025-06-12
More auth to scala

GeorgeJahad:
Fixes #
https://github.com/G-Research/spark/issues/92


richscott (approved):
Nice tightening of the code - I like the single `channel` declaration.

GeorgeJahad:
> Nice tightening of the code - I like the single channel declaration.

That was all from @EnricoMi  Thanks @EnricoMi !

client/scala/armada-scala-client/src/main/scala/io/armadaproject/armada/ArmadaClient.scala
```diff
@@ -91,31 +91,28 @@ object ArmadaClient {
       useSsl: Boolean,
       bearerToken: Option[String]
   ): ArmadaClient = {
-    (useSsl, bearerToken) match {
+    val secureMode =
+      host.startsWith("https://") || host.startsWith("grpcs://") || useSsl
```
dejanzele, line 95: Actually, given that you have the `useSsl` explicitly, better remove the `host.startsWith("https://") || host.startsWith("grpcs://")` check

EnricoMi: I did not expect `host` to contain a protocol / scheme, the naming is not the best here.

Given there is a scheme in `host`, and presuming `host` always has a scheme, it is better to remove the `useSsl` and derive secure mode from the scheme. Otherwise all kind of wild combinations would be possible like `http` and `useSsl = true`.

GeorgeJahad: There is no scheme for the armada url's at gr, just hostname/port.  (Our armada-spark code requires they prepend the psuedoscheme, "armada://", but we strip that off before calling the armada client.)

dave-gantenbein: This does allow users to configure it incorrectly, ie useSsl = false with an https:// host. Users are very silly and will do exactly this.

dave-gantenbein: Juice not worth the squeeze here, let's move on...

dave-gantenbein (approved):

PR #4380, opened 2025-06-23
Create Visual Studio Code debug configurations

mauriceyap:
This creates reliable debug configurations to run Armada services using VS Code's built-in debugging tools. They largely mirror the existing run configurations for JetBrains products.

Also create a reliable script which wraps Yarn to support developers using yarn registry mirrors without making undesired changes to the Yarn lock file.


samclark (approved):
LGTM. My only concern is the yarn wrapper doesn't look very concurrent safe so will run into issues if it's ever called multiple times in parallel.

PR #4388, opened 2025-06-27
Fix billing for away jobs scheduled and preempted in the same round

JamesMurkin:
Currently we snapshot the resource allocated to a queue when the spot price is calculated

However at this point there can be jobs scheduled that will be evicted before the end of the scheduling round

Now we snapshot the allocated + mark scheduled jctx's as "billable"
 - If these scheduled jctx's subsequently get evicted/preempted, we'll remove that from the billable resource

### DIFF
--- a/internal/scheduler/nodedb/nodedb_test.go
+++ b/internal/scheduler/nodedb/nodedb_test.go
@@ -290,11 +290,14 @@ func TestEviction(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, 0, len(node.EvictedJobRunIds))
 	assert.Equal(t, map[int32]internaltypes.ResourceList{
-		-1: testfixtures.CpuMem("30", "248Gi"),
-		0:  testfixtures.CpuMem("30", "248Gi"),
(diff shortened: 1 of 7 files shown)


internal/scheduler/nodedb/nodedb_test.go
```diff
@@ -290,11 +290,14 @@ func TestEviction(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, 0, len(node.EvictedJobRunIds))
 	assert.Equal(t, map[int32]internaltypes.ResourceList{
-		-1: testfixtures.CpuMem("30", "248Gi"),
-		0:  testfixtures.CpuMem("30", "248Gi"),
-		1:  testfixtures.CpuMem("31", "252Gi"),
-		2:  testfixtures.CpuMem("31", "252Gi"),
-		3:  testfixtures.CpuMem("31", "252Gi"),
+		-1:    testfixtures.CpuMem("30", "248Gi"),
```
JamesMurkin, line 293: These needed updating as I updated the priorities considered in testfixtures

d80tb7 (approved):

PR #4399, opened 2025-07-08
Update docs prs

suprjinx:
#### What type of PR is this?

Restores doc change PRs, but correct link issues.

Various nav and internal links were broken when these changes first landed. They were reverted temporarily, and are restored here.


suprjinx:
deleting this in favor of new branch

PR #4408, opened 2025-07-14
Simplify gang accounting in scheduler 

JamesMurkin:

PR #4421, opened 2025-07-25
Minor tidy of jobdb Job/GangInfo

JamesMurkin:
- Add a IsIsGang func as this is a common question in the code, so this helper makes it easier to answer
- Remove the need for the isGang field in GangInfo, now determined by cardinality supplied


JamesMurkin:
Largely PR feedback from - https://github.com/armadaproject/armada/pull/4408

PR #4431, opened 2025-08-06
Lookout UI: display queue names which contain the user's username first

mauriceyap:
When filtering queues on the jobs page, or when selecting a queue on the job sets page, display the queues whose names contain the logged in user's username first. This typically indicates that the queue belongs to the logged-in user, and this change would be a quality-of-life improvement.


samclark (approved):
classy

PR #4453, opened 2025-09-02
Add an optional monitoring of topic processing delay to ingester pipeline

JamesMurkin:
This will:
 - Expose metrics to show how far behind each partition of the topic being consume is in ms
 - We utilise peekMessages on the pulsar REST/admin api to get this information
 - It is optional and off by default

The purpose of this is so you can see how far behind the ingester is in time (i.e 3 minutes)
 - pulsar only exposes backlog in number of messages, so its hard to tell how much impact a delay is having (i.e 1000 messages could be 1 second behind or 10 minutes)

Also exposes a metric that shows the publish time of messages being processed by the pipeline for each partition

**Note** I've had to "fork" some of pulsar-client-go due to a bug in the client. This is being fixed in https://github.com/apache/pulsar-client-go/pull/1419. Once that is in the upstream client we can remove this internal code copy/fork


internal/common/ingest/ingestion_pipeline.go
```diff
@@ -117,12 +119,19 @@ func (i *IngestionPipeline[T, U]) Run(ctx *armadacontext.Context) error {
...
 	if i.consumer == nil {
-		consumer, closePulsar, err := i.subscribe()
+		client, consumer, closePulsar, err := i.subscribe()
 		if err != nil {
 			return err
 		}
 		i.consumer = consumer
 		defer closePulsar()
+
+		if i.pulsarConfig.DelayMonitor != nil && i.pulsarConfig.DelayMonitor.Enabled {
```
d80tb7: can we make DelayMonitor  a struct rather than a pointer.  that way we just need to check `pulsarConfig.DelayMonitor.Enabled`

internal/common/ingest/ingestion_pipeline.go
```diff
@@ -2,11 +2,13 @@ package ingest
...
 import (
 	"context"
+	"errors"
+	"fmt"
 	"sync"
 	"time"
 
 	"github.com/apache/pulsar-client-go/pulsar"
-	"github.com/pkg/errors"
+	pkgerrors "github.com/pkg/errors"
```
d80tb7: do we actually need both "errors" and "github.com/pkg/errors"?  I'm guessing we could just have the latter.

internal/common/ingest/ingestion_pipeline.go
```diff
@@ -243,11 +253,38 @@ func (i *IngestionPipeline[T, U]) Run(ctx *armadacontext.Context) error {
...
 }
 
-func (i *IngestionPipeline[T, U]) subscribe() (pulsar.Consumer, func(), error) {
+func (i *IngestionPipeline[T, U]) startProcessingDelayMonitor(ctx *armadacontext.Context, pulsarClient pulsar.Client) error {
+	if i.pulsarConfig.RestURL == "" {
+		return fmt.Errorf("cannot enable topic delay monitoring as pulsar RestURL not configured")
+	}
+	pulsarAdminClient, err := pulsarutils.NewPulsarAdminClient(&i.pulsarConfig)
+	if err != nil {
+		return pkgerrors.WithMessage(err, "Error creating pulsar admin client")
```
d80tb7: think all the errors here need to have a lower case starting letter :(

internal/common/ingest/pulsarclient/client.go
```diff
@@ -0,0 +1,73 @@
...
+	authProvider, err := auth.GetAuthProvider(config)
+	if err != nil {
+		return nil, err
+	}
+	return NewPulsarClientWithAuthProvider(config, authProvider)
+}
+
+// NewPulsarClientWithAuthProvider create a client with auth provider.
+func NewPulsarClientWithAuthProvider(config *config.Config, authProvider auth.Provider) (*PulsarClient, error) {
+	if len(config.WebServiceURL) == 0 {
```
d80tb7, line 47: Do we need this?  I guess my argument would be:
- We already have default config files.
- I don't think we do this for anything else?

internal/common/ingest/topic_delay_monitor.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (t *TopicProcessingDelayMonitor) Run(ctx *armadacontext.Context) error {
+	if len(t.partitions) == 0 {
+		err := t.Initialise(ctx)
+		if err != nil {
+			return err
+		}
+	}
+
+	wg := &sync.WaitGroup{}
```
d80tb7, line 91: can't this just be:
```
var wg sync.WaitGroup

for _, topic := range t.partitions {
    wg.Add(1)                 
    go func(topic *pulsarutils.TopicName) {
        defer wg.Done()       
        t.monitorPartitionDelay(ctx, topic)
    }(topic)              
}
wg.Wait()
```

internal/common/ingest/pulsarclient/client.go
JamesMurkin: This code is copied from the pulsar-client-go. I only modified the file by deleting functions we didn't need but otherwise didn't modify it.

I can change the logic but I'd rather not have more modifications compared to upstream than we absolutely need to.

internal/common/ingest/topic_delay_monitor.go
JamesMurkin: Yes, updated to match suggested

internal/common/ingest/pulsarclient/client.go
d80tb7: ah ok yes- in the case it doesn't matter

PR #4478, opened 2025-09-25
fix: lint errors reported after upgrading to golangci-lint v2.5.0

dejanzele:
#### What type of PR is this?

Fix for failing CI after upgrading to golangci-lint v2.5.0

From golangci-lint v2.5.0 changelog:
* gofumpt: from 0.8.0 to 0.9.1 (new rule is to “clothe” naked returns for the sake of clarity)

Latest version of `gofumpt` reports naked returns (just doing `return` when a function has named parameters) as lint issues.

There doesn't seem to be an option for disabling this rule.

Options are:
1. Auto-fix code using `golangci-lint run --fix`
2. Revert to golangci-lint v2.4.0
3. Disable `gofumpt` linter

Code was auto-fixed by running `golangci-lint run --fix`

### DIFF
--- a/internal/common/ingest/testfixtures/event.go
+++ b/internal/common/ingest/testfixtures/event.go
@@ -91,12 +91,80 @@ var Submit = &armadaevents.EventSequence_Event{
 			ObjectMeta: &armadaevents.ObjectMeta{
 				Namespace: Namespace,
 				Name:      "test-job",
+				Annotations: map[string]string{
+					"foo":                            "bar",
+					configuration.FailFastAnnotation: "true",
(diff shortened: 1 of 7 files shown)


dejanzele:
@d80tb7 

The refactor to remove named returns broke a couple of spots. In `gang_scheduler.go`, the defer augments errors and updates scheduling state, and in `reconciliation.go` it ensures the final job/run values are returned. Without named returns, the defers only touch locals and not the actual return values, which caused the test failures.

For these functions, keeping named returns is the simplest fix.

JamesMurkin (approved):

PR #4484, opened 2025-09-29
Removing named return values (#454)

MustafaI:
Removing named return values entirely

### DIFF
--- a/internal/executor/configuration/podchecks/types.go
+++ b/internal/executor/configuration/podchecks/types.go
@@ -49,8 +49,9 @@ type ContainerStatusCheck struct {
 }
 
 type FailedChecks struct {
-	Events      []PodEventCheck
-	PodStatuses []PodStatusCheck
+	Events            []PodEventCheck
+	PodStatuses       []PodStatusCheck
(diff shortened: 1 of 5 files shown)


internal/scheduler/jobdb/reconciliation.go
```diff
@@ -144,7 +144,9 @@ func markJobsAsPreemptionRequested(txn *Txn, jobIds []string, jsts map[string]Jo
 // along with a summary of the state transitions applied to the job.
 //
 // TODO(albin): Pending, running, and preempted are not supported yet.
-func (jobDb *JobDb) reconcileJobDifferences(job *Job, jobRepoJob *database.Job, jobRepoRuns []*database.Run) (jst JobStateTransitions, err error) {
+func (jobDb *JobDb) reconcileJobDifferences(job *Job, jobRepoJob *database.Job, jobRepoRuns []*database.Run) (JobStateTransitions, error) {
+	var jst JobStateTransitions
+	var err error
```
dejanzele, line 149: defers are causing issues and making the CI fail, now with the removed named params

PR #4504, opened 2025-10-22
Fix gang urgency based preemption panic

JamesMurkin:
A recent change to how we manage gang state in the scheduler has caused us to break urgency based preemption of gangs
 - This results in a panic

This is because the code tries to find the nodeId of the gang members it is evicting, however it assumes these will be in the jobDb (which is not true if the gang was scheduled in the current round)

Now we correctly look up the node id in the scheduling context if the job was scheduled in the current round

### DIFF
--- a/internal/scheduler/scheduling/preempting_queue_scheduler.go
+++ b/internal/scheduler/scheduling/preempting_queue_scheduler.go
@@ -427,7 +427,7 @@ func getNodeIdFromSchedulingContext(sctx *schedulercontext.SchedulingContext, jo
 	}
 
 	if jctx.PodSchedulingContext == nil {
-		return "", errors.Errorf("pod scheduling context missing for successfully scheduled jctx %s - %s ", job.Id(), job.Queue())
+		return "", errors.Errorf("pod scheduling context missing for successfully scheduled job %s - %s ", job.Id(), job.Queue())
 	}
 
 	if jctx.PodSchedulingContext.NodeId == "" {


internal/scheduler/scheduling/preempting_queue_scheduler.go
```diff
@@ -398,13 +398,45 @@ func (sch *PreemptingQueueScheduler) collectIdsForGangEviction(evictorResult *Ev
 		// Collect all job ids part of that gang.
 		for _, gangJob := range activeGangJobs {
 			allGangJobIds[gangJob.Id()] = true
-			gangNodeIds[gangJob.LatestRun().NodeId()] = true
+			// If the latest run is nil, it means the job was scheduled in the current scheduling round
+			// So we need to find which node it was scheduled on in the scheduling context as it won't have been persisted in the jobdb yet
+			if gangJob.LatestRun() == nil {
+				nodeId, err := getNodeIdFromSchedulingContext(sch.schedulingContext, gangJob)
+				if err != nil {
+					return nil, nil, err
```
dejanzele, line 406: nit: should the error be wrapped here for extra context?

internal/scheduler/scheduling/preempting_queue_scheduler.go
```diff
@@ -398,13 +398,45 @@ func (sch *PreemptingQueueScheduler) collectIdsForGangEviction(evictorResult *Ev
 		// Collect all job ids part of that gang.
 		for _, gangJob := range activeGangJobs {
 			allGangJobIds[gangJob.Id()] = true
-			gangNodeIds[gangJob.LatestRun().NodeId()] = true
+			// If the latest run is nil, it means the job was scheduled in the current scheduling round
+			// So we need to find which node it was scheduled on in the scheduling context as it won't have been persisted in the jobdb yet
```
dejanzele, line 402: nit: scheduler db instead of jobdb?

internal/scheduler/scheduling/preempting_queue_scheduler.go
```diff
@@ -398,13 +398,45 @@ func (sch *PreemptingQueueScheduler) collectIdsForGangEviction(evictorResult *Ev
...
+		return "", errors.Errorf("unable to find queue scheduling context for job %s - %s ", job.Id(), job.Queue())
+	}
+
+	jctx, present := qctx.SuccessfulJobSchedulingContexts[job.Id()]
+	if !present {
+		return "", errors.Errorf("unable to find successful job scheduling context for job %s - %s ", job.Id(), job.Queue())
+	}
+
+	if jctx.PodSchedulingContext == nil {
+		return "", errors.Errorf("pod scheduling context missing for successfully scheduled jctx %s - %s ", job.Id(), job.Queue())
```
dejanzele: nit: `for successfully scheduled job` (or job scheduling context) instead of `for successfully scheduled jctx`?

internal/scheduler/scheduling/preempting_queue_scheduler.go
JamesMurkin: Well both, but its the job db that matters / is being queried

internal/scheduler/scheduling/preempting_queue_scheduler.go
JamesMurkin: Done

internal/scheduler/scheduling/preempting_queue_scheduler.go
JamesMurkin: They already include the stack track, so I think that is enough context unless you had another idea

dejanzele (approved):

PR #4506, opened 2025-10-23
Remove outdated gang eviction assertions

JamesMurkin:
This used to reconcile the scheduling context with an intermediate state, however that intermediate state has since been removed

It is no longer obvious what this should reconcile as really the only way to work out what should have been evicted would be to work something that does the same logic as the original code its validating

We could update this code to do some new form of reconciliation but I think instead we should remove it as:
 - We now have better test coverage of this and test coverage makes more sense than assertions
 - The assertions code typically rots and generally is something we're deprecating


internal/scheduler/scheduling/preempting_queue_scheduler.go
```diff
@@ -483,14 +483,7 @@ func (sch *PreemptingQueueScheduler) evictionAssertions(evictorResult *EvictorRe
...
-	evictedJctxsByGangId := make(map[string][]*schedulercontext.JobSchedulingContext)
 	for jobId, jctx := range evictorResult.EvictedJctxsByJobId {
-		if jctx.Job.IsInGang() {
-			if _, present := evictedJctxsByGangId[jctx.Job.GetGangInfo().Id()]; !present {
-				evictedJctxsByGangId[jctx.Job.GetGangInfo().Id()] = []*schedulercontext.JobSchedulingContext{}
-			}
-			evictedJctxsByGangId[jctx.Job.GetGangInfo().Id()] = append(evictedJctxsByGangId[jctx.Job.GetGangInfo().Id()], jctx)
-		}
 		if !jctx.IsEvicted {
 			return errors.New("evicted job %s is not marked as such")
```
dejanzele, line 488: Although not part of this PR, this should use `fmt.Errorf` and provide the job ID, as currently `errors.New` will just print `%s` directly

PR #4517, opened 2025-11-06
fix: update armada to respect node pod limits

### LINKED ISSUE #4515: Scheduler does not respect node pod limits
**Describe the bug**

We are seeing issues where the scheduler is scheduling pods to nodes that have cpu/memory availability but don't have pod count availability and those pods just sit in a pending state and get lease returned multiple times. 

**Expected behavior**

I expect Armada to consider node pod limits

jparraga-stackav:
#### What type of PR is this?
Bug fix

This pull request updates armada to respect node limits. Without this change it is possible for Armada to try and schedule pods to nodes which do not have capacity in regards to concurrent running pods. When this happens the pods get stuck in a pending state and end up getting lease returned. Presumably Armada considers these pods towards fair share even though they cannot run. 

1. Configures default pod limits in the Armada server to include `pods`
2. Configures the Armada scheduler to consider `pods` resource in scheduling decisions by default
3. Updates the executor to sanitize pod resources before submitting pods to k8s
4. Configures the executor to sanitize pod `pods` resources by default

Fixes #4515

In order to rollout safely the executors must be updated first. After that the scheduler/server can be rolled out in any order.


internal/executor/configuration/types.go
```diff
@@ -79,6 +79,7 @@ type KubernetesConfiguration struct {
 	FailedPodChecks           podchecks.FailedChecks
 	PendingPodChecks          *podchecks.Checks
 	FatalPodSubmissionErrors  []string
+	ResourcesToSanitize       []string
```
dejanzele, line 85: Can you please add docs for the new field?

dejanzele: In our docs, we often link to config in https://pkg.go.dev, i.e. for the executor it will be https://pkg.go.dev/github.com/armadaproject/armada/internal/executor/configuration#ApplicationConfiguration, it is easier for the end-users if they can see the config field docs there.

internal/executor/job/submit.go
```diff
@@ -147,6 +153,29 @@ func (submitService *SubmitService) submitPod(job *SubmitJob) (*v1.Pod, error) {
 	return pod, err
 }
 
+// Sanitizes pod resources that may be used during scheduling but are invalid at runtime.
+func (submitService *SubmitService) sanitizePodResources(pod *v1.Pod) {
```
dejanzele, line 157: nit: this function can be simplified to a simple `sanitizePodResources` function, no need for it to be a receiver method on SubmitService, same for `sanitizeResourceList`

JamesMurkin:
Hey so generally a good direction but I there are a few things I think we'd like to consider / possibly change

1. Generally we're trying to move to a world where all mutation happens on the executor API, and the executor does less edits. This PR does not move us in that direction
2. We already basically have this functionality in the executor API
  - https://github.com/armadaproject/armada/blob/master/internal/scheduler/api.go#L158
  - I think you could achieve what you want by simply adding a config option to `supportedResourceTypes` to say if we send it to the executor
 3. I think we should debate if we add pods to more of the config:
   - indexedResources - I think add it here
   - dominantResourceFairnessResourcesToConsider - possibly add it here. It'd mean we fairshare on pods, which is probably correct in some cases (happy to leave this for now)
  4. Could we confirm if you overwrite the config it does actually remove pods (I can't remember if it merges or overwrites and I don't have time to test now)
  5. I'm somewhat debating if we should add it as `defaultJobLimits` as it means now the api + scheduler need to be configured in unison. It could just be a scheduler concern (tbh all of this PR could be just a scheduler concern) but I think this is probably the most configurable option for now.

dejanzele:
Closing in favour of https://github.com/armadaproject/armada/pull/4841

PR #4533, opened 2025-11-13
Implement reserved capacity

JamesMurkin:
This is currently implemented as special handling of taints `armadaproject.io/reservation`

A reserved node is designated a purpose (via taint currently) and currently quite a light touch feature largely for reporting purposes

This PR is largely focussed around enabling a setup where a job is scheduled onto nodes in this order:
 - reserved
 - unreserved
 - spare reserved capacity

Reservations are an experimental feature and could expand into a first class feature in Armada. Which may include different scheduling treatment, a simpler way to configure (a label rather than a taint) and possible proper integration into existing pools rather than being split into their own pool

Key new features:
 - Reserved nodes will be in their own pool (optional)
   - Pool will be pool + "-reserved"
 - Run/Node reconciliation to preempt jobs if nodes swap pool/reservation (optional)
   - This is needed as Armada is not designed to properly handle node swapping pool/reservation
 - Adding reservation details to metrics
 - Add option to disable away scheduling on a pool
 - Added the experimental concept of submission group
   - If a job passes the submit check for any pool in the submission group, it is considered schedulable on all pools in the group
 - Allow wildcard for wellknown nodes types
   - Needed so we can let away jobs run on any reserved node

### DIFF
--- a/internal/common/util/taints.go
+++ b/internal/common/util/taints.go
@@ -8,7 +8,7 @@ import (
 
 const EmptyReservationName = "unspecified"
 
-func GetReservationKey(taints []v1.Taint) string {
+func GetReservationName(taints []v1.Taint) string {
 	for _, taint := range taints {
 		if taint.Key == constants.ReservationTaintKey {
 			if taint.Value == "" {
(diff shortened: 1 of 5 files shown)


internal/common/util/taints.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/armadaproject/armada/internal/common/constants"
+)
+
+const EmptyReservationName = "unspecified"
+
+func GetReservationKey(taints []v1.Taint) string {
```
JamesMurkin: Change to value

PR #4537, opened 2025-11-17
Fix PodDefaults panics accessing Ingress object

### LINKED ISSUE #4186: Executor panics when PodDefaults or PodDefaults.Ingress is not configured
PodDefaults and PodDefaults.Ingress are defined as struct pointers which implies that these values may not be configured and may be nil. When these values are not configured (since we want to disable the ingress defaults) the executor panics in several different places because the code assumes that these pointers are non-nil. 

ie.
```

:59.221 | panic: runtime error: invalid memory address or nil pointer dereference |  
-- | -- | --
  |   | 2025-02-03 19:34:59.222 | [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1b34339] |  
  |   | 2025-02-03 19:34:59.222 |   |  
  |   | 2025-02-03 19:34:59.222 | goroutine 359 [running]: |  
  |   | 2025-02-03 19:34:59.222 | github.com/armadaproject/armada/internal/executor/job.CreateSubmitJobFromExecutorApiJobRunLease(0xc0008b3860, 0x0) |  
  |   | 2025-02-03 19:34:59.222 | /home/jparraga/code/armada/internal/executor/job/util.go:35 +0xd9 |  
  |   | 2025-02-03 19:34:59.222 | github.com/armadaproject/armada/internal/executor/service.(*JobRequester).createSubmitJobs(0xc000691140, {0xc000323660, 0x4, 0xc001311ea0?}) |  
  |   | 2025-02-03 19:34:59.222 | /home/jparraga/code/armada/internal/executor/service/job_requester.go:135 +0x166 |  
  |   | 2025-02-03 19:34:59.222 | github.com/armadaproject/armada/internal/executor/service.(*JobRequester).RequestJobsRuns(0xc000691140) |  
  |   | 2025-02-03 19:34:59.222 | /home/jparraga/code/armada/internal/executor/service/job_requester.go:66 +0x1e8 |  
  |   | 2025-02-03 19:34:59.222 | github.com/armadaproject/armada/internal/common/task.(*BackgroundTaskManager).startBackgroundTask.func1() |  
  |   | 2025-02-03 19:34:59.222 | /home/jparraga/code/armada/internal/common/task/background_task.go:73 +0x64 |  
  |   | 2025-02-03 19:34:59.222 | created by github.com/armadaproject/armada/internal/common/task.(*BackgroundTaskManager).startBackgroundTask in goroutine 1 |  
  |   | 2025-02-03 19:34:59.222 | /home/jparraga/code/armada/internal/common/task/background_task.go:71 +0x22c
```

nikola-jokic:
#### What type of PR is this?

Bug fix

Fixes nil reference panics accessing Ingress object in PodDefaults

Fixes #4186


internal/executor/util/kubernetes_object.go
```diff
@@ -29,6 +30,13 @@ func CreateOwnerReference(pod *v1.Pod) metav1.OwnerReference {
 func ExtractIngresses(job *executorapi.JobRunLease, pod *v1.Pod, executorIngressConfig *configuration.IngressConfiguration) []*networking.Ingress {
 	result := make([]*networking.Ingress, 0, 10)
 
+	var executorIngressConfigAnnotations map[string]string
```
dejanzele: Could we instantiate `annotations := make()` here, and then in the if condition do a `util.MergeMaps` if `executorIngressConfigAnnotations` are defined?

That way we keep the same access pattern everywhere.

nikola-jokic: I think this should be okay now

internal/executor/util/kubernetes_object.go
```diff
@@ -29,6 +30,11 @@ func CreateOwnerReference(pod *v1.Pod) metav1.OwnerReference {
 func ExtractIngresses(job *executorapi.JobRunLease, pod *v1.Pod, executorIngressConfig *configuration.IngressConfiguration) []*networking.Ingress {
 	result := make([]*networking.Ingress, 0, 10)
 
+	annotations := make(map[string]string)
+	if executorIngressConfig != nil && len(executorIngressConfig.Annotations) > 0 {
+		annotations = maps.Clone(executorIngressConfig.Annotations)
```
dejanzele, line 36: isn't it simpler to do `MergeMaps` also?

dejanzele (approved):

internal/executor/util/kubernetes_object.go
nikola-jokic: In my opinion, there are few reasons I did it this way:
1. The annotations is extracted to be used from ingress config. It is a starting point of its usage, so I'm cloning it to ensure that `executorIngressConfig` is not modified in-place on accident if `annotations` are modified
2. It is less readable to see `MergeMaps` with an empty map we allocated 2 lines above. This clearly states that the map is being cloned, not merged with the empty map. Having Clone makes it obvious that this is a clone of the `executorIngressConfig.Annotations`, nothing more, nothing less.
3. `util.MergeMaps` below are appropriate since we are merging the base map of annotations with `additionalObject`'s annotations.

PR #4544, opened 2025-11-21
Log start and end time on each scheduler tick, and on each cycle

nikola-jokic:
#### What type of PR is this?

Improvement

Logs when cycles start/end and when scheduling starts/ends.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -7,8 +7,8 @@ pull_request_rules:
         success_conditions:
         - or:
           - "#approved-reviews-by>=2"
-          - and: 
+          - and:
             - "#approved-reviews-by>=1"
-            - "author~=^(d80tb7|dave[-]gantenbein|dejanzele|eleanorpratt|geaere|JamesMurkin|mauriceyap|masipauskas|MustafaI|zuqq|richscott|robertdavidsmith|samclark|suprjinx|EnricoMi)$"
+            - "author~=^(d80tb7|dave[-]gantenbein|dejanzele|eleanorpratt|geaere|JamesMurkin|mauriceyap|masipauskas|MustafaI|zuqq|richscott|robertdavidsmith|samclark|suprjinx|EnricoMi|nikola-jokic)$"
         title:
           Two approvals required, or one if author is a maintainer.


internal/scheduler/scheduler.go
```diff
@@ -162,65 +162,73 @@ func (s *Scheduler) Run(ctx *armadacontext.Context) error {
...
-			if leaderToken.Leader() && leaderToken != prevLeaderToken {
-				ctx.Infof("becoming leader")
-				syncContext, cancel := armadacontext.WithTimeout(ctx, 5*time.Minute)
-				err := s.ensureDbUpToDate(syncContext, 1*time.Second)
-				if err != nil {
-					ctx.Logger().WithStacktrace(err).Error("could not become leader")
-					leaderToken = leader.InvalidLeaderToken()
-				} else {
-					fullUpdate = true
+			func() {
```
dejanzele, line 168: I am wondering at this point should we extract it in a `run` (or whatever better name) function.

internal/scheduler/scheduler.go
```diff
@@ -162,65 +162,73 @@ func (s *Scheduler) Run(ctx *armadacontext.Context) error {
...
+				// Run a scheduler cycle.
+				//
+				// If there is an error, we can't guarantee that the scheduler-internal state is consistent with what was published
+				// (scheduling decisions may have been partially published)
+				// and we must invalidate the held leader token to trigger flushing Pulsar at the next cycle.
+				//
+				// TODO: Once the Pulsar client supports transactions, we can guarantee consistency even in case of errors.
+				shouldSchedule := s.clock.Now().Sub(previousSchedulingRoundEnd) > s.schedulePeriod
+				if !shouldSchedule {
+					ctx.Info("Won't schedule this cycle as still within schedulePeriod")
```
dejanzele, line 206: We could even log when should the next schedule period happen in this log line. (We should have the timestamp from the logger already)

internal/scheduler/scheduler.go
```diff
@@ -251,6 +259,10 @@ func (s *Scheduler) Run(ctx *armadacontext.Context) error {
 //     As state transitions are persisted and read back from the schedulerDb over later cycles,
 //     there is no change to the jobDb, since the correct changes have already been made.
 func (s *Scheduler) cycle(ctx *armadacontext.Context, updateAll bool, leaderToken leader.LeaderToken, shouldSchedule bool, cycleNumber int) (scheduling.SchedulerResult, error) {
+	ctx.Logger().Infof("starting cycle at: %s", s.clock.Now().Format(time.RFC3339Nano))
+	defer func(ctx *armadacontext.Context) {
+		ctx.Logger().Infof("finished cycle at: %s", s.clock.Now().Format(time.RFC3339Nano))
```
dejanzele: nit: I'd remove the `:` as previous log lines use just `at|in %s`

internal/scheduler/scheduler.go
```diff
@@ -251,6 +259,10 @@ func (s *Scheduler) Run(ctx *armadacontext.Context) error {
 //     As state transitions are persisted and read back from the schedulerDb over later cycles,
 //     there is no change to the jobDb, since the correct changes have already been made.
 func (s *Scheduler) cycle(ctx *armadacontext.Context, updateAll bool, leaderToken leader.LeaderToken, shouldSchedule bool, cycleNumber int) (scheduling.SchedulerResult, error) {
+	ctx.Logger().Infof("starting cycle at: %s", s.clock.Now().Format(time.RFC3339Nano))
```
dejanzele: Worth to double-check should we actually log the timestamp, or the logger should automatically add a timestamp.

dejanzele: @JamesMurkin can you weigh in?

internal/scheduler/scheduler.go
nikola-jokic: What would be the next schedule period?

internal/scheduler/scheduler.go
nikola-jokic: Maybe, but we would have to return multiple results or a result struct in order to assign new values to local variables. This method does multiple things, so there are two things we can do:
1. Return  a result struct and then assign new values (Not a big fan of that because it just moves the logic that is highly dependent on the call context
2. Extract the state machine for the run function that would keep the reference to the scheduler, and run local variables (such as `previousSchedulingRoundEnd`, etc). This approach seams more appropriate to me, but anonymous function wrapper seamed good enough in this particular use-case

Let me know if you want me to take one of these paths, or just leave it like this

internal/scheduler/scheduler.go
dejanzele: I think it should be `previousSchedulingRoundEnd + s.schedulePeriod`

internal/scheduler/scheduler.go
dejanzele: it is fine as is

PR #4606, opened 2026-01-15
Scala Client: add new jobWatcher() method for watcher API

richscott:
Add new jobWatcher(queue, jobset, lastMessage) method to ArmadaClient class, based on original logic in armada-spark repository.

Fill in implementation of watch() method in ArmadaClientSuite (test) EventMockServer. Clear out job, queue, and status maps (the backing store for the ArmadaMockServer) before running each test, to avoid test pollution.

Change from using .onComplete{..} idiom to Await.result(response, timeout) for more robust async testing behavior. Add `watch` support to testing mock server, add unit test for jobWatcher().

Fixes https://github.com/G-Research/spark/issues/160

### DIFF
--- a/internal/lookout/conversions/convert.go
+++ b/internal/lookout/conversions/convert.go
@@ -41,6 +41,7 @@ func ToSwaggerJob(job *model.Job) *models.Job {
 		CancelUser:         job.CancelUser,
 		Node:               job.Node,
 		Cluster:            job.Cluster,
+		Pool:               job.Pool,
 		ExitCode:           job.ExitCode,
 		RuntimeSeconds:     job.RuntimeSeconds,
 	}
--- a/internal/lookout/gen/models/job.go
+++ b/internal/lookout/gen/models/job.go
@@ -93,6 +93,9 @@ type Job struct {
 	// Min Length: 1
 	Owner string `json:"owner"`
 
+	// pool
+	Pool *string `json:"pool,omitempty"`
+
 	// priority
 	// Required: true
 	Priority int64 `json:"priority"`
--- a/internal/lookout/gen/restapi/embedded_spec.go
+++ b/internal/lookout/gen/restapi/embedded_spec.go
@@ -671,6 +671,10 @@ func init() {
           "minLength": 1,
           "x-nullable": false
         },
+        "pool": {
+          "type": "string",
+          "x-nullable": true
+        },
         "priority": {
           "type": "integer",
           "format": "int64",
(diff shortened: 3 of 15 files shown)


client/scala/armada-scala-client/src/test/scala/io/armadaproject/armada/ArmadaClientSuite.scala
```diff
@@ -69,14 +89,103 @@ private class EventMockServer extends EventGrpc.Event {
...
+    // Find all jobs in the specified jobSet
+    val matchingJobs = jobMap.filter { case (_, job) =>
+      job.jobSetId == request.jobSetId
+    }
+
+    // Emit events for each job showing the usual (not all possible) transition states
+    matchingJobs.foreach { case (jobId, job) =>
+      // Check if job exists in statusMap
+      statusMap.get(jobId) match {
+        case Some(currentStatus) =>
```
GeorgeJahad: it doesn't appear as currentStatus is used anywhere.  Shouldn't it be?

Also, it looks like you are emiting all the "usual" states simultaneously below.  That seems odd.  Don't we want the state emitted to depend on the current state?

richscott: The `currentStatus` was just a placeholder for whatever `get(jobId)` returned, but it was ultimately just for determining if the given job ID really existed in the mock Armada. I have changed this to just use `Map.contains()` instead, which makes the code simpler, in commit 38306bb5.

The state events are emitted sequentially (but not simultaneously, strictly-speaking), mostly just to verify that they are received by the iterator that `jobWatcher(..)` returns (and is used in armada-spark).  The code just checks that all the events are received and in the order they were emitted.  There's no higher-level state-transition logic in here for ordering, valid transition states, etc - that's elsewhere in Armada (outside the client), so it didn't seem necessary to try to reproduce and test that here.

Also, the e2e test in armada-spark appears to run successfully with a change of using the new `jobWatcher()` API method, but I will wait until this is merged and published out to the world.

GeorgeJahad: I'm ok with it, but it is puzzling to see all events being emitted sequentially.  I would add some of the comments from the explanation above explaining why it is the way it is.

richscott: 👍🏼  I've added an explanatory note in commit 9b0e647c - let me know if seems incomplete or confusing.

PR #4612, opened 2026-01-19
Add preempt and cancel jobs on node for specified executor

nikola-jokic:
#### What type of PR is this?

Enhancement

Jobs may need to be preempted or canceled on a node. The server and `armadactl` should support these use cases natively.

### DIFF
--- a/internal/scheduler/database/query/query.sql
+++ b/internal/scheduler/database/query/query.sql
@@ -70,9 +70,6 @@ UPDATE runs SET run_attempted = true WHERE run_id = ANY(sqlc.arg(run_ids)::text[
 -- name: MarkJobRunsRunningById :exec
 UPDATE runs SET running = true WHERE run_id = ANY(sqlc.arg(run_ids)::text[]);
 
--- name: MarkRunsCancelledByJobId :exec
-UPDATE runs SET cancelled = true WHERE job_id = ANY(sqlc.arg(job_ids)::text[]);
-
 -- name: SelectJobsForExecutor :many
 SELECT jr.run_id, j.queue, j.job_set, j.user_id, j.groups, j.submit_message
 FROM runs jr
--- a/client/scala/armada-scala-client/pom.xml
+++ b/client/scala/armada-scala-client/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.armadaproject</groupId>
   <artifactId>armada-scala-client_2.13</artifactId>
-  <version>0.3.0-SNAPSHOT</version>
+  <version>0.4.0-SNAPSHOT</version>
   <name>Armada Scala Client</name>
   <description>A Scala client for Armada.</description>
   <inceptionYear>2025</inceptionYear>
--- a/client/scala/armada-scala-client/src/main/scala/io/armadaproject/armada/ArmadaClient.scala
+++ b/client/scala/armada-scala-client/src/main/scala/io/armadaproject/armada/ArmadaClient.scala
@@ -1,7 +1,7 @@
 package io.armadaproject.armada
 
 import api.job.{JobStatusRequest, JobStatusResponse, JobsGrpc}
-import api.event.EventGrpc
+import api.event.{EventGrpc, WatchRequest}
 import api.submit.{
   CancellationResult,
   JobCancelRequest,
@@ -73,6 +73,18 @@ class ArmadaClient(channel: ManagedChannel) {
     val qReq = QueueGetRequest(name)
     val blockingStub = SubmitGrpc.blockingStub(channel)
     blockingStub.getQueue(qReq)
+  }
+
+  def jobWatcher(
+      q: String,
+      jobSet: String,
+      lastMessage: String
+  ): scala.collection.Iterator[api.event.EventStreamMessage] = {
+    val eventStub = EventGrpc.blockingStub(channel)
+    val watchRequest =
+      WatchRequest(queue = q, jobSetId = jobSet, fromId = lastMessage)
+
+    return eventStub.watch(watchRequest)
   }
 }
 
--- a/client/scala/armada-scala-client/src/test/scala/io/armadaproject/armada/ArmadaClientSuite.scala
+++ b/client/scala/armada-scala-client/src/test/scala/io/armadaproject/armada/ArmadaClientSuite.scala
@@ -39,17 +39,37 @@ import api.job.{
 }
 import com.google.protobuf.empty.Empty
 import api.health.HealthCheckResponse
-import api.event.{EventGrpc, EventStreamMessage, JobSetRequest, WatchRequest}
+import api.event.{
+  EventGrpc,
+  EventMessage,
+  EventStreamMessage,
+  JobPendingEvent,
+  JobQueuedEvent,
+  JobRunningEvent,
+  JobSetRequest,
+  JobSucceededEvent,
+  WatchRequest
+}
+import api.event.EventMessage.Events
+
 import io.grpc.{Server, ServerBuilder, Status, StatusRuntimeException}
 import jkugiya.ulid.ULID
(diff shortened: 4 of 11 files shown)


example/example-job.yaml
```diff
@@ -6,6 +6,7 @@ jobs:
     podSpec:
       terminationGracePeriodSeconds: 0
       restartPolicy: Never
+      priorityClassName: armada-preemptible
```
JamesMurkin: I think we just added a preemptible example jobs - so I think we should remove this change

See:

https://github.com/armadaproject/armada/blob/master/example/example-preemptible-job.yaml

cmd/armadactl/cmd/preempt.go
```diff
@@ -84,6 +85,51 @@ func preemptExecutorCmd() *cobra.Command {
 	return cmd
 }
 
+func preemptNodeCmd() *cobra.Command {
+	a := armadactl.New()
+	cmd := &cobra.Command{
+		Use:   "node <name> <executor>",
```
JamesMurkin: Thinking out loud, not sure if it should be `<name> <executor>` or `<executor> <node>` or if we should make executor optional (or flagged)

I think what you've done is sensible - just a general thought - we may need to reevaluate based on usage later

_local/priority-class.yaml
```diff
@@ -5,3 +5,11 @@ metadata:
 value: 1000
 globalDefault: false
 description: "This priority class should be as a default for Armada jobs."
+---
```
JamesMurkin: Why do we need this? Not saying we don't, just how Armada _should_ work by default is that is doesn't rely on cluster priority classes

However maybe that isn't happening and is why we need these

internal/armadactl/preempt.go
```diff
@@ -56,6 +56,26 @@ func (a *App) PreemptOnExecutor(executor string, queues []string, priorityClasse
 	return nil
 }
 
+func (a *App) PreemptOnNode(node string, executor string, queues []string, priorityClasses []string) error {
+	queueMsg := strings.Join(queues, ",")
+	priorityClassesMsg := strings.Join(priorityClasses, ",")
+	// If the provided slice of queues is empty, jobs on all queues will be cancelled
```
JamesMurkin:
```suggestion
	// If the provided slice of queues is empty, jobs on all queues will be preempted
```

internal/scheduleringester/schedulerdb.go
```diff
@@ -450,6 +450,63 @@ func (s *SchedulerDb) WriteDbOp(ctx *armadacontext.Context, tx pgx.Tx, op DbOper
...
+			if err != nil {
+				return errors.Wrapf(err, "error preempting jobs on node %s on executor %s by queue and priority class", nodeOnExecutor.Node, nodeOnExecutor.Executor)
+			}
+
+			if len(jobs) == 0 {
+				continue
+			}
+
+			if len(preemptRequest.PriorityClasses) > 0 {
+				jobs, err = filterJobsByPriorityClasses(jobs, preemptRequest.PriorityClasses)
```
JamesMurkin, line 468: Could this func be used in `PreemptExecutor` ?

internal/server/node/node.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil, status.Errorf(codes.Internal, "error publishing preempt on node event for executor %s: %s", req.Name, err)
+	}
+
+	return &types.Empty{}, nil
+}
+
+// CancelOnNode implements api.NodeServer.
+func (s *Server) CancelOnNode(grpcCtx context.Context, req *api.NodeCancelRequest) (*types.Empty, error) {
+	ctx := armadacontext.FromGrpcCtx(grpcCtx)
+	err := s.authorizer.AuthorizeAction(ctx, permissions.PreemptAnyJobs)
```
JamesMurkin: CancelAnyJobs perm

internal/server/node/node.go
```diff
@@ -0,0 +1,117 @@
...
+
+	var ep *armadaerrors.ErrUnauthorized
+	if errors.As(err, &ep) {
+		return nil, status.Errorf(codes.PermissionDenied, "error preempting jobs on executor %s: %s", req.Name, ep)
```
JamesMurkin: These messages say preempted, replace with cancel

internal/server/node/node.go
```diff
@@ -0,0 +1,117 @@
+package node
```
JamesMurkin, line 1: Ideally there would be some tests on this - as its our API

internal/scheduleringester/dbops.go
```diff
@@ -82,6 +82,20 @@ type PreemptOnExecutor struct {
 	PriorityClasses []string
 }
 
+type PreemptOnNode struct {
```
JamesMurkin, line 85: Can we double check there are no sensible tests we can add for these

cmd/armadactl/cmd/preempt.go
nikola-jokic: This sounds good by me
What we can do is
1. have executor as flag `-e` for example
2. If not provided, search for nodes. If there are two executors with the single node name, we stop the command and report that `-e` must be specified
3. Otherwise, we don't require `-e`

internal/scheduler/database/query/query.sql
```diff
@@ -159,6 +159,16 @@ WHERE jr.executor = @executor
...
 
+-- name: SelectJobsByNodeAndExecutorAndQueues :many
+SELECT j.*
+FROM runs jr
+        JOIN jobs j
+             ON jr.job_id = j.job_id
+WHERE jr.node = @node
+  AND jr.executor = @executor
+  AND j.queue = ANY(@queues::text[])
+  AND jr.succeeded = false AND jr.failed = false AND jr.cancelled = false AND jr.preempted = false;
```
dejanzele, line 167: Instead of `AND jr.succeeded = false AND jr.failed = false AND jr.cancelled = false AND jr.preempted = false` we could add a `terminated` field which would be autogenerated from `succeeded OR failed OR cancelled OR preempted`. That way, an index will need to update only for 1 additional field instead of 4 additional fields and produce less index bloat.
Here is an example how we implemented it for the `jobs` table - https://github.com/armadaproject/armada/blob/master/internal/scheduler/database/migrations/027_add_terminated_column.sql

cc @masipauskas

internal/scheduler/database/query/query.sql
```diff
@@ -159,6 +159,16 @@ WHERE jr.executor = @executor
...
   AND jr.succeeded = false AND jr.failed = false AND jr.cancelled = false AND jr.preempted = false;
 
+-- name: SelectJobsByNodeAndExecutorAndQueues :many
+SELECT j.*
+FROM runs jr
+        JOIN jobs j
+             ON jr.job_id = j.job_id
+WHERE jr.node = @node
+  AND jr.executor = @executor
+  AND j.queue = ANY(@queues::text[])
```
dejanzele, line 166: Also can you use `jr.queue = ANY(@queues::text[])` so the filter can be done before the join, as that will speed up the query?

nikola-jokic: Should we update the `SelectJobsByExecutorAndQueues` as well then? I based this query on that one

internal/scheduler/database/query/query.sql
nikola-jokic: @JamesMurkin If we go with that, should we update other queries?
It makes sense to me to go with these changes, and the next PR would address jr.queue and termination.

dejanzele: actually I require this new field for some other work so I'll create a PR which will add it in a migration

internal/scheduler/database/query/query.sql
dejanzele: I'd say just update the one you added, I can prepare a PR with the migration and for the other queries

PR #4622, opened 2026-01-22
Detect jobs on deleted nodes during run reconciliation

mauriceyap:
Previously, the reconciler only checked for pool/reservation mismatches on existing nodes. Jobs assigned to nodes that no longer exist were silently ignored, leaving them in an invalid state.

Now the reconciler builds a set of current node IDs and fails reconciliation for any job whose assigned node is missing. This ensures orphaned jobs are properly marked as failed rather than remaining in limbo.

Also fixes typos in error messages ("this jobs" -> "this job's", "resevation" -> "reservation").

### DIFF
--- a/internal/lookoutui/src/pages/jobs/components/sidebar/SidebarTabJobDetails.tsx
+++ b/internal/lookoutui/src/pages/jobs/components/sidebar/SidebarTabJobDetails.tsx
@@ -2,6 +2,7 @@ import { Alert } from "@mui/material"
 
 import { formatBytes, formatCpu } from "../../../../common/resourceUtils"
 import { useFormatNumberWithUserSettings } from "../../../../components/hooks/formatNumberWithUserSettings"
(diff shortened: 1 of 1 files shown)


internal/scheduler/scheduling/reconciliation.go
```diff
@@ -35,13 +35,34 @@ func (r *RunNodeReconciler) ReconcileJobRuns(txn *jobdb.Txn, executors []*schedu
...
-	jobsToReconcile := r.getJobsToReconcileByNodeId(txn)
+	jobsToReconcileByNodeId := r.getJobsToReconcileByNodeId(txn)
 
 	configByPool := poolConfigSliceToMap(r.poolsToReconcile)
+	nodeIdSet := r.buildNodeIdSet(nodes)
 
 	var result []*FailedReconciliationResult
+
+	// Check for jobs whose nodes no longer exist
+	for nodeId, jobs := range jobsToReconcileByNodeId {
```
dejanzele, line 55: nit: as this functions a lot of iterations and processing, might be a bit more easier to understand if the processings were extracted into helpers

PR #4631, opened 2026-01-26
Bump Go dependencies

nikola-jokic:
#### What type of PR is this?

Bump

The updates didn't compile so I manually updated all of them. The rest of the updates should be handled by dependabot


internal/server/submit/validation/submit_request_test.go
```diff
@@ -657,7 +657,9 @@ func TestValidatePodSpecSize(t *testing.T) {
 		},
 	}
 
-	defaultPodSpecSize := uint(len(protoutil.MustMarshall(defaultPodSpec)))
+	b, err := defaultPodSpec.Marshal()
```
JamesMurkin: Why do we need this change?

PR #4635, opened 2026-01-27
Add node job phase metric showing active jobs on nodes

nikola-jokic:
#### What type of PR is this?

Feature

Include metric showing active jobs on nodes by executor


internal/scheduler/metrics.go
```diff
@@ -420,6 +427,15 @@ func (c *MetricsCollector) updateClusterMetrics(ctx *armadacontext.Context) ([]p
 					}
 					phaseCountByQueue[key]++
 
+					switch phase {
+					case "PENDING", "RUNNING": // Only active phases
```
dejanzele: Can we use `schedulerobjects.JobRunState_name[schedulerobjects.JobRunState_RUNNING]` and `schedulerobjects.JobRunState_name[schedulerobjects.JobRunState_PENDING]` instead of hardcoded strings?

It is not perfect, but we reuse existing logic without duplication or repetition.

internal/common/metrics/scheduler_metrics.go
```diff
@@ -166,6 +166,13 @@ var QueueLeasedPodCountDesc = prometheus.NewDesc(
 	nil,
 )
 
+var NodeJobPhaseCounterDesc = prometheus.NewDesc(
+	MetricPrefix+"node_job_phase_count",
+	"Number of jobs in a given phase on a node",
```
dejanzele, line 171: nit: I'd tweak the description a bit to explain all metric labels, for example:
`Count of jobs per node, executor, and job phase.`

PR #4640, opened 2026-01-28
feat: add preempt reason

williamvega:
#### What type of PR is this?

Feature

Currently the reason provided when preempting a job does not get stored, and the response to the UI is hardcoded for preempted jobs. This change adds the reason to the scheduler db runs table and appends the user to the run reason provided.

image


internal/scheduleringester/schedulerdb.go
```diff
@@ -228,15 +228,19 @@ func (s *SchedulerDb) WriteDbOp(ctx *armadacontext.Context, tx pgx.Tx, op DbOper
...
-				Queue:  key.queue,
-				JobSet: key.jobSet,
-				JobIds: value,
-			}
-			err := queries.MarkJobRunsPreemptRequestedByJobId(ctx, params)
-			if err != nil {
-				return errors.WithStack(err)
+		for key, jobReasons := range o {
+			// pocess each job individually since each job has its own reason
+			for jobId, reason := range jobReasons {
```
JamesMurkin: I think you could group these again by reason and in the extreme massively reduce the number of database queries and the expense of a little bit of compute in the ingester
- Possibly I'm missing something that means that isn't possible

Alternatively we could batch these so it is less round trips to the db
 - See `MarkJobsValidated`
 - It may be a bit more work to get it to work for the generated queries (I'm not 100% without looking)

internal/scheduler/jobdb/job_run_test.go
```diff
@@ -136,6 +137,7 @@ func TestDeepCopy(t *testing.T) {
 		true,
 		true,
 		true,
+		nil,
```
mauriceyap: Please could you use a non-nil value here and in the expected object below? I think this test is more effective if the entire object is populated

internal/scheduleringester/instructions.go
```diff
@@ -316,11 +316,18 @@ func (c *JobSetEventsInstructionConverter) handleJobErrors(jobErrors *armadaeven
 }
 
 func (c *JobSetEventsInstructionConverter) handleJobPreemptionRequested(preemptionRequested *armadaevents.JobPreemptionRequested, meta eventSequenceCommon) ([]DbOperation, error) {
+	// cobine the preemption reason and user
```
mauriceyap:
```suggestion
	// combine the preemption reason and user
```

internal/scheduleringester/instructions.go
```diff
@@ -316,11 +316,18 @@ func (c *JobSetEventsInstructionConverter) handleJobErrors(jobErrors *armadaeven
...
+	reasonWithUser := preemptionRequested.Reason
+	if meta.user != "" {
+		reasonWithUser = preemptionRequested.Reason + " | Preempted by " + meta.user
```
mauriceyap: nit
```suggestion
		reasonWithUser += " | Preempted by " + meta.user
```

internal/scheduleringester/schedulerdb.go
```diff
@@ -228,15 +228,19 @@ func (s *SchedulerDb) WriteDbOp(ctx *armadacontext.Context, tx pgx.Tx, op DbOper
...
-			params := schedulerdb.MarkJobRunsPreemptRequestedByJobIdParams{
-				Queue:  key.queue,
-				JobSet: key.jobSet,
-				JobIds: value,
-			}
-			err := queries.MarkJobRunsPreemptRequestedByJobId(ctx, params)
-			if err != nil {
-				return errors.WithStack(err)
+		for key, jobReasons := range o {
+			// pocess each job individually since each job has its own reason
```
mauriceyap:
```suggestion
			// process each job individually since each job has its own reason
```

internal/scheduleringester/schedulerdb.go
williamvega: I would say the batching method may be better because the preempt reason is passed as a freeform string arg by admins through the cli (and their id is appended to the end of the reason), so there would be many groups with few jobs each. 

I'll try batching.

williamvega: `MarkRunsForJobPreemptRequested` is now batching the preempt reason updates.

internal/scheduler/database/query/query.sql
```diff
@@ -53,7 +53,7 @@ SELECT run_id FROM runs;
 SELECT * FROM runs WHERE serial > $1 AND job_id = ANY(sqlc.arg(job_ids)::text[]) ORDER BY serial;
 
 -- name: MarkJobRunsPreemptRequestedByJobId :exec
-UPDATE runs SET preempt_requested = true WHERE queue = sqlc.arg(queue) and job_set = sqlc.arg(job_set) and job_id = ANY(sqlc.arg(job_ids)::text[]) and terminated = false;
+UPDATE runs SET preempt_requested = true, preempt_reason = sqlc.arg(preempt_reason) WHERE queue = sqlc.arg(queue) and job_set = sqlc.arg(job_set) and job_id = ANY(sqlc.arg(job_ids)::text[]) and cancelled = false and succeeded = false and failed = false and terminated = false;
```
JamesMurkin: This looks to be a merge conflict, you can remove

`and cancelled = false and succeeded = false and failed = false`

JamesMurkin: Actually is this even called anymore?

internal/scheduleringester/schedulerdb.go
```diff
@@ -228,17 +228,17 @@ func (s *SchedulerDb) WriteDbOp(ctx *armadacontext.Context, tx pgx.Tx, op DbOper
...
-		for key, value := range o {
-			params := schedulerdb.MarkJobRunsPreemptRequestedByJobIdParams{
-				Queue:  key.queue,
-				JobSet: key.jobSet,
-				JobIds: value,
-			}
-			err := queries.MarkJobRunsPreemptRequestedByJobId(ctx, params)
-			if err != nil {
-				return errors.WithStack(err)
+		markRunsPreemptRequestedSqlStatement := "UPDATE runs SET preempt_requested = true, preempt_reason = $1 WHERE queue = $2 and job_set = $3 and job_id = $4 and cancelled = false and succeeded = false and failed = false"
```
JamesMurkin: Please use terminated column (appeared since your PR was created) and remove `cancelled = false and succeeded = false and failed = false`

It'd also still be tempted to make it take an ANY for the job_id and group them by reason

- In cases where someone submits many job ids preempted by id (which could happen) it'll be much faster
- I suspect the common case is batches will turn up with the same reason for the same job_set

internal/scheduleringester/schedulerdb.go
```diff
@@ -228,13 +228,27 @@ func (s *SchedulerDb) WriteDbOp(ctx *armadacontext.Context, tx pgx.Tx, op DbOper
...
+			reasonToJobIds := make(map[string][]string)
+			for jobId, reason := range preemptReason {
+				reasonToJobIds[reason] = append(reasonToJobIds[reason], jobId)
+			}
+
+			// batch updates for jobs with the same reason
+			batch := &pgx.Batch{}
+			for reason, jobIds := range reasonToJobIds {
+				for i := 0; i < len(jobIds); i += batchSize {
+					end := min(i+batchSize, len(jobIds))
```
JamesMurkin, line 245: I think you could probably simplify this with https://github.com/armadaproject/armada/blob/master/internal/common/slices/slices.go#L13

However that is just a nitpick

PR #4648, opened 2026-01-29
[Lookout UI] Adding Copy All Button to Job ID Column

Sigele:
#### What type of PR is this?
Feature

This PR adds a button in the jobID column of the Lookout UI to copy all visible Job IDs on the current page. This eliminates the need for users to copy IDs individually. I made minor changes to the HeaderTableCell attributes to ensure that the button remains visible when the column is resized to the minimum width.

Fixes #

### DIFF
--- a/website/yarn.lock
+++ b/website/yarn.lock
@@ -6350,17 +6350,12 @@ minimist@^1.0.0, minimist@^1.2.0, minimist@^1.2.6:
   resolved "https://registry.yarnpkg.com/minipass/-/minipass-7.1.2.tgz#93a9626ce5e5e66bd4db86849e7515e92340a707"
   integrity sha512-qOOzS1cBTWYF4BH8fVePDBOO9iptMnGUEZwNc/cMWnTV2nVLZ7VoNWEPHkYczZA0pdoA7dl6e7FL659nX9S2aw==
 
-minizlib@^3.0.1:
-  version "3.0.2"
-  resolved "https://registry.yarnpkg.com/minizlib/-/minizlib-3.0.2.tgz#f33d638eb279f664439aa38dc5f91607468cb574"
-  integrity sha512-oG62iEk+CYt5Xj2YqI5Xi9xWUeZhDI8jjQmC5oThVH5JGCTgIjr7ciJDzC7MBzYd//WvR1OTmP5Q38Q8ShQtVA==
+minizlib@^3.1.0:
+  version "3.1.0"
+  resolved "https://registry.yarnpkg.com/minizlib/-/minizlib-3.1.0.tgz#6ad76c3a8f10227c9b51d1c9ac8e30b27f5a251c"
+  integrity sha512-KZxYo1BUkWD2TVFLr0MQoM8vUUigWD3LlD83a/75BqC+4qE0Hb1Vo5v1FgcfaNXvfXzr+5EhQ6ing/CaBijTlw==
   dependencies:
     minipass "^7.1.2"
-
-mkdirp@^3.0.1:
-  version "3.0.1"
-  resolved "https://registry.yarnpkg.com/mkdirp/-/mkdirp-3.0.1.tgz#e44e4c5607fb279c168241713cc6e0fea9adcb50"
-  integrity sha512-+NsyUUAZDmo6YVHzL/stxSu3t9YS1iljliy3BSDrXJ/dkn1KYdmtZODGGjLcc9XLgVVpH4KshHB8XmZgMhaBXg==
 
 mlly@^1.7.4:
   version "1.7.4"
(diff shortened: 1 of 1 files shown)


internal/lookoutui/src/pages/jobs/components/JobsTableCell.tsx
```diff
@@ -159,50 +169,59 @@ export function HeaderCell({
...
         >
-          {header.column.getCanSort() ? (
-            <div
-              style={{
-                display: "flex",
-                flexDirection: "row",
-                textOverflow: "ellipsis",
-                whiteSpace: "nowrap",
-                overflow: "hidden",
+          <div
```
mauriceyap: This change has removed the conditional check for `header.column.getCanSort()`. The `<TableSortLabel>` component is now rendered unconditionally for all columns, including non-sortable ones. Please could you restore the conditional rendering so that `<TableSortLabel>` is only used when `header.column.getCanSort()` returns true?

internal/lookoutui/src/pages/jobs/components/JobsTableCell.tsx
```diff
@@ -159,50 +169,59 @@ export function HeaderCell({
...
+            style={{
+              display: "flex",
+              flexDirection: "row",
+              alignItems: "center",
+              gap: 6,
+              whiteSpace: "nowrap",
+            }}
```
mauriceyap: I think we still need `overflow: 'hidden'` here to prevent potential layout issues when content overflows e.g. long header text. Was there a visual reason why you removed this?

internal/lookoutui/src/pages/jobs/components/JobsTableCell.tsx
```diff
@@ -94,6 +95,16 @@ export function HeaderCell({
...
 
+  // job ID copy values
+  const isJobIdColumn = header.column.id === StandardColumnId.JobID
+  const jobsTable = header.getContext().table
+  const visibleRows = jobsTable.getRowModel().rows
+  const visibleRowsExist = visibleRows.length > 0
+  const allJobIds = visibleRows
+    .map((row) => row.original.jobId)
+    .filter(Boolean)
+  const copyContent = allJobIds.join(", ")
```
mauriceyap: This is logically correct, but I think this would be slightly cleaner
```suggestion
  const copyContent = header.column.id === StandardColumnId.JobID
    ? header.getContext().table.getRowModel().rows
        .map((row) => row.original.jobId)
        .filter(Boolean)
        .join(", ")
    : ""
```

internal/lookoutui/src/pages/jobs/components/JobsTableCell.tsx
```diff
@@ -94,6 +95,16 @@ export function HeaderCell({
   const borderWidth = 1
   const remainingWidth = totalWidth - resizerWidth - borderWidth
 
+  // job ID copy values
+  const isJobIdColumn = header.column.id === StandardColumnId.JobID
```
mauriceyap: This would be better if it were a configuration value in `GET_JOB_COLUMNS()` in `internal/lookoutui/src/common/jobsTableColumns.tsx` - perhaps:
```
accessorColumn({
    id: StandardColumnId.JobID,
    // ...
    additionalMetadata: {
        // ...
        allowCopyColumn: true,
    }
})
```

Sigele: Addressed comment in [this commit](https://github.com/armadaproject/armada/pull/4648/commits/9d1972b9dbdd71dc920ddf49e789dd0b722afb2f)

internal/lookoutui/src/pages/jobs/components/JobsTableCell.tsx
Sigele: Addressed [here](https://github.com/armadaproject/armada/pull/4648/commits/7f5c56a7dc0ae11db921a3bb2c28fbfe31468060)

internal/lookoutui/src/pages/jobs/components/JobsTableCell.tsx
Sigele: The goal was to make sure that the copy button did not disappear when then column was resized to minimized width. Was able to restore coverflow hidden in [this commit ](https://github.com/armadaproject/armada/pull/4648/commits/e04429039cd79f828ce447cbb36a948c9d35d26f)

internal/lookoutui/src/pages/jobs/components/JobsTableCell.tsx
Sigele: Addressed in subsequent commit [here](https://github.com/armadaproject/armada/pull/4648/commits/e04429039cd79f828ce447cbb36a948c9d35d26f)

MustafaI (approved):

PR #4649, opened 2026-01-29
[Lookout UI] Submit dialogs on enter

d2burkhalter:
#### What type of PR is this?
Lookout UI enhancement

Allows users to press enter to submit dialogs for canceling and reprioritizing jobs.


internal/lookoutui/src/pages/jobs/components/ReprioritizeDialog.tsx
```diff
@@ -140,72 +164,74 @@ export const ReprioritizeDialog = ({ onClose, selectedItemFilters }: Reprioritiz
   const reprioritizableJobsCount = reprioritizableJobs.length
   const selectedJobsCount = selectedJobs.length
   return (
-    <Dialog open={true} onClose={onClose} fullWidth maxWidth="xl">
+    <Dialog open={true} onClose={onClose} fullWidth maxWidth="xl" onKeyDown={handleDialogKeyDown}>
```
mauriceyap, line 167: Could this instead be achieved by passing the props `component="form" onSubmit={handleSubmit}` to the `<Dialog>` here? If it does, the button below with `type="submit"` should hopefully take effect since it's inside this rendered `<form>` component and so we'd be able to remove `handleDialogKeyDown`.

mauriceyap (approved):

PR #4655, opened 2026-02-02
Allow pool scheduling to fail independently for some errors

JamesMurkin:
Pool scheduling can now fail independently for errors that are deemed recoverable

Currently "recoverable" errors covers
 - Error from the internal scheduling algo
 - Error from the reconciler
 - Timeouts

Largely we're trying to cover issues that are caused by bugs rather than fundamental "critical" bugs such as our job db failing to upsert our changes (as this implies something is really wrong with our job db).

Unrecoverable errors will still cause all pools to fail and typically covers completely unexpected events where we should just abort

Over time we will likely change what is deemed recoverable - but from experience what is covered now will handle nearly all loop failures - as they're typically caused by bugs due to bad state being passed to the scheduling code that just errors because it gets into an unknown state. 

This feature is disabled by default and is configured with `disableIndependentPoolFailures`

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -19,12 +19,11 @@ pull_request_rules:
       - "#approved-reviews-by>=1"
     actions:
       post_check:
-        neutral_conditions:
-          - and:
-              - "files~=^internal/lookoutui/"
-              - "-files~=^(?!internal/lookoutui/).*$"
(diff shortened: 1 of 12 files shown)


mauriceyap:
@mergifyio refresh

PR #4658, opened 2026-02-04
Add preempt job tests cases

williamvega:
#### What type of PR is this?

test suite improvement

Adds two tests for preempting jobs to the basic test suite.


internal/testsuite/eventwatcher/eventwatcher.go
```diff
@@ -231,15 +231,20 @@ func isTerminalEvent(msg *api.EventMessage) bool {
...
 		return true
+	case *api.EventMessage_Preempted:
+		return true
 	}
 	return false
 }
 
 // ErrorOnNoActiveJobs returns an error if there are no active jobs.
+// Note: The executor intentionally sends both Preempted and Failed events for preempted jobs,
+// so we allow Failed to follow Preempted without erroring.
```
JamesMurkin: These actually come from the scheduler now - but otherwise true

PR #4662, opened 2026-02-05
Feat/preempt button

dslear:
#### What type of PR is this?

Feature

This PR adds a preempt button to the Lookout UI, allowing a user to preempt selected jobs directly from the interface. This is mostly modeled after a combination of the existing reprioritize and cancel buttons to attempt to maintain consistency.

Notably, the preempt endpoint of the Armada API returns an empty response upon success, so I had to handle the response differently

Fixes #

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -19,12 +19,11 @@ pull_request_rules:
       - "#approved-reviews-by>=1"
     actions:
       post_check:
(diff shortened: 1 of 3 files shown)


mauriceyap:
@mergifyio refresh

mauriceyap (approved):

PR #4681, opened 2026-02-09
Add response for job preemption

dslear:
#### What type of PR is this?

Feature

This adds a custom response to the Armada API when a job is preempted. Instead of an empty response, the call will now report back the list of preempted jobs. This PR also updates the preempt dialog in Lookout to reflect these changes and report the preempted job ids back to the user via the snackbar message.

Fixes #

Screenshot 2026-02-11 at 1 50 14 PM

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -62,7 +62,6 @@ jobs:
         uses: golangci/golangci-lint-action@v9
         with:
           version: v2.5.0
-          only-new-issues: true
           skip-cache: true
           args: --timeout=10m --issues-exit-code=1 ./...
 


internal/lookoutui/src/pages/jobs/components/PreemptDialog.tsx
```diff
@@ -72,7 +72,17 @@ export const PreemptDialog = ({ onClose, selectedItemFilters }: PreemptDialogPro
       })
 
       if (response.failedJobIds.length === 0) {
-        openSnackbar("Successfully requested preemption of selected jobs. See table for job statuses.", "success")
+        const ids = response.successfulJobIds
+        const MAX_DISPLAY = 4
```
mauriceyap: nit: please could you move this to the top level of this file? I think `MAX_JOB_IDS_TO_DISPLAY` would be a more descriptive and clearer name

internal/lookoutui/src/pages/jobs/components/PreemptDialog.tsx
mauriceyap, line 92: Please could you add these lines, so that the status table shows the failure reasons if they are present?

```tsx
response.successfulJobIds.forEach((jobId) => {newResponseStatus[jobId] = "Success"})
response.failedJobIds.forEach(({ jobId, errorReason }) => {newResponseStatus[jobId] = errorReason})
```

internal/lookoutui/src/services/lookout/mocks/mockServer.ts
```diff
@@ -152,7 +152,7 @@ export class MockServer {
 
   setPreemptJobsResponse(status_code: number = 200, status_text: string = "") {
     this.server.use(
-      http.post(PREEMPT_JOBS_ENDPOINT, async () => {
+      http.post(PREEMPT_JOBS_ENDPOINT, async (req) => {
```
mauriceyap: This will require some small changes to `PreemptDialog.test.tsx`, but please could you change this so its function signature and logic is similar to `setReprioritizeJobsResponse()` above? That way we can make assertions that the correct error reasons are being displayed in the test.

mauriceyap (approved):

PR #4701, opened 2026-02-19
Fix bug in market driven queue iteration

JamesMurkin:
The CandidateGangIterator assumes that iterators it was passed were ordered
 - Evicted jobs
 - Non-evicted jobs

So if it ever saw a non-evicted job, it could them just drop the whole queue when `OnlyYieldEvicted` was true

However in market based scheduling, that assumption does not hold true and you can get evicted and non-evicted jobs in a mixed ordering

This meant if we ever hit something that caused `OnlyYieldEvicted` when using market based scheduling, you could end up preempting many (or all) evicted jobs, depending on the job ordering - as the CandidateGangIterator stopped considering the queue as soon as it saw a non-evicted job

Now I've pushed that logic down through the iterator stack - as they are better placed to make the right assumptions
 - I've also removed the logic about OnlyYieldEvicted dotted all over CandidateGangIterator and it is now contained to their OnlyYieldEvicted/OnlyYieldEvictedForQueue functions

### DIFF
--- a/cmd/broadside/main.go
+++ b/cmd/broadside/main.go
@@ -1,14 +1,17 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"os"
 	"strings"
 
 	mapstructure "github.com/go-viper/mapstructure/v2"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"golang.org/x/exp/slices"
+	"gopkg.in/yaml.v3"
 
 	"github.com/armadaproject/armada/internal/broadside/configuration"
 	"github.com/armadaproject/armada/internal/broadside/estimation"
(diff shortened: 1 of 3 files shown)


internal/scheduler/scheduling/market_driven_preempting_queue_scheduler_test.go
```diff
@@ -295,6 +293,29 @@ func TestMarketDrivenPreemptingQueueScheduler(t *testing.T) {
 			},
 			PriorityFactorByQueue: map[string]float64{"A": 1, "B": 1},
 		},
+		"reschedules evicted jobs if scheduling limits hit": {
```
JamesMurkin, line 296: This test was what covered the original bug

internal/scheduler/scheduling/market_driven_preempting_queue_scheduler_test.go
```diff
@@ -541,18 +562,6 @@ func TestMarketDrivenPreemptingQueueScheduler(t *testing.T) {
 			jobDb := jobdb.NewJobDb(tc.SchedulingConfig.PriorityClasses, tc.SchedulingConfig.DefaultPriorityClassName, stringinterner.New(1024), testfixtures.TestResourceListFactory)
 			jobDbTxn := jobDb.WriteTxn()
 
-			// Add all the initial jobs, creating runs for them
-			for nodeIdx, jobs := range tc.InitialRunningJobs {
```
JamesMurkin, line 545: Unused code

internal/scheduler/scheduling/jobiteration.go
```diff
@@ -13,11 +13,13 @@ import (
 
 type JobContextIterator interface {
 	Next() (*schedulercontext.JobSchedulingContext, error)
+	OnlyYieldEvicted()
```
mauriceyap, line 16: nit: `SetOnlyYieldEvicted(onlyYieldEvicted bool)` would be nicer

PR #4706, opened 2026-02-23
Suppress slow query logging in Broadside

mauriceyap:
Since the load testing tool is designed to cause queries to be slow as the system is stress-tested, preventing slow queries from being logged reduces unwanted noise during load tests.

### DIFF
--- a/internal/lookout/repository/util.go
+++ b/internal/lookout/repository/util.go
@@ -756,8 +756,9 @@ func logQueryDebug(user string, query *Query, description string) {
 		Debug(description)
 }
 
-func logQueryError(user string, query *Query, description string, duration time.Duration) {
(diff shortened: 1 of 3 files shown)


internal/lookout/repository/util.go
```diff
@@ -764,7 +765,21 @@ func logQueryError(user string, query *Query, description string, duration time.
 		Errorf("Error executing %s query", description)
 }
 
-func logSlowQuery(user string, query *Query, description string, duration time.Duration) {
+// slowQueryLoggingDisabledKey is a context jey used to suppress slow query
```
JamesMurkin:
```suggestion
// slowQueryLoggingDisabledKey is a context key used to suppress slow query
```

PR #4710, opened 2026-02-23
Introduce per queue time limit for scheduling new jobs

JamesMurkin:
We have an existing configuration `NewJobsSchedulingTimeout` which limits how long the scheduler will allow for new job scheduling before it stops allowing new jobs to be scheduled that round

That works well - however what often happens is when the scheduler is slow, it is slow for a single queue

This PR therefore adds the same limit but per queue:
 - This limits impact to a single queue
 - This means the config can be set to a more aggressive value, as it only impacts a single queue when hit

This PR also:
 - Updates how `NewJobsSchedulingTimeout` works to act like other constraints (adding logic to constraints.go)
 - Updates where scheduling duration config lives
   - Move MaxSchedulingDuration -> Scheduling.MaxSchedulingDuration
   - Move NewJobsSchedulingTimeout -> Scheduling.MaxNewJobSchedulingDuration
   - The motivation here is:
     - More consistent naming
     - Scheduling config moved to scheduling section of config
   - For now this will be backwards compatible with warning


internal/scheduler/configuration/validation.go
```diff
@@ -23,6 +32,22 @@ func (c Configuration) Validate() error {
 func SchedulingConfigValidation(sl validator.StructLevel) {
 	c := sl.Current().Interface().(SchedulingConfig)
 
+	// Validate scheduling timeouts
+	if c.MaxNewJobSchedulingDuration > 0 && c.MaxNewJobSchedulingDuration >= c.MaxSchedulingDuration {
```
dejanzele: As we already use the https://github.com/go-playground/validator, can we use struct tags to validate like https://github.com/go-playground/validator/blob/master/_examples/struct-level/main.go#L47

PR #4714, opened 2026-02-25
Improving away node metrics 

JamesMurkin:
Capacity/allocation metrics now include:
 - `physical_pool` which tells you what pool the physical hardware the capacity/allocation belongs to
 - `capacity_class` which tells you if the capacity is coming from dedicated or shared hardware

Also fixed bug where the metrics were wrong for away nodes

### DIFF
--- a/cmd/broadside/main.go
+++ b/cmd/broadside/main.go
@@ -28,6 +28,7 @@ const (
 	customConfigLocation string = "config"
 	resultsDirFlag       string = "results-dir"
 	teardownOnlyFlag     string = "teardown-only"
+	yesFlag              string = "yes"
 	teardownTimeout             = 2 * time.Minute
 )
 
@@ -44,6 +45,11 @@ func init() {
 		teardownOnlyFlag,
 		false,
 		"Connect to the configured database and tear it down, then exit. Use to clean up after a crashed load test.")
+	pflag.BoolP(
+		yesFlag,
+		"y",
+		false,
+		"Skip the confirmation prompt for data generation")
 	pflag.Parse()
 }
 
@@ -197,7 +203,7 @@ func main() {
 	}
 
 	est := estimation.Estimate(config)
-	if estimation.ShouldPrompt(est) {
+	if estimation.ShouldPrompt(est) && !viper.GetBool(yesFlag) {
 		confirmed, err := estimation.DisplayEstimationAndConfirm(est)
 		if err != nil {
 			panic(err)


internal/common/metrics/scheduler_metrics.go
```diff
@@ -183,21 +183,21 @@ var NodeJobPhaseCounterDesc = prometheus.NewDesc(
...
 
 var ClusterFarmCapacityDesc = prometheus.NewDesc(
 	MetricPrefix+"cluster_farm_capacity",
 	"Cluster capacity less usage from non-Armada pods",
-	[]string{"cluster", "pool", "resourceType", "nodeType", "reservation"},
+	[]string{"cluster", "pool", "resourceType", "nodeType", "reservation", "physical_pool", "capacity_class"},
 	nil,
 )
 
 var ClusterAvailableCapacityDesc = prometheus.NewDesc(
```
dejanzele, line 218: nit: add a func comment describing the labels or enhance the description in the second param

internal/common/metrics/scheduler_metrics.go
```diff
@@ -505,32 +505,32 @@ func NewMedianQueueAllocated(value float64, pool string, priorityClass string, q
...
+	return prometheus.MustNewConstMetric(QueueAllocatedDesc, prometheus.GaugeValue, value, cluster, pool, priorityClass, queue, queue, priceBand, resource, nodeType, reservation, physicalPool)
 }
 
 func NewQueueLeasedPodCount(value float64, cluster string, pool string, queue string, phase string, nodeType string, reservation string) prometheus.Metric {
 	return prometheus.MustNewConstMetric(QueueLeasedPodCountDesc, prometheus.GaugeValue, value, cluster, pool, queue, queue, phase, nodeType, reservation)
 }
 
-func NewClusterAvailableCapacity(value float64, cluster string, pool string, resource string, nodeType string, reservation string) prometheus.Metric {
-	return prometheus.MustNewConstMetric(ClusterAvailableCapacityDesc, prometheus.GaugeValue, value, cluster, pool, resource, nodeType, reservation)
+func NewClusterAvailableCapacity(value float64, cluster string, pool string, resource string, nodeType string, reservation string, physicalPool string, capacityClass string) prometheus.Metric {
```
dejanzele, line 537: I am wondering at this point should we make the func accept a struct instead of 8 params

internal/scheduler/internaltypes/node_test.go
```diff
@@ -151,3 +151,101 @@ func TestNode(t *testing.T) {
...
 }
+
+func TestMarkResourceUnallocatable(t *testing.T) {
+	resourceListFactory, err := NewResourceListFactory(
+		[]schedulerconfiguration.ResourceType{
+			{Name: "cpu", Resolution: resource.MustParse("1m")},
+		},
+		nil,
+	)
+	assert.Nil(t, err)
```
dejanzele: This could even be `require.NoError` as no point in proceeding the test if we get an error here, it will fail with nil pointer deref afterwards

internal/scheduler/metrics/constants.go
```diff
@@ -21,6 +21,10 @@ const (
 	priorStateLabel          = "priorState"
 	resourceLabel            = "resource"
 	reservationLabel         = "reservation"
+	schedulable              = "schedulable"
+	overAllocated            = "overAllocated"
+	physicalPool             = "physical_pool"
+	capacityClass            = "capacity_class"
```
dejanzele: Those are labels, right? Should we add a suffix `Label` like we have for the labels below

internal/common/metrics/scheduler_metrics.go
JamesMurkin: comment added

internal/common/metrics/scheduler_metrics.go
JamesMurkin: Probably a good idea - but I think I'd rather do that in a separate PR

PR #4717, opened 2026-02-25
Include all priority classes option for cancellation and preemption

nikola-jokic:
#### What type of PR is this?

Enhancement

Allow users to preempt/cancel jobs on all priority classes


cmd/armadactl/cmd/cancel.go
```diff
@@ -74,8 +74,14 @@ func cancelExecutorCmd() *cobra.Command {
 		Long:  `Cancels jobs on executor with provided executor name, priority classes, queues, and pools.`,
 		Args:  cobra.ExactArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			if err := cmd.MarkFlagRequired("priority-classes"); err != nil {
-				return fmt.Errorf("error marking priority-class flag as required: %s", err)
+			all, err := cmd.Flags().GetBool("all-priority-classes")
```
dejanzele, line 77: I am thinking should we extract `all-priority-classes` into a const as it is repeated 30 times

cmd/armadactl/cmd/cancel.go
```diff
@@ -74,8 +74,14 @@ func cancelExecutorCmd() *cobra.Command {
...
+			if err != nil {
+				return fmt.Errorf("error reading all-priority-classes flag: %s", err)
+			}
+			if !all {
+				if err := cmd.MarkFlagRequired("priority-classes"); err != nil {
+					return fmt.Errorf("error marking priority-class flag as required: %s", err)
+				}
```
dejanzele, line 84: Actually, the block from L77-L84 is repeated, might be worthwhile to extract it into a function and reuse in 12 places

PR #4724, opened 2026-02-26
Make away node types more configurable

JamesMurkin:
An away node type is now:
 - Made up of a list of well known node types
 - Each well known node type can also have conditions that must be met for it to take effect

This allows us to add gpu node taints to cpu jobs conditionally but not inadvertently add all gpu node taints to gpu jobs

### DIFF
--- a/config/scheduler/config.yaml
+++ b/config/scheduler/config.yaml
@@ -1,6 +1,5 @@
 cyclePeriod: 1s
 schedulePeriod: 10s
-maxSchedulingDuration: 5s
 executorTimeout: 1h
 databaseFetchSize: 1000
 pulsarSendTimeout: 5s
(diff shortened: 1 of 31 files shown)


internal/scheduler/nodedb/nodedb_test.go
```diff
@@ -672,6 +673,153 @@ func TestHomeNodeScheduling(t *testing.T) {
...
+			conditions:   []types.AwayNodeTypeCondition{{Resource: "cpu", Operator: types.AwayNodeTypeConditionOpEqual, Value: cpu2}},
+			jobResources: testfixtures.Cpu("4"),
+			expectMatch:  false,
+		},
+		"eq: job value below threshold does not match": {
+			conditions:   []types.AwayNodeTypeCondition{{Resource: "cpu", Operator: types.AwayNodeTypeConditionOpEqual, Value: cpu4}},
+			jobResources: testfixtures.Cpu("2"),
+			expectMatch:  false,
+		},
+		// Missing resource treated as zero
```
dejanzele: redundant comment as the test case name mentions the same

PR #4733, opened 2026-03-02
fix: handle requeued gang members in eviction and assertion logic

dejanzele:
#### What type of PR is this?

Bug fix

Fixes a gang eviction assertion crash (`gang was partially evicted: 2 out of 3 jobs evicted`) that occurs when a gang has a member that was returned by the executor and requeued.

**Root cause:** When the executor returns a failed gang job, the scheduler requeues it with `WithQueued(true)`, but `LatestRun().NodeId()` still points to the old node. When oversubscribed eviction later targets a running gang member, `collectIdsForGangEviction` reads the stale node ID from the requeued job, the filtered evictor looks at the wrong node, gang eviction is incomplete, and `evictionAssertions` panics because it expects all gang members to be evicted but the requeued member was never on the node the evictor checked.

**Changes in `preempting_queue_scheduler.go`:**

1. `collectIdsForGangEviction`: Check the scheduling context first for a node ID (handles current-round rescheduling). If the job is not in the scheduling context and is queued, skip node collection entirely (the job is not on any node). Fall back to `LatestRun().NodeId()` only for running jobs.

2. `evictionAssertions`: Exclude queued-and-not-evicted gang members from the expected eviction count. These members are not on any node and cannot be evicted, so they should not contribute to the total the assertion checks against.

-

The crash is deterministic when a gang has a mix of running and requeued members and oversubscribed eviction is triggered.

Before this fix, the node-lookup order was: `LatestRun() == nil` -> scheduling context, else -> `LatestRun().NodeId()`. This missed the case where `LatestRun()` is non-nil but stale (requeued job).

The new order is: scheduling context first -> queued with no context (skip node) -> `LatestRun().NodeId()` for running jobs -> error for anything else.

### DIFF
--- a/cmd/broadside/configs/examples/test-postgres.yaml
+++ b/cmd/broadside/configs/examples/test-postgres.yaml
@@ -9,6 +9,14 @@ databaseConfig:
     user: broadside
     password: broadside
     sslmode: disable
+  # Optional: Postgres tuning SQL statements applied after schema migrations.
+  postgresTuningSQL:
+    - "ALTER TABLE job SET (autovacuum_vacuum_scale_factor = 0.01, autovacuum_analyze_scale_factor = 0.005, autovacuum_vacuum_threshold = 1000, autovacuum_analyze_threshold = 500)"
+    - "ALTER TABLE job_run SET (autovacuum_vacuum_scale_factor = 0.02, autovacuum_analyze_scale_factor = 0.01)"
+  # Optional: Postgres revert SQL statements executed at the start of TearDown.
+  postgresTuningRevertSQL:
+    - "ALTER TABLE job RESET (autovacuum_vacuum_scale_factor, autovacuum_analyze_scale_factor, autovacuum_vacuum_threshold, autovacuum_analyze_threshold)"
+    - "ALTER TABLE job_run RESET (autovacuum_vacuum_scale_factor, autovacuum_analyze_scale_factor)"
 
 queueConfig:
   - name: queue-1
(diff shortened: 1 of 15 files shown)


internal/scheduler/scheduling/preempting_queue_scheduler.go
```diff
@@ -387,27 +387,31 @@ func (sch *PreemptingQueueScheduler) collectIdsForGangEviction(evictorResult *Ev
...
 		for _, gangJob := range activeGangJobs {
 			allGangJobIds[gangJob.Id()] = true
-			// If the latest run is nil, it means the job was scheduled in the current scheduling round
-			// So we need to find which node it was scheduled on in the scheduling context as it won't have been persisted in the jobdb yet
-			if gangJob.LatestRun() == nil {
-				nodeId, err := getNodeIdFromSchedulingContext(sch.schedulingContext, gangJob)
-				if err != nil {
-					return nil, nil, err
+
+			if gangJob.Queued() {
```
JamesMurkin, line 401: Does this not break urgency based preemption of gangs scheduled in the current cycle?

dejanzele: I don't think so.
Jobs scheduled in the current round have `Queued() == true` in jobdb (not yet updated), so they enter the `Queued()` branch, where the scheduling context lookup runs before continue. Same lookup as master [L403-408](https://github.com/armadaproject/armada/blob/99c8995884c7c6328c315036d5beb65423d90e90/internal/scheduler/scheduling/preempting_queue_scheduler.go#L403-L408), same result.

JamesMurkin: I read it wrong, you're correct

PR #4734, opened 2026-03-02
LookoutUI: Add analytics

d2burkhalter:
#### What type of PR is this?
This PR adds the ability to add website analytics to LookoutUI in a way that is provider agnostic.

This allows for insights on how LookoutUI is being interacted with by users.
Tried to make the implementation as generic as possible for providers and tested using Umami, Plausible and OpenPanel. 
Adds a new optional configuration `uiConfig.analytics`  to load analytics scripts into LookoutUI for page, event and user analytics.
Creates a new `<Analytics>` component that can wrap any HTML element or React component and inject properties for analytics.


internal/lookout/configuration/types.go
```diff
@@ -144,6 +144,24 @@ type JobLinkConfig struct {
...
+
+type Analytics struct {
+	Scripts        []ScriptTag   `json:"scripts"`
+	Method         string        `json:"method,omitempty"`
+	EventAttribute string        `json:"eventAttribute,omitempty"`
+	DataAttribute  string        `json:"dataAttribute,omitempty"`
+	UserIdentify   *UserIdentify `json:"userIdentify,omitempty"`
+}
+
 // UIConfig must match the LookoutUiConfig TypeScript interface defined in internal/lookoutui/src/lookoutUiConfig.d.ts
```
mauriceyap: This is my mistake from 7 months ago!
```suggestion
// UIConfig must match the LookoutUiConfig TypeScript interface defined in internal/lookoutui/src/config/types.ts
```

internal/lookout/configuration/types.go
```diff
@@ -144,6 +144,24 @@ type JobLinkConfig struct {
...
+	Content    string            `json:"content,omitempty"`
+	Attributes map[string]string `json:"attributes,omitempty"`
+}
+
+type UserIdentify struct {
+	Provider      string `json:"provider"`
+	IdentifyParam string `json:"identifyParam,omitempty"`
+}
+
+type Analytics struct {
```
mauriceyap: nit
```suggestion
type AnalyticsConfig struct {
```

internal/lookout/configuration/types.go
```diff
@@ -177,4 +195,7 @@ type UIConfig struct {
 	JobLinks []JobLinkConfig `json:"jobLinks"`
 
 	CustomThemeConfigs *CustomThemeConfigs `json:"customThemeConfigs,omitempty"`
+
+	// Analytics is an optional analytics configuration
+	Analytics *Analytics `json:"analytics,omitempty"`
```
mauriceyap:
```suggestion
	Analytics *AnalyticsConfig `json:"analytics,omitempty"`
```

internal/lookoutui/src/config/types.ts
```diff
@@ -40,6 +40,22 @@ export interface JobLinkConfig {
   linkTemplate: string
 }
 
+export interface ScriptTag {
+  content?: string
+  attributes?: Record<string, string>
+}
+
+export interface analyticsConfig {
```
mauriceyap:
```suggestion
export interface AnalyticsConfig {
```

We try to keep types in UpperCamelCase to make them easy to identify ([Google's style guide also prescribes this](https://google.github.io/styleguide/tsguide.html#naming-rules-by-identifier-type))

internal/lookoutui/src/config/types.ts
```diff
@@ -40,6 +40,22 @@ export interface JobLinkConfig {
...
+  scripts: ScriptTag[]
+  method?: "attribute" | "class"
+  eventAttribute?: string
+  dataAttribute?: string
+  userIdentify?: {
+    provider: string
+    identifyParam?: string
+  }
+}
+
```
mauriceyap: For consistency with the Go type, please could you extract this out into a `UserIdentify` interface? (I appreciate this pattern isn't consistently applied in this code base)

internal/lookoutui/src/analytics/AnalyticsScript.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+interface analyticsProps {
+  config: analyticsConfig | undefined
+}
+
+/**
+ * Component that dynamically injects analytics scripts into the document head
+ * based on the configuration provided.
+ */
+export const AnalyticsScript = ({ config }: analyticsProps) => {
```
mauriceyap: This _looks_ like an XSS risk to the untrained eye as it's injecting script elements into `document.head`, but it isn't as the content and attributes are controlled by the server. Please could you add a brief comment here explaining this?

internal/lookoutui/src/analytics/AnalyticsScript.tsx
```diff
@@ -0,0 +1,49 @@
+import { useEffect } from "react"
+
+import { analyticsConfig } from "../config/types"
+
+interface analyticsProps {
```
mauriceyap:
```suggestion
interface AnalyticsProps {
```

internal/lookoutui/src/pages/settings/visualTheme/components/ColourModeSelector.tsx
mauriceyap: If it's straightforward, please could you also wrap this in an `<Analytics eventName="System color mode selected">`?

internal/lookoutui/src/analytics/AnalyticsUserIdentifier.tsx
```diff
@@ -0,0 +1,65 @@
+import { useEffect, useRef } from "react"
+
+import { analyticsConfig } from "../config"
+import { useUserManager } from "../oidcAuth"
+
+export interface AnalyticsUserIdentifierProps {
+  analyticsConfig: analyticsConfig | undefined
+}
```
mauriceyap: ultranit: please could you move this down, so that it's immediately above `AnalyticsUserIdentifier`? It's helpful to the reader to be able to quickly see the prop types of the component they are looking at.

internal/lookoutui/src/analytics/Analytics.tsx
```diff
@@ -0,0 +1,86 @@
+import { ComponentPropsWithRef, ElementType, forwardRef, ReactNode, JSX } from "react"
+
+import { getConfig, analyticsConfig } from "../config"
+
+type AnalyticsProps<C extends ElementType> = {
+  component: C
+  eventName: string
```
mauriceyap: I think I'd prefer for the `eventName` to be typed rather than a free-form string. This would avoid typos and give us a complete view of all possible events. Something like this in `events.ts` in this directory:

```tsx
export const ANALYTICS_EVENTS = {
  CANCEL_JOBS_CLICKED: "Cancel Jobs Clicked",
  PREEMPT_JOBS_CLICKED: "Preempt Jobs Clicked",
  REPRIORITIZE_JOBS_CLICKED: "Reprioritize Jobs Clicked",
  SIDEBAR_TAB_VIEW: "Sidebar Tab View",
  DOWNLOAD_LOGS_CLICKED: "Download Logs Clicked",
  MARK_NODE_UNSCHEDULABLE_CONFIRMED: "Mark Node Unschedulable Confirmed",
  LIGHT_MODE_SELECTED: "Light Mode Selected",
  DARK_MODE_SELECTED: "Dark Mode Selected",
} as const

export type AnalyticsEventName = (typeof ANALYTICS_EVENTS)[keyof typeof ANALYTICS_EVENTS]
```

Then we can do `<Analytics eventName={ANALYTICS_EVENTS.CANCEL_JOBS_CLICKED} ... />`

internal/lookoutui/src/analytics/Analytics.tsx
```diff
@@ -0,0 +1,86 @@
...
+      analyticsAttributes[analyticsConfig.eventAttribute] = eventName
+    }
+    if (analyticsConfig.dataAttribute && eventData) {
+      Object.entries(eventData).forEach(([key, value]) => {
+        analyticsAttributes[`${analyticsConfig.dataAttribute}-${key.toLowerCase()}`] = value
+      })
+    }
+
+    return analyticsAttributes
+  }
```
mauriceyap: nit: remove the `else` block

internal/lookoutui/src/analytics/AnalyticsUserIdentifier.tsx
```diff
@@ -0,0 +1,65 @@
...
+    if (typeof window !== "undefined" && provider in window) {
+      const analytics = (window as any)[provider]
+      if (typeof analytics === "object" && typeof analytics.identify === "function") {
+        return analytics
+      }
+    }
+    await new Promise((resolve) => setTimeout(resolve, retryDelay))
+  }
+  return undefined
+}
```
mauriceyap, line 9: I think there's a cleaner way to do this rather than retrying - as it's the `<AnalyticsScript>` component which loads the scripts, you could get that to pass a signal to this `<AnalyticsUserIdentifier>` component when it has finished loading all scripts.

The way I'd do it is this:

Add an `onReady: () => void` prop to `AnalyticsScript` component and call it when all scripts have finished loading (or failed):
```tsx
  useEffect(() => {
    if (!config?.scripts || config.scripts.length === 0) {
      onReady() // add this
      return
    }

    // ...etc...

    let cancelled = false

    const loadPromises = scriptElements
      .filter((s) => s.src)
      .map((script) => new Promise((resolve) => script.addEventListener("load", resolve, { once: true })))

    Promise.allSettled(loadPromises).then(() => {
      if (!cancelled) {
        onReady()
      }
    })

    // Cleanup function to remove scripts when component unmounts
    return () => {
      cancelled = true // add this

      scriptElements.forEach((script) => {
        if (script.parentNode) {
          document.head.removeChild(script)
        }
      })
    }
  }, [config, onReady])
```

`Promise.allSettled()` rather than `Promise.all()` in case some of the scripts fail to load.

In `<App>`:
```tsx
const [analyticsReady, setAnalyticsReady] = useState(false)

<AnalyticsScript
  config={config.analytics}
  onReady={() => setAnalyticsReady(true)}
/>

<AnalyticsUserIdentifier
  analyticsConfig={config.analytics}
  providerReady={analyticsReady}
/>
```

Finally here in `AnalyticsUserIdentifier`, add a `providerReady: boolean` prop and the start of the `useEffect()` hook would look like this:
```tsx
useEffect(() => {
  if (!providerReady || hasIdentified.current || !analyticsConfig?.userIdentify || !userManager) {
    return
  }

  // ...etc...
}, [providerReady, analyticsConfig, userManager])
```

internal/lookoutui/src/analytics/AnalyticsScript.tsx
d2burkhalter: updated the documentation so it should be more clear

internal/lookoutui/src/analytics/Analytics.tsx
d2burkhalter: done, changed the filename to `types.ts` since my browser was blocking `events.ts`

internal/lookoutui/src/analytics/AnalyticsUserIdentifier.tsx
d2burkhalter: agree this is a much cleaner, more React way to solve this problem

internal/lookoutui/src/pages/settings/visualTheme/components/ColourModeSelector.tsx
d2burkhalter: Umami didn't automatically detect the MUI Switch element interaction so I made the `Analytics` component send a custom event to the analytics provider when clicked. This should ensure that all providers have consistent behavior.

internal/lookoutui/src/analytics/Analytics.tsx
d2burkhalter: rewrote this section so we are no longer trying to modify html attributes and css classes for analytics

internal/lookoutui/src/pages/jobs/components/sidebar/Sidebar.tsx
```diff
@@ -249,17 +250,51 @@ export const Sidebar = memo(({ job, sidebarWidth, onClose, onWidthChange, comman
...
+                    />
+                    <Analytics
+                      component={StyledSidebarTab}
                       label="Commands"
                       value={SidebarTab.Commands}
                       disabled={job.state === JobState.Queued}
+                      eventName={ANALYTICS_EVENTS.SIDEBAR_TAB_CLICKED}
+                      eventData={{ tab: "Commands", jobState: job.state }}
                     />
                   </SidebarTabs>
```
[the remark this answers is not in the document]
d2burkhalter: https://mui.com/material-ui/react-tabs/#keyboard-navigation default MUI behavior requires tabs to be selected manually which is detected by the Analytics wrapper

internal/lookoutui/src/analytics/Analytics.tsx
```diff
@@ -0,0 +1,82 @@
+import { ComponentPropsWithRef, ElementType, forwardRef, ReactNode, JSX } from "react"
+
+import { getConfig } from "../config"
+
+const config = getConfig()
+
+type AnalyticsProps<C extends ElementType> = {
+  component: C
+  eventName: string
```
mauriceyap:
```suggestion
  eventName: AnalyticsEventName
```

d2burkhalter: good catch

PR #4743, opened 2026-03-06
Add redis metrics giving better view into redis 

nikola-jokic:


internal/server/submit/validation/submit_request.go
```diff
@@ -253,7 +253,7 @@ func validateResources(j *api.JobSubmitRequestItem, config configuration.Submiss
 		maxOversubscriptionByResource = map[string]float64{}
 	}
 	for _, container := range armadaslices.Concatenate(spec.Containers, spec.InitContainers) {
-		if len(container.Resources.Requests) == 0 && len(container.Resources.Requests) == 0 {
+		if len(container.Resources.Requests) == 0 {
```
JamesMurkin: I think this is a bug but probably it was meant to say `&& len(container.Resources.Limits) == 0`

internal/server/redismetrics/collector.go
```diff
@@ -0,0 +1,334 @@
...
+			prometheus.GaugeOpts{
+				Name: "armada_redis_stream_age_seconds",
+				Help: "Age of top-N oldest Redis streams in seconds",
+			},
+			[]string{"queue", "jobset"},
+		),
+		bytesHistogram: prometheus.NewHistogram(prometheus.HistogramOpts{
+			Name:    "armada_redis_stream_size_bytes_distribution",
+			Help:    "Distribution of Redis stream sizes in bytes",
+			Buckets: prometheus.ExponentialBuckets(1024, 2, 20),
```
JamesMurkin: These buckets are too small

internal/server/redismetrics/collector.go
```diff
@@ -0,0 +1,334 @@
...
+		topNMemoryGauge: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "armada_redis_stream_memory_bytes",
+				Help: "Memory usage of top-N Redis streams by bytes",
+			},
+			[]string{"queue", "jobset"},
+		),
+		topNEventsGauge: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "armada_redis_stream_event_count",
```
JamesMurkin: It'd be good to use a constant prefix - as we do everywhere else

Possibly it should be `armada_event_redis` although I'm not 100%
 - I can envision we have more redis in future
 - So either we'd need the name to distinguish the difference or we'd need another label (instance/purpose something like that)

config/server/config.yaml
```diff
@@ -25,6 +25,25 @@ eventsApiRedis:
...
   db: 1
   poolSize: 1000
+redisMemoryMetrics:
+  enabled: false
+  collectionInterval: 5m
+  topN: 100
+  scanBatchSize: 1000
+  pipelineBatchSize: 500
+  interBatchDelay: 100ms
+  memoryUsageSamples: 5
```
JamesMurkin: Personally I'd suggest:

collectionInterval -> 15m
memoryUsageSamples -> 25

config/server/config.yaml
```diff
@@ -25,6 +25,25 @@ eventsApiRedis:
...
+  # metricsRedis:  # Optional: separate Redis for metrics collection
```
JamesMurkin: Does this redis need metrics in its name? It is already under the metrics section

config/server/config.yaml
```diff
@@ -25,6 +25,25 @@ eventsApiRedis:
   password: ""
   db: 1
   poolSize: 1000
+redisMemoryMetrics:
```
JamesMurkin: Could we make a metrics section and instead have:

```
metrics:
  redis:
   <section>
```

internal/server/redismetrics/scanner.go
```diff
@@ -0,0 +1,217 @@
...
+	for {
+		// Check for cancellation
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
+
+		// Scan for keys matching the Events: pattern
+		cmd := s.client.Scan(ctx, cursor, "Events:*", s.config.ScanBatchSize)
```
JamesMurkin: Could we use shared constant with other Events code

internal/server/redismetrics/scanner.go
```diff
@@ -0,0 +1,217 @@
...
+
+	return results, nil
+}
+
+// processBatch processes a batch of keys using pipelined commands.
+func (s *Scanner) processBatch(ctx context.Context, keys []string) ([]StreamInfo, error) {
+	var results []StreamInfo
+
+	// Process in pipeline batches
+	for i := 0; i < len(keys); i += s.config.PipelineBatchSize {
```
JamesMurkin: I think we have a utility for batching up values in this way

internal/server/redismetrics/collector.go
JamesMurkin: Discussed this

Make byte and events 30 buckets, and age 20 buckets

internal/eventingester/metrics/redis/collector.go
```diff
@@ -0,0 +1,358 @@
...
+		c.topNEventsGauge.WithLabelValues(byEvents[i].Queue, byEvents[i].JobSetId).Set(float64(byEvents[i].Length))
+		c.topNAgeGauge.WithLabelValues(byAge[i].Queue, byAge[i].JobSetId).Set(byAge[i].AgeSeconds)
+	}
+
+	// Update histograms
+	for _, s := range streams {
+		c.bytesHistogram.Observe(float64(s.MemoryBytes))
+		c.eventsHistogram.Observe(float64(s.Length))
+		c.ageHistogram.Observe(s.AgeSeconds)
+	}
```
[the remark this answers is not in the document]
nikola-jokic: @JamesMurkin I think this is a good point maybe, should we swap to gauge?

internal/eventingester/redismetrics/collector.go
```diff
@@ -0,0 +1,411 @@
...
+	return collector
+}
+
+// Run starts the background collection loop.
+func (c *Collector) Run(ctx *armadacontext.Context) error {
+	ctx.Infof("Will update Redis metrics every %s", c.config.CollectionInterval)
+
+	// Delay to guard against crash loops during startup and to prevent thundering herd on leadership changes
+	// The delay is [0, 1 minute) to ensure that in the worst case, all collectors will be staggered by at least 1 minute.]
+	initialDelay := time.Duration(rand.Int64N(int64(1 * time.Minute)))
```
nikola-jokic: @JamesMurkin here is the initial delay as random interval up to a minute. I think this might be fine, but we can change that

PR #4745, opened 2026-03-06
Wire error categories into executor event reporting

dejanzele:
## What type of PR is this?

Feature (PR 2 of 4)

## What this PR does / why we need it

Wires the classifier and FailureInfo into the executor's event reporting path.

- Constructs the `Classifier` from config at executor startup unconditionally (with no rules configured, it simply returns empty categories)
- Passes classifier to the pod issue handler and job state reporter
- Calls `ExtractFailureInfo()` + `classifier.Classify()` on every pod failure, attaching structured `FailureInfo` (exit code, termination message, categories, container name) to the `Error` events sent through Pulsar
- After this PR, every failed pod event flowing into Pulsar carries exit code, termination message, container name, and matched category names

## Which issue(s) this PR fixes

Part of #4713 (Error Categorization)

## Special notes for your reviewer

- This is PR 2 of 4: Proto + classifier (#4741) -> Wire into executor (this) -> Store in lookout DB + UI (#4755) -> e2e tests (#4760)
- Depends on #4741
- The classifier is always instantiated at startup, even with no rules configured - no conditional nil-checking needed at call sites
- Changes are in `application.go` (startup wiring), `pod_issue_handler.go`, `job_state_reporter.go`, `reporter/event.go`, `cluster_allocation.go`, and `preempt_runs.go`

Issue #4713, opened 2026-02-25
Error Categorization


dejanzele:
## Motivation

Armada has multiple mechanisms for handling job failures (executor pod checks, lease loss retries), but they don't share a consistent classification of *why* a job failed. The raw signals (exit code 74, termination message "CUDA error: GPU has fallen off the bus") flow through the system but carry no semantic meaning.

We want a single classification that serves **observability**. Instead of every consumer needing to understand raw k8s signals, the executor classifies failures into named categories that flow through the whole system.

### Classify failures at the executor

The executor already has the richest failure data: full pod status, Kubernetes events, container states. Instead of passing all k8s events and conditions up to the scheduler for it to make sense of, the executor classifies the failure into a named category and attaches it to the job run as a label.

Matching rules (exit codes, termination messages, conditions) run at the executor to set a "termination reason" category on the job. This can then flow through the whole system without every component needing to understand raw k8s signals.

Categories can live locally in the executor cluster. A GPU cluster defines `cuda_error` and `infiniband_error`; a CPU cluster doesn't need them.

### Proto schema

A `FailureInfo` message on the `Error` proto carries structured failure metadata from the executor through Pulsar to Lookout:

- `exit_code` - exit code of the first failed container
- `termination_message` - termination message from the first failed container
- `categories` - executor-assigned category labels matched from config
- `container_name` - which container the exit code and termination message were extracted from

### Categories

Configuration in executor, a category is just a name plus matching rules:

```yaml
errorCategories:
  - name: cuda_error
    rules:
      - onTerminationMessage: { pattern: ".*(CUDA error|Xid error).*" }
      - onExitCodes: { operator: In, values: [74, 75] }

  - name: infiniband_error
    rules:
      - onExitCodes: { operator: In, values: [42, 43] }
      - onTerminationMessage: { pattern: ".*InfiniBand.*link down.*" }

  - name: oom
    rules:
      - onConditions: [OOMKilled]
      - onExitCodes: { operator: In, values: [137] }

  - name: sidecar_crash
    rules:
      - containerName: istio-proxy
        onExitCodes: { operator: NotIn, values: [0] }
```

Rules within a category are OR'd. If exit code equals XX **or** termination log has YY, the category matches.

Each rule supports an optional `containerName` field. When set, the rule only matches failures from that specific container. When empty (the default), failures from any container can match. This is useful for multi-container pods where sidecars and init containers may fail independently and you want distinct categories for them.

### Data flow

[...]

PR #4755, opened 2026-03-09
Store FailureInfo in lookout DB and expose via API

dejanzele:
## What type of PR is this?

Feature (PR 3 of 4)

## What this PR does / why we need it

Stores FailureInfo in the lookout database, exposes it via the Lookout API and public event API, and displays it in the Lookout UI.

- Adds DB migration (031) to add a `failure_info` JSONB column to `job_run`
- Extracts `FailureInfo` from Error events in the lookout ingester and stores as JSONB (exit code, termination message, categories, container name)
- Adds `failureInfo` to the Lookout v2 Swagger spec and threads it through the query builder, model, and conversions layer
- Displays FailureInfo in the Lookout UI sidebar on failed job runs: container name, exit code, termination message, and matched categories
- Adds Error Categories column to the jobs table (display-only, filtering is not yet supported since categories are stored in JSONB and not a server-side filterable field)
- Propagates error categories in the public `JobFailedEvent` proto, so categories are available in the event stream (used by the e2e testsuite in PR #4760)
- Adds executor error category config to `_local/executor/config.yaml` for local development

## Which issue(s) this PR fixes

Part of #4713 (Error Categorization)

## Special notes for your reviewer

- This is PR 3 of 4: Proto + classifier (#4741) -> Wire into executor (#4745) -> Store in lookout DB + UI + public event (this) -> e2e tests (#4760)
- Depends on #4745
- JSONB round-trip means `exitCode` (int32 in proto) arrives as float64 after PostgreSQL JSON deserialization - `failureInfoToSwagger()` handles this via type assertion
- `failureInfoToMap()` only emits non-zero fields to keep the JSONB clean
- Uses `coalesce(tmp.failure_info, job_run.failure_info)` in the batch upsert so earlier partial updates aren't overwritten
- Only terminal errors have their FailureInfo persisted (`if e.Terminal` guard in the ingester) to prevent transient failures from overwriting the final classification
- Categories on `JobFailedEvent` are set in `FromInternalJobErrors()` conversion from `FailureInfo.Categories`

image
image


internal/lookout/conversions/convert.go
```diff
@@ -122,6 +123,35 @@ func PostgreSQLTimeToSwaggerTime(t *model.PostgreSQLTime) *strfmt.DateTime {
...
+	if msg, ok := failureInfo["terminationMessage"].(string); ok {
+		result.TerminationMessage = msg
+	}
+	if categories, ok := failureInfo["categories"].([]interface{}); ok {
+		for _, c := range categories {
+			if s, ok := c.(string); ok {
+				result.Categories = append(result.Categories, s)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
dejanzele: nice one!

internal/lookoutingester/instructions/instructions.go
```diff
@@ -469,6 +472,26 @@ func (c *InstructionConverter) handleStandaloneIngressInfo(event *armadaevents.S
...
+		m["exitCode"] = fi.GetExitCode()
+	}
+	if fi.GetTerminationMessage() != "" {
+		m["terminationMessage"] = fi.GetTerminationMessage()
+	}
+	if len(fi.GetCategories()) > 0 {
+		m["categories"] = fi.GetCategories()
+	}
+	return m
+}
```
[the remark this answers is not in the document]
dejanzele: nice!

internal/lookout/conversions/convert.go
```diff
@@ -122,6 +123,34 @@ func PostgreSQLTimeToSwaggerTime(t *model.PostgreSQLTime) *strfmt.DateTime {
...
+	if cats, ok := failureInfo["categories"].([]interface{}); ok {
+		result.Categories = make([]string, 0, len(cats))
+		for _, c := range cats {
+			if s, ok := c.(string); ok {
+				result.Categories = append(result.Categories, s)
+			}
+		}
+	}
+	return result
```
mauriceyap, line 159: In the case where `failureInfo` is a map with no relevant key-value pairs with (e.g. `map[string]any{"foo": -1, "exitCode": "I am not a float64"}`), I think this should return `nil` rather than an empty `*models.RunFailureInfo`. What do you think about this?

If so, the logic would look like this:
```go
populated := false

if v, ok := failureInfo["exitCode"].(float64); ok {
	result.ExitCode = int32(v)
	populated = true
}
if msg, ok := failureInfo["terminationMessage"].(string); ok {
	result.TerminationMessage = msg
	populated = true
}
if cats, ok := failureInfo["categories"].([]interface{}); ok {
	result.Categories = make([]string, 0, len(cats))
	populated = true
	for _, c := range cats {
		if s, ok := c.(string); ok {
			result.Categories = append(result.Categories, s)
		}
	}
}

if !populated {
	return nil
}

return result
```

internal/lookout/schema/migrations/031_add_failure_info_to_job_run.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE job_run ADD COLUMN failure_info JSONB NULL;
```
mauriceyap: nit: the `NULL` is redundant
```suggestion
ALTER TABLE job_run ADD COLUMN failure_info JSONB;
```

internal/lookout/conversions/convert.go
dejanzele: nice, I'll implement!

internal/lookout/schema/migrations/031_add_failure_info_to_job_run.sql
dejanzele: nice catch!

PR #4759, opened 2026-03-10
Rust client release/publish

mladjan-gadzic:
#### What type of PR is this?
CI changes for Rust publishing.

Configures CI for publishing.

Fixes #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -168,22 +168,43 @@ jobs:
     runs-on: ubuntu-22.04
     environment: crates-release
     steps:
+      - name: Checkout at release tag
+        uses: actions/checkout@v6
+        with:
+          ref: ${{ github.event.workflow_run.head_sha }}
+
       - name: Install Rust (stable)
         uses: dtolnay/rust-toolchain@stable
 
-      - name: Download crate artifact
-        run: gh run download ${{ github.event.workflow_run.id }} --repo ${{ github.event.workflow_run.repository.full_name }} --name crate-artifact --dir ./crate
+      - name: Set crate version from tag
         env:
-          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ github.event.workflow_run.head_branch }}
+        run: |
+          VERSION=$(echo "${TAG#v}" | sed 's/-patch[0-9]*$//')
+          if [[ ! "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "::error::Derived VERSION '$VERSION' is not a valid semver string"
+            exit 1
+          fi
+          sed -i "s/^version = \".*\"/version = \"$VERSION\"/" client/rust/Cargo.toml
+          grep -q "^version = \"$VERSION\"" client/rust/Cargo.toml || {
+            echo "::error::sed did not update version in Cargo.toml — pattern may not have matched"
+            exit 1
+          }
+          cargo generate-lockfile --manifest-path client/rust/Cargo.toml
+
+      - name: Vendor Armada protos
+        run: |
+          mkdir -p client/rust/proto/pkg/api
+          cp pkg/api/submit.proto pkg/api/event.proto pkg/api/job.proto pkg/api/health.proto \
+            client/rust/proto/pkg/api/
 
       - name: Publish to crates.io
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -161,3 +161,65 @@ jobs:
         run: |
           VERSION=${TAG#v}
           dotnet nuget push ./bin/client/DotNet/ArmadaProject.Io.Client.$VERSION.nupkg -k ${{ secrets.NUGET_API_KEY }} -s https://api.nuget.org/v3/index.json
+
+  push-crate:
+    name: Push crates.io client
+    needs: validate
+    runs-on: ubuntu-22.04
```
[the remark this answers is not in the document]
mladjan-gadzic: Intentionally depends only on validate, not on release.

client/rust/build.rs
```diff
@@ -28,6 +43,38 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        (
+            vec![
+                "proto/pkg/api/submit.proto",
+                "proto/pkg/api/event.proto",
+                "proto/pkg/api/job.proto",
+                "proto/pkg/api/health.proto",
+            ],
+            vec!["proto"],
+        )
+    };
```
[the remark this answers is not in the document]
mladjan-gadzic: The actual fix is reordering to ["proto", "../../"], ensuring google/ and k8s.io/ always resolve from the vendored proto/ copies first.

client/rust/LICENSE
jgiannuzzi, line 1: Is there a way to use the top-level license file instead? Maybe via symlink?

.github/workflows/release.yml
```diff
@@ -161,3 +161,66 @@ jobs:
...
+    environment: crates-release
+    steps:
+      - name: Checkout
```
jgiannuzzi: Why do we rebuild the crate entirely instead of using the CI artefact like we do for .NET?

.github/workflows/rust-client.yml
jgiannuzzi, line 1: The 2 jobs are almost identical apart from the Rust toolchain version. Consider using a matrix instead.

client/rust/LICENSE
mladjan-gadzic: thought about symlink, but that is not good the best approach for windows users as per this https://github.com/armadaproject/armada/pull/4759#discussion_r2917473079

jgiannuzzi: how about simply setting `license = "Apache-2.0"` on the package instead of packaging the file itself?

jgiannuzzi: see https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields

.github/workflows/release.yml
mladjan-gadzic: unlike .net (which produces a versioned binary .nupkg in CI), cargo publish uploads a source tarball. the version is stamped at release time after the tag is known (sed updates Cargo.toml), so the package can't be pre-built in CI with the correct version. the CI cargo package steps are verification only.

client/rust/LICENSE
mladjan-gadzic: > how about simply setting license = "Apache-2.0" on the package instead of packaging the file itself?

that was already done, it's just that i wanted file itself. will do only license property.

.github/workflows/release.yml
jgiannuzzi: but the tag is known when the CI runs for the build, so it could set the version in `Cargo.toml`

mladjan-gadzic: i've tried to replicate .net behavior now. there are differences between other clients. rust client is inlined with .net.

jgiannuzzi: ah I got confused between `build.yml` and `rust-client.yml`! I thought we could have the package produced by the same jobs that already run the tests, etc. in `rust-client.yml`.

basically I'd like to avoid duplication _if we can_ - which solution do you prefer?

mladjan-gadzic: the reason we can't easily consolidate into rust-client.yml is that release.yml triggers on the CI workflow and downloads artifacts using ${{ github.event.workflow_run.id }} (the CI run id). if packaging were in rust-client.yml (a separate workflow), the release job would have no reliable way to reference its run id.                                                                                                                                                                                
                                                                                                                                                                                       
so i'd prefer keeping pack-rust-crate in build.yml - it's the exact same pattern as pack-nuget and avoids the timing/referencing problem. the setup duplication between build.yml and rust-client.yml is the tradeoff, but i think the architectural consistency is worth it.

Issue #4683, opened 2026-02-10
Native support for retry policies


dejanzele:
## The Problem

When a job fails due to infrastructure reasons (preemption, node eviction, OOMKill), Armada permanently fails the job. Users must manually resubmit, even though the failure wasn't their fault.

The existing retry logic only retries jobs in narrow cases:
- Executor heartbeat timeout
- Executor explicitly returns the lease
- Armada preemption marks the run as "Returned"

Any other failure, including OOMKill, node eviction, or application-level transient errors, results in permanent failure.

```go
// Current logic: only "Returned" runs can retry
requeueJob = !failFast && lastRun.Returned() && job.NumAttempts() < s.maxAttemptedRuns
```

## The Solution

A configurable retry policy system. Operators define policies with rules that match failures by exit code, condition (OOMKilled, Evicted, etc.), termination message, or error category (from #4713). The scheduler evaluates failed jobs against these rules and retries when appropriate.

- User code failures still fail by default. Operators explicitly configure what to retry.
- Feature-flagged with `retryPolicy.enabled: false` by default. When disabled, pod names use legacy format.
- RetryPolicy is a CRUD resource (like Queue), managed via `armadactl` without scheduler restarts.
- Queue policies always apply. Job annotations can add more policies but cannot remove queue-level ones.

---

## RetryPolicy Resource

Managed the same way as Queue: stored in the database, cached by the scheduler.

```bash
armadactl create -f ./retry-policy.yaml
armadactl get retrypolicy infrastructure
armadactl update -f ./retry-policy.yaml
armadactl delete retrypolicy infrastructure
```

### Policy YAML

```yaml
apiVersion: armadaproject.io/v1
kind: RetryPolicy
metadata:
  name: ml-training
spec:
  retryLimit: 5
  defaultAction: Fail
  backoff:
    initialDelay: 0s
  antiAffinity:
    mode: none
  rules:
    - action: Retry
      onConditions:
        - Preempted
        - Evicted
      antiAffinity:
        mode: node
    - action: Retry
      onConditions:
        - OOMKilled
      retryLimit: 3
    - action: Retry
      onExitCodes:
        operator: In
        values: [137]
    - action: Retry
      onFailureCategory: [cuda_error, infiniband_error]
```

### Scheduler Config

Global settings only, not policy definitions:

```yaml
scheduling:
  retryPolicy:
    enabled: true
    globalMaxRetries: 20
    defaultPolicyName: default
    defaultBackoff:
[...]

PR #4804, opened 2026-03-30
Add retry policy with engine, CRUD API, scheduler integration, and OnPodError matcher

dejanzele:
> [!IMPORTANT]
> **This PR is a reference implementation, not for wholesale review.** It will be re-split into a series of small, feature-flagged PRs (schema first, behavior last, flag default off). See the [status update on #4683](https://github.com/armadaproject/armada/issues/4683#issuecomment-4891660958) for the landing plan and current v1 scope. Rebased onto master 2026-07-06. This PR closes when the series lands.

## What this PR does

Adds RetryPolicy as a first-class concept. Operators define policies via armadactl, attach them to queues, and the scheduler evaluates them when a job run fails or is preempted to decide whether to retry or terminally fail.

Everything is behind `scheduling.retryPolicy.enabled` (default off). With the flag off there is no behaviour change: the engine is never invoked and the existing pod-name format is preserved byte-for-byte.

A policy is a list of rules. Each rule has matchers (conditions, exit codes, termination message, category/subcategory) and an action (Retry or Fail). First match wins. There's a per-policy `retryLimit` (counted in retries, not failures: `retryLimit: 3` means up to 3 retries after the initial failure, 4 total attempts) and a scheduler-wide `globalMaxRetries` safety cap with the same semantics. Categories and subcategories come from the executor classifier introduced in #4891, so policies can match on the same buckets the UI already surfaces.

Pod naming changes for retried runs: when the flag is on, each retry attempt gets its own pod (`armada-<jobId>-0-<runIndex>`) so the executor doesn't have to mutate or delete the previous pod. The old `armada-<jobId>-0` format is kept exactly when the flag is off.

## Commits

1. **Retry engine and matching primitives.** Introduces the decision logic: `Policy`, `Rule`, `Action`, and `Result` types, regex pre-compilation at policy-load time, AND-within-a-rule plus first-match-wins across rules, the global cap and per-policy retry-limit checks. Pure Go with no callers; the engine is dead code at this point.
2. **CRUD API for retry policies.** Makes RetryPolicy a first-class API resource: proto messages, gRPC service with REST gateway, postgres-backed repository, authorisation hooks, an armadactl surface (`create/update/delete/get retry-policy`), a Go client library, and a `--retry-policy` flag on `armadactl create/update queue`. Custom enum unmarshalling so YAML can use friendly names (`Fail`/`Retry`, `In`/`NotIn`) instead of the canonical `RETRY_ACTION_*` strings.
3. **Run-index plumbing.** Each retry attempt should be a fresh pod, so the scheduler now tracks an attempt counter per run. Adds a `run_index` column to the `runs` table, an `optional uint32 job_run_index` field on `JobRunLease`, and a branch in the executor's pod-naming so the suffix becomes `armada-<id>-0-<runIndex>` when the field is set. Field absence reproduces today's exact pod name, which keeps the feature flag's zero-impact promise true even at the wire level.
4. **Wire the engine into the scheduler.** The actual behaviour change. Adds a policy cache that periodically refreshes from the api, a proto-to-engine converter, and three call sites that consult the engine: container-level run failure, operator-issued preemption via `armadactl preempt job`, and scheduling-algorithm preemption when a higher-priority job needs the resource. All three emit a non-terminal `JobErrors` on retry so the api event stream surfaces `JobFailedEvent.retryable = true` for intermediate failures. Gangs are skipped here.
5. **End-to-end test.** A testsuite case that submits a pod which fails the first attempt and succeeds the second, reading the run-index from inside the container via the downward API so the only way the second attempt succeeds is if the engine actually retried and the executor stamped the new index on the new lease. Testsuite validator and watcher updates to assert and tolerate `retryable: true` Failed events. The gang-retry analysis and the v1 gang exclusion rationale are in the [status update on #4683](https://github.com/armadaproject/armada/issues/4683#issuecomment-4891660958).

Each commit builds, tests, and lints clean on its own. Most of the diff volume sits in regenerated proto/swagger files; the hand-written changes are in `*.proto`, the non-`pb.go` Go files, the SQL migration, and the sqlc queries.

## How to validate

The testsuite case in commit 5 covers the full flow in CI. To exercise it against the goreman localdev stack:

1. Bring up the dependencies and components. `_local/scheduler/config.yaml` already has the retry-policy flag enabled, so no config edits are needed:
   ```bash
   docker compose -f _local/docker-compose-deps.yaml up -d
   scripts/localdev-init.sh
   goreman -f _local/procfiles/no-auth.Procfile start
   ```
2. Write a policy YAML:
   ```yaml
   apiVersion: armadaproject.io/v1beta1
   kind: RetryPolicy
   name: my-policy
   retryLimit: 3
   defaultAction: Fail
   rules:
     - action: Retry
       onConditions: ["AppError"]
     - action: Retry
       onExitCodes:
         operator: In
         values: [42, 137]
   ```
   `retryLimit` is the number of retries allowed (3 here means up to 3 retries, 4 total attempts). `defaultAction` is what happens when no rule matches. Matchers within a rule are ANDed; rules are first-match-wins. Available matchers: `onConditions`, `onExitCodes`, `onTerminationMessagePattern`, `onCategory` (with optional `onSubcategory`).
3. `armadactl create -f policy.yaml`
4. `armadactl create queue my-queue --retry-policy my-policy` (or update an existing queue with `armadactl update queue --retry-policy ...`).
5. Submit a job that fails on first attempt.

You should see one pod per attempt (`armada-<id>-0-0`, `armada-<id>-0-1`, ...) until either a Retry rule stops matching, the policy's `retryLimit` is hit, or the scheduler's `globalMaxRetries` cap kicks in. The scheduler logs a `retry decision` line per failure with the matched rule and the resulting action.

The api event stream now carries `JobFailedEvent.retryable = true` for intermediate retries so clients and the testsuite can distinguish them from terminal failures.

To run the testsuite case directly against the same stack:
```bash
go run cmd/testsuite/main.go test \
  --tests testsuite/testcases/retries/single_pod_retry.yaml \
  --config e2e/config/armadactl_config.yaml
```

## What's deferred

- Gangs are out of scope. Coordinating retries across gang members (atomic failure detection, run-index sync, killing healthy pods, retry state at gang level) is enough complexity to warrant its own PR. Gang jobs fall through to the existing behaviour: terminal failure for pod errors, lease-return retry up to `maxAttemptedRuns` when the executor returns the lease. Design notes in `notes/retry-policy/gang-retry.md`.
- Backoff between attempts. Retries currently fire on the next scheduler cycle.
- Removing the legacy lease-return retry code path. The new engine coexists with it for now; cleanup after this lands.
- Lookout UI for run-index. Database has the data; UI work is a follow-up.

## Issue

Part of #4683.


dejanzele:
@greptileai

dejanzele:
@greptileai

dejanzele:
Closing in favour of a different approach

PR #4841, opened 2026-04-16
Add respectNodePodLimits scheduler flag to enforce per-node pod capacity

### LINKED ISSUE #4515: Scheduler does not respect node pod limits

dejanzele:
## Summary

- Adds `scheduling.respectNodePodLimits` feature flag (default `false`) that enables the scheduler to track `pods` as a resource and reject scheduling to nodes that have exhausted their pod limit (`node.Status.Allocatable["pods"]`)
- When enabled, the scheduler programmatically registers `pods` in `supportedResourceTypes` and `indexedResources` at startup, and injects `pods: 1` into every job's internal resource requirements
- The executor now always reports non-Armada pod count in `NonArmadaAllocatedResources` so the scheduler can subtract system/DaemonSet pods from available capacity

Fixes #4515

This PR builds on top of https://github.com/armadaproject/armada/pull/4517 and big thanks for the initial work to @Sovietaced

## Operator upgrade notes

- **Executor change is unconditional.** After the executor upgrade, `NonArmadaAllocatedResources` gains a `pods` key in every report regardless of whether any scheduler has the flag enabled. Dashboards, metrics, or custom consumers that iterate this map generically (e.g. sum over all keys) will start including pod counts. Audit prometheus / Grafana panels before rollout.
- **Rollback is clean.** Reverting the scheduler flag to `false` stops the scheduler from tracking pods; reverting the executor binary removes the `pods` key from its reports. Neither requires data migration.
- **Rolling upgrade order is flexible.** Old scheduler + new executor is safe (scheduler's `FromNodeProto` silently drops unknown resources). New scheduler + old executor briefly overestimates free pod capacity by the count of non-Armada pods per node (~10-30 typically, DaemonSets + system pods), since the old executor does not report them. The overestimate resolves as executors are upgraded.

## Known limitations

- `pods` is **not** added to `dominantResourceFairnessResourcesToConsider`. On dense-pod nodes (e.g. GKE's 110-pod limit) a queue running many small pods can monopolize pod slots without a fair-share penalty. Deferred per reviewer request; follow-up if this becomes a problem in practice.

### DIFF
--- a/config/scheduler/config.yaml
+++ b/config/scheduler/config.yaml
@@ -83,6 +83,7 @@ scheduling:
   maxSchedulingDuration: 5s
   enableAssertions: false
   enablePreferLargeJobOrdering: false
+  respectNodePodLimits: false
   protectedFractionOfFairShare: 1.0
   nodeIdLabel: "kubernetes.io/hostname"
   priorityClasses:
--- a/internal/common/resource/resource.go
+++ b/internal/common/resource/resource.go
@@ -7,6 +7,10 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
+// PodsResourceName is the Kubernetes resource key for per-node pod capacity
+// (node.Status.Allocatable["pods"]), used by the scheduler's RespectNodePodLimits feature.
+const PodsResourceName = "pods"
+
 // FromResourceList function takes a map with keys of type ResourceName and values of type
 // "resource.Quantity" as defined in the K8s API.
 //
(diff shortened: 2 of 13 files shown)


dejanzele:
@greptileai

dejanzele:
@greptileai

JamesMurkin (dismissed):

JamesMurkin (approved):

Sovietaced:
Thanks for landing this. We'll try it out at some point

PR #4853, opened 2026-04-21
Display failure_category/failure_subcategory in Lookout UI

dejanzele:
Reads the flat failure_category and failure_subcategory columns added in #4843 instead of the failure_info jsonb blob. Updates the querybuilder SQL, the internal Go model, swagger and its generated code, the conversion layer, and the jobs sidebar.

Depends on #4843.


dejanzele:
@greptileai

masipauskas (approved):

PR #4855, opened 2026-04-21
Drop failure_info column from job_run table

dejanzele:
Drops the failure_info jsonb column from job_run. Nothing writes or reads it after #4843 and #4853, and the column was never populated in production outside the opt-in flag path anyway.

Also drops the unused FailureInfo field from the queryapi sqlc model.

Only merge after #4843 and #4853 have been deployed long enough that we are sure no consumer still depends on the column.


dejanzele:
@greptileai

mauriceyap (approved):

PR #4890, opened 2026-04-30
Add diagnostic hint catalog for well-known executor failure modes

dejanzele:
## Summary

When a job fails for a well-known reason (e.g. wrong-arch image), the user-facing error in Lookout is the raw kubelet text, often opaque. This PR adds a `diagnostics` package that maps such errors to actionable hints; the hint is prepended to the failure message at the three executor sites that build user-facing text (`reporter/event.go` PodFailed branch, `service/pod_issue_handler.go` retryable failed-pod path, and `service/pod_issue_handler.go` `createStuckPodMessage`).

The package is separate from the categorizer because the two are orthogonal: diagnostic hints are always-on and curated, categorization is opt-in and operator-configured. Splitting them avoids forcing a categorizer dependency on hint consumers and lets either side evolve without churning the other.

Adding a hint is a single `{pattern, text}` entry in `builtinHints`. Patterns compile at startup; first match wins. A future PR may move the catalog from compiled-in to config-driven once we see whether operators want deployment-specific entries.

## Validation

To reproduce on local dev:

**1. Submit a wrong-arch job** (`example/platform-mismatch.yaml`):

```yaml
queue: test
jobSetId: platform-mismatch-repro
jobs:
  - namespace: default
    priority: 0
    podSpec:
      restartPolicy: Never
      containers:
        - name: wrong-arch
          # On an arm64 cluster, force-pull the amd64 variant.
          # Flip to arm64v8/busybox if reproducing on an amd64 cluster.
          image: amd64/busybox:latest
          command:
            - sh
            - -c
            - echo should-never-run
          resources:
            requests:
              memory: 64Mi
              cpu: "0.1"
            limits:
              memory: 64Mi
              cpu: "0.1"
```

```bash
armadactl create queue test   # if not already present
armadactl submit example/platform-mismatch.yaml
```

**2. Wait** for the kubelet event-based fail check to fire (typically 1-5 minutes depending on `event_checks` grace period).

**3. Verify** the hint appears in the lookout DB. The `error` column is zlib-compressed bytea, so decode in two steps:

```bash
# Failure metadata
docker exec postgres psql -U postgres -d lookout -c \
  "SELECT job_id, run_id, finished, exit_code
   FROM job_run
   ORDER BY finished DESC NULLS LAST
   LIMIT 1;"

# Decompressed error message (the byte string the Lookout UI shows users)
docker exec postgres psql -U postgres -At -d lookout -c \
  "SELECT encode(error, 'base64') FROM job_run
   ORDER BY finished DESC NULLS LAST LIMIT 1;" \
| python3 -c 'import sys,base64,zlib; print(zlib.decompress(base64.b64decode(sys.stdin.read())).decode())'
```

### Expected (and observed) result

Decompressed `error` column begins with the curated hint, followed by the raw kubelet error preserved verbatim:

```
No compatible image was found for the target farm node's architecture; this is often due to an x64/arm64 mismatch, so check the image architecture or build and publish a compatible version.
Container ... has been in state Waiting for reason ImagePullBackOff (Failed to pull image "amd64/busybox:latest": rpc error: code = NotFound desc = failed to pull and unpack image "docker.io/amd64/busybox:latest": no match for platform in manifest: not found) for more than timeout 1ns
```

Live-validated end-to-end on macOS arm64 (M3) against a k3d cluster.

### DIFF
--- a/internal/executor/diagnostics/diagnostics.go
+++ b/internal/executor/diagnostics/diagnostics.go
@@ -1,0 +1,28 @@
+package diagnostics
+
+import "regexp"
+
+type hint struct {
+	pattern *regexp.Regexp
+	text    string
+}
+
+// builtinHints is consulted in order; first match wins, so place
+// more-specific patterns before broader ones.
+var builtinHints = []hint{
+	{
+		pattern: regexp.MustCompile(`no match for platform in manifest`),
+		text:    "No compatible image was found for the target farm node's architecture; this is often due to an x64/arm64 mismatch, so check the image architecture or build and publish a compatible version.",
+	},
+}
+
+// LookupHint returns user-facing guidance matching the failure message,
+// or "" if no built-in hint applies.
+func LookupHint(message string) string {
+	for _, h := range builtinHints {
+		if h.pattern.MatchString(message) {
+			return h.text
+		}
+	}
+	return ""
+}
--- a/internal/executor/diagnostics/diagnostics_test.go
+++ b/internal/executor/diagnostics/diagnostics_test.go
@@ -1,0 +1,35 @@
+package diagnostics
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLookupHint(t *testing.T) {
+	tests := map[string]struct {
+		message       string
+		wantSubstring string
+	}{
+		"empty message returns no hint": {
+			message: "",
+		},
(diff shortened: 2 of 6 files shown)


mauriceyap (approved):

dejanzele:
Closing in favour of https://github.com/armadaproject/armada/pull/4891 and a different approach

dejanzele:
xr

PR #4891, opened 2026-04-30
Add onPodError matcher to categorize pre-startup pod failures

dejanzele:
## Summary

This PR extends the failure categorizer with two operator-facing additions:

1. **`onPodError` matcher** - matches pod-level error text, covering pre-startup kubelet/runtime errors (image pull, missing volume, missing ConfigMap/Secret) and Armada-detected pod-level failures (stuck terminating, active deadline exceeded, externally deleted) that produce no useful container `terminationMessage`. These end up with empty `failure_category` / `failure_subcategory` in lookoutdb today.
2. **`hint` field on rules** - operator-supplied user-facing copy describing the failure mode. When set, it is appended to the failure message that lands in `lookoutdb.job_run.error`, so end users see actionable guidance alongside the raw runtime error.

The two commits are independently useful; together they let operators both classify and explain previously-opaque pod-level failures.

### `onPodError`
- New rule field on `CategoryRule`. Matches a regex against the issue's pod-level error message. `ContainerName` scoping is ignored (pod-level text has no container attribution).
- `onTerminationMessage` is unchanged - still matches container `Terminated.Message`, still honors `ContainerName`. Non-overlapping data source from `OnPodError` by design.
- `Classify` is split into `ClassifyContainerError(pod)` and `ClassifyPodError(pod, podErrorMessage)`. The pod-error variant is needed because kubelet rotates `Waiting.Reason` from `ErrImagePull` to `ImagePullBackOff` within seconds, replacing `Waiting.Message` with a generic backoff string, so by the time Armada classifies the pod the runtime error is no longer in `pod.Status` and must be passed in by the caller.
- Config validation: a rule must specify exactly one of the four matchers; regex matchers compile-check at startup so invalid patterns fail fast.

### `hint`
- Optional `Hint string` on `CategoryRule`. Empty by default - no behavior change for existing configs.
- The classifier returns the matched rule's hint alongside category/subcategory in `ClassifyResult`.
- `pod_issue_handler` and `reporter/event.go` append the hint to the user-facing message with two newlines: `"<original message>\n\n<hint>"`. Appended (not prepended) so the raw runtime error stays the lede.
- Hints flow through the existing `PodError` -> Pulsar -> lookout-ingester pipeline into `lookoutdb.job_run.error`.

## Validation

To reproduce on local dev:

**1. Add a rule with both `onPodError` and `hint`** (`_local/executor/config.yaml` under `application:`):

```yaml
application:
  errorCategories:
    defaultCategory: "uncategorized"
    defaultSubcategory: "unknown"
    categories:
      - name: infrastructure
        rules:
          - onPodError:
              pattern: "no match for platform in manifest"
            subcategory: "platform_mismatch"
            hint: "Build the image for the cluster's CPU architecture (typically x64/arm64 mismatch)."
```

Categorization is opt-in: Armada ships no default rules.

**2. Submit a wrong-arch job** (`example/platform-mismatch.yaml`):

```yaml
queue: test
jobSetId: platform-mismatch-repro
jobs:
  - namespace: default
    priority: 0
    podSpec:
      restartPolicy: Never
      containers:
        - name: wrong-arch
          image: amd64/busybox:latest
          command:
            - sh
            - -c
            - echo should-never-run
          resources:
            requests:
              memory: 64Mi
              cpu: "0.1"
            limits:
              memory: 64Mi
              cpu: "0.1"
```

```bash
armadactl create queue test
armadactl submit example/platform-mismatch.yaml
```

**3. Wait** for the kubelet event-based fail check to fire (typically 1-5 minutes).

**4. Verify** the categorization and hint landed:

```bash
docker exec postgres psql -U postgres -d lookout -c \
  "SELECT job_id, run_id, finished, failure_category, failure_subcategory,
          convert_from(decompress(error), 'UTF8') AS error_text
     FROM job_run
    ORDER BY finished DESC NULLS LAST
    LIMIT 1;"
```

Expected:

```
 failure_category | failure_subcategory
------------------+---------------------
 infrastructure   | platform_mismatch

 error_text
------------
 Pod of the job failed
 ...
 Failed to pull image "amd64/busybox:latest": no match for platform in manifest
 ...

 Build the image for the cluster's CPU architecture (typically x64/arm64 mismatch).
```

The hint appears on its own paragraph after the raw kubelet error.

Live-validated end-to-end on macOS arm64 (M3) against a k3d cluster.

### DIFF
--- a/internal/executor/categorizer/classifier.go
+++ b/internal/executor/categorizer/classifier.go
@@ -29,12 +29,26 @@ type rule struct {
 	onPodError           *regexp.Regexp
 	onConditions         []string
 	subcategory          string
+	hint                 string
 }
 
 // ClassifyResult holds the classification output for a failed pod.
 type ClassifyResult struct {
 	Category    string
 	Subcategory string
+	// Hint is operator-supplied user-facing copy attached to the matching rule.
+	// Use AppendHint to attach it to the failure message before emitting events.
+	Hint string
+}
+
+// AppendHint returns the message with this result's hint appended after a blank
+// line. Returns the message unchanged when no hint is set. Centralizing the
+// format here keeps both event-reporting call sites consistent.
+func (r ClassifyResult) AppendHint(message string) string {
+	if r.Hint == "" {
+		return message
+	}
+	return fmt.Sprintf("%s\n\n%s", message, r.Hint)
 }
 
 // Classifier evaluates pods against a set of category rules and returns
@@ -157,6 +171,7 @@ func buildRule(cfg CategoryRule) (rule, error) {
 		onTerminationMessage: terminationRegex,
 		onPodError:           podErrorRegex,
 		subcategory:          cfg.Subcategory,
+		hint:                 cfg.Hint,
 	}, nil
 }
 
@@ -194,7 +209,7 @@ func (c *Classifier) classify(pod *v1.Pod, podErrorMessage string) ClassifyResul
 			matched := ruleMatches(r, containers, podReason, podErrorMessage)
 			metrics.RecordRuleEvaluationDuration(cat.name, r.subcategory, time.Since(start))
 			if matched {
-				return ClassifyResult{Category: cat.name, Subcategory: r.subcategory}
+				return ClassifyResult{Category: cat.name, Subcategory: r.subcategory, Hint: r.hint}
 			}
 		}
 	}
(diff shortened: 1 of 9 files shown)


dejanzele:
@greptileai

mauriceyap (dismissed):

dejanzele:
@greptileai

dejanzele:
@greptileai

JamesMurkin (approved):

PR #4894, opened 2026-04-30
Add lookoutingestor metrics to track state changes and count terminal states

ianhockett:
#### What type of PR is this?

feature / observability

Adds a Prometheus counter metric to the lookout ingester to track job state updates processed by UpdateJobs(). The metric (`lookout_ingester_job_state_updates_total`) is labeled by state, allowing operators to observe transition rates and specifically monitor terminal state updates, which trigger cross-partition row movement in the Lookout database.

Also fixes error categorization schema in `_local/executor` configs that was broken by
a prior change.

The `terminal_state_updates_total` counter was initially added as a standalone metric
but was removed in favor of deriving it via PromQL from the per-state counter
(sum by state where state is terminal).


mauriceyap:
Nit: `terminal_state_updates_total` seems redundant as it's derivable from `state_updates` by summing the terminal-state labels in PromQL. Is there a reason to have both?

ianhockett:
> Nit: `terminal_state_updates_total` seems redundant as it's derivable from `state_updates` by summing the terminal-state labels in PromQL. Is there a reason to have both?

I suppose there is not a reason, other than to make queries slightly simpler. I will remove it since it is redundant and provides little-to-no additional benefit.

mauriceyap (approved):

MustafaI (approved):

PR #4899, opened 2026-05-06
Hot cold local dev

dslear:
Adding local development configurations to run the Lookout hot/cold partitioned database in parallel with the existing Lookout database. This includes the VSCode configurations in the `.vscode` directory, the goreman procfiles and config files under `_local`, and the JetBrains configurations under `.run`


config/lookouthc/config.yaml
```diff
@@ -0,0 +1,51 @@
...
+  keyPath: /certs/tls.key
+postgres:
+  connection:
+    host: postgres
+    port: 5433
+    user: postgres
+    password: psw
+    dbname: lookouthc
+    sslmode: disable
+prunerConfig:
```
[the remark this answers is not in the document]
dslear: This is intentional, this is an experimental feature that is not ready for production yet

_local/lookouthc/config.yaml
```diff
@@ -0,0 +1,24 @@
...
+    port: 5432
+    user: postgres
+    password: psw
+    dbname: lookouthc
+    sslmode: disable
+experimentalHotColdSplit: true
+uiConfig:
+  backend: "jsonb"
+  armadaApiBaseUrl: "http://localhost:8081"
+  binocularsBaseUrlPattern: "http://localhost:8084"
```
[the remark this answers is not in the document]
dslear: This is incorrect, the local configs for the server and binoculars set ports 8081 and 8084. VSCode and JetBrains do not use the configs in `_local` which is why they are not impacted.

.vscode/launch.json
```diff
@@ -91,6 +108,29 @@
...
+      "type": "go",
+      "request": "launch",
+      "mode": "auto",
+      "program": "${workspaceFolder}/cmd/lookout/main.go",
+      "cwd": "${workspaceFolder}",
+      "env": {
+          "ARMADA_APIPORT": "8091",
+          "ARMADA_METRICSPORT": "9010",
+          "ARMADA_EXPERIMENTALHOTCOLDSPLIT": "true",
+          "ARMADA_CORSALLOWEDORIGINS": "\"http://localhost:3001,http://localhost:4173,http://localhost:8091,http://localhost:10000,http://example.com:10000,http://example.com:8089\"",
```
mauriceyap: The escaped quotes here should be removed (pre-existing mistake which is my fault).

.vscode/tasks.json
```diff
@@ -19,12 +19,29 @@
       "command": "mage kind && mage startDependencies",
       "label": "Start dependencies"
     },
+    {
+      "type": "shell",
+      "label": "Wait for Postgres",
+      "command": "until docker exec postgres pg_isready -U postgres; do echo 'Waiting for Postgres...'; sleep 2; done"
+    },
```
mauriceyap, line 55: See my comments relating to this in https://github.com/armadaproject/armada/pull/4929 - we should remove this and fix the docker compose file (which that PR will do).

internal/lookoutui/package.json
```diff
@@ -102,5 +102,6 @@
     "@types/react-dom": "^19",
     "react": "^19",
     "react-dom": "^19"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
mauriceyap: Please could you remove this change? I suspect it snuck in when you ran `corepack use` or `yarn`. Pinning the package manager isn't necessarily a bad idea, but if we want to do that, it should be in a separate PR.

(Also you can add `export COREPACK_ENABLE_AUTO_PIN=0` to your shell to prevent Corepack from doing this in future.)

internal/lookoutui/vite.config.mts
```diff
@@ -54,7 +55,7 @@ export default defineConfig({
     proxy: PROXY_OPTIONS,
   },
   server: {
-    port: 3000,
+    port: parseInt(process.env.PORT || "3000"),
```
mauriceyap: nit - add radix
```suggestion
    port: parseInt(process.env.PORT || "3000", 10),
```

mauriceyap: Also, could you use `VITE_PORT` instead of `PORT`? Just because `PORT` is a very common env variable.

mauriceyap (dismissed):

d2burkhalter (approved):

mauriceyap (approved):

PR #4901, opened 2026-05-07
Show 'Last updated: X ago' alongside the refresh button

YHines004:
#### What type of PR is this?

Enhancement

Add the text 'Last updated X ago' next to the refresh button

Fixes #


internal/lookoutui/src/components/RefreshButton.tsx
```diff
@@ -8,9 +10,44 @@ type RefreshButtonProps = {
...
 
+function getRelativeTime(date: Date): string {
+  const minutes = Math.floor((Date.now() - date.getTime()) / 60000)
+  if (minutes < 1) return "just now"
+  if (minutes < 60) return `${minutes} minute${minutes === 1 ? "" : "s"} ago`
+  const hours = Math.floor(minutes / 60)
+  if (hours < 24) return `${hours} hour${hours === 1 ? "" : "s"} ago`
+  const days = Math.floor(hours / 24)
+  return `${days} day${days === 1 ? "" : "s"} ago`
+}
```
mauriceyap: Please could you use `formatTimestampRelative()` in `internal/lookoutui/src/common/formatTime.ts` instead of duplicating its functionality here? It is locale-aware (based on user settings or browser location), and is fully tested. Feel free to modify it if it doesn't quite do what you'd like it to!

internal/lookoutui/src/components/RefreshButton.css
mauriceyap: Not directly related to your change, but please could you delete this file and replace `<div className="refresh">` with MUI's `<Stack direction="row" alignItems="center">`? That's more idiomatic I think.

internal/lookoutui/src/components/RefreshButton.tsx
```diff
@@ -8,9 +10,44 @@ type RefreshButtonProps = {
...
+    }
+    prevIsLoading.current = props.isLoading
+  }, [props.isLoading])
+
+  useEffect(() => {
+    if (!lastUpdated) return
+    setRelativeTime(getRelativeTime(lastUpdated))
+    const interval = setInterval(() => setRelativeTime(getRelativeTime(lastUpdated)), 60000)
+    return () => clearInterval(interval)
+  }, [lastUpdated])
```
mauriceyap, line 31: A few things on the state shape here. I think it all works correctly, but the patterns compound on each other and are worth untangling.

`relativeTime` shouldn't be state because it's fully derived from `lastUpdated` + current time (we currently have to resync it in three places). The rule of thumb is keep in state only what you can't compute from other state. To force a re-render every 60s, use a tick counter (`setTick(t => t + 1)` or even better, make the tick a boolean and `setTick(t => !t)`) and compute the string inline.

The `prevIsLoading` ref works, but the `prev === true && current === false` logic is more than we need. All we actually want is "record a time when a load finishes" - a single ref flag (`wasLoading`) set to `isLoading` at the end of each effect, with the falling-edge check guarding `setLastUpdated`, reads more directly. (More generally: any time you find yourself tracking both previous and current values in a ref, it's worth asking whether you really need the full history or just a boolean "has this ever been true.")

Also, the two effects overlap. Both write `relativeTime`, which is the knock-on from above.

(Generally if two effects write the same state, one of them shouldn't exist, or the state shouldn't. Effects are for syncing React with something outside React. "Keep this state in sync with that state" is derived state in disguise.)

Putting that together, it should look like this:
```ts
  const [lastUpdated, setLastUpdated] = useState<Date | null>(null)
  const [, setTick] = useState(false)
  const wasLoading = useRef(false)

  useEffect(() => {
    if (wasLoading.current && !isLoading) {
      setLastUpdated(new Date())
    }
    wasLoading.current = isLoading
  }, [isLoading])

  useEffect(() => {
    if (!lastUpdated) {
      return
    }
    const id = setInterval(() => setTick((t) => !t), 60_000)
    return () => clearInterval(id)
  }, [lastUpdated])

```

internal/lookoutui/src/components/RefreshButton.tsx
```diff
@@ -8,9 +10,44 @@ type RefreshButtonProps = {
...
+      setLastUpdated(now)
+      setRelativeTime(getRelativeTime(now))
+    }
+    prevIsLoading.current = props.isLoading
+  }, [props.isLoading])
+
+  useEffect(() => {
+    if (!lastUpdated) return
+    setRelativeTime(getRelativeTime(lastUpdated))
+    const interval = setInterval(() => setRelativeTime(getRelativeTime(lastUpdated)), 60000)
```
mauriceyap: Please could you also extract the `60000` magic number?

```ts
const REFRESH_MESSAGE_UPDATE_PERIOD_MS = 60 * 1_000;
```

(I like writing out "number of seconds multiplied by 1,000" as well as underscores inside large numbers to make it really clear.)

internal/lookoutui/src/components/RefreshButton.tsx
YHines004: I will update this

internal/lookoutui/src/components/RefreshButton.css
YHines004: Yes I can

mauriceyap (approved):

Issue #4839, opened 2026-04-15
Preemption error message should dominate


robertdavidsmith:
Hi,

In same cases, a job is preempted (so ends up as state preempted as visible in lookout). However the preemption reason isn't visible in the RESULTS tab in lookout as instead another error is there.
- If a running job is preempted but the job writes to terminationMessagePath, the termination message is displayed in lookout, and the preemption reason is lost. (We have seen examples where jobs catch SIGTERM but then shut down uncleanly and so write a termination message as they are shutting down).
- If a pending job is preempted, the Kubernetes events are displayed as the error message, and again the preemption reason is lost.

Please can you ensure the preemption reason is not lost in these cases. Either append to the existing error message or just display the preemption reason.

Thanks,

Rob

robertdavidsmith:
Hi @williamvega 

We discussed the design some more, we decided to add a jsonb column to the lookout job runs table, called something like scheduler_termination_reason. For preempted/cancelled jobs, this will contain structured information as to why the job was killed. @masipauskas is going to post example json shortly, and also a sample addtion to JobPreemptedEvent.

masipauskas:
I think we should, basically add `ReasonArgs dict[str, str]` to `JobPreemptedEvent`. Then serialize new jsonb column as follows:
```
JobPreemptedEvent(ReasonArgs: dict[str, str]) // new property for now nil
{
    "reason": JobPreemptedEvent.Reason
    "args": {
        "preemptingRunId": JobPreemptedEvent.PreemptiveRunId (if not null/empty)
        ... JobPreemptedEvent.ReasonArgs
    }
}

JobCancelledEvent
{
    "reason": JobCancelledEvent.Reason
    "args": {
        "requestor": JobCancelledEvent.Requestor (if not null/empty)
    }
}
```

This enables us to do analytics in relatively simple way, as well as attaching more structured context to both Preemption/Cancellation events.

PR #4905, opened 2026-05-12
fix: job run scheduler termination column

williamvega:
#### What type of PR is this?

fix

This PR fixes the original problem in issue #4839, while further work is needed to include cancellation as a scheduler termination reason and populate the column with other reasons besides preemption.

The preemption reason stored in lookout's job_run error column was getting overwritten by other events. We now store the preemption reason and other reasons the scheduler may cancel a job in its own column, scheduler_termination_reason. 

The first event termination reason we are storing in this column is from the JobPreemptedEvent, and later prs should include the JobCancelledEvent reason as suggested in the issue.

Fair-share preemption:
image

Manual preemption:
image


pkg/api/event.proto
```diff
@@ -131,7 +131,7 @@ message JobPreemptedEvent {
     google.protobuf.Timestamp created = 4;
     string cluster_id = 5;
     string run_id = 6;
-    string preemptive_job_id = 7;
+    string preempting_job_id = 7;
```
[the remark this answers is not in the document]
mauriceyap: I agree - I think it would be better to do:
```
reserved 7;
string preempting_job_id = 10;
```

client/rust/src/gen/api.rs
```diff
@@ -1498,12 +1498,12 @@ pub struct JobPreemptedEvent {
     pub cluster_id: ::prost::alloc::string::String,
     #[prost(string, tag = "6")]
     pub run_id: ::prost::alloc::string::String,
-    #[prost(string, tag = "7")]
-    pub preemptive_job_id: ::prost::alloc::string::String,
     #[prost(string, tag = "8")]
     pub preemptive_run_id: ::prost::alloc::string::String,
     #[prost(string, tag = "9")]
     pub reason: ::prost::alloc::string::String,
+    #[prost(string, tag = "10")]
```
robertdavidsmith, line 1505: Is `preemptiveJobId`=>`preemptingJobId` just a straight rename, in which case keep the same protobuf numeric id?

williamvega: It was not a straight rename because the old field was never populated. In the protobuf we reserved the old field to eliminate confusion. https://github.com/armadaproject/armada/pull/4905#discussion_r3267684125

robertdavidsmith (approved):

mauriceyap (approved):

PR #4906, opened 2026-05-13
Split submit_message and groups into job_metadata table

tgucks:
#### What type of PR is this?

Feature

Splits the large `submit_message` and `groups` byte columns out of the `jobs` table and into a new `job_metadata` table, behind a three-phase migration flag (`jobMetadataMigrationPhase`) on both the `scheduler` and `scheduleringester`.

**Why:** `submit_message` is the hot path's largest per-row payload but is only read during leasing. Keeping it on `jobs` makes every row wide. Moving it to a sibling table keyed by `job_id` lets `jobs` stay narrow while preserving lookup for lease fetches.

**How the migration works** (controlled by `jobMetadataMigrationPhase` in both configs - these must be coordinated):

| Phase | Scheduler read path | Ingester write path |
|---|---|---|
| `legacy` (default) | `jobs.submit_message` / `jobs.groups` | writes to `jobs` only |
| `dualWrite` | `LEFT JOIN job_metadata` + `COALESCE(js.*, j.*)` so old and new rows both resolve | writes to `jobs` **and** `job_metadata` |
| `cutover` | `INNER JOIN job_metadata` (requires backfill complete) | writes to `job_metadata` only; `jobs.submit_message` is NULL |

Migration `037_add_job_metadata_table.sql` creates the table, marks both blob columns `STORAGE EXTERNAL`, and drops the `NOT NULL` constraint on `jobs.submit_message`. We will be following up this PR with another migration that backfills any data in `jobs.groups` and `jobs.submit_message` to `job_metadata.*` before setting the flags to `cutover`.

`JobMetadataMigrationPhase.Validate()` rejects unknown/empty values; both `Configuration` structs carry `validate:"required,oneof=legacy dualWrite cutover"`.

**Load test results across migration phases (submitted 2M jobs locally per phase with fresh dependencies each run):**

<details>
<summary>Load test results</summary>

| Metric              | P0 (legacy) | P1 (dualWrite) | P1 vs P0  | P2 (cutover) | P2 vs P0  |
| ------------------- | ----------: | -------------: | --------: | -----------: | --------: |
| jobs ingested       | 2,000,001   | 2,000,001      | -         | 2,000,001    | -         |
| WAL per job         | 6.06 K      | 6.98 K         | **1.15x** | 5.80 K       | 0.96x     |
| table size per job  | 2.10 K      | 2.76 K         | 1.31x     | 1.80 K       | **0.86x** |
| throughput (jobs/s) | 12,987      | 11,696         | -9.9%     | 12,658       | -2.5%     |

Totals for context (wall time, on-disk growth, WAL bytes):

| Metric           | P0 (legacy) | P1 (dualWrite) | P2 (cutover) |
| ---------------- | ----------: | -------------: | -----------: |
| wall time (s)    | 154.4       | 171.3          | 158.4        |
| jobs size growth | 4.01 G      | 4.06 G         | 2.23 G       |
| job_metadata growth | 0 B         | 1.21 G         | 1.21 G       |
| total growth     | 4.01 G      | 5.27 G         | 3.44 G       |
| WAL bytes        | 11.56 G     | 13.32 G        | 11.06 G      |

Takeaways:
- **Dual-write overhead is ~15% extra WAL / ~31% extra table bytes per job**
- **Cutover is ~14% smaller per-job footprint than legacy**

</details>

- Rollout order matters. The ingester's `WritesJobMetadata()` must be true before the scheduler reads from `job_metadata`, or new submissions become unreadable. 
  - This will largely be handled for most operators via the `jobMetadataMigrationPhase` defaults set in `config/scheduler/config.yaml` and `config/scheduleringester/config.yaml`. This PR sets them to `legacy` to start (read/write from `jobs` only, matching existing behavior).
  - We will follow up this PR with another that includes a migration to backfill `job_metadata.*` for all non-terminal legacy rows in `jobs` and then flip both services to `cutover`. At which point it will be safe to drop the "legacy" columns on the `jobs` table.
- This updates the `InsertJob` `DbOperation` to include an additional upsert to `job_metadata` based on the config flag value. This is covered by dedicated tests (`job_metadata_migration_test.go`) in both packages.


JamesMurkin (approved):

mauriceyap (approved):
The comment in the new config field - we should call that out in the notes for the release which contains this change.

tgucks:
> The comment in the new config field - we should call that out in the notes for the release which contains this change.

@mauriceyap Agreed. I've discussed this with James, including how to "progress" this migration to the point we are fully `cutover` to the new table and can remove this config. What's the usual process for including something in the release notes?

PR #4907, opened 2026-05-13
Allow sentinel client for metrics using read only mode

nikola-jokic:

Reduce connections on Redis master by using replicas for metrics


nikola-jokic:
Closing as not needed for now

PR #4911, opened 2026-05-14
Support optional schema creation in database migrators

mauriceyap:
Add an opt-in MigrationConfig to the scheduler and lookout migrators that, when configured, ensures the target schema exists and runs migrations inside it. Supports a two-role pattern where a separate "schema-creator" role runs CREATE SCHEMA and GRANT, while a least-privileged migrator role runs the DDL. Identifier interpolation goes through pgx.Identifier.Sanitize() with a non-empty + 63-byte length check.

Default behaviour is unchanged when MigrationConfig is unset.


internal/common/database/schema.go
```diff
@@ -0,0 +1,122 @@
...
+	if _, err := migratorDB.Exec(ctx, fmt.Sprintf(`SET search_path TO %s`, schemaIdent)); err != nil {
+		return errors.WithMessagef(err, "failed to set search_path to %q", cfg.Schema)
+	}
+	return nil
+}
+
+func createSchema(ctx *armadacontext.Context, migratorDB *pgx.Conn, cfg MigrationConfig) error {
+	schemaIdent := pgx.Identifier{cfg.Schema}.Sanitize()
+
+	if cfg.SchemaCreator == nil {
```
geaere: All schema creation done by schema creator

internal/common/database/schema.go
```diff
@@ -0,0 +1,122 @@
...
+	migratorRole, err := currentUser(ctx, migratorDB)
+	if err != nil {
+		return errors.WithMessage(err, "failed to determine migrator role")
+	}
+	if err := validateIdentifier(migratorRole); err != nil {
+		return errors.WithMessage(err, "unexpected migrator role name")
+	}
+	roleIdent := pgx.Identifier{migratorRole}.Sanitize()
+
+	bootstrapDB, err := OpenPgxConn(*cfg.SchemaCreator)
```
geaere: Maybe call this SchemaCreatorConfig?

internal/common/database/schema.go
geaere: OR, just more comments

mauriceyap: added a comment

internal/common/database/schema.go
mauriceyap: Gone with `SchemaCreatorPostgresConfig`

geaere (approved):

PR #4915, opened 2026-05-15
Extract leader election

nikola-jokic:
#### What type of PR is this?

Enhancement

Multiple components are using leaderelection which means it should be extracted into its own package


internal/scheduler/scheduler_test.go
```diff
@@ -3358,11 +3360,11 @@ func TestCycleConsistency(t *testing.T) {
 				return schedulerDb.Store(ctx, dbOpsWithMessageIds)
 			}
 
-			// failover swaps the leader tokens between a and b, thus swapping which scheduler is leader.
+			// failover swaps the leader tokens between a and b, thus swapping which scheduler is leaderelection.
```
JamesMurkin:
```suggestion
			// failover swaps the leader tokens between a and b, thus swapping which scheduler is leader.
```

internal/scheduler/scheduler_test.go
```diff
@@ -3432,7 +3434,7 @@ func TestCycleConsistency(t *testing.T) {
 				require.Equal(t, tc.expectedEventSequencesCycleThree, eventsCycleThree, "unexpected cycle three events")
 			}
 
-			// Test that the follower stays in sync with the leader.
+			// Test that the follower stays in sync with the leaderelection.
```
JamesMurkin:
```suggestion
			// Test that the follower stays in sync with the leader.
```

internal/scheduler/scheduler.go
```diff
@@ -276,8 +276,8 @@ func (s *Scheduler) cycle(ctx *armadacontext.Context, updateAll bool, leaderToke
 	}
 	ctx.Info("Finished syncing state")
 
-	// Only the leader may make decisions; exit if not leader.
-	// Only export metrics if leader.
+	// Only the leader may make decisions; exit if not leaderelection.
+	// Only export metrics if leaderelection.
```
JamesMurkin:
```suggestion
	// Only the leader may make decisions; exit if not leader.
	// Only export metrics if leader.
```

internal/scheduler/scheduler.go
```diff
@@ -46,8 +46,8 @@ type Scheduler struct {
 	// Responsible for assigning jobs to nodes.
 	// TODO: Confusing name. Change.
 	schedulingAlgo scheduling.SchedulingAlgo
-	// Tells us if we are leader. Only the leader may schedule jobs.
-	leaderController leader.LeaderController
+	// Tells us if we are leaderelection. Only the leader may schedule jobs.
```
JamesMurkin:
```suggestion
	// Tells us if we are leader. Only the leader may schedule jobs.
```

JamesMurkin (dismissed):

JamesMurkin (approved):

PR #4920, opened 2026-05-19
Emit JobRunTerminated event when executor observes pod terminal phase

dejanzele:
When the scheduler preempts a run, the pod takes a variable grace period (5s to 5min depending on `TerminationGracePeriodSeconds`) to actually exit. `preempted_timestamp` records when the scheduler made the decision, but nothing records when the kubelet finished stopping the containers, so any future retry-policy consumer has no per-run signal to wait on.

This PR adds that signal. It does not consume it; the retry-gate is a separate follow-up.

## What this PR does

Adds a new `JobRunTerminated` event emitted by the executor when it observes a pod reach `PodSucceeded` or `PodFailed`. The scheduleringester writes `runs.terminated_timestamp` from this event when the column is NULL, so the value reflects the kubelet-observed stop time rather than the scheduler decision time.

To preserve the existing Failed-state histogram, this PR also adds a `failed_timestamp` column (migration 039). `MarkRunsFailed` writes to `failed_timestamp` instead of `terminated_timestamp`, and `state_metrics.go` reads `FailedTime()` for the failed branch. The semantic split:

- `failed_timestamp`: when armada decided the run failed (scheduler event-time).
- `terminated_timestamp`: when the kubelet observed the pod stop (executor-observed).

For organic Succeeded or Cancelled runs the existing writers fill `terminated_timestamp` first and the `JobRunTerminated` write is a no-op. For preempted or organic-failed runs the column is NULL after the initial write, so `JobRunTerminated` fills it with the real kubelet-observed time.


dejanzele:
@greptileai

dejanzele:
@greptileai

dejanzele:
@greptileai

PR #4922, opened 2026-05-20
Lookout pruner: reconcile zombie jobs

mauriceyap:
Adds a reconciliation step to the lookout pruner that repairs "zombie" jobs left behind by a now-fixed ingester bug: rows where `job.state` is non-terminal (Queued/Leased/Pending/Running) but `latest_run_id` points to a `job_run` row in a terminal state (Succeeded/Failed/Cancelled/Preempted).

On each pruner run, before the existing deletion step, jobs whose latest run finished more than `prunerConfig.zombieRepairThreshold` ago have their `state` and `last_transition_time` updated to match the run. Default threshold is 15 minutes (a grace period for in-flight transitions and ingester lag). Set to 0 to disable this reconciliation entirely.

Repairs are batched (`prunerConfig.batchSize`), logged at info level (job ID, old -> new state), and counted in a new `lookout_pruner_zombie_jobs_repaired_total` Prometheus metric. Existing pruning behaviour is unchanged.

We log out all zombie jobs reconciled to enable further investigation. Ultimately, there should be no zombie jobs so the existence of them indicates a bug in Armada.

### DIFF
--- a/internal/lookout/pruner/reconcile_zombies.go
+++ b/internal/lookout/pruner/reconcile_zombies.go
@@ -1,0 +1,148 @@
+package pruner
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/pkg/errors"
+	"k8s.io/utils/clock"
+
+	"github.com/armadaproject/armada/internal/common/armadacontext"
+	"github.com/armadaproject/armada/internal/common/database/lookout"
+	log "github.com/armadaproject/armada/internal/common/logging"
+)
+
+// reconcileZombiesQuery is built once at package init from the lookout state
+// ordinals so the SQL stays the single source of truth for the mapping.
+//
+// Interpolating these ordinals into the SQL (rather than passing them as bind
+// parameters) is safe because they are compile-time constants from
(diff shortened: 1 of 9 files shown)


internal/lookout/pruner/reconcile_zombies.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// observeZombiesWithNullFinished counts and surfaces zombie jobs that the
+// reconciler cannot repair because their latest run has no finished timestamp.
+func observeZombiesWithNullFinished(ctx *armadacontext.Context, db *pgx.Conn) error {
+	var count int
+	if err := db.QueryRow(ctx, countZombiesWithNullFinishedQuery).Scan(&count); err != nil {
+		return errors.WithStack(err)
+	}
+	zombiesSkippedNullFinished.Add(float64(count))
```
ianhockett: It seems like this will increment `zombiesSkippedNullFinished` everytime it runs, thereby counting duplicates (unless I'm missing something). If we only care about how many zombos exist at that moment in total, then we could change the metric type from counter to guage.

ianhockett (dismissed):
lgtm

masipauskas (approved):

PR #4923, opened 2026-05-20
Add DryRunTxn to JobDb

JamesMurkin:
This is a new transaction type that allows local mutation but cannot have its changes committed back to the main db

The purpose of this is to allow side routines to have a functional (writable) transaction that they can use but has no risk of impacting the main db

This is precusor work to splitting the scheduling_algo into a separate go routine
 - It'll use jobdb snapshots to schedule with, but then communicate its desired changed back to the main thread rather than commit back to the jobdb directly to avoid conflicts. (Such as marking a job as scheduled, when it has already been cancelled)

### DIFF
--- a/internal/scheduler/jobdb/jobdb.go
+++ b/internal/scheduler/jobdb/jobdb.go
@@ -368,15 +368,15 @@ func (jobDb *JobDb) WriteTxn() *Txn {
 	}
 }
 
-// SnapshotTxn returns a writable transaction over an isolated snapshot of the current db state.
+// DryRunTxn returns a writable transaction over an isolated snapshot of the current db state.
 // Mutations are local to the snapshot and are never committed back to the db.
 // This is useful for processes that need to speculatively modify state without affecting the real db.
-func (jobDb *JobDb) SnapshotTxn() *Txn {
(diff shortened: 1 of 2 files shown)


internal/scheduler/jobdb/jobdb.go
```diff
@@ -368,6 +368,28 @@ func (jobDb *JobDb) WriteTxn() *Txn {
 	}
 }
 
+// SnapshotTxn returns a writable transaction over an isolated snapshot of the current db state.
+// Mutations are local to the snapshot and are never committed back to the db.
+// This is useful for processes that need to speculatively modify state without affecting the real db.
+func (jobDb *JobDb) SnapshotTxn() *Txn {
```
nikola-jokic: I'd personally suggest naming it something like DryRunTxn, because if I were to read the call-site code, I'd think it snapshots the current state and operates on it, mutating the db with the current state.
DryRun on the other hand tells me that this tx is not affecting the db.

Just a suggestion ☺️

nikola-jokic (dismissed):

nikola-jokic:
Maybe just update the title now and overall, all good :relaxed:

nikola-jokic (approved):

PR #4929, opened 2026-05-21
Local Development: Change VS Code to use Goreman

d2burkhalter:
#### What type of PR is this?
Enhancement for VS Code launch configurations that will align its debugger to use processes launched from Goreman with configs and docker compose in `/_local` to consolidate how we do local development. Also adds launch configurations for starting Armada with authorization and prometheus.

* add launch config for running with Prometheus
* add launch config for running with Authorization
* add launch config for running with Prometheus and Authorization
* fix race condition of migration task running before Postgres is ready
* add task for full teardown of local development resources
* consolidate local development to use `/_local`


j8169 (dismissed):

.vscode/launch.json
```diff
@@ -82,7 +110,28 @@
       "program": "${workspaceFolder}/cmd/lookout/main.go",
       "cwd": "${workspaceFolder}",
       "env": {
-        "ARMADA_CORSALLOWEDORIGINS": "\"http://localhost:3000,http://localhost:4173,http://localhost:8089,http://localhost:10000,http://example.com:10000,http://example.com:8089\"",
+        "ARMADA_CORSALLOWEDORIGINS": "http://localhost:3000,http://localhost:4173,http://localhost:8089,http://localhost:10000,http://example.com:10000,http://example.com:8089",
```
mauriceyap: good spot!

.vscode/tasks.json
```diff
@@ -19,6 +19,37 @@
...
+    {
+      "type": "shell",
+      "command": "mage kindTeardown && docker compose -f ${workspaceFolder}/docker-compose.yaml down -v",
+      "label": "Full teardown",
+      "problemMatcher": []
+    },
+    {
+      "type": "shell",
+      "command": "timeout 60 bash -c 'until pg_isready -h localhost -p 5432 -U postgres; do sleep 1; done'",
+      "label": "Wait for Postgres"
```
mauriceyap: I understand why this is necessary, but I think the real problem you've uncovered is that `mage startDependencies` doesn't wait for containers to healthy. It would be better to fix this in our docker setup rather than having this extra task in the VS Code config.

In `docker-compose.yaml`, in the `postgres` service definition, add a healthcheck (we already do this in `_local/` for the goreman setup:
```yaml
postgres:
  healthcheck:
    test: ["CMD-SHELL", "pg_isready -U postgres"]
    interval: 2s
    timeout: 5s
    retries: 30
```

Then in `magefiles/developer.go`, make `StartDependencies()` pass the `--wait` flag to Docker compose, so:
```go
command := append([]string{"compose", "up", "-d"}, services...)
```

.vscode/tasks.json
```diff
@@ -19,6 +19,37 @@
...
+    {
+      "type": "shell",
+      "command": "mage kind && ARMADA_EXTRA_DEPS=prometheus,keycloak mage startDependencies",
+      "label": "Start dependencies with Prometheus and Auth"
+    },
+    {
+      "type": "shell",
+      "command": "mage kindTeardown && docker compose -f ${workspaceFolder}/docker-compose.yaml down -v",
+      "label": "Full teardown",
+      "problemMatcher": []
```
mauriceyap, line 113: I think this is generally useful! Please could you add a Mage target which does this, and just run that instead? `mage teardownDependencies` perhaps

.vscode/tasks.json
```diff
@@ -19,6 +19,37 @@
...
+    },
+    {
+      "type": "shell",
+      "command": "timeout 60 bash -c 'until pg_isready -h localhost -p 5432 -U postgres; do sleep 1; done'",
+      "label": "Wait for Postgres"
+    },
+    {
+      "type": "shell",
+      "command": "timeout 60 bash -c 'until [ $(docker inspect -f \"{{.State.Health.Status}}\" keycloak) = healthy ]; do sleep 1; done'",
```
mauriceyap: See comment above - adding the `--wait` flag should mean you can delete this task.

docker-compose.yaml
```diff
@@ -45,6 +45,41 @@ services:
...
+    networks:
+      - kind
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    ports:
+      - "9090:9090"
+    volumes:
+      - "./developer/dependencies/prometheus.yml:/etc/prometheus/prometheus.yml"
+    command:
+      - "--config.file=/etc/prometheus/prometheus.yml"
```
mauriceyap: does this need a healthcheck?

.vscode/tasks.json
d2burkhalter: good callout

.vscode/tasks.json
d2burkhalter: reordered things so now the teardown is on line 123

mauriceyap (dismissed):

d2burkhalter:
@greptileai

magefiles/dev.go
```diff
@@ -27,24 +28,60 @@ const (
...
+			continue
+		}
+		switch token {
+		case "auth", "fake-executor":
+			if profile != "no-auth" {
+				fmt.Printf("warning: ignoring %q — profile already set to %q; only one of auth/fake-executor may be used\n", token, profile)
+			} else {
+				profile = token
+			}
+		case "debug":
```
dejanzele: Is it possible for this to be a flag (`--debug`)?

d2burkhalter: sure I'll make that change

d2burkhalter:
@greptileai

magefiles/dev.go
d2burkhalter: I can leave it as a bool `-debug` or change it to a string `--debug` if you really want there to be two dashes

_local/prometheus.yml
```diff
@@ -0,0 +1,39 @@
+global:
+  scrape_interval: 15s
+
+scrape_configs:
+  - job_name: "armada-server"
+    static_configs:
+      - targets: ["host.docker.internal:9000"]
```
dejanzele: host.docker.internal is a very OSX-specific feature, can we use something which would work on both Linux and OSX?

docs/developer_guide.md
```diff
@@ -157,29 +158,52 @@ and runs them as host processes, so you can attach a debugger to any component d
...
 
 GoLand runs the configurations in a compound in parallel, so `Run Migrations` starts alongside the components. The components retry their database and Pulsar connections until the migrations finish, so a short burst of connection errors at startup is expected.
 
-## VS Code debug configurations
+## VS Code Run and Debug configurations
 
 We similarly provide run and debug configurations for VS Code users to run each Armada service and use the debugger (provided with VS Code).
 
-The `Armada` configuration performs all required setup (setting up the Kind cluster, spinning up infrastructure services, performing database migrations) and then runs all services.
+The following compound configurations are provided, each launching all relevant service debuggers at once:
```
dejanzele, line 172: We might have many sections in the README.md which could go in this docs/developer_guide.md, could you debloat the README also?

magefiles/dev.go
dejanzele: honestly no opinion on `-` vs `--`, as long as it is a flag

_local/prometheus.yml
d2burkhalter: I think the extra_hosts property [here](https://github.com/d2burkhalter/armada/blob/d96604b1602248b705df8ec725bbe2168188335d/_local/compose/stack.yaml#L102-L103) should allow it to work the same on osx and linux
```yaml
  prometheus:
    container_name: prometheus
    image: ${PROMETHEUS_IMAGE:-prom/prometheus:v3.11.3}
    extra_hosts:
      - "host.docker.internal:host-gateway"
```

docs/developer_guide.md
d2burkhalter: I think we should have large doc changes in a separate PR. Want to try and get a consensus for what we want to do with `README.md`, `docs/` and `website/` and go from there

dejanzele (dismissed):

magefiles/dev.go
```diff
@@ -27,24 +28,61 @@ const (
...
-//	mage dev:up fake-executor  # no Kubernetes needed
-func (Dev) Up(profile string) error {
-	if profile == "" {
-		profile = "no-auth"
+//	mage dev:up ""                        # no-auth
+//	mage dev:up auth                      # OIDC
+//	mage dev:up fake-executor -debug      # fake executor + debug procfile
+//	mage dev:up auth,myservice            # auth + extra compose profile "myservice"
+//	mage dev:up "" -dap                 # no-auth + dap procfile
+func (Dev) Up(profiles string, dap *bool) error {
```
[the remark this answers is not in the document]
d2burkhalter: From https://magefile.org/targets/ 

> For *bool parameters, you can use the shorthand -name without a value, which is equivalent to -name=true

mauriceyap (approved):

dslear (approved):

PR #4931, opened 2026-05-21
LookoutUI: add analytics for custom view actions

ianhockett:
#### What type of PR is this?
  LookoutUI

  Adds analytics tracking for custom view create/load/delete actions in Lookout. Each event includes enriched metadata about the view's configuration (visible columns delta from defaults,
  filters, grouping, sort, page size, annotations, etc.) so we can understand how users configure their views.

  Also fixes a bug where orphaned annotation columns in saved preferences could inflate column counts.

  Fixes #

  - Builds on top of the `trackAnalyticsEvent` helper introduced in #4908
  - The `Analytics` component and `trackAnalyticsEvent` both accept a lazy `() => Record<string, string>` form for `eventData` to avoid computing metadata until the click actually fires
  - The `ensurePreferencesAreConsistent` fix removes stale `annotation_*` keys from `visibleColumns` when the annotation is no longer in `annotationColumnKeys` — this prevents ghost columns
  from appearing in analytics data


internal/lookoutui/src/pages/jobs/components/CustomViewPicker.tsx
```diff
@@ -94,9 +99,15 @@ export const CustomViewPicker = ({
...
+                      <Analytics
+                        component={Button}
+                        eventName={ANALYTICS_EVENTS.CUSTOM_VIEW_CREATED}
+                        eventData={() => getCurrentViewEventData(newCustomViewName)}
+                        variant="contained"
+                        onClick={addCustomView}
+                      >
                         Save
-                      </Button>
+                      </Analytics>
```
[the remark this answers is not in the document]
ianhockett: resolved in commit 59424a3e6c8b8

internal/lookoutui/src/analytics/viewMetadata.ts
```diff
@@ -0,0 +1,57 @@
...
+      added.push(colId)
+    } else if (!visible && isDefault) {
+      removed.push(colId)
+    }
+  }
+
+  // Annotation columns are always "added" (never in defaults)
+  const annotationCols = prefs.annotationColumnKeys ?? []
+  for (const key of annotationCols) {
+    const annotationColId = `annotation_${key}`
```
mauriceyap: this can be `toAnnotationColId(key)` (imported from `internal/lookoutui/src/common/jobsTableColumns.tsx`)

internal/lookoutui/.nvmrc
mauriceyap, line 1: I think it's probably okay to commit this, but please could you do it in a separate PR to keep this commit focused? It may also be good to call it out the UI's README (e.g. run `nvm use` first).

mauriceyap (approved):

PR #4933, opened 2026-05-28
Localdev: collapse setup into a single `mage dev:up` command

dejanzele:
Setting up Armada locally takes several manual steps today: install goreman, start the dependency containers, run an init script for the databases and priority class, then start goreman. This collapses that into one `mage dev:up` and moves the whole localdev surface under `_local/`.

```shell
mage dev:up                 # no-auth, goreman on the host
mage dev:up auth            # OIDC via keycloak
mage dev:up fake-executor   # no Kubernetes cluster needed
mage dev:deps               # just the dependency containers
mage dev:down               # stop everything
mage dev:full               # the whole stack in containers on Kind (what CI runs)
```

`dev:up` installs goreman to `./bin/` if it is missing (pinned to v0.3.15), brings up redis, postgres and pulsar from `_local/compose/stack.yaml`, runs `_local/scripts/init.sh` for migrations and the priority class, then runs goreman in the foreground with the chosen procfile. The init script skips the priority class when no cluster is reachable, so `fake-executor` works without one. `auth` adds `--profile auth` so keycloak comes up with the deps.

`dev:full` is the containerized counterpart: every component runs as a container against a Kind cluster, with hosts wired through environment overrides instead of `localhost`. CI uses this for the integration testsuite in place of `mage localdev minimal`, so the containerized flow is what proves the stack passing.

Most of the diff is files moving into `_local/`. The airflow compose stays a separate file under `_local/airflow/` rather than folding into the main stack: compose's `include:` cannot profile-gate the included services without editing them, which would break the standalone `docker compose -f _local/airflow/docker-compose.yaml up` that `mage Teste2eAirflow` runs.

Removed:
- The `mage localdev` and `mage localDevStop` targets (`magefiles/developer.go` and `magefiles/debug.go` deleted).
- Root `docker-compose.yaml` and `_local/docker-compose-deps.yaml`, replaced by `_local/compose/stack.yaml` and `_local/compose/full.yaml`.
- `developer/` (configs and env files, with airflow now under `_local/airflow/`).
- `e2e/setup/` and `e2e/config/`, now under `_local/kind/` and `_local/`.
- `scripts/localdev-init.sh`, now `_local/scripts/init.sh`.


dejanzele:
@greptileai

d2burkhalter (approved):

PR #4936, opened 2026-05-28
Localdev: add mage dev:* namespace (goreman-driven up/down/testsuite)

dejanzele:
Second PR in the localdev consolidation. **Depends on #4933 — please merge that first**; until then the diff here includes PR1's changes as a side effect.

Adds a `mage dev:*` namespace that gives humans and CI the same entry point into the goreman-driven flow.

`mage dev:up` installs goreman to `./bin/` if missing (pinned to v0.3.15), brings up dependencies via `_local/compose/stack.yaml`, runs `_local/scripts/init.sh` for migrations + priority class, then runs goreman in the foreground with the chosen procfile. Accepts `no-auth` (default), `auth`, or `fake-executor`. `mage dev:down` stops the dependency containers. `mage dev:deps` brings up only the deps for cases where goreman is run by hand.

`mage dev:testsuite` is the CI counterpart: same setup, but backgrounds goreman, waits for Armada to accept a job, runs the integration testsuite, and tears down. A follow-up will flip CI workflows from `mage localdev minimal` + `mage testsuite` to this single target — at which point the localhost/container-mode duality the existing setup carries goes away entirely, because Armada components are always host processes.

Old paths still work as the escape hatch. README's localdev section now leads with the single-command flow.

Validated locally: with no goreman installed, `mage dev:up` auto-installs it, brings deps up healthy, runs all migrations, and starts every component under goreman with clean shutdown on Ctrl+C.


dejanzele:
Folded into #4933 — one PR that actually delivers value (one-command Armada setup), per offline discussion.

PR #4939, opened 2026-05-31
Support Apache Airflow 2.10, 3.1 and 3.2 in the Airflow operator

dejanzele:
Before this change the operator was pinned to `apache-airflow>=2.6.3,<3.2.0`, and CI only ever tested the newest 3.x, so 3.2 was unsupported and 2.x was a claim with no test behind it.

The 3.2 ceiling was there because Airflow 3.2 moved `airflow.serialization.serde` to `airflow.sdk.serde`. The harder part was that `armada/__init__.py` registered Armada's serde types at import time. In 3.2, importing serde triggers provider discovery, which re-imports the armada provider while it is still initialising and deadlocks on a circular import. With the operator installed, this broke `import airflow` itself. Moving the registration to `model.py`, which is never imported during provider discovery, fixes it.

The version-divergent imports (the serde move, and `get_current_context` which moved to `airflow.sdk` in 3.0) now live in one place, `armada/_compat.py`, instead of being spread across modules.

CI now proves every supported combination. `third_party/airflow/scripts/run-tests.sh` takes an Airflow and Python version, builds a venv, installs Airflow against its official constraints file, installs the client and operator, and runs the unit tests plus the example-DAG imports. A new `airflow-compat` job runs it across the full grid of Airflow 2.10.5 / 3.1.8 / 3.2.1 and Python 3.10 / 3.11 / 3.12. The existing tox jobs still cover the newest 3.x. You can run the script locally for any pair, for example `run-tests.sh 3.2.1 3.12`.

`test/unit/test_triggers.py` also gains a serde round-trip test for the deferred trigger, which is the path the allow-list registration protects.

Removed or changed, for anyone depending on it:
- Floor raised from `2.6.3` to `2.10.0`. Airflow below 2.10 is no longer supported.
- Ceiling raised from `<3.2.0` to `<3.3.0`.
- The `airflow-min-version` CI job is replaced by `airflow-compat`, so the check name has changed.


dejanzele:
The operator supports both Airflow 2.x and 3.x (`apache-airflow>=2.6.3,<3.2.0`), so the example DAGs keep the `try/except` that prefers the 3.x `BashOperator` path and falls back to the 2.x one. To verify the 2.x support rather than assume it, this PR adds a CI job that runs the operator tests against the floor (Airflow 2.6.3 on Python 3.11), alongside the existing 3.x tox jobs.

dejanzele:
@greptileai

dejanzele:
@greptileai

dejanzele:
@greptileai

masipauskas (approved):

PR #4967, opened 2026-06-16
Clean up localdev consolidation leftovers

dejanzele:
#4933 moved localdev into the `mage dev` namespace under `_local/`, but a repo sweep found callers, docs, and IDE run configs still pointing at the old surface it deleted (the `LocalDev`/`runMigrations` mage targets, the root `docker-compose.yaml`, and the `developer/` configs). This PR finishes that migration: it repairs the broken references, consolidates setup logic that had drifted into several places, and refreshes the docs.

### Repaired references to removed targets and paths
- `.vscode/tasks.json` and `.vscode/launch.json` (the Scheduler and Server debug configs pointed at the deleted `developer/config/insecure-armada.yaml`).
- `.run/` GoLand configs: `Scheduler`, `Server`, `Infrastructure Services`, `Run Migrations`, and `Start Dependencies`.

### Pre-existing GoLand config bugs, found while validating
- `Run Migrations` and `Start Dependencies` used `/usr/bin/zsh` as the interpreter, which does not exist on macOS. Both now use `/bin/sh`.
- `Infrastructure Services` referenced its before-run task by the wrong name (`Docker Create Kind Network` instead of the actual `Create Kind Docker Network`), so it silently did nothing.

### Consolidated to one source per concern
- Databases are created by `_local/compose/postgres-init.sql`, now mounted by both `stack.yaml` and `full.yaml`. The duplicate `CREATE DATABASE` logic was removed from `init.sh`. This also fixes a fresh-machine bug where the host flow failed with "database does not exist".
- Priority classes come from `_local/kind/priorityclasses.yaml` alone. The apply step is now non-fatal when a class already exists with a different, immutable value.
- Migrations run through `mage dev:migrate`, which both the GoLand `Run Migrations` config and the VS Code task call. The GoLand `Start Dependencies` config likewise calls `mage kind dev:deps` directly, so the IDE configs invoke mage rather than wrapper scripts.

### Removed
- `.run/Armada Infrastructure Services`, `.run/schedulerPostgresMigration`, and `.run/lookoutPostgresMigration`. The `Armada` composite now uses `Run Migrations` for its migration step, and `mage dev:full` handles the containerized migrations.
- `scripts/infrastructure-start.sh` and `scripts/infrastructure-stop.sh`. They only wrapped `mage kind dev:deps` and `mage dev:down kindTeardown`, which the run configs now call directly.
- `_local/priority-class.yaml`, a partial duplicate of `_local/kind/priorityclasses.yaml`.

### Added
- `mage dev:migrate`, which runs the database init step on its own. It restores the standalone entry point that `mage runMigrations` used to provide.
- `mage dev:down` now passes `--profile auth --remove-orphans`, so `dev:up auth` no longer leaves the keycloak container orphaned.

### Docs
Updated the `mage localdev`/`LocalDev` commands to `mage dev:*`, fixed `_local/` paths and the GoLand setup section, removed the defunct "Pulsar-backed scheduler" and "Legacy Scheduler" references, and fixed several dead documentation links.


dejanzele:
@greptileai

d2burkhalter (dismissed):

dejanzele:
@greptileai

dejanzele:
@greptileai

mauriceyap (approved):
I have verified that this fixes the VS Code run config.

PR #4975, opened 2026-06-24
Create observability package introducing OTEL

nikola-jokic:
#### What type of PR is this?

Enhancement

Introuce base implementation for opentelemetry integration in Armada


internal/testsuite/trace_evidence.go
```diff
@@ -0,0 +1,207 @@
+package testsuite
```
JamesMurkin: Do we need this file at all?

internal/common/observability/config.go
```diff
@@ -0,0 +1,185 @@
...
+const (
+	ResourceAttributeServiceName      = "service.name"
+	ResourceAttributeServiceNamespace = "service.namespace"
+	ResourceAttributeServiceVersion   = "service.version"
+	ResourceAttributeServiceInstance  = "service.instance.id"
+)
+
+var validSamplers = map[string]bool{
+	SamplerParentBasedTraceRatio: true,
+	"trace_id_ratio":             true,
```
JamesMurkin: Why is only one of these a constant?

internal/common/observability/config.go
```diff
@@ -0,0 +1,185 @@
...
+
+import (
+	"fmt"
+	"maps"
+	"net/url"
+	"strings"
+)
+
+const (
+	DefaultOtlpHTTPEndpoint      = "http://localhost:4318"
```
JamesMurkin: Do we think we need this default code - we don't do this for other config typically (we rely on the config.yaml files to act as defaults)

internal/common/observability/config.go
nikola-jokic: It was the default, but since I removed defaults, I'll remove this as well, good call!

JamesMurkin (approved):

PR #4978, opened 2026-06-24
Improve the efficiency of updating job prices

JamesMurkin:
Before this PR updating job prices is very inefficient, we'd update the prices of ALL jobs every round if they need it or not
 - Even at modest scale if ~250k jobs, this can take 5-10 seconds

Now the code will only update jobs where prices have changed since the last set of prices
 - Price snapshots have an Id so we can quickly skip if the snapshot has not changed
   - This is the common case, as the snapshot is cached, so often it is simply returning the value we saw last cycle
 - When the snapshot has changed, we check which prices have changed and only update jobs that use those prices

This should mean we go from updating the whole db each cycle to no changes most the time, and targetted updates when prices have changed.

### DIFF
--- a/internal/scheduler/pricing/bid_service.go
+++ b/internal/scheduler/pricing/bid_service.go
@@ -3,6 +3,8 @@ package pricing
 import (
 	"time"
 
+	"github.com/google/uuid"
+
 	"github.com/armadaproject/armada/internal/common/armadacontext"
 	"github.com/armadaproject/armada/internal/scheduler/internaltypes"
 	"github.com/armadaproject/armada/internal/scheduler/queue"
@@ -31,6 +33,7 @@ func NewLocalBidPriceService(pools []string, queueCache queue.QueueCache) *Local
 
 func (b *LocalBidPriceService) GetBidPrices(ctx *armadacontext.Context) (BidPriceSnapshot, error) {
 	snapshot := BidPriceSnapshot{
+		Id:        uuid.NewString(),
 		Timestamp: time.Now(),
 		Bids:      make(map[PriceKey]map[string]Bid),
 	}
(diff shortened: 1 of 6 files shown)


internal/scheduler/pricing/bid_service.go
```diff
@@ -31,6 +33,7 @@ func NewLocalBidPriceService(pools []string, queueCache queue.QueueCache) *Local
 
 func (b *LocalBidPriceService) GetBidPrices(ctx *armadacontext.Context) (BidPriceSnapshot, error) {
 	snapshot := BidPriceSnapshot{
+		Id:        uuid.NewString(),
 		Timestamp: time.Now(),
```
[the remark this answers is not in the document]
JamesMurkin: This is fine, LocalBidPriceService should emulate real data, which will have a new id each request. 

Only used for local dev (in dev mode) so no concern for prod and minimal concern for dev.

internal/scheduler/scheduler.go
```diff
@@ -534,61 +533,64 @@ func (s *Scheduler) syncState(ctx *armadacontext.Context, initial, fullJobGc boo
...
-			if slices.Contains(s.marketDrivenPools, pool) {
-				return true
-			}
-		}
-		return false
+	// The provider caches the snapshot between refreshes, so most cycles see the same one.
+	// A matching, non-empty Id guarantees identical content, so we can skip all the work.
+	previousSnapshot := txn.GetBidPriceSnapshot()
+	if previousSnapshot != nil && previousSnapshot.Id != "" && previousSnapshot.Id == updatedBids.Id {
+		ctx.Logger().Infof("bid price snapshot %s unchanged, skipping job price update", updatedBids.Id)
```
robertdavidsmith, line 550: Will this produce a lot of spam?

robertdavidsmith (approved):

internal/scheduler/pricing/types.go
```diff
@@ -20,6 +21,10 @@ type PriceKey struct {
 }
 
 type BidPriceSnapshot struct {
+	// Id uniquely identifies this snapshot.
+	// Two snapshots with the same non-empty Id are guaranteed to hold identical Bids and ResourceUnits.
+	// An empty Id means "unidentified" and should never be treated as equal to another.
+	Id            string
```
nikola-jokic, line 27: Why not uuid.UUID

JamesMurkin: No good reason, will fix in follow-up PR shortly

PR #4992, opened 2026-07-02
Bump mage to 1.17.2 everywhere, guard dev:up against old mage versions, and fix localdev doc drift

dejanzele:
`mage dev:up` looks broken on older mage binaries. The target takes an optional `-dap` flag through a pointer argument (`dap *bool`), and optional arguments only exist since mage v1.16. An older mage (for example the 1.15.0 that Homebrew shipped through 2023) does not reject the signature. It silently skips the target during parsing, with a message that only appears under `mage -debug`. The result is that `dev:up` vanishes from `mage -l` and every documented command fails with `Unknown target specified: "dev:up"`, which reads like a bug in the magefile rather than a stale local tool.

The magefile cannot intercept that failure, because an old mage never registers the target in the first place. So this PR adds three layers instead:

- The mage pin is bumped from v1.15.0 to v1.17.2 everywhere it exists: in go.mod (the magefiles import the mage library) and in the 21 CI workflow invocations of `go run github.com/magefile/mage@<version>`. Anyone without a new enough mage binary can use that same form: `go run github.com/magefile/mage@v1.17.2 <target>`.
- `mage checkDeps` now checks the mage binary on PATH against `>= 1.16.0`, using the same semver-constraint helper as the other tool checks. Verified that checkDeps itself still runs under mage 1.15 (it takes no optional arguments), so it can explain the problem on exactly the machines that have it.
- The localdev docs say what the `Unknown target` failure means and point at both remedies.

Two small fixes in dev.go: the `no-auth` token is now treated as the default profile instead of being forwarded to docker compose as a bogus `--profile no-auth` flag (the VS Code tasks pass `no-auth` explicitly), and the doc comment examples now match the real CLI (the stale `-debug` becomes `-dap`, and `mage dev:up no-auth` replaces `mage dev:up ""`).

The localdev documentation is also restructured. The main README's Local Development section is now a short quick start (the four `mage dev:*` commands) plus links onward. The full reference (how `dev:up` works, profile tokens, the auth profile details, the fake executor, compose profiles, procfiles, and the service ports table) moved into the local dev stack section of `docs/developer_guide.md`, mostly verbatim; nothing was dropped. A new `_local/README.md` describes the directory layout: what each subdirectory and file in `_local` is for.

Doc corrections made while cross-referencing against the magefile (now mostly living in `docs/developer_guide.md`):

- All `dev:up` examples include the required profile argument (`mage dev:up no-auth`); the previously documented bare `mage dev:up` never worked on any mage version because mage has no optional positional arguments.
- Ports: the Server Metrics row said 9000 but `_local/server/config.yaml` sets 9009. Added rows for Scheduler HTTP (8080), Lookout API (8089), Pulsar admin (8090), Keycloak (8180, `auth` profile) and Prometheus (9090, `prometheus` profile).
- The compose profiles table lists the `prometheus` profile, the procfiles section documents the `-dap` variants, and the `dev:deps`, `dev:migrate`, `dev:full` and `dev:fullDown` targets are covered.
- `docs/developer_guide.md`: the "Setting up the local dev stack" section described the old localdev flow (bootstrapping tools.yaml, creating a kind cluster), which `dev:up` does not do. It now describes the goreman and containerized flows accurately and links to `_local/README.md`. The UI section says which port belongs to which flow: 3000 for the Vite dev server under goreman, 8089 for lookout under `dev:full`. The debugging section documents the `-dap` flag, and the prerequisites note the mage version and the `go run github.com/magefile/mage@v1.17.2` alternative.
- The default and auth profiles run a real executor, which needs a Kubernetes cluster, but no doc said so: only the VS Code tasks knew to run `mage kind` first, and without `KUBECONFIG` the executor falls back to the default kubeconfig and connects to whatever cluster that selects. The README quick start and the developer guide now say to run `mage kind` and set `KUBECONFIG=.kube/external/config`.
- The guide's "Testing if the local dev stack is working" and "Running the UI" sections were nested under the "port 6443" troubleshooting section, where a newcomer would not look. They are now subsections of "Setting up the local dev stack", and the guide's table of contents is regenerated (it listed a section that no longer exists and missed the current ones).
- `_local/compose/stack.yaml`: the header comment said container-mode profiles were "intentionally not in this PR" with a follow-up planned. That follow-up shipped as `full.yaml`, so the comment now points there, and the profile and image override lists include prometheus.

Removed:

- README ports table rows for OTEL Collector gRPC (4317), OTEL Collector HTTP (4318) and Jaeger UI (16686). No otel-collector or jaeger service exists anywhere in `_local`, so the rows described services you cannot reach.
- The "Choosing components to run" section in `docs/developer_guide.md`. `ARMADA_COMPONENTS` has no remaining references in the codebase.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/airflow-operator-release-to-pypi.yml
+++ b/.github/workflows/airflow-operator-release-to-pypi.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           version: '23.3'
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-      - run: go run github.com/magefile/mage@v1.15.0 -v airflowOperator
+      - run: go run github.com/magefile/mage@v1.17.2 -v airflowOperator
       - uses: ./.github/workflows/python-tests
         with:
           python-version: '3.10'
--- a/.github/workflows/airflow-operator.yml
+++ b/.github/workflows/airflow-operator.yml
@@ -92,7 +92,7 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       # Generates the armada_client proto code (typings, *_pb2) that the operator imports.
       # Without this the local client install is incomplete and the tests fail to collect.
-      - run: go run github.com/magefile/mage@v1.15.0 -v buildPython
+      - run: go run github.com/magefile/mage@v1.17.2 -v buildPython
       - name: Run operator tests against Airflow ${{ matrix.airflow }}
         run: third_party/airflow/scripts/run-tests.sh ${{ matrix.airflow }} ${{ matrix.python }}
 
@@ -119,7 +119,7 @@ jobs:
         with:
           version: '23.3'
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-      - run: go run github.com/magefile/mage@v1.15.0 -v airflowOperator
+      - run: go run github.com/magefile/mage@v1.17.2 -v airflowOperator
       - uses: ./.github/workflows/python-tests
         with:
           python-version: ${{ matrix.python }}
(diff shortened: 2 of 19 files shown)


dejanzele:
@greptileai

d2burkhalter (approved):

PR #4994, opened 2026-07-06
Add requestor info and display it for job events

nikola-jokic:
#### What type of PR is this?

Enhancement

Include requestor information on job events. Cancelling for example is generated by the scheduler rather than on the API, so requestor information should be propagated.


nikola-jokic:
@copilot resolve the merge conflicts in this pull request

PR #4998, opened 2026-07-08
Add category-based retry policy engine and RetryPolicy API resource

dejanzele:
Introduces the retry policy engine as a standalone library and the `RetryPolicy` API resource it operates on. Nothing is wired into the scheduler in this PR: the engine has no call sites yet, so merging it is a no-op at runtime. The scheduler wiring and the CRUD service land in follow-up PRs that stack on this one.

The engine decides whether a failed job run should be retried or permanently failed. It matches a run's failure category (and an optional subcategory) against a policy's rules, first match wins. A matched rule either retries or fails. If no rule matches, the policy's default action applies. Two limits bound retries: a per-policy `retryLimit` and a global cap. Preemptions are excluded from both budgets, because a preemption is the scheduler's own action rather than a job failure, so it must not consume a job's genuine-failure retry budget.

The proto changes add the `RetryPolicy`, `RetryRule`, and `RetryAction` messages plus the `RetryPolicyService` definition to `submit.proto`. The service is defined here so the whole retry proto surface lands in one reviewable place; its implementation lives in the CRUD PR that stacks on this one. `deserialize.go` learns to parse `RetryAction`.


dejanzele:
@greptileai

dejanzele:
@greptileai

dejanzele:
@greptileai

masipauskas (dismissed):

dejanzele:
@greptileai

JamesMurkin (approved):

PR #4999, opened 2026-07-08
Delete failed pods matching a Delete-action category so retries can reuse the name

dejanzele:
Adds an `action` field to executor failure categories, with two values: `Retain` (the default, today's behaviour: a failed pod is kept for debugging) and `Delete`. A failed pod whose matched category has `action: Delete` is deleted so a scheduler retry of the same job can reuse the deterministic pod name without colliding with its predecessor.

Delete-action pods are handled by the pod issue handler, the same machinery that runs failed pod checks. The state reporter registers an issue instead of emitting the terminal event itself, and the issue handler's periodic loop then deletes the pod, waits until the informer confirms it is gone, and only then reports the terminal event with the pod's failure category and subcategory. Because that report is what triggers a scheduler retry, the retry can never race the deletion: the pod name is free before the scheduler learns of the failure. A failed delete or a failed report is retried every cycle from the issue's pod snapshot.

Failed pod checks are not altered. The checker, its detection function, and both existing issue handlers have zero changes; the new detection runs only after the checker declines a pod, so checks keep first claim on everything they match today. This is deliberate: the intended use of `action: Delete` is to replace failed pod checks one at a time, by removing a check from `failedChecks` and adding an equivalent category with `action: Delete` plus a scheduler retry policy for that category. Each migration is config-only and reversible by restoring the check. The checks themselves are only removed once every check has been migrated and validated.

Two operational notes. A `Delete` category without a matching scheduler retry policy turns those failures terminal instead of retried, so the executor and scheduler config must change together when migrating a check. And the mechanism is sized for check-replacement volume, which is operator-curated infrastructure failures; `action: Delete` is not intended for broad categories such as user errors.

It also adds two `errormatch` conditions, `LeaseExpired` and `AppError`, used by the classifier to categorise the corresponding pod failures.

### DIFF
--- a/internal/common/errormatch/doc.go
+++ b/internal/common/errormatch/doc.go
@@ -8,6 +8,6 @@
 // Condition constants ([ConditionOOMKilled], [ConditionEvicted],
 // [ConditionDeadlineExceeded]) and the [KnownConditions] map are provided
 // for config validation. Non-pod conditions ([ConditionPreempted],
-// [ConditionLeaseReturned]) are also defined here for use by the retry
-// engine but are not included in [KnownConditions].
+// [ConditionLeaseReturned], [ConditionLeaseExpired], [ConditionAppError]) are also defined here
+// for use by the retry engine but are not included in [KnownConditions].
 package errormatch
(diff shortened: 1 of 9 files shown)


dejanzele:
@greptileai

masipauskas (dismissed):

dejanzele:
@greptileai

JamesMurkin (approved):

PR #5001, opened 2026-07-08
Wire the retry policy engine into the scheduler behind a feature flag

dejanzele:
## Problem

A failed run is retried today only when a failed pod check matches it. The checks answer yes or no, and a retry returns the lease silently: no failure category on the event, no per-queue control, no budget per failure kind, and no way to change the job before it runs again. Operators cannot express "retry OOM kills with more memory" or "give this queue three retries for infrastructure failures, away from the node that failed".

## Earlier work this builds on

- #4998 added the retry engine and the RetryPolicy API resource.
- #4999 made the executor delete a failed pod in a Delete-action category and report the failure only after the pod is gone, so a retry can reuse the pod name.
- #5049 and #5048 added the RetryPolicy CRUD service, client, and armadactl commands.
- #5080 and #5081 made pod-level failure messages and Kubernetes events matchable by category rules.
- #5012 added the mutate block on retry rules and the pipeline that delivers memory growth to the pod.

## This PR

This PR connects those pieces into scheduler behaviour, behind the scheduling.retryPolicy feature flag. With the flag off, the scheduler produces the same event sequences as before, byte for byte, and identity tests pin this. With the flag on, a failed run consults the policy attached to its queue. The policy decides whether the categorized failure retries, and its verdict replaces the attempt counting for those runs. A retryable failure emits an intermediate event with retryable=true. The wiring also applies the matched rule's mutations at requeue: avoidSameNode steers the retry away from every node the job failed on, and a memory bump grows the job so placement, accounting, and the retried pod all see the new size. Gangs and fail-fast jobs keep their existing behaviour.

One operator requirement: avoidSameNode expresses the avoidance through the scheduler's nodeIdLabel, so that label must be in the executor's trackedNodeLabels.

Two small event-stream changes apply with the flag off as well, both additive. A terminal lease expiry now emits a JobFailedEvent with the reason "Lease expired" instead of an empty reason. Failure category fields now appear on all failure reason kinds, not only pod errors. Consumers that match on empty reason strings will see the new values.

## Validating the full loop

The OOM-with-memory-bump loop is the best single check. Enable the flag and give a category for OOMKilled conditions action Delete, then:

```yaml
# armadactl create retry-policy -f policy.yaml
apiVersion: armadaproject.io/v1beta1
kind: RetryPolicy
name: oom-grow
retryLimit: 2
defaultAction: Fail
rules:
  - action: Retry
    onCategory: oom
    mutate:
      resources:
        memory:
          factor: 2
# armadactl create queue oom-grow-queue --retry-policies oom-grow
```

Submit a job with a 64Mi memory limit that writes 100Mi into a memory-backed emptyDir. Expect: the run fails with category oom and a retryable=true event, the pod disappears before that event arrives, and the same pod name comes back with a 128Mi limit and succeeds. Lookout shows two runs: one failed with the category, one succeeded. This sequence was validated on a live cluster, together with avoidSameNode relocating a retry to a second node and failing the job cleanly once every node had been tried.


dejanzele:
@greptileai

dejanzele:
@greptileai

dejanzele:
@greptileai

docs/retry_policies.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+Deletion is rejected while any queue still references the policy. Detach it from all queues first, then delete it.
+
+Managing policies requires the `create_retry_policy`, `update_retry_policy`, and `delete_retry_policy` permissions. Grant them through the server's permission group mapping; without them the corresponding CRUD calls return `PermissionDenied`.
+
+## Rollout guide for operators
+
+**Configure `action: Delete` on retried categories before enabling the flag.** A retry reuses the failed attempt's pod name, so the executor must delete the failed pod to free the name (see [Pod naming and collision avoidance](#pod-naming-and-collision-avoidance)). Any failure category a retry rule matches on must set `action: Delete` in the executor's categorizer config. If it does not, the retry collides with the retained pod, the lease is returned as a recoverable submit error, and the job eventually fails terminally with a misleading `MaxRunsExceeded` reason (see [Pod naming and collision avoidance](#pod-naming-and-collision-avoidance) for the full failure mode). Audit the categorizer config against your retry rules before turning the flag on.
```
masipauskas: Code should validate this is the case, and fail with retry policy being created which would violate this.

docs/retry_policies.md
```diff
@@ -0,0 +1,186 @@
...
+
+This has an operational consequence: **every failure category that a retry rule matches on must be configured with `action: Delete` on the executor.** If a retried category is left as the default `action: Retain`, the retained pod causes the retry's lease to fail with an `AlreadyExists` error. That surfaces as a recoverable submit error, so the run's lease is returned to the scheduler. A returned lease is not a categorized pod failure, so the retry engine does not decide it and it falls through to the legacy attempt-limit path. Once the legacy attempt limit is hit the job fails terminally with a `MaxRunsExceeded` reason that does not mention the collision, so the real cause is easy to miss. Collision handling also deletes the retained pod, so the debugging evidence that `Retain` was meant to preserve is gone anyway.
```
masipauskas, line 152: Code should validate this is the case, and fail with retry policy being created which would violate this.

docs/retry_policies.md
```diff
@@ -0,0 +1,186 @@
...
+        avoidSameNode: true
+  - action: Retry
+    onCategory: oom
+    mutate:
+      resources:
+        memory:
+          factor: 1.5
+```
+
+* `affinity.avoidSameNode`: when `true`, the retry avoids every node a previous run attempted. This matches the lease-return retry behaviour: the job fails if the anti-affinity makes it unschedulable. The check costs a per-job scheduling probe. Leave it off (the default) for categories where the node is not the cause, for example a plain application error. Turn it on for node-specific failures. The scheduler expresses the avoidance through its `nodeIdLabel`, so that label must be in the executor's `trackedNodeLabels`. When the label is not tracked, the scheduler cannot see it on any node, and the avoidance matches every node without effect.
```
masipauskas: Can we make this invariant in validation? Also this should also be in indexed node labels in scheduler config, as we expect a lot of jobs using this label.

internal/scheduler/configuration/validation.go
```diff
@@ -22,6 +22,10 @@ func (c *Configuration) Mutate() (config.Config, error) {
 		c.Scheduling.MaxNewJobSchedulingDuration = c.NewJobsSchedulingTimeout
 	}
 
+	if c.Scheduling.RetryPolicy.Enabled && c.Scheduling.RetryPolicy.GlobalMaxRetries == 0 {
+		log.Warnf("scheduling.retryPolicy.enabled is true but globalMaxRetries is 0: the retry engine is active but will never grant a retry (kill-switch semantics). Set globalMaxRetries above 0 to allow policy retries.")
```
masipauskas, line 27: should we not allow this? i.e. fail in this case - as config is invalid? Why would I want retries enabled, but disabled at the same time?

internal/scheduler/scheduler.go
```diff
@@ -25,13 +30,35 @@ import (
...
 )
 
+// retryPolicyDecisionsCounter counts every retry engine verdict, labeled by
+// which gate produced it (see retry.Decision for the label values).
+var retryPolicyDecisionsCounter = promauto.NewCounterVec(
+	prometheus.CounterOpts{
+		Name: "armada_scheduler_retry_policy_decisions_total",
+		Help: "Retry policy engine decisions by policy and decision outcome. The queue is recorded in the scheduler log.",
+	},
+	[]string{"policy", "decision"},
```
masipauskas: are we intentional on not exposing retries by queue+policy?

internal/scheduler/scheduler.go
```diff
@@ -633,7 +688,91 @@ func (s *Scheduler) addNodeAntiAffinitiesForAttemptedRunsIfSchedulable(ctx *arma
...
+	}
+	if newSchedulingInfo == nil {
+		ctx.Warnf("skipping memory bump for job %s: the bump kind differs from the job's accumulated record", job.Id())
+		return job, true, nil
+	}
+	bumpedJob, err := job.WithJobSchedulingInfo(newSchedulingInfo)
+	if err != nil {
+		return nil, false, err
+	}
+	results, _, err := s.submitChecker.Check(ctx, []*jobdb.Job{bumpedJob})
```
masipauskas, line 762: does it keep submit checker caching? as it can get very slow, very quickly?

masipauskas (dismissed):

docs/retry_policies.md
dejanzele: Good point on both. Config validation now fails startup when the retry engine is enabled and nodeIdLabel isn't in indexedNodeLabels. That part is inside the scheduler config, so we can enforce it. The trackedNodeLabels half is executor config the scheduler can't see, so it warns once per executor when reported nodes are missing the label. Proper validation there needs the executors to report their config first, which I'm planning as a follow-up.

internal/scheduler/configuration/validation.go
dejanzele: It's intentional. With the flag on and the cap at 0 the engine still runs and attributes failures, it just never grants anything. Turning the flag off instead changes what events new failures produce. So the 0 cap is how you freeze retries during an incident without touching event behaviour.

docs/retry_policies.md
dejanzele: Agreed, but the server can't see the executors' categorizer configs today. Validating this at policy creation needs the executors to report their categories first, probably on the lease request. That's a protocol change I'd rather do as a follow-up. Until then the rollout guide carries it.

internal/scheduler/scheduler.go
dejanzele: Yes, `Check` runs against the submit checker's cached snapshot, no live calls per job. The per-job calls under a mass failure are still a fair concern though. Check takes a batch, so the plan is a follow-up that collects a cycle's retry candidates and probes them in one call.

docs/retry_policies.md
dejanzele: Same as the other thread, this needs the executors to report their category configs first. Planning it as a follow-up.

internal/scheduler/scheduler.go
dejanzele: The scheduler's state metrics already track error classification by queue and pool, and the periodic reset keeps the cardinality bounded. I've moved this counter into the same module, so it now carries queue, pool, policy and decision.

internal/scheduler/jobdb/job.go
```diff
@@ -861,6 +864,21 @@ func (job *Job) AllRuns() []*JobRun {
...
+// which is what the retry engine charges against a policy's retry budgets. A
+// preempted or lease-returned run is marked failed but never ran to a genuine
+// failure, so it is excluded: neither is something the job did. Deriving this
+// from run history keeps it correct across scheduler restarts.
+func (job *Job) FailureCount() uint32 {
+	count := uint32(0)
+	for _, run := range job.runsById {
+		if run.failed && !run.everPreempted && !run.returned {
+			count++
+		}
```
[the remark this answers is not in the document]
dejanzele: @greptileai we're introducing the retry engine only to replace the executor's failed pod checks, so it only decides categorized run failures and lease expiries. Preemption stays terminal for the job, so this count is never read again after those paths run. And after a restart the bit is restored from the ingested JobRunPreempted event, so a job that lives past an executor-reported preemption keeps its budget.

masipauskas (approved):

internal/scheduler/scheduler.go
dejanzele: The follow-up is up: #5097. One engine evaluation per job, then at most two batched Check calls for the whole round, and the queue-policy map builds once per cycle. Validated with the e2e retry cases against a live stack.

PR #5012, opened 2026-07-13
Add avoidSameNode and memory bump mutations to retry policy rules

dejanzele:
This PR adds the mutate block to retry-policy rules and the pipeline that delivers mutations to the pod. It builds against master. Like the rest of the retry-policy model, it stays dormant until the scheduler wiring (#5001) merges. The wiring PR consumes this model and applies mutations at requeue time. For that reason #5001 builds on this PR, and the two merge in that order.

A Retry rule can carry a mutate block with two groups:

```yaml
rules:
  - action: Retry
    onCategory: oom
    mutate:
      resources:
        memory:
          factor: 1.1        # multiply memory by 1.1 on each retry
  - action: Retry
    onCategory: internal
    onSubcategory: node-failure
    mutate:
      affinity:
        avoidSameNode: true  # steer the retry away from every attempted node
```

affinity.avoidSameNode matches the lease-return retry behaviour: the retry avoids every node a previous run attempted, and the job fails if that makes it unschedulable. Rules that do not set it skip the per-job scheduling probe, which is expensive under mass failure.

resources.memory grows the retried job's memory. Set exactly one of factor (multiply, must exceed 1.0) or static (add a fixed quantity, e.g. "512Mi"). Validation rejects malformed bumps at write time. The bump compounds across retries: each retry grows the amount the previous retry produced. The per-policy retryLimit bounds the number of steps.

The delivery pipeline exists because pods are built from the submitted spec, not from the scheduler's internal view. A bump that only changed the scheduler's reservation would leave the container limit unchanged, and the OOM would repeat. The pipeline has three steps, all inside the scheduler. The job's scheduling info records the total growth: at most one of a factor product or a static sum, persisted with the job. Every run lease carries it in the PodRequirementsOverlay. The scheduler's executor API applies it to the pod spec when it serves the lease. The executor receives a finished spec and submits it. It knows nothing about mutations. The factor multiplies every container's memory. The static amount spreads across the main containers in proportion to their requests. The pod total then grows by exactly the static amount and stays consistent with the scheduler's reservation. Each init container receives the full static amount, because init containers run alone and each must fit the reserved total on its own. Requests and limits move together.

Both groups extend without new shapes: new placement mutations become fields on the affinity message, and new resources become fields on the resources message.


dejanzele:
@greptileai

williamvega (approved):

PR #5018, opened 2026-07-14
add e2e test cases for node cancel and preempt

williamvega:
#### What type of PR is this?

feature

This is the node-scoped integration test portion of expanding our integration tests.

This pr adds a node to the local kind config used in our integration tests pipeline (ci.yaml). This node has a noschedule taint added. There are some new node scoped integration tests added to testcases as well. In combination, these tests and the extra node ensure that node scoped cancel and preempt operations do not affect other jobs or tests running. Additionally, changes to ci.go run the node-scoped tests separately and produce new junit test artifacts. Later workflows pick up all junit artifacts and search for failures.

Changes to the test runner include decoupling the state of the job from the action, allowing for jobs to be acted upon when they reach a certain state. For instance, we can cancel a job when it is in the queued state instead of relying on an arbitrary delay, while preempt jobs are cancelled when they are in the running state. This should improve reliability and reproducibility of tests.

Additionally, the node client is brought in to perform node scoped api calls. Also, a single job's succeed to succeed transition is now considered to be no-op.

**Reviewer Notes**
* Added tests as 1x5 so they are more reliable in cases where all jobs need to be leased before the node client api call can be made. Can't really guarantee that even 10 tests will run concurrently on the CI runner, testing cancelling/preempting 5 jobs and 10 jobs tests the same thing. It has failed on 10 jobs before, at the 10th job.


williamvega:
@greptileai

internal/testsuite/testrunner.go
```diff
@@ -279,42 +329,28 @@ func preemptJobsWhenRunning(ctx context.Context, eventCh chan *api.EventMessage,
...
+	nodes, err := k8sClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to list nodes with label %s", labelSelector)
+	}
+	if len(nodes.Items) == 0 {
+		return "", errors.Errorf("no node found with label %s", labelSelector)
+	}
+	if len(nodes.Items) > 1 {
+		fmt.Printf("warn: multiple nodes match label %s; using %s\n", labelSelector, nodes.Items[0].Name)
+	}
```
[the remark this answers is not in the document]
williamvega: @greptileai

That is why we warn the user. One node label should equal one node.

internal/testsuite/testrunner.go
```diff
@@ -191,86 +183,144 @@ func (srv *TestRunner) Run(ctx context.Context) (err error) {
...
+				if len(triggeredJobs) == len(jobIds) {
+					time.Sleep(1 * time.Second)
+					var actionErr error
+					switch {
+					case testSpec.Action == api.TestSpec_ACTION_CANCEL && testSpec.Selection == api.TestSpec_SELECTION_BY_NODE:
+						actionErr = client.WithNodeClient(conn, func(nc api.NodeClient) error {
+							_, err := nc.CancelOnNode(ctx, &api.NodeCancelRequest{
+								Name:     nodeName,
+								Executor: testSpec.GetExecutor(),
+								Queues:   []string{testSpec.GetQueue()},
```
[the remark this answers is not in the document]
williamvega: @greptileai

These tests are run only on one node and sequentially in ci.go.

There are future considerations for adding some concept of concurrency within the test spec to solve this behavior, in a later and separate ticket.

internal/testsuite/eventwatcher/eventwatcher.go
```diff
@@ -253,7 +253,8 @@ func ErrorOnNoActiveJobs(parent context.Context, C chan *api.EventMessage, jobId
 				numActive++
 			} else if e := msg.GetSucceeded(); e != nil {
 				if _, ok := exitedByJobId[e.JobId]; ok {
-					return errors.Errorf("received multiple terminal events for job %s", e.JobId)
+					fmt.Fprintf(os.Stderr, "warn: duplicate succeeded event for job %s\n", e.JobId)
```
dejanzele: how can we get a duplicate succeeded event?

williamvega: Fair question, I was thinking from the perspective that duplicate terminal events in general shouldn't cause test failures. That said, in practice I have only seen duplicate failure events.

williamvega: I'm not sure if there is one definition of terminal but from the code/docs I see these states treated as terminal:
- cmd/broadside/README.md: "Terminal jobs (succeeded, failed, cancelled) are not affected"
- docs/developer/job-lifecycle-events.md: "the four terminal flows: succeeded, failed, preempted, cancelled."

So for testing purposes, I think a duplicate terminal event for a job that's already reached one of these states should be a no-op rather than failing the test. If not, then we are testing two things, the first that the job transitioned through the states correctly, and the second that no additional state transitions were attempted. I think we only want to test the first.

_local/kind/cluster.yaml
```diff
@@ -8,6 +8,16 @@ networking:
 nodes:
 - role: worker
   image: kindest/node:v1.28.15
+- role: worker
+  image: kindest/node:v1.28.15
```
dejanzele, line 12: I think we can move to a newer image (`v1.33.2` is latest, we should check which version might be best to use)

internal/testsuite/eventwatcher/eventwatcher.go
dejanzele: Hmmm I don't think preempted is a terminal event, I think it gets followed by a failed, which is a terminal event.

magefiles/ci.go
```diff
@@ -34,44 +56,41 @@ func TestSuite() error {
...
-	timeTaken = time.Now()
-	out, err = goOutput("run", "cmd/testsuite/main.go", "test",
-		"--tests", "testsuite/testcases/preemption/*",
-		"--junit", "junit-preemption.xml",
-		"--config", "_local/.armadactl.yaml",
-	)
-	fmt.Println(out)
-	if err != nil {
-		return err
+	runs := []testRun{
```
dejanzele: this is very nice!

_local/kind/cluster.yaml
williamvega: that bump breaks more things than expected so I'll do it in future pr

internal/testsuite/eventwatcher/eventwatcher.go
williamvega: I see, so preempted is a terminal state but not a terminal event (JobRunPreempted/JobPreemptedEvent is info, JobErrors/JobFailedEvent is the real terminal event). 

Updated the testing code to reflect this.

internal/testsuite/eventwatcher/eventwatcher.go
```diff
@@ -253,23 +252,17 @@ func ErrorOnNoActiveJobs(parent context.Context, C chan *api.EventMessage, jobId
 				numActive++
 			} else if e := msg.GetSucceeded(); e != nil {
```
dejanzele: can we use a `switch` instead of many elifs?

pkg/api/testspec.proto
```diff
@@ -62,4 +58,29 @@ message TestSpec {
     string environment = 13;
     // Value of the target label set on exported Prometheus metrics.
     string target = 14;
+    // New priority to apply when reprioritizing jobs (if reprioritization is configured).
+    double new_priority = 15;
```
dejanzele, line 62: do we need this field in this PR?

williamvega: https://github.com/armadaproject/armada/blob/feat/node-ops-testing/testsuite/testcases/reprioritization/reprioritize_by_id_1x1.yaml

We are still using it in this additive pr, but in the next pr it will be deprecated. It was added in a past pr, I just moved the location of new_priority in this pr to be consecutive.

dejanzele (approved):

d2burkhalter (approved):

PR #5019, opened 2026-07-14
Fix mage testsuite failures behind a TLS-intercepting proxy

ianhockett:
`mage testsuite` pulls several container images directly from `docker.io` inside the kind cluster instead of loading them from the host, which fails behind a TLS-intercepting corporate proxy (`x509: certificate signed by unknown authority`) since the kind node's containerd doesn't trust that proxy's root CA. This fixes the three gaps found while getting the suite green locally:

- `getImagesUsedInTestsOrControllers()` was missing `alpine:latest`, `alpine:3.16`, and `bitnamilegacy/kubectl:1.33.4`, which several testsuite cases reference but weren't preloaded into kind.
- `preempt_by_id_1x1` and `preempt_by_ids_1x5` were missing `imagePullPolicy: IfNotPresent` on their `alpine:latest` container, so Kubernetes defaulted to `Always` and re-pulled from `docker.io` on every run regardless of the kind preload.
- `docs/developer_guide.md` gets a new troubleshooting entry for the `docker buildx is not set to default context` error hit during `mage dev:full`, for anyone using a Docker runtime other than Docker Desktop.

### DIFF
--- a/magefiles/kind.go
+++ b/magefiles/kind.go
@@ -25,9 +25,6 @@ func getImagesUsedInTestsOrControllers() []string {
 	return []string{
 		"nginx:1.27.0", // Used by ingress-controller
 		"alpine:3.20.0",
-		"alpine:3.16",
-		"alpine:latest",
-		"bitnamilegacy/kubectl:1.30",
 		"bitnamilegacy/kubectl:1.33.4",
 	}
 }
--- a/magefiles/kind.go
+++ b/magefiles/kind.go
@@ -25,7 +25,10 @@ func getImagesUsedInTestsOrControllers() []string {
 	return []string{
 		"nginx:1.27.0", // Used by ingress-controller
 		"alpine:3.20.0",
+		"alpine:3.16",
+		"alpine:latest",
 		"bitnamilegacy/kubectl:1.30",
+		"bitnamilegacy/kubectl:1.33.4",
 	}
 }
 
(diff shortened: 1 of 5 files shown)


magefiles/kind.go
```diff
@@ -25,7 +25,10 @@ func getImagesUsedInTestsOrControllers() []string {
 	return []string{
 		"nginx:1.27.0", // Used by ingress-controller
 		"alpine:3.20.0",
+		"alpine:3.16",
+		"alpine:latest",
 		"bitnamilegacy/kubectl:1.30",
+		"bitnamilegacy/kubectl:1.33.4",
```
d2burkhalter: Probably better to just pin to a single version so everything uses `alpine:3.20.0` and `kubectl:1.33.4`

d2burkhalter (changes requested):

magefiles/kind.go
ianhockett: Good call. As I wasn't sure if there were reasons for the version mismatch, I just made an additive change to the preload list to make sure it included all the images referenced in the test cased. But, I've now traced the history and it does appear the version spread is all accidental drift. Pinned everything to alpine:3.20.0 + kubectl:1.33.4, which lets the preload list shrink instead of grow. Thanks for the nudge.

d2burkhalter (approved):

dejanzele (approved):

PR #5020, opened 2026-07-15
Armadactl delete executor

nikola-jokic:
#### What type of PR is this?

Enhancement

Sometimes, the listener is stale, so delete api allows deleting the executor from the command line directly which helps ops.


internal/server/submit/submit.go
```diff
@@ -6,10 +6,12 @@ import (
 
 	"github.com/gogo/protobuf/types"
 	"github.com/gogo/status"
+	"github.com/pkg/errors"
```
JamesMurkin, line 9: Move these changes to internal/server/executor.go

PR #5047, opened 2026-07-24
Move requireGrpcCode test helper into shared servertest package

dejanzele:
The executor, node, queue, and retry policy server tests each defined their own private `requireGrpcCode` helper for asserting that an error is a gRPC status with a given code. This moves the helper into the shared `servertest` package as `RequireGrpcCode` and switches all four to it.

The private copies used `assert.Equal` for the code comparison and the shared helper uses `require.Equal`, so a mismatched code now stops the test instead of letting it carry on.

The retry policy copy came in with #5049, which noted it could switch to the shared helper once both were in. That is done here, so no private copies remain.


dejanzele:
@greptileai

PR #5054, opened 2026-07-27
Include requestor information to lookout

nikola-jokic:

Fixes #


pkg/armadaevents/events.proto
nikola-jokic, line 280: This is intentional, so we can use requestor across all events. The idea was that using cancel_user is a bit verbose when the message is obviously CancelledJob, so user would suffice. Since requestor is on all other messages, we can use requestor on this one as well.

pkg/armadaevents/events.proto
```diff
@@ -264,15 +268,16 @@ enum JobState {
...
     repeated JobState states = 1;
     string reason = 2;
+    string requestor = 3;
 }
 
 // Generated by the scheduler in response to CancelJob and CancelJobSet.
 // One such message is generated per job that was cancelled.
 message CancelledJob {
-    reserved 1;
+    reserved 1, 5;
```
nikola-jokic, line 277: Reserved 5 for now if want to propagate the reason as well. We might not need that so I'll remove it if review shows it is not important.

PR #5059, opened 2026-07-28
feat: support pod-level resources (KEP-2837)

washcycle:
Adds optional Kubernetes pod-level resources (`podSpec.resources`), gated by `submission.podLevelResources` (default `false`). Containers can share a pod budget without Armada injecting per-container limits for the pooled resources.

When enabled:

- Validate resource names against the pinned `k8s.io/component-helpers` helper, reject pod-level claims, require pod requests to cover aggregate container requests, and enforce container limits within the pod limit.
- Apply resource minimums only to declared resources, including explicit zero requests, using exact quantity comparisons. A CPU-only job is not rejected for an unrelated memory or GPU minimum.
- Reserve the effective pod request and preserve resource defaults for resources absent from the pod-level block, such as ephemeral storage.
- Apply retry memory increases to the pod budget and its containers. Static increases are shared in proportion to requests within the pool, including native sidecars and init containers.
- Include pod-level resources in cordoned-node accounting and resource-specific active deadlines.

Submitting a pod-level block while the feature is disabled returns a clear validation error. The flag is declared in the shipped server configuration so environment overrides work.

Lookout now reports the same effective request used for scheduling. This also changes reporting for container-only jobs: native sidecars and classic init-container requirements are included. Regression tests cover these cases.

Compatibility: keep the existing Kubernetes dependencies, as agreed in review. The current `component-helpers v0.32.11` accepts CPU and memory at the pod level; `hugepages-*`, GPUs, and ephemeral storage are rejected there. CPU and memory must be included in the scheduler's supported resource types. Every executor cluster must have Kubernetes' `PodLevelResources` feature gate enabled.

Validation:

- Passed full tests for submission validation/conversion, common resource accounting, API helpers, Lookout instructions, and executor utilisation.
- Passed scheduler executor-API and resource-mutation tests, including pooled-only jobs, combined retry increases, native sidecars, sequential init containers, and multi-GiB budgets.
- Formatting and `git diff --check` passed.
- The broader submission/scheduler suites were attempted but their database-backed cases require PostgreSQL at `localhost:5432`; no local PostgreSQL or running Docker daemon was available.


washcycle:
Thanks for the review. Addressed in 4157310:

**1. Disabled resources still affect accounting** — fixed. Added a `dropPodLevelResourcesIfDisabled` post-processor that clears `podSpec.resources` when `submission.podLevelResources` is false, so all downstream accounting (scheduler + Lookout) ignores it when the feature is off. This keeps the accounting helpers gate-free while making disabled fully inert.

**2. Minimum checks ignore effective requests** — fixed. `MinJobResources` is now checked against the effective request `max(pod-level, sum(container requests))` via `SchedulingResourceRequirementsFromPodSpec`, so a 5-CPU container + 2-CPU pod-level request is no longer rejected under a 4-CPU minimum. Added a test for both the accept and reject cases.

**3. Filtering can erase the pod budget** — this mirrors the existing container-level behaviour: the allow-list intentionally strips floating resources before the pod reaches k8s while the scheduler retains them for accounting. Applying a different rule only to the pod-level block would be inconsistent with how container resources are already handled. Open to a follow-up if you'd prefer a revalidation step, but I kept it consistent with the current container path for now.

Tests added for 1 and 2.

internal/common/resource/resource_test.go
```diff
@@ -148,6 +148,49 @@ func TestTotalResourceRequest_ShouldCombineMaxInitContainerResourcesWithSummedCo
 	assert.Equal(t, result, FromResourceList(expectedResult))
 }
 
+// Pod-level resources (KEP-2837): TotalPodResourceRequest uses the effective
+// request max(sum(containers), pod-level).
+func TestTotalResourceRequest_PodLevelResources(t *testing.T) {
```
dejanzele, line 154: Can you refactor the test to be table-driven?

washcycle: Will do.

washcycle:
@dejanzele how does that look now?

internal/server/submit/validation/submit_request_test.go
```diff
@@ -1150,6 +1150,122 @@ func TestValidateResources(t *testing.T) {
...
+
+	tests := map[string]struct {
+		req           *api.JobSubmitRequestItem
+		expectSuccess bool
+	}{
+		// container 5cpu + pod-level 2cpu -> effective 5cpu >= 4cpu min: accepted.
+		"container total meets minimum, low pod-level ok": {
+			req:           req(&v1.ResourceRequirements{Requests: cpu("2"), Limits: cpu("2")}, container5),
+			expectSuccess: true,
+		},
```
nikola-jokic, line 1249: Hey I think greptile was correct here. If you define:
```go
container2 := []v1.Container{{Name: "main", Resources: v1.ResourceRequirements{Requests: cpu("2"), Limits: cpu("2")}}}
```
And add test:
```
"container below minimum covered by pod-level block, accepted": {
  req:           req(&v1.ResourceRequirements{Requests: cpu("4"), Limits: cpu("4")}, container2),
   expectSuccess: true,
 },
```
container 2cpu + pod-level 4cpu -> effective max(2,4)=4cpu >= 4cpu min, but the per-container minimum check only sees the container's raw 2cpu and rejects.

washcycle: You're right, and thanks for the concrete repro — I've used it verbatim as the test case.

Fixed in 886b52df. `validatePodLevelResources` already checks `MinJobResources` against the effective request `max(pod-level, sum(containers))`, which is what the scheduler actually reserves, so the per-container pass is redundant when a pod-level block is set — and, as you showed, actively wrong. It now returns early for that case, leaving the per-container check untouched for every job without a pod-level block.

I confirmed the test is load-bearing: reverting the one-line guard fails `container below minimum covered by pod-level block, accepted` and nothing else.

washcycle:
Pushed two fixes (886b52df, 1e6ece8e).

**1. Per-container minimums vs the effective request** — fixed, per @nikola-jokic's repro above. Details in the thread.

**2. `defaultJobLimits` defeated pod-level pooling** — this one wasn't reported here; I found it running the feature on a real cluster, and it made the feature ineffective in its main use case.

`defaultResource` applied `defaultJobLimits` to every container declaring nothing, including the containers of a pod that supplies its budget at the pod level. That nests a per-container ceiling inside the pod's cgroup, so each container is capped at the default rather than drawing on the pooled budget — the pool is reserved and then unusable. Measured on Kubernetes 1.35.7 with the feature enabled, a pod-level `2cpu/4Gi` job whose single container declared nothing:

```
spec.resources:                      {"cpu":"2","memory":"4Gi"}
spec.containers[0].resources:        {"cpu":"1","memory":"1Gi","ephemeral-storage":"8Gi"}   # injected
container's cgroup memory.max:       1073741824   # 1Gi, inside a 4Gi pod
```

Hand-applying the same pod shape without Armada in the path gives the container the pod's full ceiling, so the injection is the only thing in the way.

The fix skips defaulting a resource when the pod-level block carries that resource. Resources the block does not carry still default — notably `ephemeral-storage`, which KEP-2837 cannot express at the pod level. `dropPodLevelResourcesIfDisabled` already runs earlier in `podLevelProcessors`, so this is implicitly feature-gated and the disabled path is byte-for-byte unchanged.

It also closes a gap in `validateInitContainerCpu`: a bare init container is waived by validation and was then defaulted to a whole core — the exact value that check rejects when written explicitly. Verified live before the fix: an init container declaring nothing landed with `cpu: "1"`, while declaring `cpu: "1"` explicitly is rejected at submit.

**3. On the allow-list filtering concern** (`internal/scheduler/api.go`) — I want to correct my earlier reply, which was too dismissive. It is reachable, though narrower than the bot's framing: it needs a pod-level block naming *only* resources outside `scheduling.supportedResourceTypes` together with containers that declare nothing. With the default set (`memory`, `cpu`, `ephemeral-storage`, `nvidia.com/gpu`) and the fact that the apiserver only admits `cpu`, `memory` and `hugepages-*` at the pod level, the sole shape that gets there is a hugepages-only block. Anything else is either kept by the filter or rejected by the apiserver first.

So the consistency argument holds for the general case, but that one shape would reach the executor with no budget anywhere. Happy to add a targeted guard — either nil the block when filtering empties it, or reject at submit if the pod-level block would be left with nothing and no container declares resources. Say which you'd prefer and I'll push it; I didn't want to change lease-time behaviour unprompted.

nikola-jokic (dismissed):
LGTM

nikola-jokic:
@dejanzele do you want to take another look?

dejanzele:
Hey @washcycle ,

Thanks for opening this PR, we are very interested in this feature!

I am reviewing it, overall it looks good and definitely in the right direction, there are a couple of things we should fix.

## Validation

I think this function `validatePodLevelResources` must also check the resource names. Kubernetes accepts only `cpu`, `memory`, and `hugepages-*` in `spec.resources`. The API server rejects all other names, for example `nvidia.com/gpu` and `ephemeral-storage`:
* https://github.com/kubernetes/kubernetes/blob/v1.34.0/staging/src/k8s.io/component-helpers/resource/helpers.go#L60-L71
* https://github.com/kubernetes/kubernetes/blob/v1.34.0/pkg/apis/core/validation/validation.go#L7152-L7163

Here are some examples which would pass Armada validation but not Kubernetes validation:
- A pod-level `nvidia.com/gpu: 8` block. Kubernetes supports only `cpu`, `memory`, and `hugepages-*` at the pod level. The scheduler reserves 8 GPUs, and then the pod creation fails.
- A pod-level `ephemeral-storage: 8Gi` block. Kubernetes rejects it for the same reason. This mistake is easy to make, because a user copies all three resources from a container block.
- A container with 5 CPU and a pod-level block with 2 CPU. Kubernetes requires the pod-level request to be greater than or equal to the sum of the container requests. The test `container total meets minimum, low pod-level ok` expects Armada to accept this spec.
- Two containers with 3 CPU each and a pod-level block with 4 CPU. Each container fits the pod limit, but the sum is 6 CPU, so Kubernetes rejects the pod. The validation has no check that compares the container sum with the pod-level request.
- A pod-level block with valid requests and a `claims` entry. Kubernetes forbids `claims` at the pod level, and the validation does not read the field.

## Retry Policies

Retry Policies will need to be aware of this work as a retry policy can be configured to bump memory on a retry. `applyResourceMutations` in `internal/scheduler/api.go` changes only the container memory. It does not read the pod-level block.

## Minimal resources for a Job

The default config has `minJobResources: memory: 1Mi`. Validation runs before `defaultResource`. The result is different for two similar jobs:
- A container with only `cpu: 1` passes, and then receives the default memory.
- A pod-level block with only `cpu: 1` and a bare container fails with `effective memory requests (0) below server minimum (1Mi)`.

The new check tests each resource in `MinJobResources`. The container check tests only the resources that the container declares. I think the two paths should agree.

## Other

- When the feature is off, `dropPodLevelResourcesIfDisabled` removes the block without a message. A job with a pod-level block of 64 CPU and a container with 1 CPU runs with 1 CPU. Would a rejection with a clear error be better here?
- `eff.Value()` rounds CPU up, so a pod-level value of `3500m` passes a minimum of 4 CPU. `Cmp` gives the correct result. The container check has the same pattern, so this can be a separate fix.
- The change to `getJobResources` in the Lookout ingester applies to all jobs, not only to jobs with pod-level resources. The reported value now includes the init containers and the sidecars. I think the change is correct, but can you mention it in the PR description and add a test?
- `getCordonedResource` in `internal/executor/utilisation/cluster_utilisation.go` still reads only the container limits. A job with pod-level resources on a cordoned node counts as zero.

washcycle:
@dejanzele on the `hugepages-*` point — it isn't reachable at the moment for a versioning reason worth flagging.

This repo pins `k8s.io/component-helpers v0.32.11` (and `k8s.io/api v0.32.11`), where the supported set is only cpu and memory:

```go
// component-helpers@v0.32.11/resource/helpers.go:56
var supportedPodLevelResources = sets.New(v1.ResourceCPU, v1.ResourceMemory)
```

The sources you linked are `v1.34.0`, which does include `hugepages-*` — so we're both right, just about different releases. Validating against the helper we actually vendor means pod-level `hugepages-*` is rejected today.

Two ways to go, and I'd rather you pick:

1. Validate against the vendored helper, as now. `hugepages-*` is rejected, and it starts working on its own whenever the k8s deps are bumped — no code change needed then, and no hardcoded resource list to go stale.
2. Bump `component-helpers` to >= 0.34 as part of this PR, which makes `hugepages-*` work immediately but widens the change.

I've gone with (1) so the error text is generated from `SupportedPodLevelResources()` rather than a literal, which keeps it honest across bumps. Happy to switch to (2) if you'd prefer this land with hugepages support.

dejanzele:
@washcycle 1. is fine

dejanzele:
@greptileai

internal/common/resource/resource.go
```diff
@@ -264,6 +264,13 @@ func TotalPodResourceRequest(podSpec *v1.PodSpec) ComputeResources {
 			totalResources.Max(containerResource)
 		}
 	}
+
+	// Pod-level resources (KEP-2837): the effective request is
+	// max(sum of container requests, pod-level request) per resource. Inert when
+	// podSpec.Resources is unset.
```
dejanzele: The code already shows the max. The useful part is why the max is right.

```suggestion
	// Kubernetes requires the pod-level request to be at least the sum of the container requests.
	// Max keeps the accounting correct for a pod that breaks that rule.
```

pkg/api/util.go
```diff
@@ -69,6 +69,13 @@ func SchedulingResourceRequirementsFromPodSpec(podSpec *v1.PodSpec) *v1.Resource
 			maxResourcesToList(rv.Limits, c.Resources.Limits)
 		}
 	}
+
+	// Pod-level resources (KEP-2837): max with the pod-level request/limit so the
+	// scheduler reserves the pod-level budget. Inert when podSpec.Resources is unset.
```
dejanzele: Same as in resource.go, so I would use the same sentence in both places.

```suggestion
	// Kubernetes requires the pod-level request to be at least the sum of the container requests.
	// Max keeps the accounting correct for a pod that breaks that rule.
```

internal/lookoutingester/instructions/instructions.go
```diff
@@ -587,18 +587,21 @@ func getJobResources(job *api.Job) jobResources {
...
-	for _, container := range podSpec.Containers {
-		resources.Cpu += getResource(container, v1.ResourceCPU, true)
-		resources.Memory += getResource(container, v1.ResourceMemory, false)
-		resources.EphemeralStorage += getResource(container, v1.ResourceEphemeralStorage, false)
-		resources.Gpu += getResource(container, "nvidia.com/gpu", false)
-	}
+	// Use the canonical effective-request computation so the reported footprint
+	// matches what the scheduler bills: sum of main containers + native sidecars,
+	// max over classic init containers, and max with the pod-level block (KEP-2837).
+	// This also fixes the prior undercount that summed only main containers.
```
dejanzele: The last sentence describes the past, so it will confuse a reader in a year. The list of terms repeats the function it calls.

```suggestion
	// Lookout reports the same request that the scheduler reserves.
```

internal/server/configuration/types.go
```diff
@@ -99,6 +99,11 @@ type SubmissionConfig struct {
 	AddGangIdLabel bool
 	// Controls whether custom service names are allowed
 	AllowCustomServiceNames bool
+	// When true, honour Kubernetes pod-level resources (KEP-2837, podSpec.Resources):
+	// a container may omit its own resources if the pod-level block is set, and
+	// accounting uses max(sum(container requests), podLevel.Requests). Default false
+	// preserves container-only behaviour.
```
dejanzele: This is the comment operators read, so I would keep the KEP reference here and add the cluster requirement. The last sentence is the zero value of a bool, so it can go.

```suggestion
	// PodLevelResources enables Kubernetes pod-level resources (KEP-2837, podSpec.resources).
	// A container may then omit its resources when the pod-level block declares them.
	// Every executor cluster must have the PodLevelResources feature gate enabled.
```

internal/server/submit/conversion/post_process.go
```diff
@@ -110,10 +111,35 @@ func defaultPriorityClass(spec *v1.PodSpec, config configuration.SubmissionConfi
...
+
 // Adds resources defined in config.DefaultJobLimits to all containers in the podspec if that container is missing
 // requests/limits for that particular resource. This can be used to e.g. ensure that all jobs define at least some
 // ephemeral storage.
+//
+// A resource carried by the pod-level block (KEP-2837) is not defaulted into the containers. Defaulting it would nest
+// a per-container ceiling inside the pod's, capping each container below the pooled budget the block exists to grant,
+// so the pool would be reserved but unusable. Resources the block does not carry still default -- notably
+// ephemeral-storage, which KEP-2837 cannot express at the pod level. dropPodLevelResourcesIfDisabled runs earlier in
+// podLevelProcessors and nils the block when the feature is off, so this is implicitly feature-gated.
```
dejanzele: This one earns its place, because a reader who removes the continue gets a pod that cannot use its budget. It is just long.

```suggestion
// A resource in the pod-level block gets no container default. A container default becomes a container limit
// below the pod-level budget, and the containers cannot use that budget. Resources that Kubernetes does not
// support at the pod level, for example ephemeral-storage, still get the default.
// This relies on dropPodLevelResourcesIfDisabled to run first.
```

internal/server/submit/validation/submit_request.go
```diff
@@ -308,6 +330,62 @@ func validateResources(j *api.JobSubmitRequestItem, config configuration.Submiss
 	return nil
 }
 
+// validatePodLevelResources validates a pod-level resources block (KEP-2837),
+// mirroring the per-container checks: requests and limits must be non-negative,
+// cover the same resource set, satisfy limit >= request within the
+// max-oversubscription ratio, and meet MinJobResources.
```
dejanzele: The list of rules will go stale when you add the name check and the consistency rules. A one line doc comment survives that.

```suggestion
// validatePodLevelResources applies the container resource rules and the Kubernetes pod-level rules to the pod-level block.
```

internal/server/submit/validation/submit_request_test.go
```diff
@@ -1150,6 +1150,129 @@ func TestValidateResources(t *testing.T) {
 	}
 }
 
+// TestValidateResources_PodLevel covers Kubernetes pod-level resources (KEP-2837),
+// gated by SubmissionConfig.PodLevelResources.
```
dejanzele: The test names and the case names already say what each case proves. I would drop the comments above the test functions and the ones inside the tables, here and in post_process_test.go, resource_test.go and util_test.go.

```suggestion
```

internal/server/submit/validation/submit_request_test.go
```diff
@@ -1150,6 +1150,129 @@ func TestValidateResources(t *testing.T) {
...
+			req:           req(&v1.ResourceRequirements{Requests: cpu("2"), Limits: cpu("2")}, container5),
+			expectSuccess: true,
+		},
+		// pod-level 2cpu, empty container -> effective 2cpu < 4cpu min: rejected.
+		"pod-level below minimum with empty container rejected": {
+			req:           req(&v1.ResourceRequirements{Requests: cpu("2"), Limits: cpu("2")}, []v1.Container{{Name: "main"}}),
+			expectSuccess: false,
+		},
+		// container 2cpu + pod-level 4cpu -> effective 4cpu >= 4cpu min: accepted. The
+		// per-container check used to see only the raw 2cpu and reject.
```
dejanzele: "used to" describes the old code, and the case name already says the container is below the minimum.

```suggestion
```

config/server/config.yaml
```diff
@@ -66,6 +66,10 @@ submission:
   defaultActiveDeadlineByResourceRequest:
     nvidia.com/gpu: "336h" # 14 days.
   assertInitContainersRequestFractionalCpu: true
+  # Honour podSpec.resources (KEP-2837). Must be listed here even when false, or
```
dejanzele, line 69: I'd even remove this comment

dejanzele:
@greptileai

PR #5061, opened 2026-07-28
Move job accounting from NodeDb into Node methods

tgucks:
#### What type of PR is this?

Refactor

- Moves job-accounting logic out of `NodeDb` and onto the `internaltypes.Node` type it operates on.
- Adds `Node.AddJob`, `Node.EvictJob`, and `Node.RemoveJob` (plus a new `SchedulableJob` interface so `Node` depends only on the four job accessors it needs, not on `jobdb.Job`), and rewrites `NodeDb`'s `BindJobToNode`/`EvictJobsFromNode`/`UnbindJobsFromNode` to delegate to them. The superseded `*InPlace` helpers and the accounting functions (`markAllocated`, `markAllocatable`, `priorityCutoffFor`) are removed from `nodedb.go`.
- NodeDb keeps ownership of the `scheduledAtPriorityByJobId` map: it now resolves each job's scheduled priority and passes it into the new Node methods. Behavior should be unchanged. The strict "job not mapped to a priority" error on unbind of a bound, non-evicted job is preserved via the `unbindResolvingPriority` wrapper.

- No behavior change intended, this should be a pure refactor
- `SchedulableJob` is satisfied by `*jobdb.Job` structurally; the new `internaltypes` tests exercise the accounting paths through a lightweight `testSchedJob` fake rather than a full job.


internal/scheduler/nodedb/nodedb.go
```diff
@@ -1045,52 +950,28 @@ func (nodeDb *NodeDb) UnbindJobsFromNode(jobs []*jobdb.Job, node *internaltypes.
...
-	allocatable := node.AllocatableByPriority
-	if isEvicted {
-		// Evicted jobs are always tracked at EvictedPriority regardless of preemptibility.
-		markAllocatable(allocatable, internaltypes.EvictedPriority, requests)
-	} else {
-		priority, ok := nodeDb.GetScheduledAtPriority(jobId)
-		if !ok {
+	priority, ok := nodeDb.GetScheduledAtPriority(jobId)
+	if !ok {
+		_, isEvicted := node.EvictedJobRunIds[jobId]
```
JamesMurkin: Make these 2 functions on the node object

JamesMurkin (dismissed):

internal/scheduler/nodedb/nodedb.go
tgucks: Done

JamesMurkin (approved):

PR #5069, opened 2026-07-29
ci: pin GitHub Actions to full-length commit SHAs

pavlovic-ivan:
## What
Pins every third-party GitHub Actions `uses:` reference in `.github/workflows/` to a full-length commit SHA, with the original tag preserved as a trailing comment (e.g. `actions/checkout@3d3c42e… # v7`).

## Why
Floating references (`@v4`, `@main`, branch/tag) are mutable — a compromised or retagged action would silently run new code in our pipelines (the class of supply-chain attack seen across GitHub in 2025). Pinning to an immutable SHA closes this. This aligns armadaproject with the org-wide "require actions pinned to a full-length commit SHA" policy.

## Scope / safety
- **Only `uses:` lines changed** — no formatting, comments, or logic touched (verified: every changed line is a `uses:` line).
- Local (`./…`) and reusable-workflow references are left as-is (exempt from the requirement).
- Each SHA is the commit the tag currently points to, so behaviour is unchanged today; Dependabot will keep the SHAs (and `# vX` comments) updated going forward.

## Note for reviewers
- `Mo-Fatah/ci-alerts@v2` is a third-party personal-account action; it's now locked to `4ee9ace…`.

### DIFF
--- a/.github/workflows/python-tests/action.yml
+++ b/.github/workflows/python-tests/action.yml
@@ -17,14 +17,14 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/setup-python@v5
+    - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
       with:
         python-version: ${{ inputs.python-version }}
     # Tox to run tests; build to build the wheel after tests pass
     - run: pip install 'tox>=4,<5' build twine
       shell: bash
     - name: Install Protoc
-      uses: arduino/setup-protoc@v3
+      uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b # v3
       with:
         version: '23.3'
         repo-token: ${{ inputs.github-token }}
--- a/.github/workflows/python-tests/action.yml
+++ b/.github/workflows/python-tests/action.yml
@@ -17,14 +17,14 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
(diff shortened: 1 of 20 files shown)


pavlovic-ivan:
Updated: this now also pins third-party actions **inside local composite actions** (`python-tests/action.yml (setup-python, arduino/setup-protoc)`), not just top-level workflows. The `uses: ./…` references themselves are unchanged (local paths, can't be SHA-pinned); it's the actions *inside* those composite action files that are now pinned. Still verified: every changed line is a `uses:` line.

pavlovic-ivan:
Also pinned the one cross-repo reusable workflow in `release.yml`: `G-Research/charts/.github/workflows/invoke-push.yaml@master` → `da173017…` (kept `# master`). It was on a moving **branch** ref (the most mutable kind), so a re-pointed `master` on `G-Research/charts` would run here with this repo's token/secrets. Note: because it tracks a branch rather than a semver tag, Dependabot won't auto-bump it — the SHA will need occasional manual refresh.

dave-gantenbein (dismissed):

.github/workflows/python-tests/action.yml
```diff
@@ -17,14 +17,14 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/setup-python@v5
+    - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
```
dejanzele: in the comment, can we pin it to the full semver of this commit? (same for others)

pavlovic-ivan: Addressed. There are a couple of refs left that are not written as full semver. Those don't have it

dejanzele (approved):

dave-gantenbein (approved):

PR #5079, opened 2026-07-30
add integration tests for jobset cancellation and terminal jobset no-op 

Sigele:
#### What type of PR is this?

improvement (test suite)

Adding two integration tests:
- jobset_cancel_inactive asserts that cancelling a jobset when all jobs within are in a terminal state results in a successful no-op with no error.
- jobset_cancel_by_set_1x5 asserts that, when a jobset is cancelled, all jobs within return "cancelled" status afterwards. 

To support jobset_cancel_inactive this also adds:
- a conditional block in testrunner.go to ensure the test fails if an error is returned on cancellation 
- a cancel jobset function
- a boolean to restrict the above cancel behavior to relevant test cases

Fixes #


Sigele:
@greptileai

Sigele:
@greptileai

Sigele:
@greptileai

williamvega (dismissed):

testsuite/testcases/basic/jobset_cancel_by_set_1x5.yaml
```diff
@@ -0,0 +1,34 @@
...
+          resources:
+            limits:
+              memory: 25Mi
+              cpu: 100m
+            requests:
+              memory: 25Mi
+              cpu: 100m
+---
+cancelSet:
+  request: {}
```
dejanzele: nit: can you remove the `{}` and keep the same format as elsewhere, just blank after `:`

dejanzele (dismissed):

testsuite/testcases/basic/jobset_cancel_by_set_1x5.yaml
Sigele: fixed in latest commit

williamvega (approved):

dejanzele (approved):

PR #5080, opened 2026-07-31
Classify delete-action failed pods with the pod-level failure reason

dejanzele:
The delete-action detection classified failed pods with ClassifyContainerError. That call passes an empty pod-error message. As a result, onPodError rules could never match on this path. Pods that never started a container were unclassifiable. Such pods have no exit codes, no termination messages, and no matching conditions.

The concrete case: the scheduler places a job on a node with a GPU that just became unhealthy. The kubelet rejects the pod at admission with "Allocate failed due to requested number of devices unavailable for nvidia.com/gpu". No container ever existed. The pod-level status message is the only description of the failure. Other admission rejections have the same shape, for example resource races with DaemonSet pods, or DiskPressure that appears after scheduling. Kubelet evictions have it too. There, the message separates a node that ran low on a resource from a pod that exceeded its own limit. The coarse Evicted condition cannot make that distinction.

The detection now classifies with ClassifyPodError. It passes the same extracted failure reason that it already stores on the issue. ExtractPodFailedReason returns pod.Status.Message first, so these messages become matchable with onPodError rules. ClassifyPodError evaluates all other rule types exactly as before. Existing categories behave identically. onPodError rules simply become reachable on this path.

This closes one of the two categorizer gaps that block the migration of failedPodChecks to Delete-action categories. A pod-status check migrates like this:

```yaml
# legacy failedPodChecks entry
podStatuses:
  - regexp: "Pod was rejected.*"
    reason: "UnexpectedAdmissionError"

# becomes a category (paired with a scheduler retry policy for it)
errorCategories:
  categories:
    - name: node_admission_failure
      action: Delete
      rules:
        - onPodError: { pattern: "Pod was rejected.*" }
```

We validated the change on a live cluster. An onPodError rule matched an evicted pod. The executor deleted the pod and reported the categorized failure only after the pod was gone. It emitted exactly one terminal event.

### DIFF
--- a/internal/executor/service/pod_issue_handler.go
+++ b/internal/executor/service/pod_issue_handler.go
@@ -194,7 +194,11 @@ func (p *PodIssueHandler) DetectAndRegisterDeleteActionIssue(pod *v1.Pod) (bool,
 	if !util.IsManagedPod(pod) || pod.Status.Phase != v1.PodFailed {
 		return false, nil
 	}
-	classification := p.classifier.ClassifyContainerError(pod)
+	// Classify with the extracted failure reason. This lets onPodError rules
+	// match pods that never started a container, for example kubelet admission
+	// rejections. Such pods have no exit codes and no termination messages.
+	failedReason := util.ExtractPodFailedReason(pod)
+	classification := p.classifier.ClassifyPodError(pod, failedReason)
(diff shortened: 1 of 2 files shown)


dejanzele:
@greptileai

JamesMurkin (approved):

PR #5085, opened 2026-08-03
Add RetryUntilSuccessOrExhausted and NonRetryableError to internal/common/util

mauriceyap:
Add a bounded-retry helper which retries `performAction` up to `maxAttempts` times. It calls `onExhausted` with the last error if the budget runs out. It short-circuits immediately if the error is wrapped in the new `NonRetryableError`.

This will be used in the ingestion pipeline to decide when to dead-letter a message after retrying it.

### DIFF
--- a/internal/common/util/retry.go
+++ b/internal/common/util/retry.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/armadaproject/armada/internal/common/armadacontext"
 )
@@ -14,6 +15,9 @@ type NonRetryableError struct {
 }
 
 func NewNonRetryableError(err error) *NonRetryableError {
+	if err == nil {
+		err = errors.New("non-retryable error")
+	}
 	return &NonRetryableError{err: err}
 }
 
@@ -53,6 +57,8 @@ func RetryUntilSuccess(ctx *armadacontext.Context, performAction func() error, o
 // must distinguish "gave up due to shutdown" from "exhausted attempts" via ctx.Err().
 // If performAction returns an error wrapped with NewNonRetryableError, remaining
 // attempts are skipped and onExhausted is called immediately with that error.
+// If maxAttempts is non-positive, performAction is never called and onExhausted
+// is called with a placeholder error describing this.
 func RetryUntilSuccessOrExhausted(
 	ctx *armadacontext.Context,
 	maxAttempts int,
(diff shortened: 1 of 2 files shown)


internal/common/util/retry.go
```diff
@@ -1,9 +1,39 @@
...
+	"errors"
+	"fmt"
+
 	"github.com/armadaproject/armada/internal/common/armadacontext"
 )
 
+// NonRetryableError wraps an error to signal that it should not be retried, even if
+// remaining attempts are available. Sinks can wrap errors with this to short-circuit
+// RetryUntilSuccessOrExhausted straight to onExhausted.
+type NonRetryableError struct {
```
nikola-jokic: Just out of curiosity, why not have something like this:
```go
type commonError string

var NonRetryableError = commonError("non-retryable error")
```

Then you can easily check with errors.Is, and if you want to wrap the error, you just use fmt.Errorf("%w: %w", NonRetryableError, innerError)

Instead of using NewNonRetryableError(nil) where nothing is wrapped, you can just return NonRetryableError. The type of the error is private, so nothing can overwrite it outside of this package. I think it would be much easier

internal/common/util/retry.go
```diff
@@ -19,3 +49,46 @@ func RetryUntilSuccess(ctx *armadacontext.Context, performAction func() error, o
...
+// RetryUntilSuccessOrExhausted behaves like RetryUntilSuccess but gives up after
+// maxAttempts consecutive failures, calling onExhausted with the last error
+// instead of continuing. Returns true on eventual success, false otherwise.
+// onExhausted is NOT called if ctx is cancelled first (shutdown case) - callers
+// must distinguish "gave up due to shutdown" from "exhausted attempts" via ctx.Err().
+// If performAction returns an error wrapped with NewNonRetryableError, remaining
+// attempts are skipped and onExhausted is called immediately with that error.
+// If maxAttempts is non-positive, performAction is never called and onExhausted
+// is called with a placeholder error describing this.
+func RetryUntilSuccessOrExhausted(
```
nikola-jokic, line 40: The retry doesn't support backoff right now, I don't know if that is desired.
Have you considered using https://github.com/avast/retry-go ?
If you prefer not to use id, I'd suggest onError to get the ctx as the first argument

nikola-jokic (commented):
Added few suggestions that I think might be useful ☺️

internal/common/util/retry.go
mauriceyap: Good shout! This is an old-school way of doing this which I copied from `internal/common/armadaerrors/errors.go`, but you are right that the modern idiomatic way of doing this is to use built-in wrapping and unwrapping! I'll update.

internal/common/util/retry.go
mauriceyap: the idea would be to have the caller implement the backoff by waiting in `onError` - see `internal/common/ingest/ingestion_pipeline.go` in #5091

nikola-jokic: Then I'd prefer getting the ctx as the first argument so if you decide to do:
```go
select {
case <-ctx.Done():
case <- time.After(d):
}
```
you can easily do it. But definitely not a blocker

nikola-jokic (approved):

PR #5091, opened 2026-08-04
Retry Sink.Store with backoff and dead-letter on exhaustion, across the pipeline and every sink

mauriceyap:
When an ingester repeatedly fails to write a message to its sink (events Redis, the lookout postgres database, or the scheduler postgres database), it currently either retries forever or drops the message. This PR gives every ingester a proper dead letter queue: failures are retried with jittered backoff up to a configurable limit, and if the message still can't be stored, it's serialized and published to a dead-letter topic for later inspection or replay, instead of being retried indefinitely or silently lost. It also fixes a related bug where certain lookup errors caused a whole batch to be dropped rather than retried.

This touches both the shared pipeline and every sink because the two are compile-coupled. `Sink[T]` gains a `Serialize` method that every implementation must provide before anything builds, so the pipeline change and the per-sink work land together in one PR.

### Details

`Sink[T]` gains a `Serialize(msg T) ([]byte, error)` method, implemented by `RedisEventStore`, `LookoutDb`, and `SchedulerDb` (each renders its instructions as JSON for DLQ inspection/manual replay), and the pipeline and every sink land together so the tree never sits in a red, non-compiling intermediate state.

The pipeline's store loop now retries `Store` up to `DeadLetterMaxAttempts` times (falling back to a `defaultDeadLetterMaxAttempts` of 5 when unset, so an unconfigured `PulsarConfig` still retries rather than dead-lettering on the first failure) with jittered exponential backoff (via the new `newBackOff()`/`backoff/v4` helper), and on exhaustion serializes the message and publishes it to the configured dead-letter topic instead of dropping it, then still ACKs so the consumer doesn't redeliver forever.

If `DeadLetterTopic` is unset, no dead-letter publisher is created at all and an exhausted message is left unacked for redelivery instead (same as the shutdown case), rather than attempting to publish to an empty topic. Two new metrics (`RecordPulsarMessageStoreRetry`, `RecordPulsarMessageDeadLettered`) track this. Removes the now-redundant `ingest.WithRetry` helper (superseded by `util.RetryUntilSuccessOrExhausted` from PR 1). Context-cancellation during retry is handled by leaving the message unacked for redelivery rather than dead-lettering it.

Each sink's own internal retry loop is removed in the same change. Retry-then-dead-letter policy now lives entirely in the shared `IngestionPipeline` ack-path, and each `Store` implementation is updated to attempt once and classify unretryable failures via `util.NewNonRetryableError` so the pipeline skips straight to dead-lettering instead of burning its retry budget. `NewLookoutDb` and `NewSchedulerDb` drop their now-unused backoff/retry-count constructor parameters; every call site is updated accordingly. `SchedulerDb.Serialize` includes a type-switch (`serializeDbOperation`) covering every concrete `DbOperation`, guarded by a coverage test so new operation types can't silently fall through to an empty payload.

`filterEventsForTerminalJobs` now propagates lookup errors instead of silently ingesting unfiltered updates. Under the old pipeline this still had a gap: an ordinary, non-cancellation lookup error returned before any update was written, then hit the pipeline's ordinary-error branch, which acked and dropped the whole batch. That gap is closed structurally here, not with an extra special case - every `Store` error (including this one) now goes through the retry-then-dead-letter path, so the batch is retried and, if the error persists, dead-lettered and preserved rather than acked and silently dropped.


internal/scheduleringester/schedulerdb.go
```diff
@@ -28,56 +29,177 @@ const (
...
+	return json.Marshal(struct {
+		Ops        []serializedOp
+		MessageIds []string
+	}{
+		Ops:        ops,
+		MessageIds: pulsarutils.MessageIdsToStrings(instructions.MessageIds),
+	})
+}
+
+func serializeDbOperation(op DbOperation) any {
```
d2burkhalter: What if we merged all of the identical cases `return o` into one or  added an optional `SerializeForDlq` interface implemented by the operations that need them to bring it down to
```go
func serializeDbOperation(op DbOperation) any {
        if s, ok := op.(dlqSerializable); ok {
                return s.SerializeForDlq()
        }
        return op
  }
```
and move the serialization logic to dbops.go?

internal/common/ingest/ingestion_pipeline.go
```diff
@@ -214,32 +279,95 @@ func (i *IngestionPipeline[T, U]) Run(ctx *armadacontext.Context) error {
...
+				},
+				func(attempt int, err error) {
+					i.metrics.RecordPulsarMessageStoreRetry()
+					wait := storeBackoff.NextBackOff()
+					log.WithError(err).Warnf("%s - Error inserting %d messages (ids: %v); will retry after %s",
+						i.pulsarTopic, len(msg.GetMessageIDs()), msg.GetMessageIDs(), wait)
+					// This sleep is not ctx-aware: RetryUntilSuccessOrExhausted only checks ctx
+					// before the next performAction call, not during this wait. On shutdown, the
+					// sleep runs to completion before cancellation is noticed, delaying shutdown
+					// by up to `wait`. Deemed acceptable since wait is bounded by backoff config.
```
nikola-jokic, line 302: Why not use:
```go
select {
case <-ctx.Done():
  return ctx.Err()
case <-time.After(wait):
```
The sleep will just delay as you pointed out the shutdown, which doesn't make much sense to wait for the full duration just to return after it. Also, time.After no longer has a leak where the counters are not de-allocated, as of go 1.23

internal/eventingester/store/eventstore.go
```diff
@@ -88,27 +98,28 @@ func (repo *RedisEventStore) doStore(ctx *armadacontext.Context, update []*model
 		return nil
 	}
 
-	return ingest.WithRetry(func() (bool, error) {
-		var data []eventData
-		uniqueJobSets := make(map[string]bool)
+	var data []eventData
```
nikola-jokic, line 101: If I read this correctly, you can preallocate here with len(update)

internal/eventingester/store/eventstore.go
```diff
@@ -77,6 +76,17 @@ func (repo *RedisEventStore) Store(ctx *armadacontext.Context, update *model.Bat
 	return result.ErrorOrNil()
 }
 
+// Serialize renders update as JSON for the dead-letter topic.
+func (repo *RedisEventStore) Serialize(update *model.BatchUpdate) ([]byte, error) {
+	return json.Marshal(struct {
```
nikola-jokic, line 81: Not too important: I'd suggest extracting the struct as private package level struct for easier tests later. It doesn't have to be exported, but it would avoid having tests defining:
```go
var data struct {
  // redefined struct fields
}
```

It is unnecessary duplication when you can simply do:
```go
type redisEventStoreBatchUpdateMessage struct {
   // fields
}
```

nikola-jokic (approved):
Overall looks good, I have few comments but they are not blockers

PR #5098, opened 2026-08-05
Replace deprecated strings.Title in scheduler metrics

### LINKED ISSUE #4776: Replace deprecated strings.Title() in scheduler metrics
**Is your feature request related to a problem? Please describe.**

`strings.Title()` is deprecated since Go 1.18 and we still use it in two places in the scheduler metrics code.

**Describe the solution you'd like**

Replace both calls with `cases.Title(language.English).String(...)` from `golang.org/x/text/cases` and `golang.org/x/text/language`.

**Files to change**

- `internal/scheduler/metrics.go:484` - `strings.Title(strings.ToLower(phase))`
- `internal/scheduler/metrics.go:493` - `strings.Title(strings.ToLower(phase))`

**Example**

```go
// Before
phase: strings.Title(strings.ToLower(phase)),

// After
phase: cases.Title(language.English).String(strings.ToLower(phase)),
```

The `cases.Title` caser is safe for concurrent use, so you can create it once as a package-level variable rather than creating a new one on every call.

**Acceptance criteria**

- Both `strings.Title()` calls replaced
- No `strings.Title` import remains
- `go test ./internal/scheduler/...` passes
- `golangci-lint run ./internal/scheduler/...` passes

hh:
strings.Title has been deprecated since Go 1.18. Replace both call sites in internal/scheduler/metrics.go with cases.Title(language.English) from golang.org/x/text/cases. These are the only two uses of strings.Title in the repository.

The caser is created once per updateClusterMetrics call rather than at package level. cases.Caser is documented as potentially stateful and not safe for sharing between goroutines, and MetricsCollector.Collect may be called concurrently by the Prometheus registry. The x/text authors note concurrency safety only for cases.Fold, and deliberately not for Title, Upper or Lower; cases.titleCaser does mutate its embedded context on every Transform, and sharing one across goroutines is reported by the race detector.

This matters more than it looks, because sharing a caser corrupts output silently rather than failing loudly. Alertmanager shipped exactly this bug in its template FuncMap and it garbled user-visible notification text in production for months before it was tracked down; see prometheus/alertmanager#3278, fixed the same way, with a separate caser per call. Here the affected strings are Prometheus label values, where corruption would fragment time series rather than merely look wrong.

Creating the caser outside the node/executor loops also avoids reallocating it on every iteration.

Because these values are public Prometheus label values, output was verified unchanged for every JobRunState_name value - UNKNOWN, PENDING, RUNNING, SUCCEEDED and FAILED all produce byte-identical results under both implementations.

golang.org/x/text was already a direct dependency, so go.mod is unchanged.

Fixes #4776

### DIFF
--- a/internal/scheduler/metrics.go
+++ b/internal/scheduler/metrics.go
@@ -7,6 +7,8 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/exp/maps"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 	"k8s.io/utils/clock"
 
 	"github.com/armadaproject/armada/internal/common/armadacontext"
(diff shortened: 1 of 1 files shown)


hh:
Closing as a duplicate — I missed that #4783 already covers #4776 and has been open since March. Apologies for the noise; #4783 is the one to review.

PR #5102, opened 2026-08-06
replace ingress-nginx with Traefik for the local kind cluster

williamvega:
This change continues to use the Ingress API, and Traefik supports what is needed for local development without making armada code changes.

#### What type of PR is this?

enchancement

ingress-nginx is retired so it was replaced with Traefik v3.7.9 for local development.

  * The Traefik chart does not render app.kubernetes.io/component, so kindWaitUntilReady in kind.go needed a new selector.
  * namespace.yaml referenced the ingress-nginx ServiceAccount in its RoleBinding, now refs traefik.
  * Drop the nginx:1.27.0 preload as the ingress test case uses alpine:3.20.0.
  * The Traefik dashboard is exposed on localhost:5002 for local debugging.


_local/kind/traefik.yaml
```diff
@@ -0,0 +1,335 @@
+# Traefik ingress controller for the local kind cluster.
```
dejanzele, line 1: Should we helm template into a file? Any reason not to just helm install?

williamvega: No particular reason to vendor in the helm charts other than that is how it was done before with ingress-nginx, so we don't have to install helm in the ci workflow.

PR #5111, opened 2026-08-11
Fix scheduler missing job updates

JamesMurkin:
**Context**

The scheduler loads data from schedulerdb by fetching all changed job/run rows where `serial > previously highest serial`.

This works as long as job/run serials are monotonically increasing in commit order — i.e. a later committed change cannot have a lower serial than an earlier committed change.

This is enforced by only allowing one active writer to job/run data at a time, using a DB lock. The lock ensures that only one transaction can generate new serial numbers at a time, so once a transaction commits, its serial numbers will always be higher than those of any previously committed rows.

**Issue**

We've added new operations that were categorised as `ControlPlaneOperation` and therefore didn't require this lock. This was originally fine, as these operations only modified the executor settings table.

However, subsequent control plane operations were added (preempt/cancel node, preempt/cancel executor) which do modify job rows.

This resulted in job data being updated outside of the lock, breaking the guarantee that job/run serials are monotonically increasing in commit order.

As a result, the scheduler could miss updates, leaving some jobs stuck in an incorrect state — typically running, because the terminal update was missed.

**Fix**

This PR correctly categorises operations based on whether they modify job data and therefore require the DB lock. The operation types have also been renamed to make this distinction more explicit.


internal/scheduleringester/dbops.go
```diff
@@ -749,119 +747,119 @@ func definesRun[M ~map[string]V, V any](a M, b DbOperation) bool {
...
 }
 
 func (a DeleteExecutor) GetOperation() Operation {
-	return ControlPlaneOperation
+	return ExecutorDataImpactingOperation
 }
 
 func (pe PreemptExecutor) GetOperation() Operation {
-	return ControlPlaneOperation
+	return JobDataImpactingOperation
```
JamesMurkin, line 842: Note the last 6 operations have had their type changed to `JobDataImpactingOperation`

eleanorpratt (approved):

PR #5115, opened 2026-08-12
WIP: add RBAC testing to e2e suite

d2burkhalter:
Work in progress on adding role based access control to the e2e test suite starting with the global permissions. Added configs to keycloak for service accounts with different groups as well as human roles for local development. Added `auth_context` field to testspec to define what context to use when taking actions and `expectErrorCode` for actions denied when not authorized. Still working on how we want to define queues in testing because some persist between tests and other are created and destroyed for individual tests, would like everything to be defined within the testspec instead of using mage for setup but if it makes things too complicated. Also working on how we want to separate running the basic unauthorized tests and authorized tests since it will require updating the config that the services are run with and we are encountering wanting to modify the config for certain services within tests.


d2burkhalter:
@greptileai

d2burkhalter:
@greptileai

d2burkhalter:
@greptileai

PR #5118, opened 2026-08-13
Add auth-fake-executor profile for local development

nikola-jokic:
#### What type of PR is this?

Dev experience

makes it easier to test job transformations without running executor while still having auth logic.


d2burkhalter (approved):
LGTM

PR #5128, opened 2026-08-21
Simplfy how requestor is set in scheduler

JamesMurkin:
- Don't set UserId on sequence generated by the scheduler, as this can have the wrong user in the case multiple events happen at once that could set the user
   - Probably we should make all sequences generated by scheduler have user id "armada-scheduler" or just left blank as now
 - Use Deref as it is neater


nikola-jokic (approved):
Much better!

PR #5151, opened 2026-09-02
fix(lookout): use kubectl exec `--` separator in default command template

### LINKED ISSUE #5150: Lookout's default exec command template uses removed kubectl syntax
The COMMANDS pane in Lookout suggests an exec command of the form:

```
kubectl --context <cluster> -n <namespace> exec -it armada-<jobId>-0 /bin/sh
```

which comes from the default `commandSpecs` template in `config/lookout/config.yaml` (line 40 on current master) and `config/lookouthc/config.yaml` (line 36):

```yaml
template: "kubectl --context {{ runs[runs.length - 1].cluster }} -n {{ namespace }} exec -it armada-{{ jobId }}-0 /bin/sh"
```

Current kubectl rejects that form outright:

```
error: exec [POD] [COMMAND] is not supported anymore. Use exec [POD] -- [COMMAND] instead
See 'kubectl exec -h' for help and examples
```

The `exec [POD] [COMMAND]` syntax was deprecated and later removed from kubectl, so the suggested command fails on every recent kubectl release. The fix is inserting `--` before `/bin/sh` in both templates:

```yaml
template: "kubectl --context {{ runs[runs.length - 1].cluster }} -n {{ namespace }} exec -it armada-{{ jobId }}-0 -- /bin/sh"
```

Observed with the armada-operator kind quickstart (Lookout on :30000) and a current kubectl; the corrected command execs into the pod fine. (The `--context <cluster id>` part is understood to be by design, working when the operator names kubectl contexts after Armada cluster ids.)

pujitha24:
#### What type of PR is this?

Bug fix

Lookout's default "Exec" command template (in `config/lookout/config.yaml` and
`config/lookouthc/config.yaml`) generates a command of the form:

```
kubectl --context <cluster> -n <namespace> exec -it armada-<jobId>-0 /bin/sh
```

The `exec [POD] [COMMAND]` syntax (without `--` separating kubectl's own
arguments from the in-container command) was deprecated and has since been
removed from kubectl. Running the suggested command on any current kubectl
release fails with:

```
error: exec [POD] [COMMAND] is not supported anymore. Use exec [POD] -- [COMMAND] instead
```

This PR inserts `--` before `/bin/sh` in both templates so the generated
command uses the syntax current kubectl requires.

Fixes #

This is a two-line config change (one line per file) with no code paths
affected. Validation performed:
- `python3 -c "import yaml; yaml.safe_load(open(...))"` on both edited files
  to confirm they remain valid YAML.
- Repo-wide grep for the old template string and for other occurrences of
  `exec -it armada-` / `bin/sh` to confirm no other config file, doc, or test
  contains the same stale syntax.
- Confirmed no Go or TypeScript test hardcodes or snapshots this template
  string.

These files are default/example configs (used by the local dev quickstack)
and are not wired into any `go build`/`go test`/lint CI job, so there is no
additional local build/test command applicable beyond the YAML syntax check
above.

Fixes #5150

### DIFF
--- a/config/lookout/config.yaml
+++ b/config/lookout/config.yaml
@@ -37,7 +37,7 @@ uiConfig:
     - name: Logs
       template: "kubectl --context {{ runs[runs.length - 1].cluster }} -n {{ namespace }} logs armada-{{ jobId }}-0"
     - name: Exec
-      template: "kubectl --context {{ runs[runs.length - 1].cluster }} -n {{ namespace }} exec -it armada-{{ jobId }}-0 /bin/sh"
+      template: "kubectl --context {{ runs[runs.length - 1].cluster }} -n {{ namespace }} exec -it armada-{{ jobId }}-0 -- /bin/sh"
       descriptionMd: Execute a command on the job's container.
       alertMessageMd: |
         This will only work if the container is still running.
--- a/config/lookouthc/config.yaml
+++ b/config/lookouthc/config.yaml
@@ -33,7 +33,7 @@ uiConfig:
     - name: Logs
       template: "kubectl --context {{ runs[runs.length - 1].cluster }} -n {{ namespace }} logs armada-{{ jobId }}-0"
     - name: Exec
-      template: "kubectl --context {{ runs[runs.length - 1].cluster }} -n {{ namespace }} exec -it armada-{{ jobId }}-0 /bin/sh"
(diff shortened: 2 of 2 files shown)


dejanzele (approved):

nikola-jokic (approved):

nojaf:
Thanks everyone!

PR #5155, opened 2026-09-03
Categorize preempted jobs under a static preemption category

dejanzele:
The scheduler stamps `FailureCategory: "preemption"` on both terminal errors of a preempted run, and the preemption type string as the `FailureSubcategory`: `fairshare`, `urgency`, `optimiser`, `api`, `reconciler`, `unknown`, or `unknown-gang`. One function, `createEventsForPreemptedJob`, builds those errors for all three preemption paths: the scheduling round passes the job context's preemption type, the node reconciler passes `reconciler`, and a preemption request through the API passes `api`. An empty type falls back to `unknown`, the same default the preemption metrics use. The category constant lives in `internal/common/errormatch`, next to the `internal` category, and follows the same rule: the Error-construction site sets it, and the operator categorizer never does.

The Lookout ingester needs its own change, because `handleJobRunErrors` skips run-level preemption errors: the `JobRunPreempted` event owns the run state and the error text. The preempted-error case now appends an update that carries only the failure classification, so the `job_run` row gets the category and subcategory without touching state or error.

The subcategory reuses the scheduler's `PreemptionType` strings, the same values the preemption metrics expose, so one vocabulary covers the metrics and the Lookout rows.

Verification: unit tests cover the scheduler events, the type fallback, and the ingester passthrough. On a local stack with a real executor on a kind cluster, `armadactl preempt job` on a running job produced a Lookout run row with `failure_category=preemption` and `failure_subcategory=api`, run state `RUN_PREEMPTED`, and job state `PREEMPTED`. A second scenario saturated the cluster from one queue with preemptible jobs and submitted a non-preemptible job from another queue. The scheduler evicted a running pod by fair-share preemption, and the evicted run's row read `failure_category=preemption` and `failure_subcategory=fairshare`.

### DIFF
--- a/internal/common/errormatch/doc.go
+++ b/internal/common/errormatch/doc.go
@@ -10,4 +10,10 @@
 // for config validation. Non-pod conditions ([ConditionPreempted],
 // [ConditionLeaseReturned], [ConditionLeaseExpired], [ConditionAppError]) are also defined here
 // for use by the retry engine but are not included in [KnownConditions].
+//
+// The failure-category constants ([CategoryInternal], [CategoryPreemption])
+// and the Subcategory constants name the classifications that Armada
+// hard-codes on self-generated errors. The operator categorizer never
+// assigns them. Errors in [CategoryPreemption] carry the scheduler's
+// preemption type string as their subcategory.
 package errormatch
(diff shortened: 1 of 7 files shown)


dejanzele:
@greptileai

masipauskas (approved):

JamesMurkin (approved):

PR #5156, opened 2026-09-03
remove debug events and messages

nikola-jokic:
#### What type of PR is this?

Stability improvement

Reduce the payload size in redis to reduce memory usage

### DIFF
--- a/internal/eventingester/convert/conversions.go
+++ b/internal/eventingester/convert/conversions.go
@@ -56,6 +56,8 @@ func (ec *EventConverter) Convert(ctx *armadacontext.Context, eventsWithIds *uti
 
 		// Remove cancellation reason as it's not needed for public event store
 		clearCancellationReason(es)
+		clearDebugMessages(es)
+		removeDebugOnlyEvents(es)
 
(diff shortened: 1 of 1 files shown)


internal/eventingester/convert/conversions.go
```diff
@@ -112,3 +114,37 @@ func clearCancellationReason(es *armadaevents.EventSequence) {
...
+// Debug info is only consumed by Lookout, which reads directly from Pulsar, so it
+// is dead weight in Redis.
+func clearDebugMessages(es *armadaevents.EventSequence) {
+	for _, e := range es.Events {
+		switch event := e.GetEvent().(type) {
+		case *armadaevents.EventSequence_Event_JobRunErrors:
+			for _, err := range event.JobRunErrors.GetErrors() {
+				switch reason := err.Reason.(type) {
+				case *armadaevents.Error_PodError:
+					reason.PodError.DebugMessage = ""
```
JamesMurkin, line 128: Please double check these info doesn't currently end up exposed via the public events

JamesMurkin (approved):

